PR #88, opened 2014-07-08
Update noobs.md

ericthered:
SD card size greater than 32gb causes problems


ghollingworth:
Also, try setting the volume label...  We found a situation in the past where parted didn't like a filesystem formatted by a mac with no label

lurch:
Apologies for the delay, I've finally got round to documenting how to format an SDXC card for use with NOOBS, with https://github.com/raspberrypi/documentation/commit/320d77b2d8bf400055e296e25288bb91a5e4261a

PR #133, opened 2014-12-07
Update to bash scrip for VNC in boot (vncboot)

magamig:
- Option to set name of the user for vncserver.
- Easy to change geometry

### DIFF
--- a/remote-access/vnc/README.md
+++ b/remote-access/vnc/README.md
@@ -74,6 +74,7 @@ cd /etc/init.d/
 - Create a new file here containing the following script:
 
 ```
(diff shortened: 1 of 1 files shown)


ebenupton:
Any reason you've changed the script to explicitly use tightvncserver?

magamig:
I think it is the most used software and is easier to work with.
And changes I made allow a more simple and customizable way of running the VNC server. And, to have that options it was a need the selection of the software, so I chose the most common one.

lurch:
Should this be closed now, in light of other changes that have been made to the VNC page?

PR #139, opened 2014-12-23
Add extra dd example, for beginners

listlfa:
Some command line beginners, like me, need to see useable command line examples before we can fully understand the explanation paragraphs.


bennuttall:
We could do with an explanation of this command - it's for seeing the progress, right?

lurch:
Yeah, the page actually _already_ says:

> Instead of `dd` you can use `dcfldd`; it will give a progress report about how much has been written.

just a few paragraphs further down, so I'm not sure if this repetition is necessary?

LornaLynch:
Closing this, as the change seems repetitious.

PR #145, opened 2015-01-10
Update dt-blob-cam1.dts

element14matt:
Changed the NUM_CAMERAS from 2 to 1, for the single camera dts


XECDesign:
Well spotted, thank you.

PR #153, opened 2015-01-26
Update updating.md

Gadgetoid:
Removed erroneous reference to "rpi-update" which is being trained into users who shouldn't be using it and causing a support nightmare.

### DIFF
--- a/raspbian/updating.md
+++ b/raspbian/updating.md
@@ -18,21 +18,7 @@ However, there are occasional changes made in the Foundation's Raspbian image th
 
 ## Updating the kernel and firmware
 
-The kernel and firmware are installed as a Debian package, and so will also get updates when using the procedure above. These packages are updated infrequently (after extensive testing); if you want to try more recent experimental software, it's also easy to update to the latest available version using the [rpi-update](https://github.com/Hexxeh/rpi-update) tool.
-
-To run this update, simply run from the command line:
-
-```bash
-sudo rpi-update
-```
-
-You'll need the latest version of the tool, so make sure you've recently run `apt-get update` and `upgrade` first.
-
-Your system date/time will need to be accurate too, or it will fail to communicate with the server. To manually set your system time, run the following command (with the *actual* time and date):
-
-```bash
-sudo date -s "8 JAN 2015 12:00:00"
-```
+The kernel and firmware are installed as a Debian package, and so will also get updates when using the procedure above. These packages are updated infrequently and after extensive testing.
 
 ## Running out of space
 


Gadgetoid:
To lend some weight to this ongoing issue:

https://github.com/pimoroni/skywriter-hat/issues/2#issuecomment-71555975
http://www.reddit.com/r/raspberry_pi/comments/2tpfvz/piglow_users_check_this_before_updating_your/
http://forums.pimoroni.com/t/display-o-tron-3000/29/79?u=gadgetoid

lurch:
I'm not going to offer any opinion on how or if this should get fixed; but just as an FYI, `rpi-update` is *also* mentioned in [linux/kernel/updating.md](https://github.com/raspberrypi/documentation/blob/master/linux/kernel/updating.md) and [troubleshooting/hardware/camera.md](https://github.com/raspberrypi/documentation/blob/master/troubleshooting/hardware/camera.md)

Gadgetoid:
Good observation! It could be argued that you shouldn't *need* the very latest "firmware" in order to make something as common as the camera function.

I'm not entirely sure it should be completely removed, but it shouldn't be posed as a magic solution to problems and should come with an appropriate warning where it's mentioned.

The updating the kernel page is, if you subscribe to the typical "spirit" of Linux, just flat out incorrect.

asb:
For what it's worth, my personal view is that updating via rpi-update should be de-emphasised. I don't think removing any mention as this PR does is correct, but some bold text that says something to the effect of "Only run rpi-update if you've been advised to do so, are having problems you believe it may fix, and are happy checking on the forums for any issues that may arise from the update".

waveform80:
Just a quick clarification, if possible. It's good to hear that `rpi-update` is being "de-emphasised" in favour of straight packages (it's definitely the Right Way to do things). Unfortunately I've got several references to `rpi-update` in the picamera docs, e.g. [here](http://picamera.readthedocs.org/en/release-1.9/fov.html#camera-modes) (see the Note in blue). Should I be replacing those with `sudo apt-get upgrade` (as seems to be suggested by the PR), or `sudo apt-get dist-upgrade` (which is the method I'm generally used to under Ubuntu for kernel upgrades, as the straight `upgrade` sub-command misses out kernel packages)? For that matter is there any difference in these two sub-commands on Raspbian?

asb:
@waveform80: my suggestion would be that once the package at archive.raspberrypi.org includes the fix you mention, you stop encouraging rpi-update (e.g. last firmware package is 19th Dec). `apt-get dist-upgrade` will intelligently handle cases where as part of the upgrade some packages are removed. This isn't the case with the standard firmware/kernel packaging we use on the Foundation Raspbian image. `apt-get dist-upgrade` is perfectly safe though. I think the first time I had to suggest someone use it was with the recent UI changes because we imported a newer pcmanfm and other associated packages, which meant some packages had to be removed. This would only happen with a dist-upgrade rather than a plain 'upgrade'.

waveform80:
@asb: actually, that's just it - it looks like the firmware package is perfectly well up to date for the camera. I've just tried a stock NOOBS raspbian and run `apt-get upgrade` on it, vs. my development Pi which obviously I've been using `rpi-update` on, and the stock NOOBS one is now happily on firmware 730 which is more than recent enough to include everything important for the camera.

If `dist-upgrade` is tolerably safe on Raspbian I think I'll start converting all the references to `rpi-update` to that for the picamera 1.10 release. One further question though: if someone's been using `rpi-update` historically (as many camera users will have) is `apt-get dist-upgrade` going to be enough to get them back on the right path or are there going to be other steps required? If it's simplest to suggest re-installing NOOBS I can always do that in the installation/upgrade section of the docs.

Gadgetoid:
@waveform80 @asb @lurch I really appreciate your feedback and mindfulness of this topic, thank you.

@waveform80 I've been using "sudo apt-get install --reinstall raspberrypi-bootloader" to get my Raspbian installs back into a stable state- I have been bouncing back and forth between bleeding edge and normal recently to feel out upcoming problems. I don't know if this is the right solution, but it seems to work.

waveform80:
@Gadgetoid that's good to know. I'll try that on my development Pi along with `dist-upgrade` and see what happens. If it all works smoothly, I should probably include something mentioning that in the upgrade sections of the picamera manual. Many thanks!

lurch:
Yeah, there's all a big bunch of overlaps, and it can be confusing at first which files get updated by what... ;-)

Going back to the "very start", there's [raspberrypi/linux](https://github.com/raspberrypi/linux) (sourcecode for Pi Linux [kernel](https://www.kernel.org/)) and [raspberrypi/userland](https://github.com/raspberrypi/userland) (sourcecode for Pi [libraries](http://www.raspberrypi.org/open-source-arm-userspace/) and applications - vcgencmd, tvservice, [raspivid](http://www.raspberrypi.org/documentation/raspbian/applications/camera.md), [hello_pi](http://www.raspberrypi.org/documentation/usage/demos/README.md), etc.).
Both of these then get [compiled](http://www.raspberrypi.org/documentation/linux/kernel/building.md), and get added to [raspberrypi/firmware](https://github.com/raspberrypi/firmware), along with the binary-only bootcode.bin, start.elf and fixup.dat files ([boot files](https://github.com/raspberrypi/noobs/wiki/Standalone-partitioning-explained#standalone-booting) / low-level firmware).

[raspberrypi/firmware](https://github.com/raspberrypi/firmware) in turn then gets mirrored to [Hexxeh/rpi-firmware](https://github.com/Hexxeh/rpi-firmware) (same files, just in a slightly re-arranged directory hierarchy), and it's this latter repository that is the source of files when you use [rpi-update](https://github.com/Hexxeh/rpi-update). So using `rpi-update` updates the low-level firmware, the kernel (& modules) as well as the Pi-specific libraries (including header files) and applications. And this all exists 'outside' of any package manager and blindly overwrites any pre-existing files - the only way of telling what rpi-update has 'done' is to look at `/boot/.firmware_revision` to see what git-hash from [rpi-firmware](https://github.com/Hexxeh/rpi-firmware/commits/master) you're currently at.

Using `apt-get upgrade` OTOH installs .deb packages from e.g. [archive.raspberrypi.org](http://archive.raspberrypi.org/debian/pool/main/r/raspberrypi-firmware/) which of course *does* integrate properly with the package manager. And they're also properly split into neat sub-packages (libraspberrypi-bin, libraspberrypi-dev, libraspberrypi-doc, libraspberrypi0 & raspberrypi-bootloader) rather than the all-or-nothing that rpi-update gives you. So I guess at least in *theory* you ought to `apt-get install --reinstall` **all** of the packages that rpi-update 'tramples over' ;-)

And because rpi-update and apt-get aren't aware of each other, if you run them in the wrong order (e.g. running rpi-update before apt-get upgrade) it's possible for apt-get upgrade to actually overwrite the kernel installed by rpi-update with an older one, but if you try running rpi-update again it will believe that everything is already all up-to-date (until you manually delete /boot/.firmware_revision).

Using rpi-update now includes [DeviceTree](http://www.raspberrypi.org/documentation/configuration/device-tree.md) by default, whereas apt-get still only 'knows' about the pre-device-tree kernel included in the [Christmas Eve](http://www.raspberrypi.org/merry-christmas-got-a-new-pi/) release of Raspbian. I *suspect* that the next release of Raspbian (whenever that happens) might also include the DT kernel (at which point it'll be available to apt-get too) but @asb obviously knows more about that than I do.

And as if all that wasn't confusing enough, rpi-update used to be a [manual install](https://github.com/Hexxeh/rpi-update/blob/master/README.md) but it too is now installable from [archive.raspberrypi.org](http://archive.raspberrypi.org/debian/pool/main/r/rpi-update/) ;-)

(And I dare not mention http://nightly.raspberrypi.org/ !! But AFAIK that's (currently?) only used for detecting build-failures, rather than producing 'user installable' packages.)


Sorry for such a long comment... hopefully it helps and hasn't just confused you all even more! I'm sure @asb or @popcornmix will point out any errors in what I've written.

waveform80:
@lurch ah, okay - that's some good detail there. I'll have to have a think about how to word this in the picamera docs so as not to horribly confuse new/existing users and what advice I should be giving for anyone who's been following `rpi-update` (my gut instinct at this point is to start with something like "if you want to keep things simple, just create a fresh Raspbian install and skip the rest of this section" ;-)

bennuttall:
Thanks for the info @lurch. What do we think of the PR?

Gadgetoid:
From what I've read here, and in the Raspberry Pi forums, there seems to be an overwhelming body of evidence to suggest that if you do want a bleeding-edge kernel, rpi-update is most definitely not the right way to go about it. How and when it's retired from the bug reporting and testing process that the Kernel undergoes during development is a different discussion, but that this discussion is even happening is a pretty clear indicator that it's not suitable as a tool for anyone else.

I refer to: http://www.raspberrypi.org/forums/viewtopic.php?f=66&t=80023

These are people who have every reason to use rpi-update, saying there's no reason to be using rpi-update.

I appreciate that this is a phenomenally brutal PR with no attempt to even point people towards an alternative, but I operate on the assumption that anyone who should be using a bleeding-edge kernel can figure out how to get one on their own.

On the flip side, the documentation would not be comprehensive without these instructions, so perhaps they need only appropriate warnings.

Gadgetoid:
I also realised that I did not read the contributing guidelines and filed this as a pull request before an issue. Sorry. I've retrospectively raised an issue to draw attention to the wider problem: https://github.com/raspberrypi/documentation/issues/155

lurch:
@bennuttall I'm happy to agree with what @asb [says](#issuecomment-71621205)

(for anybody not aware - it tends to be @asb that updates [archive.raspberrypi.org](http://archive.raspberrypi.org/debian/pool/main/r/raspberrypi-firmware/) (the RPF-maintained .deb packages) as used by apt-get, and @popcornmix that updates [Hexxeh/rpi-firmware](https://github.com/Hexxeh/rpi-firmware) as used by rpi-update. And then @plugwash (who appears in the forum thread that @Gadgetoid linked to) maintains [archive.raspbian.org](http://raspbian.org/RaspbianRepository) - the pure Raspbian base packages, before the RPF packages get added on top.)

waveform80:
I'm happy with the PR as it stands, given that these docs have an emphasis on being as simple as possible (so there shouldn't be tons of prose defining every corner case) and an emphasis on defining best practice (which I think everyone seems to be agreed, is using apt-get to manage as much as possible, up and including the kernel).

For the picamera docs, I'm writing something rather more extensive as I know there are numerous camera users out there who've used rpi-update in the past (sometimes at my behest - sorry!) because during the last half of 2014 there was a furious amount of development on the camera and rpi-update was heavily used by those wanting to test (for example) long exposures, motion vector data, etc. Feel free to nick anything from there (I should have it all done and pushed by the weekend), but as mentioned above I think the existing PR is fine as is.

plugwash:
On 29/01/15 11:37, Andrew Scheller wrote:
>
> @bennuttall <https://github.com/bennuttall> I'm happy to agree with 
> what @asb <https://github.com/asb> says <#issuecomment-71621205>
>
> (for anybody not aware - it tends to be @asb <https://github.com/asb> 
> that updates archive.raspberrypi.org 
> <http://archive.raspberrypi.org/debian/pool/main/r/raspberrypi-firmware/> 
> (the RPF-maintained .deb packages) as used by apt-get, and @popcornmix 
> <https://github.com/popcornmix> that updates Hexxeh/rpi-firmware 
> <https://github.com/Hexxeh/rpi-firmware> as used by rpi-update. And 
> then @plugwash <https://github.com/plugwash> (who appears in the forum 
> thread that @Gadgetoid <https://github.com/Gadgetoid> linked to) 
> maintains archive.raspbian.org 
> <http://raspbian.org/RaspbianRepository> - the pure Raspbian base 
> packages, before the RPF packages get added on top.)
>
I just got this email but it's kinda difficult to tell what you are 
talking about without context.

bennuttall:
@plugwash See the issue thread on GitHub: https://github.com/raspberrypi/documentation/pull/153

It's regarding a suggestion to drop `rpi-update` from the documentation.

PR #157, opened 2015-02-03
Update command line instruction in graphical-mode

memmaker650:
Changed 1m to 1M in dd command. 1m doesn't means anything in dd command.


ebenupton:
This usage is correct for Mac OS X.

bennuttall:
Weird auto-reopening caused by waffle: https://twitter.com/ben_nuttall/status/573301869000445952

PR #172, opened 2015-02-20
update some Device Tree link

eshizhan:


pelwell:
Thank you for contributing these changes. Could you modify the SPI and I2C raspi-config sections to say that the options are to enable or disable the respective (SPI or I2C) interface (which is in addition to controlling the loading of the module)?

It would be better if you could squash the last two commits.

PR #177, opened 2015-02-24
Add hdmi_cvt and filters

uecasm:
hdmi_cvt docs taken from http://www.raspberrypi.org/forums/viewtopic.php?f=29&t=24679.
Filter docs taken from https://github.com/raspberrypi/firmware/issues/320 and https://github.com/raspberrypi/firmware/issues/361.
I think there may be another filter but I don't know the details enough to document it.

### DIFF
--- a/configuration/config-txt.md
+++ b/configuration/config-txt.md
@@ -379,6 +379,33 @@ Your HDMI monitor may support only a limited set of formats. To find out which f
 
 The `edid.dat` should also be provided when troubleshooting problems with the default HDMI mode.
 
+### Custom mode
+
+If your monitor requires a mode that is not in one of the tables above, then it is possible to define a custom [CVT](http://en.wikipedia.org/wiki/Coordinated_Video_Timings) mode for it instead:
+
+    hdmi_cvt=<width> <height> <framerate> <aspect> <margins> <interlace> <rb>
+
+| Value | Default | Description |
+| --- | --- | --- |
+| width | (required) | width in pixels |
+| height | (required) | height in pixels |
+| framerate | (required) | framerate in Hz |
+| aspect | 3 | aspect ratio 1=4:3, 2=14:9, 3=16:9, 4=5:4, 5=16:10, 6=15:9 |
+| margins | 0 | 0=margins disabled, 1=margins enabled |
+| interlace | 0 | 0=progressive, 1=interlaced |
+| rb | 0 | 0=normal, 1=reduced blanking |
+
+(Fields at the end can be omitted to use the default values.)
+
+Note that this simply _creates_ the mode (group 2 mode 87); in order to make the Pi use this by default you must add some additional settings.  As an example, the following selects an 800x480 resolution and enables audio drive:
+
+    hdmi_cvt=800 480 60 6
+    hdmi_group=2
+    hdmi_mode=87
+    hdmi_drive=2
+
+This may not work if your monitor does not support standard CVT timings.
+
 ### Generic display options
 
 #### hdmi_force_hotplug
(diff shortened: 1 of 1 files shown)


lurch:
Looks good! @popcornmix will be able to give more accurate info than me, but here's my 2 pence... (whoops, ended up being more like 50pence!)

The documentation about `hdmi_cvt` seems to be lacking the critical point that the new mode defined always becomes the new 'mode 87', and it *always* needs to be accessed via DMT i.e. 'group 2'.

And it's probably worth incorporating the "The first three arguments are required. The remaining ones are optional. aspect defaults to 16:9 if not specified." info from the forum post too? (in a nicer format) Presumably the other settings default to 0, @popcornmix ?

And for reference purposes, perhaps including the wikipedia CVT link into this page would be useful too?

I think the "Filters" header should be renamed to "Conditional Filters" to make it clear these aren't e.g. anti-alias or blur/sharpening filters ;)

To match the formatting of the rest of the page, could you put 'config.txt' always in backquotes, so it appears as `config.txt`. And I think 'card' ought to be consistently clarified as 'SD card'. Not sure if the 'card image' is relevant?

IMHO "However if different monitors are connected to the same Pi" makes it sound like the Pi has multiple simultaneous video outputs? ;) Perhaps "However if the same Pi gets swapped between different monitors"? And similarly perhaps "swapped between" when referring to the SD card sounds better than "used on"?

Is it worth mentioning in the filter-intro-paragraph that filters always have to be enclosed in [square brackets] ? (although it's probably obvious from the examples) Or maybe the headers should use square-brackets rather than back-ticks.

I'm not sure if 'processed on' is the right verb to use, perhaps 'applied to'? (this is probably just a personal preference thing). And it's even more nit-picky but I'm not sure if "any lines below" or "any settings below" sounds better than "anything below"?

I think after the initramfs example, I might be inclined to add "(remembering to apply the [all] filter afterwards, so that subsequent settings aren't applied only to Pi2 hardware)" or similar.

Like mentioned earlier, perhaps "When using a single card with multiple monitors" would be better as "When using the same SD card with different monitors"?

In the Combining Filters section, I think "# settings here are applied..." is clearer than "# settings applied..."


**Please** bear in mind these are all just my personal opinion / suggestions (feel free to ignore them) and please don't let it put you off submitting future PullRequests :-)

P.S. I've now fixed https://github.com/raspberrypi/documentation/commit/f4cbd2d8eeeed06dfd647dbd04fb9e1fac1163a3 so if you rebase from master, the Travis CI build for this PR should then succeed.

popcornmix:
Thanks for this. Looks good. I agree with @lurch's comments. Additionally:

> Note that this setting has a space instead of an equals sign like most other settings

Actually all config.txt settings can take a space or an equals. The equals is preferred so I'd use that in the examples and not mention the space.

The other config.txt filter is by serial number. E.g.
```
[0x12345678]
arm_freq=1000
[all]
```
would only apply the overclock to board with serial number 0x12345678.

lurch:
> Actually all config.txt settings can take a space or an equals. The equals is preferred so I'd use that in the examples and not mention the space.

@popcornmix Oooh, never knew that! Does that mean that the **NOTE** on https://github.com/raspberrypi/documentation/blob/master/configuration/config-txt.md#initramfs is also wrong, or is that still an exception to the rule? Guess I ought to update it to mention the `followkernel` option too...

Never knew about the serial number filter either - I guess a useful application of it would be if you had a particular Pi where you knew the warranty-voided bit was already set, and so that's the only Pi that you'd want to apply your "pushing the limits" overclock settings too :-)

popcornmix:
The things with rules, is there's always an exception...
Looking  at the code, there are two entries parsed outside the common parsing code, they are "include" and "initramfs" (they were added by someone other than me), so the initramfs documentation is correct.

You're probably going to say next you don't know what "include" does...

pelwell:
Not to mention the device tree additions (but they do accept spaces).

lurch:
> You're probably going to say next you don't know what "include" does...

I can guess, but to avoid user confusion, perhaps it should remain undocumented? ;-)

uecasm:
> The documentation about hdmi_cvt seems to be lacking the critical point that the new mode defined always becomes the new 'mode 87', and it always needs to be accessed via DMT i.e. 'group 2'.

I tried to imply that with the example below that selected the mode, but you're right, it should probably be more explicit.  (I did consider that at the time, but thought the example might have been sufficient.)

> P.S. I've now fixed f4cbd2d so if you rebase from master, the Travis CI build for this PR should then succeed.

Done.  I think I've addressed all your other comments as well.

I've done the re-edits as a separate commit so that you can check before and after, but let me know if you would prefer I squash it down to a single commit, or make separate commits for the hdmi_cvt and filters.

> Actually all config.txt settings can take a space or an equals. The equals is preferred so I'd use that in the examples and not mention the space.

Huh.  I thought I remembered reading somewhere that this particular one didn't work with an =.  I agree that using the = is nicer, so I've amended accordingly.

> The other config.txt filter is by serial number.

Ok, I can add something about that too.  One question though: where do you find the serial number of a Pi?

uecasm:
So, while I'm messing with this file, are there any other things that ought to be added?

In particular, am I correct in thinking that `kernel=foo.img` will make "foo.img" load for both Pi1 and Pi2, without other filters?  Or does Pi2 need `kernel7=foo.img` or something instead?

configuration/config-txt.md
```diff
@@ -612,6 +637,96 @@ It's generally a good idea to keep the core temperature below 70 degrees and the
...
+This is the most basic filter: it resets all previously set filters and allows any settings listed below it to be processed on all hardware.
+
+    [all]
+
+It is usually a good idea to add an `[all]` filter at the end of groups of filtered settings to avoid unintentionally combining filters (see below).
+
+### The `[pi1]` and `[pi2]` filters
+
+Any settings below a `pi1` filter will only be applied to Pi1 (A, A+, B, B+) hardware.
+Any settings below a `pi2` filter will only be applied to Pi2 hardware.
```
lurch: Square brackets around the filter names?

configuration/config-txt.md
```diff
@@ -612,6 +637,96 @@ It's generally a good idea to keep the core temperature below 70 degrees and the
...
+
+    [pi1]
+    [pi2]
+
+These are particularly useful for defining different `kernel`, `initramfs`, and `cmdline` settings, as the Pi1 and Pi2 require different kernels.  For example, to define separate initramfs images for each:
+
```
lurch: Dunno if it'd be worth squeezing in something here about being useful for overclocking too, as the Pi1 and Pi2 run at different default frequencies ?

@popcornmix I guess the "Overclocking options" table (elsewhere in this page) ought to be updated with the separate default values for Pi2?

configuration/config-txt.md
```diff
@@ -379,6 +379,31 @@ Your HDMI monitor may support only a limited set of formats. To find out which f
...
+| height | (required) | height in pixels |
+| framerate | (required) | framerate in Hz |
+| aspect | 3 | aspect ratio 1=4:3, 2=14:9, 3=16:9, 4=5:4, 5=16:10, 6=15:9 |
+| margins | 0 | 0=margins disabled, 1=margins enabled |
+| interlace | 0 | 0=progressive, 1=interlaced |
+| rb | 0 | 0=normal, 1=reduced blanking |
+
+(Fields at the end can be omitted to use the default values.)
+
+Note that this simply creates the mode (group 2 mode 87); in order to make the Pi use this by default you must add some additional settings.  As an example, the following selects an 800x480 resolution and enables audio drive:
```
lurch: Just a stylistic thing, but would this look better as "...this simply *creates* the mode..." ? (added italics)

configuration/config-txt.md
```diff
@@ -612,6 +637,96 @@ It's generally a good idea to keep the core temperature below 70 degrees and the
...
+    [EDID=VSC-TD2220]
+    hdmi_group=2
+    hdmi_mode=82
+    [all]
+
+This forces 1920x1080 DVT mode for this monitor, without affecting any other monitors.
+
+Note that these settings apply only at boot, so the monitor must be connected at boot time and the Pi must be able to read its EDID information to get the correct name.  Hotplugging a different monitor after boot will not reselect different settings.
+
+### The serial number filter
```
lurch: I guess this is the only 'filter header' which doesn't use square brackets, but in this case there's probably nothing to suitable to put between square brackets anyway!

configuration/config-txt.md
```diff
@@ -612,6 +637,96 @@ It's generally a good idea to keep the core temperature below 70 degrees and the
...
+
+Sometimes settings should only be applied to a single specific Pi, even if you swap the SD card to a different one.  Examples include license keys and overclocking settings (although the license keys already support SD card swapping a different way).  You can also use this to select different display settings even if the EDID identification above is not possible for some reason (provided that you don't swap monitors between your Pis).
```
lurch: I guess you missed out the "in" from "support SD card swapping in a different way" ?

And I guess this is also useful for configuring different composite video output modes (overscan etc.) on a per-Pi basis, as composite obviously doesn't provide any EDID info.

configuration/config-txt.md
```diff
@@ -612,6 +637,96 @@ It's generally a good idea to keep the core temperature below 70 degrees and the
...
 
+## Conditional Filters
+
+When a single SD card (or card image) is only being used with one Pi and one monitor, it's easy to simply set `config.txt` as required for that specific combination and keep it that way, amending only when something changes.
+
+However if one Pi is swapped between different monitors, or if the SD card (or card image) is being swapped between multiple Pis, a single set of settings may no longer be sufficient.  Conditional filters allow you to make certain sections of the config file used only in specific cases, allowing a single `config.txt` to create different configurations when read by different hardware.
+
+### The `[all]` filter
+
+This is the most basic filter: it resets all previously set filters and allows any settings listed below it to be processed on all hardware.
```
lurch: This line still uses "processed on" rather than "applied to"

lurch:
Brilliant, thanks @uecasm :+1: 

> Also, am I correct in thinking that kernel=foo.img will make "foo.img" load for both Pi1 and Pi2, without other filters?

Yes. After all, that's the reason https://github.com/raspberrypi/firmware/issues/361 was requested ;-)

configuration/config-txt.md
uecasm: Yes, I added square brackets around the names in the headings, as per your suggestion.  Or did you mean something else?

configuration/config-txt.md
uecasm: Omitting "in" still makes sense in spoken English, though you're right that in a technical document it should probably be fully spelled out.

uecasm:
>> Also, am I correct in thinking that kernel=foo.img will make "foo.img" load for both Pi1 and Pi2, without other filters?

> Yes. After all, that's the reason raspberrypi/firmware#361 was requested ;-)

I was a little worried about this because I tried it out and it didn't seem to be working.  Turned out I just had a typo when I renamed the `kernel7.img`.  All good now.

Committed some more amendments.  Let me know if there's anything else to change, or if you want me to merge/rearrange the commits. :+1:

popcornmix:
Thanks @uecasm. Can you squash down - I think a single commit is best.

uecasm:
Done.

popcornmix:
Looks good to me. @lurch feel free to merge if you are happy.

lurch:
Merged. Thanks very much Gavin.

PR #178, opened 2015-02-26
Fixed code example @ Python usage page

hatarist:
There was a slight mistake at the first example of Python's interpreter usage.


bennuttall:
Thanks!

PR #183, opened 2015-03-02
Better instructions for verifying data on SD card

Kagee:
Just dd'ing the SD card will in many cases give a image that is larger than the original file, thus both sha1sum and md5sum will fail. Replaced diff with sha1sum, since the latter is used on other pages.


lurch:
Hmmm, I wonder if the 'dd *from* the card' instructions should even be in there in the first place? I certainly don't expect it'll be something that needs to be done by many users.
What do ya reckon @asb ? If we do keep that line, then this PR does indeed make sense.

@Kagee just as an aside: using `diff -q file1 file2` will stop after the first difference has been detected, and so may be much quicker than md5-ing both files, which obviously has to read the entirety of both files.

Issue #170, opened 2015-02-19
Raspbian MacOS installation uses slow device


rhiller:
In http://www.raspberrypi.org/documentation/installation/installing-images/mac.md the instructions specify the use of /dev/diskn in the `sudo dd bs=1m if=path_of_your_image.img of=/dev/diskn`.

This runs much much faster if instead the raw disk device is used: `sudo dd bs=1m if=path_of_your_image.img of=/dev/rdiskn`.

An internet search of "os x dd very slow" turns up lots of relevant links, including: http://daoyuan.li/solution-dd-too-slow-on-mac-os-x/

ebenupton:
Please propose a pull request.

flashspys:
this issue can be closed. PR is merged

Issue #58, opened 2014-05-11
Use /dev/rdiskn when installing images on OS X


lukasbestle:
Hi,

I just read your documentation on how to install the disk image for Raspbian to a SD card on OS X.
It says that the user should use `dd` with `/dev/diskn`. On OS X, `/dev/rdiskn` is a lot faster with SD cards, so I guess that would make the process simpler and faster for users.

You write about it in the alternative method description, but it should be used in the first place.

Cheers,
Lukas

ebenupton:
Closed, having merged PR #170.

PR #192, opened 2015-03-09
Use raw disk device on mac since it is much faster

rhiller:
In http://www.raspberrypi.org/documentation/installation/installing-images/mac.md the instructions specify the use of /dev/diskn in the sudo dd bs=1m if=path_of_your_image.img of=/dev/diskn.

This runs much much faster if instead the raw disk device is used: sudo dd bs=1m if=path_of_your_image.img of=/dev/rdiskn.

An internet search of "os x dd very slow" turns up lots of relevant links, including: http://daoyuan.li/solution-dd-too-slow-on-mac-os-x/

### DIFF
--- a/installation/installing-images/mac.md
+++ b/installation/installing-images/mac.md
@@ -11,7 +11,7 @@ On Mac OS you have the choice of the command line `dd` tool or using the graphic
 - From the terminal run:
 
     ```
-    sudo dd bs=1m if=path_of_your_image.img of=/dev/diskn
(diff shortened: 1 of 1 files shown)


lurch:
Is this a duplicate of #58 and #170 ?

rhiller:
As I understand it, I believe that #58 was closed on March 4 and  #170 was closed after my PR was merged on March 23.

XECDesign:
I see two issues being closed, but no associated PR. In any case, this seems fine.

XECDesign:
Fixed in PR #239

PR #194, opened 2015-03-11
Update remote-access/vnc/README.md

Basilmir:
Updated the shell script to reflect that on a fresh install people will pe using the user "pi" for all their setup needs. Setting this up as root, which is not even set by default kind of defeats the whole purpose of the pi user.

### DIFF
--- a/remote-access/vnc/README.md
+++ b/remote-access/vnc/README.md
@@ -87,8 +87,8 @@ cd /etc/init.d/
 #! /bin/sh
 # /etc/init.d/vncboot
 
-USER=root
-HOME=/root
+USER=pi
+HOME=/home/pi
 
 export USER HOME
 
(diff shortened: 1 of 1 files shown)


bennuttall:
Looks sensible to me. Anyone know of a reason it should be root?

AndrewFromMelbourne:
Is there any way we could get this pull request merged? The question of starting vnc on boot comes up in the forums very often and normally I would point them to the documentation. Unfortunately, I can't point them there at the moment because of this issue.

AndrewFromMelbourne:
I have now tested the changes myself and can confirm they work.

DirkUK:
Can somebody please commit this pull request.
Problems with these instructions keep coming in on the forum on a regular basis
@bennuttall ?

bennuttall:
@lurch @XECDesign anyone?

lurch:
I've not tested it myself, but other people say they've tested the proposed fix and it works okay, so I can't see any reason why it shouldn't be merged.

waveform80:
Just given it a try, and it works nicely for me

waveform80:
(on a fresh NOOBS install on a Pi 2 I should add)

Basilmir:
Works on the Pi 2 too. This is where I tested it, I don't own any of the older versions.

PR #200, opened 2015-03-21
Fix mysqldump syntax

charleswhchan:
According to MySQL documentation, there is a space between -u <username>. 
(On the other hand, for password, there is no space -p<password>) is correct.

Reference:
http://dev.mysql.com/doc/refman/5.1/en/mysqldump.html#option_mysqldump_user
http://dev.mysql.com/doc/refman/5.1/en/mysqldump.html#option_mysqldump_password


bennuttall:
The space isn't mandatory. `mysql -uroot -ppass` does work.

charleswhchan:
You are right. I just tested it. 

Not problem.

PR #201, opened 2015-03-21
Reference (Debian) File System Hierarchy Standard

charleswhchan:


bennuttall:
That's really helpful, thanks.

I can't see a licence on that page. If it's Creative Commons I'd be happy to copy the contents over too - still referencing the page, of course.

Issue #130, opened 2014-11-16
Instructions about starting and reloading NGINX


ethanyoung:
It would be better to specify the starting and reloading steps for NGINX in [NGINX setup page](https://github.com/raspberrypi/documentation/blob/master/remote-access/web-server/nginx.md). Starting is after the installation and reloading is after php configuration. Otherwise there will be "page not available" and 403 errors in the browser (at least I encountered these problems).

ebenupton:
PR please.

ethanyoung:
The starting has been solved in  2f74577.

The reloading was in #204.

PR #204, opened 2015-03-28
Update nginx.md

fmirko:
On my raspberry pi, I need to restart the nginx  http server before to be able to see index.php

### DIFF
--- a/remote-access/web-server/nginx.md
+++ b/remote-access/web-server/nginx.md
@@ -115,4 +115,8 @@ Add some dynamic PHP content by replacing the current content:
 <? echo phpinfo(); ?>
 ``` 
 
-Save and refresh your browser. You should see a page with the PHP version, logo and current configuration settings.
+Save and restart the nginx http server
+``` bash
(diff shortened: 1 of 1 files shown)


XECDesign:
1) Do you need to restart, or does a reload work?
2) You shouldn't need to do this after adding the index file, only after editing the config file.

Would it not be better to move this to the Enable PHP in NGINX section and change restart to reload? Or can you confirm that you actually need to restart after renaming the file?

fmirko:
Le 28/03/2015 15:35, XECDesign a écrit :
>
> 1) Do you need to restart, or does a reload work?
>
I will check again, but reloading/refreshing the page didn't work.
I did : /etc/init.d/nginx restart, but didn't try the /etc/init.d/nginx 
reload.

I let you when I redo the test.

ethanyoung:
I had a similar problem. A reload was needed after I edited PHP config file. I mentioned this in #130 .

bennuttall:
Fixed in #236

PR #208, opened 2015-04-04
Added information on formatting your SD card

LOVE2CODE123:
I have added information on how to format the SD card before flashing it.

### DIFF
--- a/installation/installing-images/mac.md
+++ b/installation/installing-images/mac.md
@@ -2,6 +2,23 @@
 
 On Mac OS you have the choice of the command line `dd` tool or using the graphical tool ImageWriter to write the image to your SD card.
 
+## Formaating your SD card first (Optional)
+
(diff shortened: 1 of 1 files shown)


XECDesign:
If the card is getting imaged, there is no point formatting it.

lurch:
> If the card is getting imaged, there is no point formatting it.

Agreed - imaging the card always overwrites the *entire* card contents, regardless of what was on there before. You only need to format your card first when using NOOBS, and there's already [instructions](https://github.com/raspberrypi/documentation/blob/master/installation/noobs.md#how-to-format-an-sd-card-as-fat) for that.

Should this PR be closed as invalid?

XECDesign:
I don't know. I've seen people who should know what they're talking about insist that formatting is a good idea anyway, but the reasons given were a bit vague and sketchy. Up to you.

PR #212, opened 2015-04-17
CMIO second camera clarifications

torkeldanielsson:
I made some clarifications on how to setup and connect the second
camera. I also added a note about a reported issue with GPIO pin 2 and
3.

### DIFF
--- a/hardware/computemodule/cmio-camera.md
+++ b/hardware/computemodule/cmio-camera.md
@@ -116,3 +116,5 @@ pin_define@CAMERA_1_SCL_PIN { type = "internal"; number = <29>; };
 [Enable a CAM1 only](dt-blob-cam1.dts)
 
 [Enable both cameras](dt-blob-dualcam.dts)
+
+*As of 2015-04-17 some users are reporting issues with GPIO pin 2 and 3 for camera control. Pin 4 and 5 is reported to work. The issue is being investigated.*


XECDesign:
The dt-blob.bin linked in the quickstart section is already the dualcam version, so there is no need to modify anything. So there is no need to "Also follow the instructions under 'How to attach two cameras' below for info on how to set pin configurations for the second camera."

XECDesign:
Do you think the "to change pin configuration for GPIO pins 0-3 so they can be used with the camera" is necessary? It's a little misleading, because it does more than that. I'll merge it if you think it's clearer than the way it is now, but I worry that it might be a little confusing since it's not 100% accurate.

torkeldanielsson:
Nah, maybe that's not the most important addition ever written... But the note at the end would have saved me some time googling and forum-reading when I was banging my head against the wall wondering why the second camera wouldn't work. Until that's fixed my recommendation would be to merge this change.

Here's the forum post with background:
https://www.raspberrypi.org/forums/viewtopic.php?f=98&t=89876

XECDesign:
Alright, remove "to change pin configuration for GPIO pins 0-3 so they can be used with the camera.", squash into a single commit and I'll pull it in. Thank you.

torkeldanielsson:
Squashing commits was a first for me, hope I did it without messing something up :)

XECDesign:
Yup, looks good. Thank you.

PR #213, opened 2015-04-22
Update building.md

GeVanCo:
Make distinction between Pi 1 and Pi 2 default configuration in case of cross compilation (currently, only Pi 1 configuration was mentioned for cross compilation).

### DIFF
--- a/linux/kernel/building.md
+++ b/linux/kernel/building.md
@@ -76,12 +76,21 @@ To build the sources for cross-compilation there may be extra dependencies beyon
 
 Enter the following commands to build the sources:
 
+For Raspberry Pi 1:
+
 ```
 $ cd linux
 $ make ARCH=arm CROSS_COMPILE=arm-linux-gnueabihf- bcmrpi_defconfig
 $ make ARCH=arm CROSS_COMPILE=arm-linux-gnueabihf-
 ```
 
+For Raspberry Pi 2:
+
+```
+$ cd linux
+$ make ARCH=arm CROSS_COMPILE=arm-linux-gnueabihf- bcm2709_defconfig
+$ make ARCH=arm CROSS_COMPILE=arm-linux-gnueabihf-
+```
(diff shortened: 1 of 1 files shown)


XECDesign:
Looks good to me. Something about mkknlimg would be nice too, but it's not entirely crucial.

GeVanCo:
It's the first time I'm committing something on a GitHub repo.  I don't think I have the status of a *project collaborator* (at least, I'm not aware of it...), so I don't think I can do the final step myself.  Who's taking this up?

JamesH65:
You commit to you own repo, then send a Pull Request to the origin (the Pi
repo). Then it is peer reviewed and merged if all is OK.

XECDesign:
If you would like, this can be merged as is. There was talk of moving the mkknlimg script to the kernel repo, So the instructions for that step might change. Up to you how you want to proceed. Anyway, I'll leave this to the kernel people. Just thought the mkknlimg thing was worth mentioning.

GeVanCo:
@JamesH65: in my forked repo, there's already a button "View pull request".  When I press that button, I end up here.  Doesn't that mean the pull request is already sent to the origin?  See also the top of this page here (next to the button " Open").

lurch:
@GeVanCo You should see a branch (`patch-1`) in your repo which corresponds to this pull request. If you push additional commits to that branch, this pull-request will get automatically updated.

GeVanCo:
Yes, that is clear to me.   I just want to understand the mechanism: I still see the green "Open" button at the top of this page here, saying ...wants to merge 1 commit... and I was wondering how that commit would eventually arrive at the master.  Not that this change is so important (not at all, in fact), but I just want to put my little contribution to the raspberry community.  And I have to start and learn somewhere, isn't it...

pelwell:
What you are seeing is that the discussion for a pull request happens in the repo to which the PR is submitted. When the maintainer are happy with the PR then they press a "Merge pull request" button (that you can't see because you aren't a maintainer), possibly edit the commit message and then a final "Confirm merge" button.

You can probably (I've not tried this) get a feel for this by creating another branch in your own repo and then submitting a pull request from one to the other.

Issue #211, opened 2015-04-16
Camera Documentation typos


electronicsguy:
A. "Allows the insertion of specific EXIF tags into the JPEG image. You can have up to 32 EXIF tge entries."

Don't you mean 32 EXIF "tags" entries?

B. --exif GPS.GPSLongitude=5/1,10/1,15/100
would set the longitude to 5degs, 10 minutes, 15 seconds.

This seems inconsistent. Either it should be 15/1, or 0.0015 seconds.

PR #214, opened 2015-04-27
fix typos in GPS tags

electronicsguy:


XECDesign:
@JamesH65 Is that right?

lurch:
I have no idea if this is correct or not, but there's a bit more detail in #211

JamesH65:
It looks OK to me. Straining my memory about the /100 but nothing popping out...

XECDesign:
Thank you

electronicsguy:
:+1:

PR #215, opened 2015-05-03
changed reccomended editor for rc.local to nano instead of vim

blessedbiped:
nano is generally the recommended editor in the rpi docs, just making this page match.

### DIFF
--- a/linux/usage/rc-local.md
+++ b/linux/usage/rc-local.md
@@ -9,7 +9,7 @@ An alternative for scheduled task management is cron / crontab.
 On your Pi, edit the file `/etc/rc.local` using the editor of your choice. You must edit with root, for example:
 
 ```
-sudo vim /etc/rc.local
+sudo nano /etc/rc.local
 ```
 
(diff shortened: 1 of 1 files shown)


bennuttall:
:(

blessedbiped:
I know, sorry. I love it now, I just remember the first time I accidentally opened vim- it wasn't a pleasant experience as a new user.

bennuttall:
I'm a Vim user, I don't get Nano. But Nano is pre-installed and Vim is not, so this PR is fine.

While Vim is difficult to use for the first time, I don't think Nano is all that much better. If you're booted to desktop you're probably better off opening Leafpad, but you can't assume that as we don't boot to desktop by default.

bennuttall:
Note: Vi is preinstalled, but Vim is not.

Vim allows use of arrow keys in input mode, whereas Vi does not (and annoyingly enters escape characters when you try) which is a real pain, even for seasoned Vim users.

PR #218, opened 2015-05-09
Include Pi 2 in the cross-compiling instructions

argilo:


XECDesign:
This seems to already be on the page.(line 35)

argilo:
That's in the "local building" section. I thought it would be good to repeat it in the cross-compiling section as well. When following the instructions I went straight to the cross-compiling section, and ended up building myself a kernel that wouldn't boot.

XECDesign:
Ah then yeah, that seems sensible.

pelwell:
Any addition for Pi 2 should also say that the correct name for a Pi 2 kernel is kernel7.img. And the kernel copy stage should be replaced by the use of the mkknlimg script, which has now been added to the kernel source tree as scripts/mkknlimg. Without this step, you can't use Device Tree on a Pi.

clayton-shopify:
@bennuttall There was no reason to merge this as the information was already added in https://github.com/raspberrypi/documentation/commit/5be41aee520a8995930f678c30aab3231ca1deb6. You should probably revert it.

argilo:
@bennuttall There was no reason to merge this as the information was already added in https://github.com/raspberrypi/documentation/commit/5be41aee520a8995930f678c30aab3231ca1deb6. You should probably revert it.

PR #220, opened 2015-05-12
Specify the toolchain without modify PATH

NoTengoBattery:
In some cases, we need to keep the Raspberry toolchain away from PATH, because we need both of them.
So, specify the toolchain in the `make` command. This is for noobs, the people that had made this (I guess) already know it.


NoTengoBattery:
@pelwell You are absolutely rigth, some typos, and I'm not a native speaker. Sorry about that. However, I really think that this is useful for some people (because it happend to me, you know, a headache, researching in Google).

PR #222, opened 2015-05-19
Update camera.md

meepsalot:
It's surprisingly easy to mistake the display port for the camera port.  After diligently following all the above instructions for debugging my undetected camera, flashing a new copy of Raspbian, and even trying a second camera, I finally found the source of my issue.  The display port even lights up the camera LED, making debugging more confounding!  Adding this seemingly simple CSI vs DSI port verification check to the debug instructions may help others get their cameras working sooner.

### DIFF
--- a/troubleshooting/hardware/camera.md
+++ b/troubleshooting/hardware/camera.md
@@ -2,6 +2,8 @@
 
 If the camera is not working correctly, there are number of things to try:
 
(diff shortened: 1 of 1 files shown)


XECDesign:
Thanks

pelwell:
I think you should change Ethernet to HDMI, since some models have no Ethernet and the HDMI is closer on others.

XECDesign:
Done

PR #226, opened 2015-05-30
replaces bs=1m with bs=1M

bglnelissen:
On OS X Yosemite 10.10.3 with `dd (coreutils) 8.23` installed, the lower case `m` gave an error: 

    dd: invalid number ‘1m’

Uppercase `M` works just file.


lurch:
Isn't this already mentioned on that page?
"This may result in an dd: invalid number '1m' error if you have GNU coreutils installed. In that case you need to use 1M:"

XECDesign:
Closing for inactivity

PR #228, opened 2015-06-02
Update building.md

skullandbones:
Use zImage rather than Image and build the modules after building the kernel, the INSTALL_MOD_PATH is not relevant for building the modules so fixed it as well.

At a later date add a description of installing onto a running system by using sftp or scp etc.


pelwell:
Thanks. I have a larger update awaiting the merge of a commit to 3.18 that does this and more, but I see no harm in merging this now.

Issue #230, opened 2015-06-19
Add helpful SIGALERT (CNTL+T) to Mostly Graphical Install section of Mac doc


willingc:
The Command line section of the Mac installation document has this helpful message:

"This will take a few minutes, depending on the image file size. You can check the progress by sending a SIGINFO signal pressing Ctrl+T."

It would be helpful to include this useful tip in the Graphical Install section too. Since writing the image to the SD card takes a while, this tip helps the user know whether writing is progressing.

pranav16255:
https://github.com/raspberrypi/documentation

JamesH65:
Fixed by #445

PR #231, opened 2015-06-19
Update mac.md to include status tip #230

willingc:
Fixes Issue #230. Adds the helpful tip for checking write status to the Graphical Interface section. It's a helpful tip and users may overlook the tip in the command line section.

### DIFF
--- a/installation/installing-images/mac.md
+++ b/installation/installing-images/mac.md
@@ -15,6 +15,9 @@ On Mac OS you have the choice of the command line `dd` tool or using the graphic
     ```
 
     Remember to replace `n` with the number that you noted before!
+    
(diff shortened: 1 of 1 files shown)


XECDesign:
I see the same information repeated later down the page (line 92).

XECDesign:
Since I don't like leaving PRs and issues hanging for too long, if nobody merges this, I'll close it in a few days. My reasons for not merging it would be that it's duplicating data on the page which results in extra maintenance work in the future. It's easy to accidentally make changes to one copy instead of both, if such changes need to be made. It also requires the reader to pick out information they want (how to install the OS) from the optional information.

PR #238, opened 2015-07-08
Update cmio-camera.md GPIO pin warning

samskiter:
Is there anywhere I can discuss this or provide debug information?

After power up, the GPIO pins are set differently from the device tree - the kernel must be overriding the setting.

2&3 are set to be SDA1. I tried to setup GPIO 44&45 to SDA1 but SDA1 just appears on 44/45 AND 2/3.

Separately, I cannot get the single cam to work no matter what I do.

### DIFF
--- a/hardware/computemodule/cmio-camera.md
+++ b/hardware/computemodule/cmio-camera.md
@@ -117,4 +117,4 @@ pin_define@CAMERA_1_SCL_PIN { type = "internal"; number = <29>; };
 
 [Enable both cameras](dt-blob-dualcam.dts)
 
-*As of 2015-04-17 some users are reporting issues with GPIO pin 2 and 3 for camera control. Pin 4 and 5 is reported to work. The issue is being investigated.*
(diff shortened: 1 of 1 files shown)


XECDesign:
That sounds like the setup itself is wrong. If you're using GPIO2 and GPIO3 for power enable and the LED, you can't also use them for SDA1/SCL1. Do you have I2C1/I2C_ARM enabled?

If not, could you please install `device-tree-compiler`, post the content of config.txt and the output of `dtc -I fs /proc/device-tree`

samskiter:
Gpio pin 2&3 not working was already mentioned here as a problem. I think it's because the device tree is being overridden somewhere for these two pins. After boot, I2C1 will always appear on those two pins no matter how I change the device tree

XECDesign:
Could you please provide the output mentioned earlier to help debug the issue?

XECDesign:
Closing due to inactivity and because it appears to be more of a general support query. Please post the requested output on the forum, in the compute module section, and we'll do our best to help.

samskiter:
Apologies, I've been pulled off this project for a few weeks. When I get back to it, I'll pull of that debug information.

PR #241, opened 2015-07-22
add instructions for gzipping image data.

MicahStevens:
gzip can be used to reduce the file size of system images on the fly. Added some instructions to do this. Hope it helps.


bennuttall:
Cool thanks

lurch:
I'm not sure about the 'restore' line - wouldn't it result in `gunzip` being run as root, and `dd` being run as non-root, which is the opposite way around to what you actually want? I think moving the `sudo` to after the pipe should fix it? (i.e. `gunzip --stdout rasbian.img.gz | sudo dd bs=4m of=/dev/sdb`)

(apologies for not spotting this pre-merge!)

PR #242, opened 2015-07-23
Typo in README

rmcaninch:
amperage


XECDesign:
Thank you for pointing out the typo. IMHO, 'amperage' is no better than 'kilogramage' or 'kilometers-per-hourage'. It has always rubbed me the wrong way. I'd go with 'rated current' or just 'current' since, the rated part is implied. Maybe it's a regional thing, but while studying electrical engineering, I've never heard 'amperage' used outside the context of high power transmission systems. To quote The Art of Electronics, `don't call current "amperage"; that's strictly bush league. Unless you're a power engineer working with giant 13kV transformers and the like - those guys are allowed to say amperage`. Maybe someone will correct me on this.

PR #247, opened 2015-08-21
Update ip-address.md

ErnstNeger:
Comment on line 11: hostname -I is not implemented f.e. on Arch Linux.
Comment on line 31: Same thing


XECDesign:
The documentation is written with Raspbian in mind, as that's the officially supported distro. Under raspbian, using the -i option gives 127.0.1.1 on wheezy and jessie, which isn't very useful

XECDesign:
Sorry, ignore the last comment. Under Mint, running hostname -i gives me 127.0.1.1, which isn't very useful. I am also not sure about the change on line 11. You've left it as -I, but added a note that it will only work from the Pi's command line. Could you please clarify a bit?

PR #252, opened 2015-09-11
Extra description for formatting SD card on Windows

conan:
When formatting an SD card using the sdcard.org tool on Windows, the documentation already says that you need to select a specific option in order to use all the space on the card.  It's a little confusing however, because the tool doesn't show the full size until after the format is complete, so it's not clear that you've done it correctly until then.  This is just some reassuring text to say that this will happen.

### DIFF
--- a/installation/noobs.md
+++ b/installation/noobs.md
@@ -31,7 +31,7 @@ To set up a blank SD card with NOOBS:
 
 ##### Windows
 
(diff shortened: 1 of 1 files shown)


Ruffio:
@lurch Can this PR be merged or should it be adjusted/rejected?

PR #253, opened 2015-09-16
Update config-txt.md

fredericofs:
Added overscan scale and cec name settings info

### DIFF
--- a/configuration/config-txt.md
+++ b/configuration/config-txt.md
@@ -31,6 +31,8 @@ overscan_right=12
 overscan_top=10
 overscan_bottom=10
 ```
+# Force overscan settings to non-framebuffer layers.
+overscan_scale=1
 
 ## Memory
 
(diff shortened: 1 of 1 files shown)


bennuttall:
@XECDesign is this correct?

lurch:
AFAICT there are lots of additional `config.txt` settings that get added (dunno if they ever get removed too?) with various different [firmware](https://github.com/raspberrypi/firmware) releases of `start.elf`, and unfortunately the `configuration/config-txt.md` documentation rarely gets updated to match.
I believe some of the extra settings are 'debug' stuff that doesn't need documenting (e.g. some of the stuff @popcornmix does for fixes with video playback), but there are probably other options that should be documented (with the `ignore_lcd` from https://www.raspberrypi.org/documentation/hardware/display/README.md being just one example picked at random).

And on a side-note I guess the "OVERCLOCKING OPTIONS" in the `config.txt` documentation need updating to cover the different default frequencies used on the Pi2 and PiZero?

XECDesign:
The information seems accurate, but the formatting it's quite right.The first option should probably be a level down. I am also not sure if it's so important that it should be so close to the top of the page.

Ruffio:
@lurch @XECDesign what does it take for this PR to be merged?

JamesH65:
Technically happy for this to be merged, will forward to comms for copy editing/format fixing.

PR #264, opened 2015-10-09
script didn't work for me on 2015-09-24-raspbian-jessie

chahld:
needed to move the line #!/bin sh to the first line of the file. Can't be past the comments. Otherwise, get an exec error on startup and it doesn't start the server.


XECDesign:
Thank you

PR #268, opened 2015-10-13
Add raspi-config to camera setup

nielsswinkels:
Would be nice to remind people to enable the camera in the raspi config settings.


bennuttall:
Thanks

PR #283, opened 2015-11-27
Correct exit nano command

nielsswinkels:
Changed ctrl+c to ctrl+x for closing a file in nano.

Added the suggestion that you should check your sd card if your screen stays black. Happened to me that I forgot to put in a card.


bennuttall:
Thanks

Issue #23, opened 2014-04-02
SSH passwordless authentication for Remote access: Consider using ssh-copy-id instead of cat | ssh


tommybobbins:
When appending to the authorized_keys file, I find it easier to use the built in command:

```
 $ ssh-copy-id -i ~/.ssh/id_rsa.pub pi@192.168.1.137
```

Which gives the following output 

```
pi@192.168.1.137's password: 
Now try logging into the machine, with "ssh 'pi@192.168.1.137'", and check in:

  ~/.ssh/authorized_keys

to make sure we haven't added extra keys that you weren't expecting.
```

The advantage of this command is it has no backticks or pipes to worry about.

PR #287, opened 2015-12-05
The built-in 'ssh-copy-id' command is a bit simpler.

drawpause:
ssh-copy-id installs your public key in a remote machine's authorized_keys automatically. No need to mess with permissions etc.


lurch:
Duplicate of #23 ?

drawpause:
Ooops. Sorry about that.

Issue #182, opened 2015-03-02
Add Pi2 details and specs


lurch:
Just like #132, but for the Raspberry Pi 2 Model B ;)

Also, where things used to say "Model A", "Model B+", etc. should these now be updated to "Pi1 Model A", "Pi1 Model B+", or is there some other official nomenclature that should be used?  Something to add to https://github.com/raspberrypilearning/creating-resources/blob/master/style.md ?

lurch:
Also needs adding to http://www.raspberrypi.org/documentation/hardware/raspberrypi/models/

ebenupton:
Ben to clean up

lurch:
After seeing #199 I went looking for the "model specs" page on http://www.raspberrypi.org/documentation/ and was confused when I couldn't find it. I eventually discovered that I couldn't find it because of https://github.com/raspberrypi/documentation/commit/f68bfc961423aa200bb7fea610f0a255bf32f73a ;-)

I can see how `hardware/raspberrypi/models/README.md` (originally added by @JamesH65 ) is a bit of a duplication of http://www.raspberrypi.org/products/ , but IMHO the `hardware/raspberrypi/models/specs.md` page **is** still useful, which now has no links to it. When Ben gets round to tackling this issue (I know he's still busy with resources stuff) maybe he could add a link to `hardware/raspberrypi/models/specs.md` from somewhere else?

ebenupton:
I think the conclusion was to retire the specs page completely, and to move any useful information onto the main website product pages.

________________________________

flashspys:
Is the Peripheral specification for the BCM2836 (Pi2's cpu) already somewhere in the internet? We need it because some adresses were changed.

P33M:
See https://github.com/raspberrypi/documentation/commit/b0639f5a3ff6ba9b5006c62d643403d97408b1f7 for details on how the Cortex-A7 cluster on BCM2836 has been implemented.

Other than that, Pi2 is basically identical to a B+ in terms of mechanical layout.

bennuttall:
Are we dropping specs or updating them? They ought to be on the product pages really...

ebenupton:
We should retire this page.

PR #288, opened 2015-12-09
Add Pi Zero to specs page

JamesNewton:


lurch:
AFAIK there aren't actually any links to the specs page that you've edited?
See https://github.com/raspberrypi/documentation/issues/182#issuecomment-83872783 for background info.

JamesNewton:
Goggle found it. And it's the best comparison table I've seen after about 10 min of searching except for lacking the Zero and pi2 . Sorry to see it go unused.... where else can you find guidance on which to buy / use?

lurch:
I guess 2a46bba69e3e0171f32906867a5566eff742ad91 means that this PR will need to be closed?

JamesH65:
Sorry to leave a comment on a closed page, but this has just come up on the forum - the products page only has current models, there is no information on older models that people may come across. Since these may come up for sale at various points, people may wish to know what they have got. This is where a table might be useful, as it could flag whether it's a current model or not, as well as giving a quick comparison between models.

JamesNewton:
Honestly, this also bugs me... As JamesH65 notes (good name) there is no other place that lists all the models side by side. I would really suggest retaining this information somewhere.

PR #292, opened 2015-12-24
Instructions how to change mouse cursor from "x" to arrow

DSA101:
Added instructions how to change mouse cursor from "x" (which is default with tightvnc) to an arrow.

### DIFF
--- a/remote-access/vnc/README.md
+++ b/remote-access/vnc/README.md
@@ -57,6 +57,12 @@ chmod +x vnc.sh
 ./vnc.sh
 ```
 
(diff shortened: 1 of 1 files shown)


t3chguy:
:+1:

raamana:
That didn't work for me. Any other clues?

remote-access/vnc/README.md
```diff
@@ -57,6 +57,12 @@ chmod +x vnc.sh
 ./vnc.sh
 ```
 
+- If your prefer your mouse pointer in the VNC client to appear as an arrow as opposed to an "x" which is default, in ```./vnc/xstartup``` add the following parameter to ```xsetroot```
```
an732001: where?

PR #295, opened 2016-01-06
Update cron.md

nielsswinkels:
Added how to start processes at reboot, and how to do it in the background.


lurch:
Does this have any pros/cons when compared to https://github.com/raspberrypi/documentation/blob/master/linux/usage/rc-local.md ?

Also, 'everytime' should be two words i.e. 'every time'.

bennuttall:
I think it's worthwhile having this as option as well as rc.local - we ought to have init.d too...

PR #297, opened 2016-01-14
Update nginx.md

DSdavidDS:
Clarified where the root directory is. Since I installed nginx through the Raspbian repo manually, my version of Raspbian had the root directory located in /var/www/html/.


LornaLynch:
Thanks!

PR #300, opened 2016-01-22
Fix instructions for Linux installs: remove truncate bit, add expand filesystem notes

azizLIGHT:
Remove old instructions, and mention the ability to expand the filesystem

### DIFF
--- a/installation/installing-images/linux.md
+++ b/installation/installing-images/linux.md
@@ -30,17 +30,11 @@ Please note that the use of the `dd` tool can overwrite any partition of your ma
 
 - You can check what's written to the SD card by `dd`-ing from the card back to another image on your hard disk, truncating the new image to the same size as the original, and then running `diff` (or `md5sum`) on those two images.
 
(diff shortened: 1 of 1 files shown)


waveform80:
Looks good to me; the existing truncate instructions are incorrect and `raspi-config` is by far the most "user friendly" way of expanding the FS up to the size of the chosen SD card.

XECDesign:
The bit about truncating the image is about verifying that what was written was correct. If you remove that, then the previous paragraph is useless.

waveform80:
Ah, missed that the former paragraph was about copying back - thought this was to with file-system expansion (serves me right for skimming stuff late at night). Is verification worth having in the instructions? I note there's nothing similar in the Mac/Win sections.

For that matter, would it be worth having `raspi-config` mentioned at the end (or a link to the [raspi-config](https://www.raspberrypi.org/documentation/configuration/raspi-config.md) page?)

JamesH65:
Unless I hear otherwise, I'll close this in a week, I don't think it's relevant. If there is an issue with the current truncate instructions (should they even be there?) I'd suggest raising an issue.

JamesH65:
No responses so closing.

PR #301, opened 2016-01-27
Update README.md

teschlg:
Some fixes for the latest version

### DIFF
--- a/remote-access/vnc/README.md
+++ b/remote-access/vnc/README.md
@@ -24,7 +24,7 @@ tightvncserver
 vncserver :1 -geometry 1920x1080 -depth 24
 ```
 
-- Note: Since by default an X session is started on display zero you will get an error in case you use ```:0```.
+- Since by default an X session is started on display zero you will get an error in case you use ```:0```.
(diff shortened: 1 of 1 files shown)


XECDesign:
Why do you need to touch the lightdm service for this?

teschlg:
If lightdm is running on :0 you can't start vnc on :0! Of course you could have lightdm on :0 and vnc on :1 but given the limited resources of the pi that does not seem a good idea?

XECDesign:
Ah yes, makes sense. I think I would prefer to run vnc on :1 anyway and mention that lightdm can be disabled to save resources. However, this is definitely an improvement on the instructions so looks good to me as is.

Minor typo: "Since a desktop **ist** typically..."

teschlg:
OK, I changed this to :1. See if you like it.

XECDesign:
Looks good, thank you.

bennuttall:
I'm not fussed about squashing commits but I'd prefer better commit messages than `Update README.md` :)

XECDesign:
Heh yeah, good point. Noted.

Issue #217, opened 2015-05-09
I think the page on passwordless SSH access has a mistake.


I-mess-up:
[Link to page](https://github.com/raspberrypi/documentation/blob/master/remote-access/ssh/passwordless.md)

I'm a beginner, so it's possible that I'm just confused, but I think there is a mistake on this page. In one line, it suggests using the code

    cat ~/.ssh/id_rsa.pub | ssh <USERNAME>@<IP-ADDRESS> 'cat >> .ssh/authorized_keys'

to copy the public key over to the pi. This didn't work for me, but once I added the path to the home directory (not sure if this is the right way to say this) it worked. My modified entry was 

    cat ~/.ssh/id_rsa.pub | ssh <USERNAME>@<IP-ADDRESS> 'cat >> ~/.ssh/authorized_keys'

Notice the extra "~/". 

Am I right that this is an issue or am I just a misguided user?

XECDesign:
When you pass a command like that, it runs in the home directory, so the two commands should behave exactly the same.
```
echo test | ssh xec-pi 'cat >> blah'
echo test | ssh xec-pi 'cat >> ~/blah2'
echo test | ssh xec-pi 'ls blah*'
  blah
  blah2
```

I-mess-up:
Alright. For me it didn't work but the change made it work. I can now connect to my Pi without a password. Is it ok if I leave the issue open for a bit just in case others want to chime in?




Another thing: is it possible I got some sort of error because I didn't have a directory named .ssh?

XECDesign:
Sure

Yeah, it's important to pay attention to error messages and paste them when reporting issues. However, adding a ~/ in that scenario would not have helped:
```
echo test | ssh serge-pi 'cat >> not-a-dir/blah'
  bash: not-a-dir/blah: No such file or directory
echo test | ssh serge-pi 'cat >> ~/not-a-dir/blah2'
  bash: /home/pi/not-a-dir/blah2: No such file or directory
```

I-mess-up:
Alright, I guess that I solved my issue by creating the directory then.  Thanks for explaining. I'll close the issue now.

PR #307, opened 2016-02-04
Missing instructions about creating .ssh on Pi

jophde:
The copy command chain will fail if .ssh hasn't been created on the Pi.

### DIFF
--- a/remote-access/ssh/passwordless.md
+++ b/remote-access/ssh/passwordless.md
@@ -58,7 +58,7 @@ If your Pi does not have an .ssh directory you will need to set one up so that y
 
 ```
 cd ~
-mkdir .ssh
+install -d -m 700 ~/.ssh
 touch .ssh/authorized_keys
 ```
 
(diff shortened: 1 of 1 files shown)


XECDesign:
I would recommend adding `install -d -m 700 ~/.ssh`

lurch:
semi-duplicate #217 #285 and #286 ?

jophde:
Wouldn't you want it created with the permissions of whatever user you are logged in as? @XECDesign

XECDesign:
When I tried it it made it 777 permissions (read and write for owner, group and other) 700 means read and write for owner only. If it's left as 777 other logged in users can take your keys, which isn't recommended.

lurch:
The `touch` isn't necessary either - if the file doesn't exist then trying to append something to it with `>>` will automatically create it.

XECDesign:
Thanks

I'd recommend one more change - `chmod 600` the authorized keys file, but that's a separate issue.

PR #309, opened 2016-02-07
Fix permissions for authorized_keys

jophde:


XECDesign:
The idea isn't that it's non-executable, but to make sure that only you can read and write it.

The way it's currently written, it looks like the user should run the chmod command on the current machine, rather than the remote one. Perhaps something like this would work better?
`ssh <USERNAME>@<IP-ADDRESS> 'chmod 600 .ssh/authorized_keys'`

I think if the permissions are wrong, ssh might refuse to work, so I am not sure about this.
http://www.openssh.com/faq.html#3.14

jophde:
Isn't the directory already created with with ready, write, and execute permission for only the current user because of the `-m 700` in the `install` command?  The authorized_keys file would inherit the permissions of the .ssh directory right?

I think adding ssh <USERNAME>@<IP-ADDRESS> would be better too..

XECDesign:
Tested it a little and yeah, it seems like as long as others don't have the 'x' permission on the directory, they can't read the file, even if the file itself is readable. However, every reliable source (openssh website, ssh manpage and ubuntu documentation) recommends that it's 600 anyway. In practice it doesn't really seem to matter.

jophde:
So should we change the install command to use 600 or add an explicit `chmod 600 .ssh/authorized_keys` to the end?

XECDesign:
.ssh still needs to be 700, not 600, so the second option is the way to go.

lurch:
If you changed the `install` command to 700, then the directory wouldn't have execute permissions, and you'd be unable to list its contents!

lurch:
...and it might be better to remove the earlier `cd ~` command, and instead always use `~/.ssh/...` everywhere?

XECDesign:
It still looks like the instructions say that you should run `chmod 600 ~/.ssh/authorized_keys` on the local machine rather than the remote machine.

lurch:
It's a little bit OffTopic for this specific PR, but given that the first section talks about `/home/pi/...` I guess it's also not clear that ssh-keygen should be run on the local machine (your computer) and not the remote machine (your Raspberry Pi) ?

XECDesign:
Yeah, I think it's written as if you're trying to ssh from one pi to another, which leaves a lot of room for ambiguity.

PR #310, opened 2016-02-12
Update ftp.md

Jagps0:
I removed the semicolon from the groupadd ftpgroup command. I also fixed the hyphen before the d in the useradd ftpuser -g ftpgroup -s /sbin/nologin -d /dev/null command. It would not work.


XECDesign:
Thanks

Jagps0:
Can I make another change?

PR #319, opened 2016-02-29
Link to the recommended SD card not working

SamJongenelen:


LornaLynch:
I'll update the link in the document: thanks.

PR #330, opened 2016-03-28
Update README.md

Erikun:
Changed Screaming Jelly Baby link to Burping Jelly Baby as the previous does not seem to exist on raspberrypi.com/learning anymore.


lurch:
Good spot! https://github.com/raspberrypilearning/burping-jelly-baby/issues/13

@bennuttall Any idea why this wasn't picked up by the Travis-documentation-check?

bennuttall:
Ah, it's supposed to redirect URLs from resources which were renamed, but that wasn't in my lookup table (screaming-jellybaby was).

It probably didn't show in Travis because the page it lands on isn't a real 404. I need to fix that!

PR #191, opened 2015-03-05
Add backup method with compression

pickfire:
Hi, I add some [backup method with compression](https://github.com/raspberrypi/documentation/issues/140#issuecomment-68245203). I hope you can merge it. Thanks.

PR #336, opened 2016-04-04
Update backup.md

mshaub:
The restore from zipped backup didn't work. I had an error due to an uncapitalized 'm' and missing 'sudo' after the 'pipe' but it got me close. Thanks!

### DIFF
--- a/linux/filesystem/backup.md
+++ b/linux/filesystem/backup.md
@@ -67,7 +67,7 @@ sudo dd bs=4M if=/dev/sdb | gzip > rasbian.img.gz
 To restore, pipe the output of `gunzip` to `dd`:
 
 ```bash
-sudo gunzip --stdout rasbian.img.gz | dd bs=4m of=/dev/sdb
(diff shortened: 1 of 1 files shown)


lurch:
Yeah, this matches the syntax used in https://github.com/raspberrypi/documentation/pull/191/commits/ee802351ae3c6487917a223e308e08105786ee96 (which didn't get merged, because https://github.com/raspberrypi/documentation/commit/4a970592099ad1d230d8de5366f3ae1380a3c87a did instead)

I believe the 'sudo' is _only_ needed after the pipe, i.e. gunzip doesn't require sudo - would you mind testing (and updating this PR as appropriate) ?

I suspect the upper-case / lower-case 'm' used with dd might be a Mac-issue, like as described on https://github.com/raspberrypi/documentation/blob/master/installation/installing-images/mac.md ? But as the 'backup' command uses 4M, then the 'restore' command should use 4M too.

lurch:
fixed in #374

PR #339, opened 2016-04-06
multi-job make

mynameismevin:
Changed the RPi3 make command to include the `-j4` command by default, updated the note section to explain the flag instead of mark it as optional. I wouldn't want anyone to suffer through compiling the kernel on a RPi without using multiple jobs, especially newbies.


bennuttall:
@XECDesign?

XECDesign:
Yeah, I see no problem with this. It's probably sane to assume that most people are not compiling their kernels on a pi 1.

mynameismevin:
Awesome!

PR #341, opened 2016-04-08
Added graphical method to check dd progress...

gbowman85:
... using the bytes read column in Activity Monitor.


JamesH65:
AIUI, this doesn't work unless the activity monitor is told to monitor at su level. Can you update the PR to reflect that? @gbowman85

JamesH65:
Replaced by #538

gbowman85:
Thanks for modifying the instructions and closing the issue James.

PR #350, opened 2016-04-16
dd now has status use "status=progress"

paulpoco:
"dd bs=4M if=2016-03-18-raspbian-jessie.img of=/dev/sdd" has no status
"dd bs=4M if=2016-03-18-raspbian-jessie.img of=/dev/sdd status=progress" has a status of the progress


lurch:
This option is only supported by _some_ versions of `dd`. On Ubuntu 14.04 it reports:
```
dd: invalid status flag: ‘progress’
Try 'dd --help' for more information.
```

(perhaps this PR should be closed as invalid?)

einstein95:
As the latest Ubuntu LTS is 15.04 which has this functionality, shouldn't this be merged?

lurch:
See also https://github.com/raspberrypi/documentation/pull/381#issuecomment-239769164

JamesH65:
Technically this is fine - but i wonder if there needs to be a disclaimer after the example saying to omit the status option on older version of Linux if the command doesn't work?

If the originator/others could comment and/or update to reflect this, that would be great.

JamesH65:
replaced by PR #539

PR #357, opened 2016-05-16
Updated kernel headers information.

jai-bits:
Signed-off-by: jai-bits jaishiv5@gmail.com

### DIFF
--- a/linux/kernel/headers.md
+++ b/linux/kernel/headers.md
@@ -1,1 +1,5 @@
 # Kernel Headers
+You can use the following command for downloading the headers only.
+```sh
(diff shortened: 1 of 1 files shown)


XECDesign:
Those aren't the correct headers.

raspberrypi-kernel-headers would contain the headers for the kernel in raspberrypi-kernel.

jai-bits:
But we can port any kernel on Raspberry board.
And the above command is correct in that context.


*With Best Regards,Jai Shiv*
*81061-22410*

JamesH65:
This page has been updated so closing.

PR #364, opened 2016-05-20
Fix new typos found by codespell

stweil:
Signed-off-by: Stefan Weil <sw@weilnetz.de>

### DIFF
--- a/configuration/device-tree.md
+++ b/configuration/device-tree.md
@@ -631,7 +631,7 @@ The loading of overlays at runtime is a recent addition to the kernel, and so fa
 
 * Applying or removing some overlays may cause unexpected behaviour, so it should be done with caution. This is one of the reasons it requires `sudo`.
 
-* Unloading the overlay for an ALSA card can stall if something is actively using ALSA - the LXPanel volume slider plugin demonstrates this effect. To enable overlays for sound cards to be removed, the `lxpanelctl` utility has been given two new options -- `alsastop` and `alsastart` -- and these are called from the auxilliary scripts dtoverlay-pre and dtoverlay-post before and after overlays are loaded or unloaded, respectively.
+* Unloading the overlay for an ALSA card can stall if something is actively using ALSA - the LXPanel volume slider plugin demonstrates this effect. To enable overlays for sound cards to be removed, the `lxpanelctl` utility has been given two new options -- `alsastop` and `alsastart` -- and these are called from the auxiliary scripts dtoverlay-pre and dtoverlay-post before and after overlays are loaded or unloaded, respectively.
 
 * Removing an overlay will not cause a loaded module to be unloaded, but it may cause the reference count of some modules to drop to zero. Running `rmmod -a` twice will cause unused modules to be unloaded.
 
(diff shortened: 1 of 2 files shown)


bennuttall:
Ooh that's very useful. Thanks. Perhaps we should be using this ourselves...

bennuttall:
@LizUpton @lauraclay @LornaLynch @helenlynn

stweil:
Yes, it is useful and easy to get: https://github.com/lucasdemarchi/codespell.

I suggest to use a tagged version – the latest Git version which added Travis support also confused the code a little bit.

LizUpton:
Sounds good to me - but make sure any automated spellchecking catches are signed off by @lauraclay or @LornaLynch please!

bennuttall:
I don't think automated corrections would work here (too many technical terms), but it would be good for spotting typos (suggesting corrections, not making them).

lurch:
I also recall seeing codespell mentioned in typo-fixing PRs on the userland and/or firmware repos.

lauraclay:
Have now got round to looking at this, and several thoughts spring to mind:

- The dictionary they're using from Wikipedia uses American spelling. Unless we're radically changing the documentation repo, we're still on UK spelling.

- Am I correct in thinking that, as in the example ' clas->class, disabled because of name clash in c++', it'll fall over if a resource uses words reserved by programming languages?

- Automated spellcheckers should be taken with a massive cellar of salt, especially user-maintained ones. For instance, the most recent commit to the dictionary has misspelled something in the OED. ('preexisting' should be hyphenated.)

- The documentation is unclear as to whether it _suggests_ changes or just straight-up corrects them. If the latter, it would definitely need a second pass from @LornaLynch and me to check it wasn't making mistakes.

I'm generally wary of 'X is an automated solution to doing it manually,' for which you can blame years of QA. So, colour me sceptical of codespell.

LizUpton:
In this instance I believe it suggested changes. I absolutely agree: automated spellcheck which does not allow for a human stage in the middle is a TERRIBLE idea. 

Anecdote that proves the point: My sister-in-law wrote a journal paper called "Bonfire of the Quangos" that ended up being published as "Bonfire of the Guanos" because the editor subbed his job out to a spellchecker. True story. All instances of Aenurin Bevan in it were also replaced with with Anuran Bean.

stweil:
@lauraclay, as far as I know, codespell does not require that text must use American spelling. I'm not a native speaker, so please correct me if I am wrong. Regarding your questions about codespell: it is primarily a tool to detect possible typos. It can also fix typos, but only if there is only one possible fix and only if this fix does not introduce potential problems (like changing a C++ name from `clas` to `class`). So yes, `clas` will be reported as a typo, but no, it won't be fixed even if you tell codespell to do so. 

@LizUpton, yes, you need someone who decides which typos are typos and which of them should be fixed. You even need a second person who reviews the modifications. Typical scenarios where typos should not be fixed include code which was copied from other projects (those typos should be fixed in the original project) or typos in patch files which would break the patch.

Nevertheless automated spell checking (without automated fixes) can be used in continuous integration, and the automated fixes which can be done with codespell can help a human.

PR #366, opened 2016-05-24
Fix command line typo

williamriancho:


JamesH65:
Investigation shows that rdisk is the correct option. rdisk means raw disk and is unbuffered and allegedly much faster. Recommendation is to use rdisk for external devices i.e. SD cards. 

 Either do work, so I'm going to close this PR without including.

PR #368, opened 2016-05-30
correction on updating.md

airtime166:
running sudo apt-get dist-upgrade will not update the distribution version . Unless the sources.list file has been updated

### DIFF
--- a/raspbian/updating.md
+++ b/raspbian/updating.md
@@ -9,9 +9,17 @@ sudo apt-get update
 Next, **upgrade** all your installed packages to their latest versions with the command:
 
 ```bash
-sudo apt-get dist-upgrade
+sudo apt-get upgrade
 ```
 
+If you wish to upgrade the distribution version. Ex. From Wheezy to Jessie
+```bash
+sed -i 's/wheezy/jessie/g' /etc/apt/sources.list
+sudo apt-get update
+(if you receive a error similar to E: Release file expired. run apt-get -o Acquire::Check-Valid-Until=false update )
+sudo apt-get upgrade
+sudo apt-get dist-upgrade
+```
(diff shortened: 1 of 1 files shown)


XECDesign:
This page isn't about upgrading the distribution. Following the instructions as you've written them, starting from the last wheezy image, would not result in a correctly working system.

The reason dist-upgrade is used instead of upgrade is to pick up new dependencies. Otherwise, apt holds the updates back. Again, this results in a system that may not function properly.

airtime166:
The process I wrote are the steps i used on my fully functional servers and local desktop. I am testing this out on the PI now and will post the results. Can this be added to the page as when i search on google for "Raspbian system upgrade "  The first result is " https://www.raspberrypi.org/documentation/raspbian/updating.md "  The second result is " https://linuxconfig.org/raspbian-gnu-linux-upgrade-from-wheezy-to-raspbian-jessie-8 " which list the steps similar to my changes.  dist-upgrade is not the standard upgrade option but can be used occasionally  .

XECDesign:
The test has been done several times when we migrated to jessie.
The correct instructions are here https://www.raspberrypi.org/forums/viewtopic.php?f=66&t=121880 and linked from https://www.raspberrypi.org/blog/raspbian-jessie-is-here/.

On standard debian, there is no reason to run dist-upgrade on the same release because they rarely go changing dependencies in a stable release. In the foundation's release of Raspbian, there is a lot of ongoing development and new dependencies are added quite often, so dist-upgrade is the recommended way to keep up to date.

airtime166:
Understood, Thanks for the clarification  @XECDesign . Looks like the instructions are geared towards the GUI which i don't use on the PI often.  Maybe consider adding a link from this page to  "https://www.raspberrypi.org/forums/viewtopic.php?f=66&t=121880 " as a reference.

JamesH65:
Closing as not relevant.

PR #375, opened 2016-06-08
Add share clipboard support for vnc server config documentation

zhaocon:


JamesH65:
No comment from originator, so closing. Can be reopened if OP feels there is still some relevance.

PR #377, opened 2016-06-15
Add warning to not use rpi-update anymore to docs

dragetd:
rpi-update does not provide a way of rolling back changes. Generally it causes more support issues from people messing up their kernel config than it solves problems. It might be useful for different distributions or use-cases, but not for a regular noobs/raspbian user! Update docs accordingly.
See https://github.com/Hexxeh/rpi-update/pull/201


dragetd:
Oh, the very same minute there seems to be some activity and life back in the project! Okay, let's see how this turns out. This makes this PR obsolete! :-)

PR #381, opened 2016-06-26
Update wording for unmount

OjasA:
Change install instructions/tutorial to use "umount -R" instead of unmounting each partition individually.

### DIFF
--- a/installation/installing-images/linux.md
+++ b/installation/installing-images/linux.md
@@ -10,9 +10,7 @@ Please note that the use of the `dd` tool can overwrite any partition of your ma
 
 - Now that you've noted what the device name is, you need to unmount it so that files can't be read or written to the SD card while you are copying over the SD image.
 
-- Run `umount /dev/sdd1`, replacing `sdd1` with whatever your SD card's device name is (including the partition number).
-
-- If your SD card shows up more than once in the output of `df` due to having multiple partitions on the SD card, you should unmount all of these partitions.
+- Run `umount -R /dev/sdd`, replacing `sdd` with whatever your SD card's device name is.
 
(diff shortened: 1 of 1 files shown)


lurch:
The `-R` flag of `umount` isn't available in the version included with Ubuntu 14.04...
```
umount: invalid option -- 'R'
Usage: umount -h | -V
       umount -a [-d] [-f] [-r] [-n] [-v] [-t vfstypes] [-O opts]
       umount [-d] [-f] [-r] [-n] [-v] special | node...
```

See also #350 for a similar situation.

Ruffio:
The -R flag is available on Raspbian:

```
pi@raspberrypi:/dev $ umount
Usage:
 umount [-hV]
 umount -a [options]
 umount [options] <source> | <directory>

Options:
 -a, --all               unmount all filesystems
 -A, --all-targets       unmount all mountpoints for the given device in the
                           current namespace
 -c, --no-canonicalize   don't canonicalize paths
 -d, --detach-loop       if mounted loop device, also free this loop device
     --fake              dry run; skip the umount(2) syscall
 -f, --force             force unmount (in case of an unreachable NFS system)
 -i, --internal-only     don't call the umount.<type> helpers
 -n, --no-mtab           don't write to /etc/mtab
 -l, --lazy              detach the filesystem now, clean up things later
 -O, --test-opts <list>  limit the set of filesystems (use with -a)
 -R, --recursive         recursively unmount a target with all its children
```

So the question is which OS should the documentation be for? AFAIK Raspbian is the official OS for RPi, so the documentation should IMHO be written for that.

XECDesign:
These are the commands you need to image an sd card for use with a raspberry pi, not commands to run on the raspberry pi. The instructions are intended to be run on your main desktop. For some people that might be a pi running Raspbian, but for most linux users it will be ubuntu or mate.

Ruffio:
My fault, I didn't read it in context only the change :-(

waveform80:
Ubuntu 16.04 upgrades the available version of `umount` to include support for `-R`. Still, I'd probably close this for now as `-R` appears to be a recent addition and we shouldn't assume such a feature is ubiquitous until a good year has passed since availability.

bennuttall:
Agreed

PR #388, opened 2016-07-06
First stub document for Windows based network booting

OldBoyCoder:
Very much just a stub.  I'll flesh this out over the next few days


ghollingworth:
I would suggest keeping this PR open and squashing the commits all together when you've finished and are happy...

OldBoyCoder:
Thanks Gordon,  I've attempted to flesh out the document so I think it stands better as a first draft.  It'll definitely need more work from more informed people than me but at least it's a start.

PR #393, opened 2016-07-09
correct display of mount point of MSD device

kduvekot:
Just a minor layout specific change .. nothing material


lurch:
Looking at the 'raw' view of `msd.md` it looks like each of the "indented sections" have a different indent level? Might be nice of they were all the same ;-)

PR #395, opened 2016-07-11
Rewrite netboot and mass storage instructions

liamfraser:


lurch:
I see you've changed the netboot instructions to connect the SERVER and CLIENT Pis to an existing (home) network, rather than having them only directly-connected over an ethernet cable. But doesn't this then mean that there's two DHCP servers on the same network? IIRC it's then a matter of random luck which DHCP server-reply the Client will use, and if you're unlucky it'll use the response from the router, which won't contain the extra-info needed for TFTP booting.
(I haven't tried netbooting one of my Raspberry Pi 3s yet (I will when I get the time), but I ran into this problem many many years ago when I was experimenting with netbooting on an old Sun Javastation).

liamfraser:
Gordon said the bootloader on the Pi will wait until it has both an IP address and the pxe boot info, so it will only start to boot once it has a reply from the pxe dhcp server.

lurch:
Oooh, that's clever :-)

ghollingworth:
#blush

lurch:
"the address of your DNS server, which is the same address as your gateway"

That's _often_ the case with (home) routers, but not always! At a previous house (I think it was using Virgin Media at the time) the gateway address was different from the DNS server addresses (which always made static IP address configuration a bit more awkward, trying to remember which was which).
To avoid confusion, IMHO the guide here ought to treat them separately.

PR #408, opened 2016-08-08
New boot modes: GPIO pull up typo.

ED6E0F17:
OTP fuses must be blown to enable new boot modes . GPIO pull-ups can then be used to disable these modes.


ghollingworth:
No,

You've got the sense the wrong way around...

Enabling the gpio_bootmode OTP in the chip and leaving the GPIOs as they are (read as 0) all the bootmodes will be disabled.

To then enable a boot mode you apply a single pull up to the required GPIO

Gordon

ED6E0F17:
Ok. I thought it might be that.

ghollingworth:
If you think it's not clear enough it might need a little more clarification?

Thanks

Gordon

PR #413, opened 2016-08-13
Move note about dcfldd up

C-Bouthoorn:
Put `dcfldd` before the actual command, as most users will probably first run the `dd` command and then continue reading, so they find out about `dcfldd` too late. (Or at least I did)


JamesH65:
No longer required when using the status=progress flag on dd.

Issue #414, opened 2016-08-15
Add details for access-over-internet service


timfernando:
My first contribution here, I'm proposing an addition/deletion to [access-over-internet](https://github.com/raspberrypi/documentation/blob/master/remote-access/access-over-Internet/internetaccess.md) to better explain the already-listed remote access service Dataplicity and how it works. 

I'm adding an issue here as per contribution instructions, but as the change is minor I've also the PR for it. Hope that's OK 👍

lurch:
There's ongoing discussion on the relevant PR, so closing this issue.

timfernando:
I believe the discussion is complete as the requested change has been made. Should the merge request go in the PR or in the issue?

lurch:
The PR (Pull Request) *is* the merge request ;)

PR #415, opened 2016-08-15
Add details to access-over-internet

timfernando:
Added a short intro to remote access service Dataplicity to explain how it works and what it does. Removed the small link from the bottom too.

Relates to issue #414


phiney0711:
Is there a valid reason to change the listed order (from top-of-page to bottom-of-page) for the remote access alternatives to port-forwarding?  (disclosure: I am a Weaved employee)

timfernando:
It seemed sensible for it to be alphabetical below the overall heading of port-forwarding (as it was lumped under "alternatives" before). Is there a better order that you would suggest?

pelwell:
But Dataplicity appears after port-forwarding, so it is already out of order. To avoid any suggestion of trying to give Dataplicity greater prominence over Weaved (I have no connection to either and no preference) then you could leave the order as it was. Alternatively, a declaration of your own interests might help to clear the air.

timfernando:
@pelwell no problem, I've adjusted the order. 

Disclosure: I've written a [blog article using Dataplicity](http://medium.com/@timfernando/how-to-build-a-raspberry-pi-thermometer-you-can-access-anywhere-a-beginner-s-guide-4ad44ce9f4c9) and am connected to the team behind it, but I am not an employee.

timfernando:
@pelwell @phiney0711 Thought and suggestion: a stub page with a list of one-liners with links to separate pages for each? Perhaps that would be reasonable for all concerned?

bennuttall:
Port forwarding should be at the top because it's probably the first thing a user should try. Let's leave it. The other options should really be under the heading "Alternatives" or similar, with subheadings "Weaved" and "Duplicity".

I don't think there's a problem with the current presentation, but if the list of examples was longer it would make sense to cut it to a list of links to other pages for each alternative solution.

bennuttall:
I'm going to merge and fix the headings.

phiney0711:
looks good.  thanks!

PR #416, opened 2016-08-19
Display/port wrong on ElCap & Sierra

flipphillips:
When I use the macOS built-in VNC viewer, I have to specify the -full- port, not just the display number. So, 59xx where xx = the display number.


lurch:
Duplicate of #373 ?

PR #426, opened 2016-09-13
OS X Screen Sharing problem due to screen number

jcdumas:
After encountering an issue while following this documentation, I found the problem and the solution. I've added to the documentation accordingly.

The OS X built in VNC client (Screen Sharing) doesn't appear to like screen numbers (:0, :1, :2), and only understands explicit port numbers (:5900, :5901, :5902). Tested on OS X 10.9.5.


lurch:
Duplicate of #373 ?

It would be great if somebody who has commit-access to this repo, and access to a Mac for testing, is able to close all these duplicate "VNC on OSX" documentation issues & PRs in one fell swoop...

waveform80:
Sorry, I do - haven't checked up on this repo in a while (my github notifications list is *huge*). I'll get #373 committed in a mo.

PR #430, opened 2016-09-21
Update msd.md

realdmitchell:
See line 5: it says 

Before a Pi will network boot

Did you mean 

Before a Pi will USB boot


JamesH65:
@ghollingworth I believe this change is correct- can you confirm and forward to copy edit as appropriate.

AisforAstronaut:
@JamesH65 File checked and approved. No changes made. Cannot merge until the conflict is resolved.

PR #431, opened 2016-09-22
Update linux.md

ovular:
truncate command wrong?!

### DIFF
--- a/installation/installing-images/linux.md
+++ b/installation/installing-images/linux.md
@@ -34,7 +34,7 @@ Please note that the use of the `dd` tool can overwrite any partition of your ma
 
     ```bash
     dd bs=4M if=/dev/sdd of=from-sd-card.img
-    truncate --reference 2016-05-27-raspbian-jessie.img from-sd-card.img
+    truncate --reference from-sd-card.img 2016-05-27-raspbian-jessie.img
     diff -s from-sd-card.img 2016-05-27-raspbian-jessie.img
     ```
 


pelwell:
Please explain why you think it's wrong (hint: I don't).

ovular:
The SD-Image is changing the size not the raspbian image.

pelwell:
Yes, it's supposed to. The purpose of this sequence of commands is to verify that the card has been written correctly. It does so in a rather clumsy way, but it works:

1) Read back the whole card into an image.
2) Truncate that image to the size that was written.
3) Compare the original image with the read back image.

The document says:

> The SD card might be bigger than the original image, and `dd` will make a copy of the whole card. We must therefore truncate the new image to the size of the original image.

dd creates an image of the whole card called `from-sd-card.img`. We then use truncate to discard any of the image that is beyond the size of the Raspbian image. This changes the size of `from-sd-card.img` but not the Raspbian image. Then we use diff to check they are the same.

lurch:
I agree with @pelwell 
In an ideal world, you'd be able to tell `dd

pelwell:
Actually you can, using a combination of "bs" (block size) and "count" number of blocks. You would need to use `stat` to get the size of the Raspbian image, and it might be tricky to choose a suitable block size to be efficient and yet be a factor of the total image size... Ah-ha, the "count_bytes" iflag changes "count" to be in bytes. So it should be possible to say:
 ```
dd bs=4M if=/dev/sdd of=from-sd-card.img iflag=count_bytes count=`stat -c %s 2016-05-27-raspbian-jessie.img`
```
(although you might want to split it in two for clarity:
```
raspbian_size=`stat -c %s 2016-05-27-raspbian-jessie.img`
dd bs=4M if=/dev/sdd of=from-sd-card.img iflag=count_bytes count=$raspbian_size
```

**Comment by lurch** [comment]
> Ah-ha, the "count_bytes" iflag changes "count" to be in bytes. So it should be possible to say...

Oooh, cunning! And I guess this approach would also be _much_ faster if you've e.g. written the Raspbian image onto a 32GB microSD card :)

**Comment by pelwell** [comment]
Or we might be able to simplify everything and just use:
```
cmp --bytes=$src_size /dev/sdd 2016-05-27-raspbian-jessie.img
```
If not, we should definitely delete the temporary file.

And while we're at it, we should put the name of the image in a variable to save the duplication.

lurch:
I guess using `cmp` directly _might_ be slower, because it probably uses a smaller block read size than `dd`? \*shrug\*  I don't have time to test myself ATM.

I wonder if using variables for the image names might be too confusing for newbies?

ovular:
Ok understand.
I thought that the image i have to write to the sd card must be the same size as the sd card.
But it is only a check if there was no erros with dd.

My problem is that the root partition doesn't fit to the sd card. With gparted I can see, that there are unpartitioned space on the sd card.

Sorry for prematurely change the docu.

lurch:
> With gparted I can see, that there are unpartitioned space on the sd card.

Check the Expand Filesystem option in raspi-config :-)  https://www.raspberrypi.org/documentation/configuration/raspi-config.md

PR #432, opened 2016-09-26
Replace existing VNC instructions with RealVNC-specific ones

minkustree:
An updated set of VNC instructions for a world where RealVNC's VNC Server and VNC Viewer are the default and pre-installed. 

Creating a pull request for these changes as discussed with various members of the Raspberry Pi team. 

Clearly it doesn't make sense to merge these changes until such a time as the RealVNC VNC Server & Viewer are the default. 😄


minkustree:
I'm not sure why the build check failed - looking at the logs, it seems to be more to do with the environment than these changes? 

Now that it's been announced that RealVNC is part of Raspbian, it would be good to get these changes merged as soon as possible, please?

lurch:
> I'm not sure why the build check failed - looking at the logs, it seems to be more to do with the environment than these changes?

Oh yeah. I tried restarting the build (in case it was a temporary failure) and got the same error.
Hopefully a fix won't take too long... https://github.com/gjtorikian/html-proofer/issues/357

carlosperate:
Sounds like you need ruby 2.2 or higher for some of the dependencies and the travis worker is using 1.9.x.

lurch:
Thanks @carlosperate I just tried [updating](https://github.com/raspberrypi/documentation/commit/a903a176bc05b49d1f7f5ed2fdf1a6288ecae780) the Ruby version to 2.2 (which looks like the highest version that Travis supports), and it seems it's still not new enough :-(

carlosperate:
Yeah, it seems like the default `2.2` is `2.2.0` and the dependencies need `2.2.2`.

Based on https://docs.travis-ci.com/user/languages/ruby/#Choosing-Rubies-that-aren%E2%80%99t-installed it looks like you might be able to install other versions as long as they are available in the Ubuntu 12.04 packages. If 2.2.2 is not available in 12.04 you can ask travis to use 14.04, and that should probably have the version available.
https://docs.travis-ci.com/user/trusty-ci-environment/

Alternatively you can probably specify older versions of the dependencies that worked with earlier ruby versions.

lurch:
https://github.com/gjtorikian/html-proofer/issues/357#issuecomment-250145792
:-)

PR #435, opened 2016-10-03
Made it safer

UnderFact:
I replaced the "/dev/sdd" with "/dev/cats" so that people who aren't paying attention wont break something.


XECDesign:
I would suggest /dev/sdX and make sure the paragraph before it is also modified accordingly.

UnderFact:
good point, anybody with 26 storage devices will now what their doing
anyway.

lurch:
> anybody with 26 storage devices will now what their doing anyway.

24 ;-)
And even then, it'd still be `/dev/sdx` (all lowercase) and not `/dev/sdX`

Frodox:
> anybody with 26 storage devices will now what their doing anyway.

24 ? On Raspberry Pi? Seriously? :D I think this man would be triply careful with partitions :)
+1 for `sdX`, because  it is **very important** to note, that `sdX`, nor `sdxN`.

JamesH65:
@underfact Would it be possible to rebase this PR against the latest documentation, taking in to account the suggestions above, then I can push to copy edit and get it merged.

JamesH65:
Actually, I'm going to fix this up myself, and will close this PR when have done so. see #536

PR #441, opened 2016-10-12
Update README.md

lancelocey:
Someone who is new to Linux may not know the zip/unzip procedure.

### DIFF
--- a/installation/installing-images/README.md
+++ b/installation/installing-images/README.md
@@ -10,7 +10,8 @@ Official images for recommended operating systems are available to download from
 
 Alternative distributions are available from third-party vendors.
 
-After downloading the `.zip` file, unzip it to get the image file (`.img`) for writing to your SD card.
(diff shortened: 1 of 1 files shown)


waveform80:
Looks good to me. The only thing I'd point out is that the images usually just have the suffix `.zip` rather than `.img.zip` but I don't think that matters much.

lancelocey:
A new person would not know how to unzip.  That's all

Sent from my iPhone

On Oct 15, 2016, at 8:33 AM, Dave Jones <notifications@github.com<mailto:notifications@github.com>> wrote:


Looks good to me. The only thing I'd point out is that the images usually just have the suffix .zip rather than .img.zip but I don't think that matters much.

bennuttall:
A command line instruction wouldn't help new people. New people (or people who need help with this step) would likely be using Windows or Mac. And we recommend NOOBS for complete beginners, rather than writing images to SD cards, and we provide a [guide for beginners](https://www.raspberrypi.org/learning/software-guide/) in the help pages.

I don't think this change particularly helps. Thanks though!

PR #446, opened 2016-10-28
Update linux.md

peterwillis:
You can use ctrl+t to send a SIGINFO to dd whilst it is running to see the progress.


XECDesign:
Which Linux distros does this work on? I thought it was just MacOS and a few *BSDs, not any common Linux.

peterwillis:
Yeah it looks like you are right sorry. Wikipedia says:

> Ctrl-T (not supported on all UNIXes) sends an INFO signal (SIGINFO); by default, and if supported by the command, this causes the operating system to show information about the running command.

PR #453, opened 2016-11-15
Update configuring.md

GameDevelopmentCompany:


pelwell:
The original is correct - your amended version is ungrammatical.

PR #454, opened 2016-11-19
add note about newest upstream toolchain download

cgzones:


Ruffio:
@lurch Should this get merged?

JamesH65:
There is no guarantee from RPF that the latest Linaro toolchain is going to work (and it won't be Pi specific which could problems elsewhere in the documentation)  so I'm going to reject this PR.

PR #456, opened 2016-11-22
Fix what exactly should be done to force wpa-supplicant  take config changes into account

mvasin:
According to manual https://w1.fi/cgit/hostap/plain/wpa_supplicant/README : "Changes to configuration file can be reloaded be sending SIGHUP signal to wpa_supplicant ('killall -HUP wpa_supplicant'). Similarly, reloading can be triggered with 'wpa_cli reconfigure' command."

### DIFF
--- a/configuration/wireless/wireless-cli.md
+++ b/configuration/wireless/wireless-cli.md
@@ -37,6 +37,6 @@ network={
    
 Now save the file by pressing **Ctrl+X** then **Y**, then finally press **Enter**.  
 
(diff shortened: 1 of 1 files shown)


JamesH65:
This appears to be technically valid. Tried it on a recent Raspbian install and it did appear to do what it says.

mvasin:
We are working with `wpa_supplicant` here, and I suggested the way to take config changes into account recommended in the `wpa_supplicant` documentation. This passage "If it does not, restart..." is there because maybe current approach doesn't work as good. At least I had no troubles with `sudo wpa_cli reconfigure`.

JamesH65:
Thanks, this has been passed on for checking and inclusion.

LornaLynch:
Merged: thanks.

PR #461, opened 2016-11-27
dd showing write progress status

twikedk:
As of GNU Coreutils 8.24+ dd is able to show progress with the option status=progress

### DIFF
--- a/installation/installing-images/linux.md
+++ b/installation/installing-images/linux.md
@@ -17,7 +17,8 @@ Please note that the use of the `dd` tool can overwrite any partition of your ma
 - In the terminal, write the image to the card with the command below, making sure you replace the input file `if=` argument with the path to your `.img` file, and the `/dev/sdd` in the output file `of=` argument with the right device name. This is very important, as you will lose all data on the hard drive if you provide the wrong device name. Make sure the device name is the name of the whole SD card as described above, not just a partition of it; for example, `sdd`, not `sdds1` or `sddp1`, and `mmcblk0`, not `mmcblk0p1`.
 
     ```bash
-    dd bs=4M if=2016-09-23-raspbian-jessie.img of=/dev/sdd
+    dd bs=4M if=2016-09-23-raspbian-jessie.img of=/dev/sdX && sync
+    dd bs=4M if=2016-09-23-raspbian-jessie.img of=/dev/sdX status=progress && sync
     ```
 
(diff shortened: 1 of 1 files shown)


XECDesign:
What percentage of the top 10 distros ship with that version of dd in their stable repos?

twikedk:
Taken from distrowatch.com?? I would say more than 50% of the top 10 and
basically every new distro from this year.

--
Ole

2016-11-27 15:35 GMT+01:00 XECDesign <notifications@github.com>:

XECDesign:
Then this would error out for a lot of people. Given that the documentation is for beginners, that probably wouldn't be the best start for them.

twikedk:
No you are probably right I guess it is better that the page recommend the
use of 'dcfldd' which is not even part of a default installation.
If you look at my suggestion I didn't remove the old command just added
another with status=progress option and sync included.
By the way Linux Mint, all Ubuntu derivatives, Fedora and openSUSE holds
the new coreutils today which probably also is the start
point for a lot of new comer, but never mind just reject it and close the
request again.

--
Ole

2016-11-27 17:14 GMT+01:00 XECDesign <notifications@github.com>:

XECDesign:
I am not accepting or rejecting anything, only commenting. I don't have write access to this repo.

Adding a command without removing the old one does not make it any clearer.
Just a note that mentions the use of status= would probably be useful.

Changing sdd to sdX without changing the references to sdd is another issue.

Is the added sync needed? AFAIK, dd won't return until all the data is actually written.

twikedk:
The sdX was a last minute add, sorry for not completing it but I always
believe having example code that isn't directly destructible is a good
thing. For the sync command flushing your cache before removing drives just
seems like a reasonable ending as far as I know even dd uses cache unless
clearly set with an option not to do so.

--
Ole

2016-11-27 17:56 GMT+01:00 XECDesign <notifications@github.com>:

gelliott181:
Just a thought, but the `dd` command outputs it's current state with Ctrl-T. If I'm not mistaken that's the command for a SIGINFO?

gelliott181:
I figure if we mention `status=progress` the old school way could also make an appearance.

lurch:
@gelliott181 see #446

JamesH65:
Duplicate of #350. Please use that one for further comments.

I think I'd happy to have the status=progress added as the default, with a disclaimer to remove if it doesn't work.

PR #371, opened 2016-06-01
Update docs for nginx install

GarethOates:
Having recently installed nginx on a raspberry pi, I noticed the default location for the web page is now /usr/share/nginx/html  not /usr/share/nginx/www like previously described.  Also I clarified that the options are visible in the file named "default" in the /etc/nginx/sites-available directory.

Closes #370

PR #462, opened 2016-11-28
Default HTML test file location updated

ahogen:
When I start with a fresh install of nginx, the test HTML file is in `/var/www/html/ `instead of the previously mentioned `/usr/share/nginx/www`. It could be nice to indicate that to viewers even though, like the article mentions, they can check the currently set path themselves in the config file.

### DIFF
--- a/remote-access/web-server/nginx.md
+++ b/remote-access/web-server/nginx.md
@@ -28,7 +28,7 @@ Browse to the default web page either on the Pi or from another computer on the 
 
 ### Changing the default web page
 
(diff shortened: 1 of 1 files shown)


lurch:
Does this mean that #371 is now outdated?

ahogen:
@lurch, I would guess so. Unless I did something wrong during my setup, which could very well be the case since I am brand new to using nginx.

I just wanted to report what my default setup came out looking like, in case it was helpful.

JamesH65:
Just checked, oddly that the html files in both those locations are the same on a fresh install, although they have different names. However, the default config file refers to "root /var/www/html" so I would suggest that is the required location for the index,

I'll close the other PR and let this one through.

AisforAstronaut:
File checked and approved. No changes made.

Is this ready to merge, or does the failing check need fixing?

JamesH65:
I'm ignoring the Travis check at the moment, theory being get all the PR's out of the way, then fix the test.

PR #470, opened 2016-12-13
add a note to the expand-filesystem option

bamarni:
Currently, Raspbian passes a custom init script to the kernel during first boot (`/usr/lib/raspi-config/init_resize.sh`) which already expands the root partition. I assume this config command is not much relevant anymore.

### DIFF
--- a/configuration/raspi-config.md
+++ b/configuration/raspi-config.md
@@ -55,7 +55,7 @@ Generally speaking, `raspi-config` aims to provide the functionality to make the
 <a name="expand-filesystem"></a>
 ### Expand filesystem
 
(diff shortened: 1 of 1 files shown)


JamesH65:
Good point, but I'm not keen on the wording - do we need anything in there at all about resizing? Also worth noting that raspi-config does have the expand option still in it - does it need to be there?

Going to pass on for copy editing to see if they have better ideas.

LornaLynch:
I've edited the text a little, but would agree that it seems a little redundant. Is the changed version any better, or should the text be cut altogether?

LornaLynch:
I've changed the text in line with our discussion: OK to merge?

JamesH65:
Reads fine. Merged.

PR #476, opened 2016-12-22
Update VNC option location

waveform80:
VNC option has moved from "Advanced Options" to "Interfacing Options"


LornaLynch:
Thanks!

PR #477, opened 2016-12-24
Fix misspelling of 'sudo'

Aaronduino:
Was 'sudp' (a 'p' instead of an 'o')


Ruffio:
@LornaLynch can this be merged?

LornaLynch:
Thanks for the spot.

PR #482, opened 2017-01-07
Update README.md

MelvinLoos:
Updated option names for enabling SSH through raspi-config


XECDesign:
Looking at the latest version of raspi-config, this change seems wrong.
https://github.com/RPi-Distro/raspi-config/blob/master/raspi-config#L1561
https://github.com/RPi-Distro/raspi-config/blob/master/raspi-config#L1465

SSH is configured from interfacing options.

It's best to leave out menu entry numbers like A4, since these are likely to change as options are added/removed.

InfopactMLoos:
Yeah sorry it seemed I needed to run an update, I assumed the image I downloaded contained the latest version.

MelvinLoos:
Yeah sorry it seemed I needed to run an update, I assumed the image I downloaded contained the latest version.

PR #483, opened 2017-01-10
Removed a mistake

Dreamzilla:
Serverauditor changed its name to Termius, I've updated the documents
that mention Serverauditor


LornaLynch:
Thanks for the spot!

PR #491, opened 2017-01-20
Fixes #490 by adding small introduction to using wpa_passphrase with wpa_supplicant configuration

### LINKED ISSUE #490: Using `wpa_passphrase` for generating secret token instead of entering password as plain text
Many tutorials (including the official documentation here) store the password for a profile inside `wpa_supplicant.conf` as plain text. This is wrong since it not only doesn't mention the security concerns related to this but also doesn't nurture a more security-oriented approach from the user when handling such information. In addition `wpa_passphrase` ensures a minimal length for a password (8..63 characters) which adds to the security enhancement it already offers.

I have added some information about this in my edit (check my fork of the documentation and the **Wireless CLI** article in particular).

aleksandaratanasov:
Most tutorials (including the official documentation for the Raspbian) use the very simple way of entering a password in a `wpa_supplicant` profile as **plain text**. Depending on the scenario (for example a server) the user might not want to do that. Using `wpa_passphrase` ensures that we use a secret token instead of plain text password in our configuration file. This fixes the issue #490 .

### DIFF
--- a/configuration/wireless/wireless-cli.md
+++ b/configuration/wireless/wireless-cli.md
@@ -11,7 +11,7 @@ To scan for WiFi networks, use the command `sudo iwlist wlan0 scan`. This will l
 
 1. `IE: IEEE 802.11i/WPA2 Version 1` is the authentication used; in this case it's WPA2, the newer and more secure wireless standard which replaces WPA. This guide should work for WPA or WPA2, but may not work for WPA2 enterprise; for WEP hex keys, see the last example [here](http://www.freebsd.org/cgi/man.cgi?query=wpa_supplicant.conf&sektion=5&apropos=0&manpath=NetBSD+6.1.5). You'll also need the password for the wireless network. For most home routers, this is found on a sticker on the back of the router. The ESSID (ssid) for the network in this case is `testing` and the password (psk) is `testingPassword`.
 
-1. If `security` is a concern entering a password as plain text is not recommended. You can use the `wpa_passphrase` to create a unique token and enter it instead. As arguments the command requires the ESSID and the password. With the example from above calling the command will be `wpa_passphrase "testing" "testingPassword` which will output a ready-to-use wpa-compliant network profile:
+1. If `security` is a concern, entering a password as plain text is not recommended. You can use `wpa_passphrase` to create a unique token and enter it instead. As arguments, the command requires the ESSID and the password. With the example from above, calling the command will be `wpa_passphrase "testing" "testingPassword` which will output a ready-to-use wpa-compliant network profile:
 
   ```
   network={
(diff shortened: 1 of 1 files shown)


configuration/wireless/wireless-cli.md
```diff
@@ -34,6 +47,8 @@ network={
     psk="testingPassword"
 }
 ```
+
+If you are using `wpa_passphrase` you can redirect its output to your configuration file by calling `wpa_passphrase "testing" >> /etc/wpa_supplicant/wpa_supplicant.conf`. Note that this requires you to change to `root` (by executing `sudo su`) or find another way to redirect the output since the file we write to can be altered only by a user with administrative privileges. Last but not least make sure you use `>>` (which is used to append text to an existing file) since `>` will erase all contents and **then** append the output to the specified file.
```
lurch: Based on the other examples, it seems the `wpa_passphrase` invocation here is missing the password?

lurch:
Rather than make the existing instructions (which are aimed at beginners) more complex, I wonder if it _might_ make sense to add your "how to use wpa_passphrase for extra security" instructions to a separate page, linked to from the current page?

configuration/wireless/wireless-cli.md
swerveshot: Don't forget to purge the wpa_passphrase command from the bash_history or you will still have the psk saved in clear text.

swerveshot:
@lurch I agree a link would make more sense as it raises awareness of a more secure config without making things too complex for a novice user.

configuration/wireless/wireless-cli.md
aleksandaratanasov: Yes, it seems I have forgotten to add the password for the piped call. Will fix it.

aleksandaratanasov:
I didn't know that it was possible to add a whole new page as a contributor. Can someone tell me how to do that? I do not however agree that using `wpa_passphrase` is not beginner-friendly. Leaving the fact that it offers a more secure way of doing the `wpa_supplicant` configuration it actually gives a very fast and easy-to-follow way of creating a whole new profile just by entering the command itself followed by the ESSID and password.

lurch:
> Can someone tell me how to do that?

https://help.github.com/articles/adding-a-file-to-a-repository/

JamesH65:
I like the idea of this section - I don't think it need to be a new page though - I'd be inclined to add it as a new section at the end of the current page, titled something like "How to make you wifi password more secure" or similar. Make it clear this is optional.

@aleksandaratanasov Will you be able to do that? Once done I can forward to the copy editors and get it merged.

aleksandaratanasov:
I can do either one of these but it will be nice to know which one is the preferred solution. :)

JamesH65:
Sorry! I'd suggest a new section at the end of the current page, marked as optional.

JamesH65:
Fixes #491

LornaLynch:
Edited and merged. Thanks!

PR #492, opened 2017-01-21
Updated documentation for enabling ssh

jlian:
1. The interface described in the docs for `raspi-config` was different than what appears in the latest version of Raspian. I've updated the instructions to reflect that change. Like from here http://www.raspberrypi-spy.co.uk/2012/05/enable-secure-shell-ssh-on-your-raspberry-pi/

2. Moore context for the extension-less 'ssh' file you could place on the boot partition to enable ssh. I was having some network issues with my router and was confused why the 'ssh' file would disappear all the time. I found the answer via https://www.raspberrypi.org/blog/a-security-update-for-raspbian-pixel/


Hum4n01d:
Oh lol I just did the same thing #496

JamesH65:
This is fine - should be merged asap.

LornaLynch:
Edite and merged. Thanks!

PR #496, opened 2017-01-22
Fixed instructions

Hum4n01d:
Interfacing options has apparently been renamed to advanced options so I fixed it


Hum4n01d:
Never mind, someone else already did this #492

PR #508, opened 2017-02-13
Update VNC docs for upcoming new version

minkustree:
Later on this week a new version of VNC is scheduled to hit the Raspbian repositories (currently aiming for 18th Feb).

We'd like to update the VNC docs to reflect the changes, but for obvious reasons it shouldn't happen before the new version hits the repositories!

Please let me know if there are any changes that we need to be made before that can happen.

Thanks!

Andy


minkustree:
This update has now been announced, so it would be great to get these documentation changes merged, please?

LornaLynch:
All done: thanks!

minkustree:
Great, thanks!

lurch:
See also #514

minkustree:
Drat, excellent point. #514 should be merged too in order to fix our mistake. Thanks @lurch !

Issue #506, opened 2017-02-07
Feature request: How to mount hard drives 


SWLF:
I've written some documentation on mounting a hard drive in Raspbian. If you think that this information will be a useful addition to the documentation, I can submit a pull request so you can review it.

JamesH65:
We are always happy to look at PR's. Send it in and we can take a look.

JamesH65:
Please update this issue with a link to the PR when you submit it.

SWLF:
PR: https://github.com/raspberrypi/documentation/pull/516

JamesH65:
Fixed by #516

PR #516, opened 2017-02-22
#506: Added documentation for connecting an external hard drive over USB

SWLF:


configuration/external-storage.md
```diff
@@ -0,0 +1,83 @@
...

+    ```
+    lsof /mnt/PIHDD
+    ```
+
+2. End all programs using the mount point. Run the following command:

+    ```
+    sudo kill 13827
+    ```
```
JamesH65: Really not happy with using kill. Better instructions would be to manually close down each app that is using the mount point.

configuration/external-storage.md
```diff
@@ -0,0 +1,83 @@
...
+For more information on the Linux commands, refer to the specific manual pages using the `man` command. For example, `man fstab`.
+
+## Unmounting a HDD
+Before you unmount your HDD, ensure that there are no programs accessing the HDD.
+
+1. Run the following command to get the list of programs using the mount point:

+    ```
+    lsof /mnt/PIHDD
+    ```
```
JamesH65: lsof is not installed in raspbian by default - you will need to install it for this to work.

configuration/external-storage.md
```diff
@@ -0,0 +1,83 @@
+# External Storage configuration
+You can connect your external hard disk drive (HDD) to any of the two USB2.0 ports on the Raspberry Pi and mount it to access the data stored on your HDD. 
+
+**NOTE:** The following instructions require that your Raspberry Pi is running Raspbian.
+
```
JamesH65: I think that all our documentation assumes Raspbian, so this could be omitted.

configuration/external-storage.md
```diff
@@ -0,0 +1,83 @@
...
+    ```
+    sudo blkid
+    ```
+
+3. Note the location of the disk partition. For example, `/dev/sda1`.
+4. Create a target folder to be the mount point of the HDD. Run the following command:

+    ```
+    sudo mkdir /mnt/PIHDD
+    ```
```
JamesH65: Needs to be clear that the PIHDD is user defined - it doesn't need to be this.

configuration/external-storage.md
```diff
@@ -0,0 +1,83 @@
+# External Storage configuration
+You can connect your external hard disk drive (HDD) to any of the two USB2.0 ports on the Raspberry Pi and mount it to access the data stored on your HDD. 
+
```
JamesH65: There are 4 ports on the 3, might be worth just saying any USB port.

JamesH65:
@SWLF Are you going to be able to make the changes as suggested by the reviews?

SWLF:
@JamesH65 yes. Please review the latest commit.

JamesH65:
This looks technically OK to me but I don't have the equipment to actually test the entire process. I think we can go to copy edit and merge it - if issues turn up later we can fix then.

AisforAstronaut:
Document checked. Minor changes made. Pushed back for technical checking.

procount:
Would it be a good idea to add information about changing ownership and permissions of the mount point? Typically, mounting a USB HDD via sudo leaves it owned by root, so user pi cannot write to it without using sudo, especially a problem for newbies using the desktop file manager.

JamesH65:
Sounds like a good idea, but not my field - would be happy to try out any PR's though....

PR #518, opened 2017-02-27
Updated progress options for mounting raspbian using dd

chuchro3:
There's a bulky paragraph explaining how to get visual feedback when running the `dd` operation. We can actually just append `status=progress` as a simple solution to be able to monitor the progress and make sure things aren't frozen.


lurch:
Duplicate of #461 ?

JamesH65:
Yes, should be closed. Awaiting the right privileges to do so, unless OP wishes to close themselves.

PR #520, opened 2017-02-27
Update raspi-config.md

JamesH65:
Removed reference to rastrak as has been removed from raspi-config.

NOTE: There is still an image referenced from this page that includes the older screenshot with rastrak still present so that still needs to be updated.


JamesH65:
I'm going to merge this now as there is a Travis break that this will help with, |I'll dd n issue that flags up the incorrect screen grab

PR #504, opened 2017-02-02
Update ftp.md

hdillies:
Hello,

I've followed this documentation and I have to "sudo" the "ln" command.

Bye.

Herve.

PR #537, opened 2017-03-03
Update ftp.md - missing sudo's

JamesH65:
Update ftp instructions to use sudo where it is missing. Partly from #504


LornaLynch:
Thanks!

PR #539, opened 2017-03-03
Update linux.md - add use of status=progress to dd.

JamesH65:
Removed the previous rather clunky process, added use of status=progress to dd, which should now be fairly standard on most distro's.

Replaces PR #350


LornaLynch:
Text edited: merging.

Issue #185, opened 2015-03-03
Win32DiskImager now downloads as a .exe file instead of a .zip


lurch:
[installation/installing-images/windows.md](https://github.com/raspberrypi/documentation/blob/master/installation/installing-images/windows.md) says "Download the Win32DiskImager utility from the Sourceforge Project page (it is also a zip file); you can run this from a USB drive." but the link to SourceForge now downloads `Win32DiskImager-0.9.5-install.exe` by default, which means the documentation now needs updating.
(It means it can't be run from a USB drive, but there's now no need to **Run as administrator** as it requests elevated privileges itself)

JamesH65:
Fixed by #544

JamesH65:
Fix merged.

PR #544, opened 2017-03-06
Update windows.md - win32diskimager instructions

JamesH65:
Win32DiskImager now comes with an installer and does not need "Eun as admin" as it elevates privileges itself. Updated docs to match.


JamesH65:
fixes #185

AisforAstronaut:
File checked. Changes made. Pushing back for technical checking.

PR #562, opened 2017-03-08
Jamesh65 patch 1#87

JamesH65:
Update the Wireless config to remove some outdated links, and some outdated files.


AisforAstronaut:
File checked. No changes. Squashed and merged.

PR #564, opened 2017-03-08
Update linux.md

JamesH65:
Added information for linux installs that do not auto mount SD cards.


AisforAstronaut:
File checked. Changes made for grammar and style. Pushing back for technical checking.

PR #579, opened 2017-03-13
Slight change to config readme

JamesH65:
Wireless option now specifies Pi3 inbuilt Wireless as well as a wireless dongle


LornaLynch:
Copy edit done - tweaked the text to add reference to the Pi Zero W, which also has wireless connectivity.

PR #422, opened 2016-09-05
Revised scratch gpio doc

timrowledge:
Add new commands and links to frizzing diagrams for example wiring.

Let's see if this does things right...

PR #581, opened 2017-03-14
Scratch gpio changes

JamesH65:
Modified PR from #422, had to removed some unnecessary files.


JamesH65:
This is a modification of #422 with some extraneous files removed. @LornaLynch Can you give it a once over to make sure I haven't missed anything?

PR #588, opened 2017-03-17
Show redirect rather than 404 for moved config.txt page

bennuttall:


JamesH65:
Done an immediate merge on this, no need for copy edit.

PR #593, opened 2017-03-21
Update raspberrypi.org URLs

bennuttall:
- Update raspberrypi.org URLs from http -> https
- Update raspberrypi.org forum URLs /phpBB3/ -> /forums/

If this doesn't flag up any broken URLs when the test suite runs, it should be good to merge.


bennuttall:
Yep - it's good: https://travis-ci.org/raspberrypi/documentation/builds/213412025

PR #613, opened 2017-04-09
Update linux.md - raspbian img name correction

milankni:
linux.md was only partiall updated to reflect the current image file name (2017-03-02).


JamesH65:
No copy edit needed....merging.

Issue #455, opened 2016-11-19
Include instructions for Etcher 


johannrichard:
[Etcher](https://etcher.io) is a graphical, cross-platform tool that is by far the easiest way to flash disk images to SD cards (also for Mac OS X which has been short of suitable, graphical installers so far).

It would probably be helpful to have a short instruction on using Etcher in the documentation.

![product](https://cloud.githubusercontent.com/assets/189003/20457964/f7909b30-ae97-11e6-9e6f-c44b3cb2f1e9.gif)

Ruffio:
@lurch what is the opinion of introducing Etcher here?

brianjking:
@Ruffio -- I highly suggest it! It makes it very easy to add the `ssh` folder to the `boot` partition to enable headless connection via `ssh` to the Pi with the recent changes to the Raspbian distribution releases.

I opened a PR #511 for this on the MacOS page and as I mentioned in the pull request if this is merged in I'm happy to write instructions for Windows as well as this is also proposed in #450.

Thanks!

Ruffio:
@brianjking I'm not within the foundation and have no ruling :-( You must convince them in order to get your RP accepted.

brianjking:
@Ruffio Sorry - updated post & tagged @lurch

JamesH65:
The MAC docs for Etcher have been merged - would it be possible for a Windows version to be written as suggested above? @johannrichard

JamesH65:
Actually, Etcher on Linux should also be documented if you have the time.

lurch:
Only needing one set of instructions for all three platforms, was part of the original reason behind Etcher ;-)
https://resin.io/blog/introducing-etcher/

JamesH65:
I wonder if we should have one page for Etcher, and link to it from the other OS specific pages?

JamesH65:
After thinking about it for another 2 minutes, Yes, I think that is the way to go - one Etcher page, then link to it from the individual OS installation pages. So if @brianjking fancies knocking up a PR to do that, that would be splendid.

pimterry:
Looks like this has gone quiet - anybody object if I pick this up?

JamesH65:
Feel free!

pimterry:
@lurch @JamesH65 This is now done, see #614.

JamesH65:
Fixed by #614

PR #614, opened 2017-04-11
Include Etcher instructions as default for image writing

pimterry:
This solves #455.

I've gone for including the instructions in the top-level readme, rather than a separate file, to keep the happy path obvious and because they're very short anyway.

I've also dropped a couple of unnecessary steps from the Etcher instructions. If you're using Etcher you don't need to unzip your image first, or preformat the SD card, or eject it manually when you're done, that all happens automatically.

I've made the intro for the mac/windows/linux instructions all point back to the main readme, since I think it's likely there'll be quite a few existing direct links in to these pages, and I've tweaked their format and intro to be all consistent with one another.

### DIFF
--- a/installation/installing-images/README.md
+++ b/installation/installing-images/README.md
@@ -10,7 +10,7 @@ Official images for recommended operating systems are available to download from
 
 Alternative distributions are available from third-party vendors.
 
-After downloading the `.zip` file, unzip it to get the image file (`.img`) for writing to your SD card.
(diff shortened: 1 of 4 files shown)


brianjking:
Sorry, I got busy and forgot about this. I'll take a look at the PR and see if there's any improvements I can put forth. If not, thanks to @pimterry for finishing this up.

JamesH65:
Looks good to me, I'll forward to the copy editors for final once over.

AisforAstronaut:
Minor changes made. Pushing back for technical checking and conflict resolution.

PR #623, opened 2017-04-21
James h65 patch 1#158

### LINKED ISSUE #158: Document the bcm_host helper functions
For programs wanting to poke peripheral addresses directly (for lack of a better interface), they need to know where in the virtual memory map the peripheral bus segment has been placed. This used to be tied to the platform (or rather had the same address since day 0), now is variable between Pi1 and Pi2.

Perhaps we need a one-pager in the hardware section detailing how to use this api:

https://github.com/raspberrypi/firmware/blob/master/opt/vc/include/bcm_host.h

JamesH65:
Add a new peripherals address page

Fixes #158


LornaLynch:
Copy edit done: pushing back for final technical check.

PR #627, opened 2017-04-25
Jamesh65 patch 1#574

JamesH65:
Adds a screensaver page to fix issue 574


LornaLynch:
Copy edit done. Made a few tweaks to the text, so please make sure it is all still OK.

PR #636, opened 2017-05-08
Add images of connectors to monitor-connection.md page

gbaman:
In relation to long standing open issue https://github.com/raspberrypi/documentation/issues/41.


JamesH65:
I think rather than wait for perfection on this, it can go to copy edit as is, we can add any extra images later.

PR #637, opened 2017-05-08
Update backup.md

vikeri:
The capital `M` didn't work for me, so changed it to a lower case `m` which worked.

I'm on macOS Sierra


JamesH65:
Unfortunately, this would not work on Linux, so I have changed the PR to have a Mac specific line that explains to replace the 4M with 4m. @vikeri Are you happy with that?

vikeri:
Sorry didn't know that macOS and Linux had different notations. Looks good to me!

JamesH65:
@vikeri Cool, will push to the copy editors for a final check.

JamesH65:
Thanks for the PR BTW.

vikeri:
@JamesH65 Haha yeah, there are similar issues with grep regex. Np!

LornaLynch:
Tweaks made to copy. Pushing back for final check.

PR #652, opened 2017-05-25
Update wireless-cli.md

### LINKED ISSUE #650: Bug in documentation regarding wpa_supplicant psk entry?
The documentation under https://github.com/raspberrypi/documentation/blob/master/configuration/wireless/wireless-cli.md#adding-the-network-details-to-the-raspberry-pi states, that wpa_passphrase returns a config with unquoted "psk" value and states:

> Note that, although wpa_passphrase returns the psk value unquoted, the wpa_supplicant.conf file requires this value to be quoted, otherwise your Pi will not connect to your network.

This is not true for the raspian i setup today on my Raspberry PI Zero (2017-04-10-raspbian-jessie-lite.zip).

The pi did not connect to my WPA2 secured WiFi network when i added quotes, it only worked without quotes.

Maybe this is outdated information, i am just starting to work with raspbian so i don't know? I can provide a pull request with updated text if the documentation is wrong here.

Thanks!

JamesH65:
Fixes #650


LornaLynch:
I have this now

LornaLynch:
Copy looks OK - pushing back for checking.

PR #655, opened 2017-05-28
Update building.md

realies:
Making the cross-compilation tools installation steps slightly more obvious.

### DIFF
--- a/linux/kernel/building.md
+++ b/linux/kernel/building.md
@@ -60,18 +60,22 @@ You can either do this using VirtualBox (or VMWare) on Windows, or install it di
 
 ### Install toolchain
 
-Use the following command to install the toolchain:
+Use the following command to download the toolchain to the home folder:
 
 ```bash
-git clone https://github.com/raspberrypi/tools
+git clone https://github.com/raspberrypi/tools ~/
 ```
 
-Linking the toolchain to the root directory makes later command lines easier:
+Updating the $PATH environment variable makes the system aware of file locations needed for cross-compilation. On a 32-bit host system you can update and reload it using:
+```bash
+echo PATH=\$PATH:~/tools/arm-bcm2708/gcc-linaro-arm-linux-gnueabihf-raspbian/bin >> ~/.bashrc
+source ~/.bashrc
 ```
-sudo ln -rs ./tools/ /
+If you are on a 64-bit host system you should use:
+```bash
+echo PATH=\$PATH:~/tools/arm-bcm2708/gcc-linaro-arm-linux-gnueabihf-raspbian-x64/bin >> ~/.bashrc
+source ~/.bashrc
 ```
-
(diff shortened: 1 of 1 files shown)


JamesH65:
Sorry about delay, busy. 

I like the idea, could you add some information on the link command you use to explain the flags? I had to look them up, so adding would be of educational benefit, at least to me!

realies:
@JamesH65, it might make more sense to clone `https://github.com/raspberrypi/tools` to the root of the filesystem or the home folder and reflect that in $PATH instead of linking. Thoughts?

JamesH65:
I think cloning it to root is prone to making mistakes, so not sure what the best approach is here. @pelwell Do you have any thoughts on best practice here?

realies:
@JamesH65, why is cloning to root prone to making mistakes? The other option I suggested is to clone to `~/` and expand to an absolute path in $PATH.

JamesH65:
I'm just a bit wary of doing anything in root since you need to be sudo to do it, and you might clone over something you didn't mean to. Probably just me being paranoid though.

pelwell:
I am equally wary. Cross compilation increases the chances that the user won't have root/sudo access.

I would rather introduce a shell variable to store the place where the toolchain should go, then use that for the clone and in constructing the path.

realies:
@pelwell, the guide already contains `sudo`. I would think most people would fire up a VM just for the purpose of cross-compiling making them the owner. On the other side, I appreciate that the process can be completed without root permissions.

pelwell:
Of course - the escaping is just because of the shell evaluating the parameters of the "echo" command. Without it, the current value of `PATH` is substituted when the line is added to `.bashrc`.

It would be clearer and more effective to make the whole line an unevaluated string using single quotes (`'`). As written now, the `~` will be expanded to `/home/pi` which, although not wrong, probably isn't what you wanted.

realies:
@pelwell, I have tested the code before committing. The cross-compilation tools are cloned to `~/` which gets expanded before being passed to `git clone`. Since the `.bashrc` file would be setting up the environment for the currently logged user, expanding should not be a problem and would keep the absolute path list of locations consistent.

JamesH65:
Sorry about the delay. I'll push to copy edit, and if no-one comment here they don't like the change I will merge it.

realies:
Wondering why the process of updating the documentation is happening so slow.

pelwell:
Wondering why you think that is a good motivational strategy.

realies:
I was unaware that motivational strategy is what is needed. For the good of education, please be better than the speed of bureaucracy. If you require extra hands, let the people know.

JamesH65:
Once a pr is submitted, I do a technical check on it to ensure it it ok. Then it goes to our copy editors who will translate it from engineering speak to English. Then it comes back for another technical check to ensure copy editing hasn't broken it. Then it gets merged
 Every point in the chain can have delays. For example I have been on holiday, and have had higher priority work to do, so unable to look at docs for about a month. Copy editors also have other work to do. Hope that explains it. 

Things are actually a lot quicker nowadays....

PR #662, opened 2017-06-06
No root required for GPIO

Epic0ne:
Newer versions of Raspbian allow the `pi` user to access the GPIO pins without root privileges, so update the documentation accordingly.


Epic0ne:
Thank you very much @bennuttall !

bennuttall:
Thanks @Epic0ne

PR #666, opened 2017-06-18
ENC28J60 interrupt should be GPIO25 not GPIO12

motters:
Tested on a CM3, on a computer module IO V3 board.

### DIFF
--- a/hardware/computemodule/cm-peri-sw-guide.md
+++ b/hardware/computemodule/cm-peri-sw-guide.md
@@ -216,7 +216,7 @@ will return with the hardware clock time, and not an error.
 
 ##Example 2 - Attaching an ENC28J60 SPI Ethernet Controller on BANK0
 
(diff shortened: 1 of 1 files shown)


JamesH65:
@pelwell @popcornmix Can either of you guys comment on this? I don't know whether this is correct or not.  Failing that, who would know the answer?

pelwell:
The README entry for the overlay says:
```
pi@raspberrypi:~$ dtoverlay -h enc28j60
Name:   enc28j60

Info:   Overlay for the Microchip ENC28J60 Ethernet Controller on SPI0

Usage:  dtoverlay=enc28j60,<param>=<val>

Params: int_pin                 GPIO used for INT (default 25)

        speed                   SPI bus speed (default 12000000)
```
Amazingly, the overlay agrees:
```
	eth1_pins: eth1_pins {
		brcm,pins = <25>;
		brcm,function = <0>; /* in */
	}
```
So @motters is correct - the guide is wrong.

PR #677, opened 2017-06-26
add the html rake task to help deploy this documentation on personal …

wangzhizhou:
…website

### DIFF
--- a/Rakefile
+++ b/Rakefile
@@ -8,6 +8,22 @@ BUILD_DIR = "./_build"
 task :build => [:clean]
 task :test => [:build]
 task :default => [:test]
+task :html => [:build]
+
+desc "change markdown files to html files for website deploy"
+task :html do
+  #get markdown files in BUILD_DIR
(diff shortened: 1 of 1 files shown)


JamesH65:
@bennuttall Ben, I have no idea what this is about - have you any thoughts?

bennuttall:
Ah, well on the server we use a little PHP app to translate requests into content, grabbing the appropriate md file and rendering into the template. This doesn't include the template so it's not as useful to us (it would be nice if we could output html rather than rely on a PHP app, but it's not caused us any problems in the 3 years it's been running).

It might be useful to (a small subset of) contributors to be able to fire up an HTML (template-less) version, if they have the right tooling to do so. I don't see any harm in this but it's not something we really want to support. Let me have a chat with the web team and see what they think.

@wangzhizhou any thoughts to add? (thanks for the effort btw)

waveform80:
I did build a little Python script that did this (generated an offline copy of the docs) way back when (several years ago). I might still have it somewhere if it's wanted?

bennuttall:
Merged! Thanks

PR #678, opened 2017-06-26
Correct descriptions of some boot options

pelwell:
I think we should also move arm_64bit here from the Misc section, but that is more a matter of opinion so I've left it there for now.


JamesH65:
Pushed to copy edit.

PR #681, opened 2017-06-29
Addition of some information on using systemd to start scripts at boot

buxtonpaul:
Based on some discussion on the forums I have created some guidance on using systemd to start scripts as once configured properly is a better way of handling services. My first attempt at contributing to things :-)

### DIFF
--- a/linux/usage/systemd.md
+++ b/linux/usage/systemd.md
@@ -25,6 +25,7 @@ User=pi
 WantedBy=multi-user.target
 ```
 So in this instance the service would run python3 from our working directory /home/pi/myscript which contains our python program to run myscript.py
+In this case we are running a python3 script but you could run anything you like here. 
 
 Copy this file into /lib/systemd/system as root e.g.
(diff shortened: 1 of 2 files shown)


JamesH65:
Sorry about delay, need to tech check this when I get the time.

linux/usage/systemd.md
```diff
@@ -0,0 +1,59 @@
...
+On your Pi, create a .service file for your service e.g.
+
+myscript.service
+
+```[Unit]
+Description=My service
+After=network.target
+
+[Service]
+ExecStart=/usr/bin/python3 -u main.py
```
JamesH65: One thing, you probably need to explain that service you are starting is, in this example, a Python program, but that you could have any sort of application start up here.

buxtonpaul: Agree, that makes sense.

JamesH65:
Tech check done, sending to copy edit

PR #684, opened 2017-07-01
corrects slightly misleading wording in config static IP section

TheDiveO:
* corrects mixing dhcpd (server!) with dhcpcd (client) with dhcpcd.
* removes misleading reference to "another" DHCP server, caused by mixing up dhcpd.


JamesH65:
Thanks, no need to copy edit this fixup I don't think, the entire document has already been CE'd anyway. I'll merge it.

PR #685, opened 2017-07-02
NOOBS: Partition name must be in all caps

mgaffney:
Document the need for the partition name to be in all caps when creating
the install SD card for NOOBS installation.

### DIFF
--- a/installation/noobs.md
+++ b/installation/noobs.md
@@ -35,7 +35,7 @@ For Windows users we recommend formatting your SD card using the SD Association'
 
 ##### Mac OS
 
-The [SD Association's Formatting Tool](https://www.sdcard.org/downloads/formatter_4/) is also available for Mac users, although the default OS X Disk Utility is also capable of formatting the entire disk. To do this, select the SD card volume and choose `Erase` with `MS-DOS` format.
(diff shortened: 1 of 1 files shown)


XECDesign:
Why would this be needed?

mgaffney:
Installation failed until I renamed the partition using all caps. Unfortunately, I don't have the error message(s) in front of me but they did mention the case of the partition name.

XECDesign:
I'll double check whether the presence and case of the label makes any difference when using something other than Disk Utility. This feel like a bit of a cargo cult fix.

XECDesign:
Just checked and it seems like NOOBS itself is fine with lowercase FAT32 labels. Don't have a Mac handy to check what's actually going on.

lurch:
@XECDesign Sounds like this is just another occurrence of https://github.com/raspberrypi/noobs/issues/174#issuecomment-51334441 which (despite lots of effort) I never got to the bottom of, and was never able to reproduce :-(

I suggest the issue here gets closed.

JamesH65:
Any consensus yet? Out of my knowledge arena!

XECDesign:
Simon has just formatted two cards with upper and lower case labels and they both work fine. I'd close this one, @JamesH65.

JamesH65:
@mgaffney It would be interesting to know what circumstances caused this to fail, and the errors messages, but they seem to not be available, so closing for the moment. If you feel there is still an issue please reopen.

PR #692, opened 2017-07-13
'sync' after running dd

XECDesign:
The forum user 'jojopi' has observed that some files are placed at the very end of the partition on the latest image.
https://www.raspberrypi.org/forums/viewtopic.php?f=63&t=187936#p1185337

Some users are taking out the card as soon as `dd` finishes and are finding that those files are corrupted.
https://www.raspberrypi.org/forums/viewtopic.php?f=91&t=187880#p1185702

The `conv=fdatasync` parameter could also be used, but `; sync` is shorter and arguably more intuitive.


JamesH65:
No copy edit required on this one, merging.

maxnet:
On Linux using `conv=fsync` is typically better because it is calls `int fsync()` on the file descriptor at the end and that is able to report any storage write errors back to the application, which then can display an error message to the user.

`sync` command calls `void sync()` which provides no feedback, so user will never know...

XECDesign:
Good call. I'll send an updated PR.

PR #695, opened 2017-07-17
installing-images: dd - use conv=fsync and conv=sync for Linux and Mac

### LINKED ISSUE #693: Mac dd command needs suffix conv=sync
If it doesn't have that then apt-get upgrades will fail. I am not enough of a Linux expert to just edit this, but I think you guys should do that, append it to all dd commands on this page.

This is what clued me in:

https://www.raspberrypi.org/forums/viewtopic.php?f=63&t=187936&p=1186098#p1186098

XECDesign:
See https://github.com/raspberrypi/documentation/issues/693 and https://github.com/raspberrypi/documentation/pull/692#issuecomment-315710731

Fixes #693


JamesH65:
No CE required

PR #696, opened 2017-07-20
corrected 3x  ../mailbox to /vcmailbox

hilschernetpi:


JamesH65:
Good spot, thanks - will push to copy edit as not sure that page has had a full CE before. Might take a few days to get through.

PR #701, opened 2017-08-01
Missing word

Charliemowood:


JamesH65:
You are right, missing word, but you have added the word in the wrong place!! Move it a word back and I'll merge it!

Charliemowood:
Hi James, good catch. Should be alright now :)

PR #705, opened 2017-08-04
linux: fix statement that rpiboot "completes" (it doesn't)

knitzsche:
And add tip to use dmesg to check it the new device appears


ghollingworth:
This is no longer required, the rpiboot has been fixed to complete

PR #706, opened 2017-08-07
Update wpa_supplicant

lne1030:
The default /etc/wpa_supplicant/wpa_supplicant.conf will cause error. Check the [post](https://www.raspberrypi.org/forums/viewtopic.php?p=326151) 
> Failed to connect to non-global ctrl_ifname: wlan0  error: No such file or directory.

### DIFF
--- a/configuration/wireless/wireless-cli.md
+++ b/configuration/wireless/wireless-cli.md
@@ -17,6 +17,16 @@ Open the `wpa-supplicant` configuration file in nano:
 
 `sudo nano /etc/wpa_supplicant/wpa_supplicant.conf`  
 
+Update the second line from
+```
+ctrl_interface=DIR=/var/run/wpa_supplicant GROUP=netdev
+```
+to
+```
+ctrl_interface=/var/run/wpa_supplicant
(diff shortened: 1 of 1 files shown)


XECDesign:
@lne1030 Does this apply only to Arch or Raspbian also? I haven't seen any reports of problems with the default wpa_supplicant.conf in Raspbian and it works just fine for me.

lne1030:
This for Raspbian.

Yesterday I upgraded my Raspberry Pi 3 to latest Raspbian Jessie Lite  July 2017 and got this problem. 
It's weird that no reports of this issue but it really exists on my Raspberry Pi 3 and Raspbian Jessie Lite  July 2017.

XECDesign:
Thanks, I'll take a closer look.

XECDesign:
Works as expected without this change. The existing syntax is directly from the wpa_supplicant.conf manpage.

lne1030:
Sorry, my mistake. The default /etc/wpa_supplicant/wpa_supplicant.conf  is ok.

XECDesign:
Thanks for confirming

PR #707, opened 2017-08-08
VNC password command-line config instructions

### LINKED ISSUE #674: Need command-line methods for Zero
On the page at https://www.raspberrypi.org/documentation/remote-access/vnc/ we see this paragraph:

> If you're connecting from a non-RealVNC Viewer app, you'll first need to downgrade VNC Server's authentication scheme, specify a password unique to VNC Server, and then enter that instead. To do this, open the VNC Server dialog on your Raspberry Pi, select Menu > Options > Security, and choose VNC password from the Authentication dropdown.

This cannot be done on a headless Zero. What file contains the VNC Server's authentication settings?  Can this be done with a little 'sudo nano' editing?  There seems to be nothing in the default /etc/vnc/config or /etc/vnc/config.d/common of use here.

Also worth a mention, the MacOS built-in VNC viewer is a non-RealVNC Viewer, so this is a pretty typical scenario. Mentioning MacOS in that sentence may help when googling.

minkustree:
Added command-line instructions for configuring a VNC password for
authentication.

Fixes #674


JamesH65:
Pushed to copy edit.

PR #709, opened 2017-08-08
correcting command for upgrading packages

xXP4trickXx:


XECDesign:
Could you please explain why this is needed?

JamesH65:
@xXP4trickXx As @XECDesign says, can you explain why this change is required. I've always used dist-upgrade.

JamesH65:
Closing as no reply from OP. Please reopen if you think this PR is necessary and give reasons.

PR #722, opened 2017-09-02
Typo?

eugf:
mon --> moy (for "month of year")


JamesH65:
merged, thanks.

bennuttall:
This is not a typo. That's pasted from crontab. It was correct before.

JamesH65:
It doesn't matter, it's only some text to describe the order isn't it? The typo correct a fairly heinous inconsistency in the docs.

JamesH65:
Ah, actually its the line above that is wrong, it uses "month of year" instead of just month, which is how its described in the crontab. Which is a bit bad but its probably been like that for years. Why use day of week, day of year, then not month of year. Oh well.

PR #726, opened 2017-09-08
Note multiple partitions may need to be unmounted

janicki:
With the proposed changed, the instructions then worked for me.


JamesH65:
I have no Mac to test this one, so will have to assume that the update does actually work! I'll send this to copy editing with a view to merging in the near future.

PR #739, opened 2017-10-04
Doc: update bootcode.bin link to current

macmpi:
previous link to (old) next branch has quite old version now.
`bootcode.bin` from current `master` works great.


JamesH65:
@ghollingworth Gordon, not sure where the latest version is likely to be...any comment?

PR #740, opened 2017-10-06
Using the Fing app to get Raspi's IP adress

Lukas-J:


bennuttall:
Useful suggestion, thanks.

I think maybe it should go under a heading of "more tools" (maybe just the links without the text), or at least underneath nmap.

Lukas-J:
I think, that the Fing app should be a the position I put it, because as a beginner, I don't want the first problems before I started to use my Raspi. I know it from my own experience.

JamesH65:
I think Fing should have its own section, BUT that it should be after the nmap section. The title should also be improved to show its a phone based approach, so people can find the section easily. There's also a few typos/misspellings in the text. Fix up those and I will push it to copy edit.

Lukas-J:
Hi, unfortunately I am not a native speaker. Please could someone else improve the title and the typos and misspellings in the text.
Thanks.

JamesH65:
OK, will do it.

JamesH65:
OK, have fixed up any issues i can see and pushed to copy edit.

PR #743, opened 2017-10-11
Fix logic-typo

lurch:
Change "If no SD card is found or `bootcode.bin` is found," to "If no SD card is found or `bootcode.bin` is not found,"


JamesH65:
@ghollingworth  Gordon, I just wanted to check that this is the now the correct logic - I am unfamiliar...

PR #744, opened 2017-10-12
Add note about using 64GB SD cards with NOOBS

lurch:
To try and prevent users getting caught out by the SDXC / exFAT "problem".

### DIFF
--- a/installation/sd-cards.md
+++ b/installation/sd-cards.md
@@ -4,7 +4,7 @@ The Raspberry Pi should work with any compatible SD card, although there are som
 
 ##SD card size (capacity). 
 
-For installation of NOOBS, the minimum recommended card size is 8GB. For image installations we recommend a minimum of 4GB. some distributions, specifically LibreELEC and Arch, can run on much smaller cards.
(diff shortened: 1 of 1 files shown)


lurch:
(see also https://github.com/raspberrypilearning/software-guide/issues/13 )

JamesH65:
Made a change to the default size of Rasbpian as well. It now requires 8GB. Read OK?

lurch:
I wonder if "installation of NOOBS and the full version of Raspbian" should be reworded as "installation of NOOBS or the image installation of Raspbian" ?
Otherwise it might sound like it's only the _combination_ of NOOBS and Raspbian that needs 8GB ;)

And on the subject of NOOBS and SD-card sizes: there's some other page(s) which talks about the pre-installed version of NOOBS on an 8GB card - seems like most (all?) vendors are now supplying pre-installed NOOBS on 16GB cards?

JamesH65:
No idea about NOBS, never use it. But rewording of paragraph sounds fine. Do you want to do it? I'll push to copy edit as and when.

lurch:
Done.

PR #747, opened 2017-10-21
Add link to elinux BCM2835_datasheet_errata

MaPePeR:
It just took me a few days and lending a frequency counter from a friend to figure out, that the [fractional divider for a clock needs to be specified with base 4096 instead of 1024](https://elinux.org/BCM2835_datasheet_errata#p105_table), how it says in the PDF file.

Only to then find out that this has been already documented somewhere.

I think there should at least be a link to that page.
It would be even better if it would be possible to add information like this as PDF-Annotations to the file that is hosted on the website.


JamesH65:
@jimbojr @ghollingworth Should we produce our own  errata sheet rather than rely on eLinux?

JamesH65:
OK, have made some minor changes - now passing to the copy editors.

JaninaPi:
Checked, looks fine.

PR #754, opened 2017-10-24
clarify how to enable ssh

quixand:


JamesH65:
Looks good - pushing to copy editors. May take a few days to work its way through the system.

JamesH65:
Damn, hit wrong button.

PR #758, opened 2017-11-02
Fix markdown formatting for raspberrypi.org systemd documentation

Conroman16:
The documentation on https://www.raspberrypi.org/documentation/linux/usage/systemd.md seems to be messed and not rendering the markdown properly because there isn't a new line after the first codeblock opening tag (\`\`\`) symbols.  This tweak should fix it


bennuttall:
Thanks :+1:

Conroman16:
My pleasure!

PR #769, opened 2017-11-17
Update sd-cards.md

semicolo:


JamesH65:
Good spot, thanks!

lurch:
@JamesH65  I've been wondering for a while whether this section should be reworded, given that none of the Pis currently being sold have TP1 and TP2 test-pads?  (unfortunately I'm too busy right now to PR anything myself, but I can add an issue if you think that'd be helpful?)

semicolo:
What about pins 2(or 4) and 6 of the header to check the voltage?

JamesH65:
Hmm, that is a bit outdated. I'll try and find time to update it, unless someone else volunteers.

PR #771, opened 2017-11-22
Improvements on the passwordless ssh tutorial

tecaspt:
I've never been able to setup passwordless ssh with the provided instructions on several Raspbian installs. It's always missing the step I added. Today, upon setting up a new machine, I remembered this tutorial and decided to try and correct it.

### DIFF
--- a/remote-access/ssh/passwordless.md
+++ b/remote-access/ssh/passwordless.md
@@ -67,7 +67,7 @@ To copy your public key to your Raspberry Pi, use the following command to appen
 cat ~/.ssh/id_rsa.pub | ssh <USERNAME>@<IP-ADDRESS> 'cat >> .ssh/authorized_keys'
 ```
 
-Now you'll have to edit /etc/ssh/sshd_config for it to accept connections via this method. Open the file and find the following parameter:
+Now you'll have to edit /etc/ssh/sshd_config on your Pi for it to accept connections via this method. Open the file and find the following parameter:
 ```
 #PasswordAuthentication yes
 ```
@@ -77,7 +77,7 @@ Replace it with
 PasswordAuthentication no
 ```
 
-Now restart the SSH service
+Then restart the SSH service
 ```
 sudo /etc/init.d/ssh restart
 ```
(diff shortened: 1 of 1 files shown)


texadactyl:
I take an approach that only uses ssh commands.

Goal: userA@MachineA ==ssh==> userB@MachineB

If you wish to login to the **root** on MachineB, 
then make sure that MachineB /etc/ssh/sshd_config has the following:

     PermitRootLogin yes

and then restart the sshd daemon on MachineB (or reboot it).

While logged into userA@MachineA, do the following:

     cd $HOME
     ssh-keygen # Be sure to use a nil password!
     ssh-add
     ssh-copy-id userB@MachineB

Test:

     ssh userB@MachineB

Note that If the ssh-add step above fails, ssh-agent is probably not running:

     eval `ssh-agent -s`
     ssh-add # Try again

tecaspt:
That's a possibility too. But that's up to folks on the Raspberry Pi website to see if they want a complete overhaul of the tutorial... :)

bennuttall:
Thanks for the contribution.

This step is not necessary - I do this all the time - if you have an SSH key, you can get in without a password, but if you don't, it asks for a password. If you set `PasswordAuthentication no` then you can only get in with an SSH key (no password prompt is available), but it's not a requirement for passwordless SSH.

Personally, I use `ssh-copy-id` to get the key on the Pi, but we decided not to include it here because it's not available on Mac. I would like to add it anyway (just saying it's available on Linux).

@tecaspt if you still have a problem, can you please post a screenshot or something to show what happens?

DG12:
Mac OS HighSierra has it.
>ls -l /Volumes/HighSierra/usr/bin/ssh-copy-id
-rwxr-xr-x  1 root  wheel  10658 Oct 24 21:19 /Volumes/HighSierra/usr/bin/ssh-copy-id*

JamesH65:
@bennuttall Since this does appeat to be present on Mac, can we do an update to include the copy option?

bennuttall:
Sure

PR #774, opened 2017-11-28
Alter commandline to discourage foot-shooting

mikemccabe:
Use /dev/sdz instead of /dev/sda to help linux machine users not destroy their root volumes.

Suggest lsblk.


burtyb:
Maybe using /dev/sdX would be more consistent with other docs (like the install docs).

mikemccabe:
Makes sense!  Changed.

JamesH65:
Thanks, pushed to copy edit.

PR #778, opened 2017-12-01
Added extra remote access service to the list

dimraft:
I want to add an additional remote service to the list called PiTunnel (www.pitunnel.com). It allows remote access to Raspberry Pi devices for Terminal, Device Monitoring and also Custom Network Tunnels to your project's web interfaces.
I think we should add to the list. Could you please accept my changes?

### DIFF
--- a/remote-access/access-over-Internet/README.md
+++ b/remote-access/access-over-Internet/README.md
@@ -10,6 +10,7 @@ One method is to set up port forwarding on your router. To do this, you must cha
 
 Rather than using port forwarding, there are a number of alternative online services available. These provide varying levels of functionality - see their websites for more details.
 
+- [PiTunnel](https://www.pitunnel.com)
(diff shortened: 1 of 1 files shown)


JamesH65:
@bennuttall Any comment? I'm reluctant to add too many third party commercial references here, but this does have a free option.

dimraft:
It has quite a good free option, is specifically for RaspberryPi where as a number of the others listed are not, and I think it’s good to give users choice and increase competition.

bennuttall:
BOD?

JamesH65:
@bennuttall BOD? (sorry about the delay!)

bennuttall:
benefit of doubt

JamesH65:
Although our docs do link to some third-party products, at present, we aren't as a rule adding further such links.

PR #783, opened 2017-12-17
The Raspi-Config doc page doesn't match

DougieLawson:
The documentation page no longer matches the menu structure for the current version of the raspi-config tool.


DougieLawson:
https://github.com/raspberrypi/documentation/blob/master/configuration/images/raspi-config.png needs to change as well.

JamesH65:
Pushed the text to copy edit - need to see who can change the PNG. @bennuttall ??

PR #787, opened 2017-12-27
mDNS method for discovering IP address

pietrodn:
One of the easiest ways to determine the IP address of a Raspberry Pi on the LAN is by pinging the `raspberrypi.local` hostname.
This method does not require to install additional software and is very simple, so I thought it was worth mentioning.


bennuttall:
LGTM. We could perhaps do with a note about that being the PI's hostname, and that all Pis running Raspbian will have the same one by default.

pietrodn:
I just updated the PR with a note about that. :-)

bennuttall:
Thanks! Happy for this to be merged, just needs a copy check.

PR #789, opened 2017-12-29
Update README.md

jkboxomine:
Revised a few editorial typos. 
By the way, is it correct that VideoCore IV runs at 400 MHz? It says differently on Wikipedia. 
 (https://en.wikipedia.org/wiki/VideoCore)


JamesH65:
Thanks.

400Mhz for this version of the chip is correct AFAIK.It was 250 on the 2835/6 IIRC.

jkboxomine:
Thank you.
I could found pages with more detailed information indicating 400MHz/300MHz on GPU.
https://www.raspberrypi.org/blog/raspberry-pi-3-on-sale/
https://en.wikipedia.org/wiki/Raspberry_Pi

PR #790, opened 2017-12-30
Update wireless-cli.md

ribbox:
The instruction to create a wpa_supplicant need to be updated to reflect changes in Stretch. 

The following need to be added to the start of the file;

```
country=US
ctrl_interface=DIR=/var/run/wpa_supplicant GROUP=netdev
update_config=1
```

Where country code relates to the country you use the RPi in.

### DIFF
--- a/configuration/wireless/wireless-cli.md
+++ b/configuration/wireless/wireless-cli.md
@@ -17,7 +17,16 @@ Open the `wpa-supplicant` configuration file in nano:
 
 `sudo nano /etc/wpa_supplicant/wpa_supplicant.conf`  
 
-Go to the bottom of the file and add the following:   
+For RASPBIAN STRETCH only Go to the top of the file and add the following:   
+```
+country=US
+ctrl_interface=DIR=/var/run/wpa_supplicant GROUP=netdev
+update_config=1
+```
+
+Where country is the code the RPi is being used in.
+
+Then below add the following:
 ```
 network={
     ssid="testing"


bennuttall:
This is an important point, but I don't think this is Stretch only.

XECDesign:
Under Jessie, wpa_supplicant would still run, but applications which communicate with wpa_supplicant wouldn't work. Under Stretch, wpa_supplicant doesn't start if dhcpcd can't communicate with it. So yes, you should have a valid wpa_supplicant.conf, regardless of whether you're running Jessie or Stretch.

I think the documentation says 'append' or 'add' the network block to the existing wpa_supplicant.conf, rather than replace wpa_supplicant.conf with just the network block. However, enough people try to overwrite the existing wpa_supplicant.conf that it should be made clearer.

JamesH65:
Not sure exactly what to do here, any one have any comments?

XECDesign:
I could send a PR which would have the full wpa_supplicant.conf and explain what the lines are and why they're required.

JamesH65:
That would be splendid!

XECDesign:
Since this page was written, support for this has been added to raspi-config.

Do we need to describe how to do everything manually with all of the details or can we just explain how to use raspi-config to set wifi-country and connect to wifi? Otherwise this could turn into a wpa_supplicant.conf manpage.

JamesH65:
Hmm, not sure what to do with this now that raspi-config has the required options.

mingaldrichgan:
Came here to +1 this since this is a crucial piece missing from the [headless setup documentation](https://www.raspberrypi.org/documentation/configuration/wireless/headless.md).

JamesH65:
This has been fixed by a different change - but thanks for the heads up.

PR #792, opened 2018-01-01
Fix typos in sshd option names

Haprog:
The two options  
`ChallengeResponseAuthentication` and `PasswordAuthentication`  
were incorrectly written as  
`ChallengeResponseAuthentification` and `PasswordAuthentification`  
(notice the extra "fi" in "Authentication").

Reference:
https://www.ssh.com/ssh/sshd_config/
https://linux.die.net/man/5/sshd_config


JamesH65:
Thanks.

PR #812, opened 2018-01-29
Update minimal-cm-dt-blob.dts for CM3

6by9:
It was slightly too cut down (copy/pasted from CM1), and
could result in an unstable system due to no SMPS control.

Copy/paste in the required bits from the main dt-blob.dts


6by9:
https://www.raspberrypi.org/forums/viewtopic.php?f=98&t=203919

PR #820, opened 2018-02-09
setup: readme: Fix typo in file README.md

miyatsu:
Relpace "mice" with "mouse".

Signed-off-by: Ding Tao <miyatsu@qq.com>


JamesH65:
We are using the plural of keyboards in that sentence, so really we should be using the plural of mouse (mice) as well I would have thought, so I am inclined to drop this PR. Thoughts anyone?

PR #829, opened 2018-02-21
initramfs: Document "followkernel" option to initramfs

kees:
The important "followkernel" option was not documented.

### DIFF
--- a/configuration/config-txt/boot.md
+++ b/configuration/config-txt/boot.md
@@ -51,7 +51,7 @@ Set `kernel_old` to `1` to load the kernel to the memory address `0x0`.
 
 ## initramfs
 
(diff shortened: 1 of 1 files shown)


JamesH65:
@pelwell Phil, is this something you can confirm or deny? I'm not familiar with the this stuff.

pelwell:
"followkernel" is indeed a magic address (equivalant to 0), which means "load the initramfs to any address that will work". Kernel compression makes it hard to choose an address immediately after the kernel, so instead we load other components (device tree, initramfs) to a high address that is safe for the platform.

Remove the work "immediately" and I'm happy with the change.

JamesH65:
@kees If you can update to reflect @pelwell comment, I'll merge this.

kees:
Updated! (And included the note about '0' being the same.)

PR #830, opened 2018-02-21
misc: Document arm_control and that it shouldn't be used

kees:
Include this for completeness and provide a note on avoiding it.


JamesH65:
Not sure of the necessity of this. Why document something that you shouldn't be using and most people wouldn't know existed if it was in the docs.

kees:
My rationale is that it is mentioned in the forums and in countless blogs about how to enable 64-bit booting. I think it makes sense to provide an "official" declaration about this existing and the better option to use, in case people look it up.

PR #831, opened 2018-02-21
kernel: Note that the image must be uncompressed

kees:
Image.gz is not supported. Note that.


JamesH65:
Whilst I am not convinced this is strictly necessary - the img suffix should be adequate, I'll push it just make things really clear.

kees:
Is it possible to load compressed images? Is it controlled by the filename suffix?

lurch:
I guess that's a question for @popcornmix ?

PR #832, opened 2018-02-22
Don't use vendor location for local service

kolewu:
https://www.raspberrypi.org/documentation/linux/usage/systemd.md

`/lib/systemd` is only meant to be used by vendor provided packages and should not be used for local units.

See
* [systemd.unit :: Unit File Load Path](https://www.freedesktop.org/software/systemd/man/systemd.unit.html#Unit%20File%20Load%20Path)
* [file-hierarchy](https://www.freedesktop.org/software/systemd/man/file-hierarchy.html#)


JamesH65:
Thanks for the PR and thanks @XECDesign for checking.

PR #834, opened 2018-02-23
Custom systemd config should be in /etc/systemd

eloydegen:


JamesH65:
already fixed by #832

PR #836, opened 2018-02-26
Update "SSH USING iOS" page adding a new client

ColdGrub1384:
Hey, I just added an SSH and SFTP client to the "SSH USING iOS" page. It's available on the AppStore for free and it's open source


JamesH65:
@bennuttall Ideas on whether we should accept this?

JamesH65:
Although our docs do link to some third-party products, at present, we aren't as a rule adding further such links.

PR #837, opened 2018-02-27
Fixed typo

8408323:
Fixed a typo in the dimension.  The display is 110.76 mm, not 112.76 mm on the mechanical drawing and in reality.


JamesH65:
Thanks - good spot.

PR #852, opened 2018-03-14
Update Pi 3B+ conformity URL

grega:
To match the hosted PDF URL.


JamesH65:
Sorry about delay - travelling in to office.

PR #858, opened 2018-03-18
Add oflag=sync to the dd command line

fwyzard:
Using only `status=progress conv=fsync` will result in a seemingly very fast copy, much faster than the write speed of the SD card, followed by a long delay as the data is actually written to the SD card.  
Using `status=progress conv=fsync oflag=sync` will show a slower progress, roughly matching the write speed of the SD card, but once it is done the command will complete immediately.  

I think the second approach is more intuitive and less likely to cause users to think the copy is hung.


JamesH65:
Hmm, nice idea. Once I've confirmed it's all OK will merge.

fwyzard:
In fact, I suspect `conv=fsync` can be dropped altogether if using `oflag=sync`,

And by the way, `dd(1)` suggests also `oflag=direct`, but I haven't tried it.

lurch:
@fwyzard As you've noted, this is (apparently) only an issue when using the `status=progress` option in newer versions of `dd`. So perhaps this `oflag=sync` should only be used with the last `dd` command, and not with all three `dd` commands as this PR is currently doing?

(from what you're describing, it sounds like the `conv=fsync` does one big file-sync at the end, but the `oflag=sync` does lots of smaller file-syncs as the file is being written? Note that I've not tested any of these options myself)

fwyzard:
OK.

Though it may be harder to explain why one uses different flags when showing the progress.

aallan:

aallan:

PR #862, opened 2018-03-21
Changed 'useradd' command to 'adduser'

### LINKED ISSUE #779: security.md: adduser should be used instead of useradd
[Link to the page to which the issue relates - in "Changing your username section"](https://www.raspberrypi.org/documentation/configuration/security.md)

The command "useradd" is deprecated and "adduser" should be used whenever it's possible.

The use of useradd brings undesirable consequences like sh being the login shell instead of bash, these problems are avoided if adduser is used instead.

From adduser manual page:
>        They  [adduser and addgroup] are  friendlier  front ends to the low level
>        tools like useradd, groupadd and usermod programs, by default  choosing
>        Debian  policy conformant UID and GID values, creating a home directory
>        with skeletal configuration, running a custom script,  and  other  fea-
>        tures.

From useradd manual page:
>        useradd is a low level utility for adding users. On Debian,
>        administrators should usually use adduser(8) instead.

[Link to the proposed change](https://github.com/raspberrypi/documentation/commit/3814dfeb83ce55e856729315ee81857f66cb2530)

MaxVMH:
Proposed change to fix #779 
I removed the instructions to change the password since 'adduser' asks to enter a password by itself.

### DIFF
--- a/configuration/security.md
+++ b/configuration/security.md
@@ -33,14 +33,10 @@ You can, of course, make your Raspberry Pi even more secure by also changing you
 To add a new user with the same permissions as the `pi` user:
 
 ```
-sudo useradd -m fred -G sudo
+sudo adduser fred && sudo adduser fred sudo
 ```
 
-This adds a new user called `fred`, creates a home folder, and adds the user to the `sudo` group. You now need to set a password for the new user:
-
-```
-sudo passwd fred
-```
+This adds a new user called `fred`, creates a home folder, and adds the user to the `sudo` group.
 
 Log out and log back with the new account details. Check your permissions are in place (i.e. you can sudo) by trying the following.
 


configuration/security.md
```diff
@@ -33,14 +33,10 @@ You can, of course, make your Raspberry Pi even more secure by also changing you
 To add a new user with the same permissions as the `pi` user:
 
 ```
-sudo useradd -m fred -G sudo
+sudo adduser fred && sudo adduser fred sudo
 ```
```
JamesH65: Should that second part read addgrp?

MaxVMH: TBH, I have no idea how to use addgrp in combination with adduser.

'sudo adduser fred && sudo adduser fred sudo' works fine. It creates a new user and adds it to the sudo group.

I tried 'sudo adduser fred && sudo addgrp sudo', it creates a new user but returns 'sudo: addgrp: command not found'.

'sudo adduser fred2 addgrp sudo' returns 'adduser: Only one or two names allowed.'

MaxVMH: From https://manpages.debian.org/jessie/adduser/adduser.8.en.html: 
In synopsis:
'adduser [options] user group'

Further in the document:
'Add an existing user to an existing group
If called with two non-option arguments, adduser will add an existing user to an existing group.'

bennuttall: I'm just testing this - will push it soon!

PR #867, opened 2018-03-26
Add notes about distro upgrades

cashpipeplusplus:
I found this page while looking for instructions on upgrading to stretch.  This fixes the "dist-upgrade" command to simply "upgrade" so as not to mislead the reader into thinking it's a distro upgrade.  This also adds new heading with a link to an example of how to upgrade to a new version of Raspbian.

### DIFF
--- a/raspbian/updating.md
+++ b/raspbian/updating.md
@@ -13,7 +13,7 @@ sudo apt-get update
 Next, **upgrade** all your installed packages to their latest versions with the command:
 
 ```bash
-sudo apt-get dist-upgrade
+sudo apt-get upgrade
 ```
 
 Generally speaking, doing this regularly will keep your installation up to date, in that it will be equivalent to the latest released image available from [raspberrypi.org/downloads](https://www.raspberrypi.org/downloads/).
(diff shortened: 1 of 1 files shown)


XECDesign:
`dist-upgrade` is the correct command to be using.

lurch:
@cashpipeplusplus See also #368 for explanations of why Raspbian updates work differently to Debian updates.

cashpipeplusplus:
Any objections to keeping the "Upgrading to a newer version of Raspbian" section?  Either as a link to that blog post or with explicit instructions and a warning that the in-place upgrade is not supported?

JamesH65:
Hmm, that probably a useful new section, better to pull in the stuff from the blog to make it implicit. Anyone know if those instructions (https://www.raspberrypi.org/blog/raspbian-stretch/) need updating in any way to work with the latest release? If no changes, I'll just do a quick C&P and incorporate them.

XECDesign:
As long as it's clear that it won't 'just work' for everybody. You generally need to know what you're doing when upgrading releases. Debian's own instructions on upgrading to stretch are quite lengthy:
https://www.debian.org/releases/stable/armhf/release-notes/ch-upgrading.en.html

PR #876, opened 2018-04-06
Add information about rfkill blocking the wifi scan.

robindegen:
The rfkill application blocks enabling wlan0, which also makes scanning not work.

### DIFF
--- a/configuration/wireless/wireless-cli.md
+++ b/configuration/wireless/wireless-cli.md
@@ -11,6 +11,13 @@ To scan for WiFi networks, use the command `sudo iwlist wlan0 scan`. This will l
 
 1. 'IE: IEEE 802.11i/WPA2 Version 1' is the authentication used. In this case it's WPA2, the newer and more secure wireless standard which replaces WPA. This guide should work for WPA or WPA2, but may not work for WPA2 enterprise. For WEP hex keys, see the last example [here](http://www.freebsd.org/cgi/man.cgi?query=wpa_supplicant.conf&sektion=5&apropos=0&manpath=NetBSD+6.1.5). You'll also need the password for the wireless network. For most home routers, this is found on a sticker on the back of the router. The ESSID (ssid) for the examples below is `testing` and the password (psk) is `testingPassword`.
 
+If you receive an error saying that the interface doesn't support scanning, run the following commands and try again:
+
+```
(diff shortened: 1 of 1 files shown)


JamesH65:
Might be worth adding a little bit of explanatory text to explain what is actually going on here, and whether its  likely to make a difference.

robindegen:
@JamesH65 I literally installed the latest version of raspbian lite cleanly on a 3B+ and rfkill blocked it instantly. The wifi device simply won't come online until it's "unblocked" apparently. Maybe this is due to some regulations regarding RF, but I feel it shouldn't be blocked by default on a clean install.

What more do you think I should explain on this?

XECDesign:
The idea is that the user would not ignore the message they see when they log in telling them that wifi is disabled until they set the country, instead of overriding the mechanism until next boot. IMHO, the page should tell people to use raspi-config to set wifi country and connect to wifi. Maybe it could have a description of how to do it without raspi-config underneath.

It would probably be a good idea for raspi-config to set the wifi country when the user tries to connect to wifi as well.

JamesH65:
BTW, this is a regulatory requirement - we are not allowed to turn on Wifi on 5G without the country code being set.

robindegen:
@XECDesign Hmm I didn't know you had to set a country in raspi-config. That makes more sense to put in the documentation indeed. I just googled for the error message and it brought me to rfkill

It seems to have disabled it permanently though, because after I configured the wifi, it stays up, even after reboot. Is it expected to block again at reboot? Because it doesn't

XECDesign:
@robindegen it should disable wifi each time you boot until wifi country is set. What does your `/etc/wpa_supplicant.conf`looks like and what's the output of `sudo journalctl -u wifi-country`?

Also, out of curiosity and to maybe help us improve the mechanism, any idea why you didn't notice the message that tells you to set wifi country in raspi-config? It should show up on every login shell. Is it just because it appears in a block of text that people usually ignore?

JamesH65:
Add a section to cover setting the country code on the commandline here.

https://github.com/raspberrypi/documentation/pull/879

JamesH65:
Also the desktop version 

https://github.com/raspberrypi/documentation/pull/880

JamesH65:
Both PR's have now been merged so closing.

PR #880, opened 2018-04-10
Add country code setting information - desktop

JamesH65:


JaninaPi:
I've changed PIXEL to Raspberry Pi Desktop, because I think that's what we're calling it now (Simon Long will be able to confirm/deny).

JamesH65:
That's corrects, thanks.

PR #890, opened 2018-04-20
Add Jessie to Stretch upgrade instructions

JamesH65:
Copied from the blog post describing same.


JamesH65:
Thanks for the comment, it's now in the hands of our copy editor - may take a while, think she is on holiday in the nr future!

PR #894, opened 2018-04-24
Update conditional.md

JamesH65:
Added new models, and section on GPIO filtering

### DIFF
--- a/configuration/config-txt/conditional.md
+++ b/configuration/config-txt/conditional.md
@@ -22,7 +22,9 @@ Any settings below a `[pi2]` filter will only be applied to Pi 2 hardware. The `
     [pi1]
     [pi2]
     [pi3]
+    [pi3+]
     [pi0]
+    [pi0w]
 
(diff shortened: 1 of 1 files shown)


JamesH65:
@6by9 @pelwell Does this seem correct to you guys? If so I will merge it.

6by9:
Let me check. As noted in the forum thread I reverse engineered the gpio syntax from the code but haven't tried it.

pelwell:
The gpio section looks correct from reading the code, but I've not tried it myself.

It's worth mentioning that a Zero W will see the contents of `[pi0w]` _and_ `[pi0]`. Likewise, a 3B+ sees `[pi3+]` _and_ `[pi3]`. If you want a setting to apply only to Pi Zero or Pi 3B you need to follow it (order is important) with a setting in the `[pi0w]` or `[pi3+]` section that reverts it.

JamesH65:
Have added @pelwell comments almost verbatim.

6by9:
Confirmed that gpio works as described. Looks all good to me.

pelwell:
Agreed.

PR #898, opened 2018-04-29
Suggest running apt-get update before installing apache2

Pikamander2:
The `sudo apt-get install apache2 -y` command will fail on a fresh installation. Users will generally need to run `apt-get update` first.

### DIFF
--- a/remote-access/web-server/apache.md
+++ b/remote-access/web-server/apache.md
@@ -6,7 +6,13 @@ On its own, Apache can serve HTML files over HTTP, and with additional modules c
 
 ## Install Apache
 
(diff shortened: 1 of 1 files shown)


JamesH65:
I have no objections. @XECDesign, do we need to investigate why it doesn't work correctly after a fresh install?

XECDesign:
Just flashed a card with the latest image and apache was installable without running `apt-get update`.

There are a few possible explanations why OP had to run it, but they're not going to affect most users. In general, it's a good idea to run `sudo apt update` before installing things anyway.

JamesH65:
OK, thanks, will merge as is, since its a good idea to do it.

PR #899, opened 2018-04-29
Update systemd.md

jorijnsmit:
Two different filenames used which created confusion for me. Also the line about fully referencing the directory is redundant since we are using `WorkingDirectory`.


JamesH65:
Nice spot. Thanks for the PR.

PR #900, opened 2018-05-04
header H2 for Etcher

yahsevf:


bennuttall:
I see what you're suggesting but it doesn't quite work structurally. I'll merge it and adjust the copy.

PR #905, opened 2018-05-14
iptables should run as root

rayfoss:
youll get an obscure error about being unable to initialize the nat table if you dont use root.


lurch:
?  The instructions are suggesting adding that line to `/etc/rc.local` and rebooting. When you boot `/etc/rc.local` is always run by the root user, so there shouldn't be any need to add `sudo`?

@JamesH65 Perhaps the instructions need clarifying a bit?

JamesH65:
Clarify as in the ensure its clear that this is a file edit rather than entered on the command line? 

Pretty sure you shouldn't need sudo in rc.local.

rayfoss:
strange... I recall just typing on tty. okay.

PR #908, opened 2018-05-20
Update v2 max framerate according this github commit avaliable with M…

Hermann-SW:
…arch 2018 Raspbian Stretch:

https://github.com/Hexxeh/rpi-firmware/commit/f6f7e04d01f95b86633c1722a179a5fce426c24e
"firmware: imx219: Increase max frame rate at 640x480 to 200fps"

### DIFF
--- a/raspbian/applications/camera.md
+++ b/raspbian/applications/camera.md
@@ -299,7 +299,9 @@ Version 2.x (IMX219)
 |4|1640x1232|4:3|0.1-40fps|Full|2x2|
 |5|1640x922|16:9|0.1-40fps|Full|2x2|
 |6|1280x720|16:9|40-90fps|Partial|2x2|
-|7|640x480|4:3|40-200fps|Partial|2x2|
+|7|640x480|4:3|40-200fps<sup>1</sup>|Partial|2x2|
+
+<sup>1</sup>For frame rates over 120fps, it is necessary to turn off AEC using `-ex off` and AWB using `-awb off`. This will help maintain higher frame rates but result in lower quality images. We do not guarantee framerates over 120fps.
 
 ```
 	--camselect,	-cs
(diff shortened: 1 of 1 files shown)


JamesH65:
@6by9 Is this a valid change - i.e. do we get this using default software?

6by9:
It needs a comment to say that >120fps is only achievable with ```-ex off``` (and possibly ```-awb off```).
The sensor can frame that fast, but the tuner doesn't keep up.
AE certainly limits things. AWB is in a separate thread and therefore shouldn't block things, but I couldn't say for certain that it won't cause issues without trying it.

JamesH65:
I can put the caveats on a superscripted footnote i.e.

"For frame rates over 120fps, it is necessary to turn off  AEC using `-ex off` and AWB using `-awb off`. This will help maintain higher frame rates but result in lower quality images. "

I'll knock up a PR for review.

JamesH65:
OK, have updated the original PR, any comments?

raspbian/applications/camera.md
```diff
@@ -299,7 +299,9 @@ Version 2.x (IMX219)
 |4|1640x1232|4:3|0.1-40fps|Full|2x2|
 |5|1640x922|16:9|0.1-40fps|Full|2x2|
 |6|1280x720|16:9|40-90fps|Partial|2x2|
-|7|640x480|4:3|40-90fps|Partial|2x2|
+|7|640x480|4:3|40-200fps<sup>1</sup>|Partial|2x2|
+
+<sup>1</sup>For frame rates over 120fps, it is necessary to turn off AEC using `-ex off` and AWB using `-awb off`. This will help maintain higher frame rates but result in lower quality images. We do not guarantee framerates over 120fps.
```
6by9: I've just checked - AWB doesn't have an effect.

"For frame rates over 120fps, it is necessary to turn off automative exposure and gain control off using `-ex off`. Doing so should achieve the higher frame rates, but exposure time and gains will be set to fixed values as supplied by the user."

6by9:
Ship it.

Hermann-SW:
Thank you!

PR #917, opened 2018-06-07
Update linux.md - lsblk -p

littlebaka:
Added full path option


lurch:
On Ubuntu 14.04 `lsblk -p` says `lsblk: invalid option -- 'p'`
:-/

JamesH65:
How backwards in time do we want to remain compatible with things like Ubuntu? 14.04 is pretty old now, 16.10 seems to support it, and the Pi itself works with -p. 

The -p option does match up better with the docs.

lurch:
> How backwards in time do we want to remain compatible with things like Ubuntu?

\*shrug\* That's entirely up to you.

Brief aside: the laptop I bought a few years ago came with Ubuntu 14.04 pre-installed (it was the LTS release at the time), I've never bothered to upgrade it to a newer distro, and https://wiki.ubuntu.com/Releases says it's supported until April 2019.

JamesH65:
Not sure what to do with this one.  It does make the process a little clearer, but clearly not available if you go back a few years. Hmmm.

@LizUpton Liz, any thoughts on what we should do here?

JamesH65:
Since 14.04 is now out of LTS, I think this can be merged.

PR #925, opened 2018-06-24
Make the contents of the Linux section consistent

waveform80:
Currently systemd is missing from the Linux section, but is present in Linux/Usage, while Scripting is present in Linux/Usage but missing from Linux.


bennuttall:
Thanks @waveform80

PR #926, opened 2018-06-25
Fix documentation about access point mode

GuLinux:
I tried following the original documentation, and most of the time the wifi interface would go up, then down after some seconds.
I spent some time debugging this, and I finally discoverer that just putting a static IP address in `dhcpcd.conf` is not sufficient, as `wpa_supplicant` is still launched on that interface, putting it in managed (client) mode.

The simplest and most effective solution I found is to disable the `wpa_supplicant` hook in dhcpcd.

### DIFF
--- a/configuration/wireless/access-point.md
+++ b/configuration/wireless/access-point.md
@@ -35,6 +35,7 @@ Go to the end of the file and edit it so that it looks like the following:
 ```
 interface wlan0
     static ip_address=192.168.4.1/24
+    nohook wpa_supplicant
 
 ```
 Now restart the dhcpcd daemon and set up the new `wlan0` configuration:


JamesH65:
@SurferTim I'm out of the loop on the networking stuff - have you any comments?

SurferTim:
@JamesH65 I haven't had any trouble with the AP docs. I've tested RPi Zero W, Rpi3 and RPi3B+, and all work with the current documentation and Stretch 2018-04-18.
wpa_supplicant does start ok. If you want more than one wifi interface, you must split the wpa_supplicant.conf file. The wpa_supplicant.conf file for the AP interface must not have a network section in it. Split like:
/etc/wpa_supplicant/wpa_supplicant-wlan0.conf
/etc/wpa_supplicant/wpa_supplicant-wlan1.conf

GuLinux:
@SurferTim it might be a race condition. I tried on a fresh raspbian installation, and almost every time the interface starts, then stops.

But more to the point, `wpa_supplicant` actually *shouldn't* start.

The interface should be managed only by `hostapd`, not by `wpa_supplicant` (which, again, tries to set it in managed mode).

I guess what happens in your system is that `wpa_supplicant` starts *before* `hostapd`, hence it gets overridden by the latter.

SurferTim:
@GuLinux Yes, wpa_supplicant starts, and starts ok. The wifi interface for the AP MUST NOT have a network section in it. That will cause the interface to go into client mode causing hostapd to fail.

If you plan on using more than one wifi interface, you must split the wpa_supplicant.conf file as I mentioned above. I have tested this with two wifi devices (onboard and usb), and it works fine.

GuLinux:
I'll retest when I'm home, but I think sure wpa_supplicant shouldn't start for the interface where hostapd is running, that was my whole point (I remember it didn't start, when I was using previous raspbian without dhcpcd)

Now of course, if it starts, but there's no network configured, it won't try to connect to that network in managed mode, but why starting it at all, if it's not gonna do anything?

SurferTim:
It will work as long as you have no network section in it. If you stop the wpa_supplicant service, it makes using two wifi devices (like I do) very difficult. I have one device (wlan0) as the AP, and the other (wlan1) as the WAN client.

GuLinux:
The `nohook` setting applies per interface, same as `static ip_address`.
I've already tested this in a dual wlan system, with that config wpa_supplicant doesn't start for `wlan0` (the AP interface), but it correctly starts on `wlan1`

SurferTim:
Here is the thing: Why add complexity to an otherwise simple setup? Remove any network sections from wpa_supplicant.conf (or start with a fresh install as I recommend) you won't have a problem at all. If you want to use two wifi devices at this point, split the wpa_supplicant.conf file as I suggested above, ensuring there is no network section in the AP file.

GuLinux:
The "extra complexity" you're referring to is just an extra line in `dhcpcd.conf`. I doubt anyone would really be bothered by it, people usually do this things by copy/pasting these "code" sections, and if by any chance they forget the line, and they're on a single network setup, no harms done, it works exactly as it is now.
Plus, the intent of this extra line is pretty clear: "don't use wpa_supplicant on this interface".

On the contrary, I find the current approach more complex:
 - it relies on the user *not* having any network in `wpa_supplicant.conf`. Suppose I'm not setting up a dual network interface, but on a more common scenario I just want to setup my raspberry pi (ssh into it, download packages from network, etc), and *then* switch the same interface to AP mode at the next reboot after finishing configuring everything. In this case, I also need to remember to remove all networks from `wpa_supplicant.conf`, which seems to add more complexity to me.

- It relies on an undocumented (and possibly unintentional) behaviour of`wpa_supplicant`, which although improbable, might still change in the future.

GuLinux:
Oh, and of course, the current approach *also* adds the extra complexity of having the dual `wpa_supplicant` configuration (one of them just a placeholder).

SurferTim:
I don't care. Whatever the admins want to do. Bear in mind that the dhcpcd service could also change. The only constant is change. Post the recommended change and I'll test it.

GuLinux:
Of course everything can change, but an undocumented behaviour can change without any warning.
My point is: everything should be as clear as possible. Right now there's no mention about the dual `wpa_supplicant` setup, but even if this was clearly stated, there would be no explanation on *why* one of them should be empty. Basically, you're trying to do the same thing I'm doing (disable `wpa_supplicant` on the hostapd interface), but without *explicitely* disabling it, relying instead on the fact that an empty configuration will not reset the interface.

Anyway, the change is already posted, this is a PR...

SurferTim:
@JamesH65  It's good on a RPi3B+. I like it. An easy update.

JamesH65:
OK, I'll merge it. If we have any problems we can always revert, but it does sound like a decent change.

GuLinux:
Thanks

PR #927, opened 2018-06-25
Removing insecure examples of wpa_passphrase

domdetre:
The command `wpa_passphrase ssid password` should never be used or suggested to any noob, since shell will put the line into the history, from which the password can be easily recovered along with the ssid. wpa_passphrase is not designed to be used in that way, since it will ask for the password if not added as paramter, and that input will not be stored anywhere. The ability to be able to use commands with passwords only designed for scripts they SHOULD NOT BE USED in a terminal (where history of commands being logged).


bennuttall:
There are times when network passwords are not secret, i.e. in a public venue, or perhaps in the home.

We could add a note to point out that this command will be logged, and suggest they prepend a space to the command if they want to keep it out of the history.

JamesH65:
I'm inclined to change this to the proposed solution - it's no more typing, whereas adding stuff about adding a space at the start to avoid history just adds text and makes it all more confusing. Any other thoughts, or OK to push to copy edit?

domdetre:
basicaly a noob will think it is okay to have it like that, and unless someone tells them not to include password on the command line, they won't learn it. It is a bad practice to include passwords in terminal commands and that's fact. Noobs will think it is okay, while it isn't ever.
I agree with a solution to put a note that "it is a good practice to not include passwords on the command line because it's being appended to the history file". or something similar.

My proposal is not adding stuff like spaces at the start of history or anything like that. My proposal is about a general good practice every admin should be aware about. We the users with bigger knowledge than the noobs, when teaching the noobs, we have the responsibility to teach them to the good practices, and try to make them avoid bad practices.

My proposal is more like removing stuff to make them safe, instead of adding confusing stuffs (like space ow whatever). I still think we should teach the noobs to use wpa_passphrase without the password, and putting a note that it will ask for the testing password.

lurch:
On a side-note, if this "confusing stuffs" is aimed at "teaching the noobs" perhaps the bit about either using `sudo su` first _or_ using `sudo tee -a` is potentially a bit confusing, and it'd be better to just give one alternative rather than two? \*shrug\*

bennuttall:
Yeah the PR is fine :+1:

domdetre:
I understand it's not perfect. I guess putting a note there saying you shouldn't use password on terminal because of the history logging could be enough. 
But to be fully honest, I think the password can be simply removed without any comments as the wpa_passphrase asks for the password in a very clear way: `# reading passphrase from stdin`

configuration/wireless/wireless-cli.md
```diff
@@ -37,7 +37,7 @@ Note that the plain text version of the code is present, but commented out. You
...
+If you are using the `wpa_passphrase`–encrypted PSK, you can either copy and paste the encrypted PSK into the `wpa_supplicant.conf` file, or redirect the tools output to your configuration file by calling `wpa_passphrase "testing" >> /etc/wpa_supplicant/wpa_supplicant.conf` and then enter the password when asked. To do the latter, you will need the necessary administrative privileges. To get them, either change to `root` by executing `sudo su`, or you can use `wpa_passphrase "testing" | sudo tee -a /etc/wpa_supplicant/wpa_supplicant.conf > /dev/null` and give it the testing password when asked; this will append the passphrase without you having to change to `root`. Lastly, **make sure you use `>>`, or use `-a` with `tee`** — either will append text to an existing file. Using a single chevron `>`, or omitting `-a` when using `tee`, will erase all contents and **then** append the output to the specified file. Note that the redirection to `/dev/null` at the end of the second form simply prevents `tee` from **also** outputting to the screen (standard output).
```
JaninaPi: I edited this, but it's still not very readable. I suggest:

To use the `wpa_passphrase`–encrypted PSK, you can either copy and paste the encrypted PSK into the `wpa_supplicant.conf` file, or redirect the tools output to the configuration file in one of two ways:
- Either change to `root` by executing `sudo su`, then call `wpa_passphrase "testing" >> /etc/wpa_supplicant/wpa_supplicant.conf` and enter the password when asked
- Or use `wpa_passphrase "testing" | sudo tee -a /etc/wpa_supplicant/wpa_supplicant.conf > /dev/null` and give it the testing password when asked; the redirection to `/dev/null` prevents `tee` from **also** outputting to the screen (standard output)

If you want to use one of these two options, **make sure you use `>>`, or use `-a` with `tee`** — either will **append** text to an existing file. Using a single chevron `>`, or omitting `-a` when using `tee`, will erase all contents and **then** append the output to the specified file.

JaninaPi:
I've edited and also left a comment suggesting a version of line 40 that's more readable. If someone could check it and let me know whether it's OK, I'll add it in.

configuration/wireless/wireless-cli.md
JamesH65: That reads considerable better, so lets go with @JaninaPi  new version.

JaninaPi:
Have added that version in now.

configuration/wireless/wireless-cli.md
lurch: Should `the tools output` be `the tool's output`

PR #930, opened 2018-07-01
Update updating.md

kenmercer:
Reversing a change made by XECDesign two years ago
sudo apt-get update / upgrade should be run regularly, as stated, to upgrade packages but dist-upgrade is more aggressive and will upgrade the kernel.
see https://askubuntu.com/questions/81585/what-is-dist-upgrade-and-why-does-it-upgrade-more-than-upgrade
for full details

### DIFF
--- a/raspbian/updating.md
+++ b/raspbian/updating.md
@@ -13,7 +13,7 @@ sudo apt-get update
 Next, **upgrade** all your installed packages to their latest versions with the command:
 
 ```bash
-sudo apt-get dist-upgrade
+sudo apt-get upgrade
 ```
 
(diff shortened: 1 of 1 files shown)


XECDesign:
Our distribution of Raspbian is not Debian or Ubuntu. We sometimes change dependencies, which would mean that 'upgrade' would result in updates being held back.

lurch:
This has already been discussed (and closed as invalid) several times in the past. See https://github.com/raspberrypi/documentation/search?q=dist-upgrade&type=Issues for more details.

JamesH65:
Closing as a Won't Fix. This means that although the issue is acknowledged, it is felt that no further action is likely. This may be down to a bad benefit vs cost analysis, or it may actually not be possible to fix.

Issue #877, opened 2018-04-09
Electrical Characteristics missing details for VDD_IO=3.3V


6by9:
As noted in forum thread https://www.raspberrypi.org/forums/viewtopic.php?f=29&t=138215, the DC electrical characteristics quoted in the Compute Module datasheet are for VDD_IO at 1.8V and 2.7V. The Pi is running VDD_IO at 3.3V, so the details are missing.
Can we add those details?

And are there any details about the level of hysteresis (it's always enabled by the firmware with no supported mechanism to disable it, so it is the appropriate mode of operation to be quoting values against).

Ping @Roger-Thornton.

Roger-Thornton:
@mike-stimson will be characterising this at some point.

JamesH65:
Fixed. Now on the GPIO page https://www.raspberrypi.org/documentation/hardware/raspberrypi/gpio/README.md

lurch:
The `Output high current` section lists `VDD IO = 2.7V` twice - presumably the last one was supposed to be `VDD IO = 3.3V` ? :mag:

JamesH65:
I thought I had fixed that.....maybe that was in the PDF, anyway, good spot. I'll fix it today.

JamesH65:
Ah, already done. Thanks.

seamusdemora:
Perhaps I have missed this, but I cannot find a definition of `VDD IO`. I have searched for it in the [online documentation](https://www.raspberrypi.com/documentation/computers/raspberry-pi.html#voltage-specifications), the [Compute Module datasheet](https://www.raspberrypi.com/documentation/computers/compute-module.html#datasheets-and-schematics), and the PDF [Compute Module 4 Datasheet](https://datasheets.raspberrypi.com/cm4/cm4-datasheet.pdf). 

If I've missed it, please provide a link; if not, could someone add a brief note that defines `VDD IO`? 

Thank you

seamusdemora:
FWIW: I also followed the resolution of #1463 & #2081 to [these changes](https://github.com/raspberrypi/documentation/pull/2081/files). There is a change: 

>NOTE: On the Compute Module devices, it is possible to change the VDD IO from the standard 3V3. In this case, V~OL~ and V~OH~ will change according to the table on the linked page. 

[...]

PR #932, opened 2018-07-02
Fix typo

lurch:
See https://github.com/raspberrypi/documentation/issues/877#issuecomment-401955414
Updated using https://www.raspberrypi.org/documentation/hardware/computemodule/datasheets/rpi_DATA_CM_2p0.pdf as reference.


JamesH65:
Ah, you've already done it! Thanks!

Issue #929, opened 2018-06-30
SSH on Windows


LunarWatcher:
Specifically talking about [the Raspberry Pi docs][rpidoc] (emphasis mine):

> On Windows you will **need** to download an SSH client.

Windows 10 (at least, not tested on other versions) supports `ssh` like UNIX and Mac. See [this][1] and [this][powershell]

Also, [cygwin][cyg] can come with OpenSSH too. AFAIK, it works on Win7 and 8 as well. Which means, if OpenSSH is installed (which does come with Cygwin or since the fall creators update 2017), there's no need to install an SSH client. 

[1]: https://github.com/PowerShell/Win32-OpenSSH/wiki/Install-Win32-OpenSSH
[cyg]: https://superuser.com/a/1111661/623128
[powershell]: https://blogs.msdn.microsoft.com/powershell/2017/12/15/using-the-openssh-beta-in-windows-10-fall-creators-update-and-windows-server-1709/
[rpidoc]:  https://www.raspberrypi.org/documentation/remote-access/ssh/windows.md

JamesH65:
So there is a command line SSH install on my Win10 box, not sure whether its been there from the start or whether its been installed as a result of something else. I going to hazard a guess that if someone has installed OpenSSH then they are probably savvy enough to understand about SSH clients already, so only a slight change to the docs might be necessary i.e. "Depending on your version of windows, you may need to install a SSH client" or similar. We'll still recommend putty I think.

lurch:
Also worth noting that the `Install-Win32-OpenSSH` wiki link that @LunarWatcher provides above refers to installing the SSH *server*, which obviously isn't required if all you're doing is connecting to a Raspberry Pi!

JamesH65:
OK, will make some minor changes as I outlined above.

JamesH65:
Proposed change in #933

lurch:
@LunarWatcher #933 has been merged - are you happy for this to be closed?

bsimmo:
Just an fyi.
The April 2018 Update has the OpenSSH install by default.
Before that it was still in testing and had to be installed by yourself.
 
It should behave like on Linux from the commandline.  sftp and scp are there to.
So anyone on that version can just follow the standard versions.

Of course the Ubuntu Linux subsystem could also be used on Windows10 as another method


A universal-ish method would be to use Google Chrome and their Secure Shell Extension (as used/shown by them with the Google AIY products).

lurch:
[...]

PR #933, opened 2018-07-03
Update windows.md

JamesH65:
Slight change in SSH installation requirements


lurch:
I'd get rid of the first "may" (IMHO sounds a bit strange otherwise) but LGTM.

lurch:
@MattDuffyJr Not all Windows users will be running Windows 10 - that's why this PR says "Depending on the version of Windows you are using..." :wink:  More discussion on this in the linked #929

PR #941, opened 2018-07-11
[WIP] Add an index tree, close #827

### LINKED ISSUE #827: Add an index tree
As suggested in #801, we should add an index tree containing the docs index.

From #801:

> For the docs index tree, is it possible to genreat that automatically?

That's what I had in mind

> Woudl it also be possible to have metadata per page to act as a brief content descriptor that could be used in the index?

Probably not. I think it'll be possible to use the h1 from each page rather than the directory name, but getting a meaningful description of the page would be tricky. I think the page title should be sufficient.

I could either do it in PHP, and it'll purely live on the web server (not on GitHub), but will be generated and updated automatically, or I could do it in Python (probably easier) and the output can live on GitHub too - but that'll need re-running to keep it updated as the index changes (not very often tbh). Manual re-run wouldn't be EOTW but post-commit script would require some setup. I'll have a think. First: glossary.

bennuttall:
WIP - closes #827

### DIFF
--- a/index.md
+++ b/index.md
@@ -1,0 +1,69 @@
+# Raspberry Pi Documentation index
+
+- [Configuration](configuration)
+    - [config.txt](configuration/config-txt/)
+    - [Wireless](configuration/wireless/)
+- [FAQs](faqs/)
+- [Hardware](hardware/)
+    - camera
+        - mechanical
+        - schematics
+    - computemodule
+        - datasheets
+        - designdata
+        - schematics
+    - display
+    - industrial
+    - raspberrypi
+        - bcm2835
+        - bcm2836
+        - bcm2837
+        - bootmodes
+            - pxetools
+        - compliance
+        - dpi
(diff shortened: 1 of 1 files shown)


bennuttall:
@JamesH65 first draft of this is done - it is just the output of `tree -D` converted into a nested list with links. The titles could be improved as they're just the link names not in all lowercase. And we might be missing some pages which aren't README files in folders. Bit of a pain to automate as it's a little inconsistent.

https://github.com/raspberrypi/documentation/blob/827-index-tree/index.md

Not perfect but I think it's useful though! Certainly better than nothing.

JamesH65:
If there is stuff that is inconsistent, we should probably fix that first, because I reckon we ought to be able to automate this.

aallan:

aallan:
After consideration, I **think** this is superceeded by #1911 so closing.

PR #942, opened 2018-07-12
Using a proxy server

martinohanlon:
I was in a school yesterday helping to get their Raspberry Pi's connected to the internet via their proxy server.

I thought it would be a good idea to include in our documentation.


bennuttall:
LGTM - @JamesH65 you think it's ok to go to copy edit?

JamesH65:
Very nice, thanks Martin.

PR #947, opened 2018-07-20
James h65 segment filenames 1

JamesH65:
Update to camera docs to include the new segment time based naming


JamesH65:
Ah, got some unnnecessary changes in here. Abandoning.

PR #964, opened 2018-08-12
Prefixed iptables-restore with sudo

takz012:
When running this on a new Pi Zero the above line would not step through with error:
iptables-restore v1.6.0: iptables-restore: unable to initialize table 'nat'
Error occurred at line: 2


lurch:
This seems to be a duplicate of #905 ?
(i.e. you're not quite following the instructions correctly)

JamesH65:
@takz012 Can you confirm if this is a problem in the docs, or whether it's simply a mistake when following the instructions? You should be editing the file, not typing this stuff in the terminal.

aallan:

PR #969, opened 2018-08-21
corrected typo `-adev` to `--adev`

statueofmike:
I noticed that the instructions to use `-adev` for audio returned the error `invalid option -- 'a'`.


JamesH65:
Good spot, thanks!

Issue #976, opened 2018-09-02
Booting Raspberry Pi 3 from mass storage


getz53:
The documentation  at
https://github.com/raspberrypi/documentation/blob/master/hardware/raspberrypi/bootmodes/msd.md
does not state whether the SD card still needs to be inserted after the OTP bit is set. The Pi does not seem to boot from the USB mass storage device if the SD card is not present. Is this the expected behavior?

getz53:
After further research it appears that the SD card should not be needed. The Pi must still be booting from the SD card. I will look for other solutions (e.g., delaying the boot from USB so the disk has time to spin up) and boot with the SD removed.

dgoadby:
No, the SD card is not required if you got the configuration right. I use SSD's for booting using a USB3 adaptor with no problems. That said, I did have issues with some HD/USB adaptors which just didn't work at al. I found that USB3 adaptors that use the ASMEDIA chipset work best. The problem seems to be related to the time the USB/Disk adaptor is ready after power up. The on-board LEDS's and adapter LEDS often provide a clue. My SSD's are formatted with a 100 MB FAT partition (/dev/sda1) for the boot and EXT4 for the root partition (/dev/sda2). And finally, I prepared the SSD on the RPi using parted _not_ a Linux PC. I don't know why but, disks prepared on other PC's don't' always work whereas disks prepared on a RPi work fine. I hope some of this helps.

JamesH65:
Made some clarification in the USB boot docs. #982

JamesH65:
Fix merged.

Issue #944, opened 2018-07-17
USB Boot Mode OTP Bit Programming Fail 


psymonwhite:
Using the USB Boot Mode instructions at: 

https://github.com/raspberrypi/documentation/blob/master/hardware/raspberrypi/bootmodes/msd.md 

If the OTP Bit fails to be programmed - what is the corrective action to fix this? 

The documentation states that it can fail, but does include a resolution path.

JamesH65:
@ghollingworth Is there any corrective action available?

ghollingworth:
Although in 'theory' it can fail, we'd not expect to see this happen unless there is a hardware problem or you're Pi in one in ten million...

No there's no 'corrective action'

Gordon

JamesH65:
I can update the text. Currently " If it is not, then the OTP bit has not been successfully programmed." to something like "If the bit has not been programmed, go through the programming procedure again. If the bit is still not set this would indicate a HW fault with the Pi itself."  ??

JamesH65:
@ghollingworth OK to go ahread with change suggested in previous post, or shall we just leave it?

ghollingworth:
Yes that sounds fine

JamesH65:
#982 Combined with some other changes to the same page.

JamesH65:
Fixed by #982

PR #982, opened 2018-09-12
USB boot mode updates

JamesH65:
Slight rewording, added extra statement on no SD card being needed.


JamesH65:
fixes #944

JamesH65:
fixes #976

PR #1005, opened 2018-10-19
more consistent capitalization of acronyms

jolan:
1) Capitali(s|z)e ARM consistently
2) Capitalize single occurrence of SCART


JamesH65:
Arm themselves do not capitalise their name, so in all their webpages they use Arm. It's not longer a abrieviation of Acorn Risc Machines!!

The SCART correction is accurate though, I'd accept a PR with just that change.

PR #1009, opened 2018-10-21
Modernize access-point.md documentation

rbisewski:
Not a big changeset; mostly just improving some of the command usage to reflect the newer commands, such as `apt` instead of `apt-get` and `ip addr` instead of `ifconfig`.


helenlynn:
@JamesH65 I have made only those two changes described in my commit.

Issue #986, opened 2018-09-26
Could probably do with some NFS docs


JamesH65:
How to connect to an NFS server and how to set up an NFS server on a Pi.

Might be there already but not found anything.

bennuttall:
I did this recently on Ubuntu Server following their guide: https://help.ubuntu.com/community/SettingUpNFSHowTo

If I get chance I'll have a go on a Pi and write it up. But if you want to do it, go ahead.

JamesH65:
Community effort started here https://github.com/raspberrypi/documentation/pull/1010

aallan:
Looks like #1010 got merged. Close?

JamesH65:
Good spot, closing.

PR #1010, opened 2018-10-21
Add a first draft of a NFS guide to the Pi repository

rbisewski:
As per issue #986, there is currently no official NFS setup guide for the Raspberry Pi documentation repository. This branch attempts to borrow content from the official Ubuntu wiki and adapt it slightly for newer versions of Raspian and rpcbind and systemd.

Not the most elegantly written, however, I believe this could serve as a "first draft" for a future guide.

### DIFF
--- a/configuration/README.md
+++ b/configuration/README.md
@@ -38,4 +38,5 @@ Some basic guides to configuring your Raspberry Pi.
     - How to configure screen blanking/screen saver
 - [The boot folder](boot_folder.md)
     - What it's for and what's in it
-
+- [Network File System](nfs.md)
+    - How to setup a NFS and connect clients to it
--- a/configuration/README.md
+++ b/configuration/README.md
@@ -38,5 +38,5 @@ Some basic guides to configuring your Raspberry Pi.
     - How to configure screen blanking/screen saver
 - [The boot folder](boot_folder.md)
     - What it's for and what's in it
-- [Network File System](nfs.md)
+- [Network File System (NFS)](nfs.md)
     - How to set up a NFS and connect clients to it
--- a/configuration/README.md
+++ b/configuration/README.md
@@ -39,4 +39,4 @@ Some basic guides to configuring your Raspberry Pi.
 - [The boot folder](boot_folder.md)
     - What it's for and what's in it
 - [Network File System](nfs.md)
-    - How to setup a NFS and connect clients to it
+    - How to set up a NFS and connect clients to it
--- a/configuration/nfs.md
+++ b/configuration/nfs.md
@@ -17,7 +17,7 @@ Ubuntu Before deploying NFS you should be familiar with:
 
 Install the below packages:
 
-```
+```bash
 sudo apt install nfs-kernel-server
 ```
 
@@ -25,13 +25,13 @@ For easier maintenance we will isolate all NFS exports in single directory, wher
 
 Suppose we want to export our users' home directories in /home/users. First we create the export filesystem:
 
-```
+```bash
 sudo mkdir -p /export/users
 ```
 
 Note that /export and /export/users will need 777 permissions as we will be accessing the NFS share from the client without LDAP/NIS authentication. This will not apply if using authentication (see below). Now mount the real users directory with:
 
-```
+```bash
 sudo mount --bind /home/users /export/users
 ```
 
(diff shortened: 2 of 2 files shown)


configuration/nfs.md
```diff
@@ -0,0 +1,304 @@
...
+Ubuntu Before deploying NFS you should be familiar with:
+
+1. Linux file and directory permissions
+1. Mounting and unmounting filesystems
+
+### Setup a basic NFS server
+
+Install the below packages:
+
+```
```
bennuttall: use `bash` after the three backticks where appropraite

bennuttall:
Looks good. I would like to give it a test (other testers welcome).

configuration/nfs.md
rbisewski: I have since done so :)

JamesH65:
Should probably remove the work in progress line at the start, but otherwise LGTM.

JamesH65:
Pushed to copy edit.

configuration/README.md
```diff
@@ -38,4 +38,5 @@ Some basic guides to configuring your Raspberry Pi.
     - How to configure screen blanking/screen saver
 - [The boot folder](boot_folder.md)
     - What it's for and what's in it
-
+- [Network File System](nfs.md)
+    - How to setup a NFS and connect clients to it
```
andrum99: Sentence doesn't make sense. I'm guessing it is means to read "how to set up an NFS server and connect clients to it".

configuration/nfs.md
```diff
@@ -0,0 +1,303 @@
+# NFS - Network File System
```
andrum99: The page heading should be the same as the link text from the main page. So "Network File System".

configuration/nfs.md
```diff
@@ -0,0 +1,303 @@
...
+
+Now add the following line to `/etc/hosts.allow`:
+
+```
+rpcbind : <NFS server IP address>
+```
+
+Where `<NFS server IP address>` is the IP address of the server.
+
+## NFS Server with complex user permissions
```
andrum99: Camel caps is wonky. Make the 's' on server lower case to match rest of sentence

configuration/nfs.md
```diff
@@ -0,0 +1,303 @@
...
+
+After setting up `/etc/exports`, export the shares:
+
+```bash
+sudo exportfs -ra
+```
+
+You'll want to do this command whenever `/etc/exports` is modified.
+
+### Restart Services
```
andrum99: For consistency please change capitalisation so it says "Restart services".

configuration/nfs.md
```diff
@@ -0,0 +1,303 @@
...
+```
+
+Create a symbolic link inside your home, pointing to the actual mount location. For example, in our case delete the 'Music' directory already existing there first:
+
+```bash
+rmdir /home/user/Music
+ln -s /nfs/music/ /home/user/Music
+```
+
+## Author Information
```
andrum99: Again - capitalisation. Change to "Author information"

configuration/nfs.md
```diff
@@ -0,0 +1,304 @@
...
+
+This way, server and client do not need the users to share same UID/GUID. For those who use LDAP-based authentication, add the following lines to the `idmapd.conf` of your clients:
+
+```
+[Translation]
+
+Method = nsswitch
+```
+
+This will cause `idmapd` to know to look at `nsswitch.conf` to determine where it should look for credential information. If you have LDAP authentication already working, `nsswitch` shouldn't require further explanation.
```
helenlynn: Should this read "`nsswitch` shouldn't require further information" instead of "`nsswitch` shouldn't require further explanation"?

configuration/nfs.md
```diff
@@ -0,0 +1,304 @@
...
+If any changes are made, rpcbind and NFS will need to be restarted:
+
+```bash
+sudo systemctl restart rpcbind
+sudo systemctl restart nfs-kernel-server
+```
+
+### Security items to consider
+
+Aside from the UID issues discussed above, it should be noted that an attacker could potentially masquerade as a machine that is allowed to map the share, which allows them to create arbitrary UIDs to access your files. One potential solution to this is IPSec; see also the NFS and IPSec section below. You can set up all your domain members to talk to each other only over IPSec, which will effectively authenticate that your client is who it says it is.
```
helenlynn: "see also the NFS and IPSec section below": the section this refers to does not seem to exist.

andrum99: I would just remove that sentence

configuration/nfs.md
andrum99: I think what the author is getting at is that if the user has already set up LDAP authentication, then they should know about what nsswitch does since configuring nsswitch is required to enable it.

configuration/nfs.md
helenlynn: Someone made and I think then deleted the sensible-seeming suggestion to just delete this phrase. I shall do so.

PR #1011, opened 2018-10-21
Adding some technical details regarding the Broadcom BCM2837B0 chip

rbisewski:
As the title mentions, this adds a page for the newer chip used in the latest Pi 3B+ models.

### DIFF
--- a/hardware/raspberrypi/bcm2837b0/README.md
+++ b/hardware/raspberrypi/bcm2837b0/README.md
@@ -1,12 +1,12 @@
 # BCM2837B0
 
-This is the Broadcom chip used in the Raspberry Pi 3B+. The underlying architecture of the BCM2837B0 is identical to the BCM2837A0 chip used in other versions of the Pi. The ARM core hardware is the same, only the frequency is rated higher.
+This is the Broadcom chip used in the Raspberry Pi 3B+ and 3A+. The underlying architecture of the BCM2837B0 is identical to the BCM2837A0 chip used in other versions of the Pi. The ARM core hardware is the same, only the frequency is rated higher.
 
- The ARM cores are capable of running up to 1.4GHz, making the device about 17% faster than the original Raspberry Pi 3. The VideoCore IV runs at 400MHz. The ARM core is 64-bit while the VideoCore IV is 32-bit.
+The ARM cores are capable of running at up to 1.4GHz, making the 3B+/3A+ about 17% faster than the original Raspberry Pi 3. The VideoCore IV runs at 400MHz. The ARM core is 64-bit, while the VideoCore IV is 32-bit.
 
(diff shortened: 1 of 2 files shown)


hardware/raspberrypi/bcm2837b0/README.md
```diff
@@ -0,0 +1,16 @@
+# BCM2837B0
+
+This is the Broadcom chip used in the Raspberry Pi 3B+. The underlying architecture of the BCM2837B0 is identical to the BCM2836 and BCM2837. The only significant difference is the replacement of the ARMv7 quad core cluster with a quad-core ARM Cortex A53 (ARMv8) cluster.
+
```
JamesH65: This isn't quite right. The B0 is basically exactly the same as the A0, it's just the packaging that is different, which is one reason why the thermals are better. There is no ARM core change between the A0 and the B0.

hardware/raspberrypi/bcm2837b0/README.md
```diff
@@ -0,0 +1,16 @@
...
+The ARM cores run at 1.4GHz, making the device about 17% faster than the original Raspberry Pi 3. The VideoCore IV runs at 400MHz.
+
+Regarding the BCM2837B0 chip itself, it is an updated version of the 64-bit Broadcom application processor used in Raspberry Pi 3B, which incorporates power integrity optimisations, and a heat spreader. These allow higher clock frequencies (or to run at lower voltages to reduce power consumption), and to more accurately monitor and control the temperature of
+the chip.
```
JamesH65: The ARM cores are 64bit, but the Videocore4 is 32bit, so this statement is a bit misleading.

rbisewski:
Good points, yes I see how that could be misleading. Thanks for the comments :)

I have since corrected the paragraphs to something more precise. You can see the corrected file [here](https://github.com/rbisewski/pi_documentation/tree/add_raspberry_pi_3plus_hardware_info/hardware/raspberrypi/bcm2837b0).

Let me know what you think, and feel free to tweak it and merge once you think it is complete.

bsimmo:
Technically
" The ARM cores run at 1.4GHz "
Should be
" The ARM cores runs at upto 1.4GHz "

It can only sustain that, in a normal setup, for short periods of time before it is back down at 1.2GHz.

Side note /
I think the main advantage is it (B0) can sustain 1.2GHz longer than the A0 (or more the pi3B+/3A+ than the pi3B, as I'm sure it's more than just the soc allowing it), this is also unverified as not scientifically tested, just my personal observations.

rbisewski:
Sounds good @bsimmo I have since made the requested changes.

Now as a follow up @JamesH65, I believe I have committed all of the requested changes. Feel free to merge.

JamesH65:
OK, it's flagged up for copy edit, once done I can merge it.

PR #1015, opened 2018-10-24
Country code in wpa_supplicant.conf

Vanov:
Added section in wireless-cli.md to specify ISO 3166 country codes in wpa_supplicant.conf, since some cards require that parameter to be set.

### DIFF
--- a/configuration/wireless/wireless-cli.md
+++ b/configuration/wireless/wireless-cli.md
@@ -34,15 +34,15 @@ network={
     psk="testingPassword"
 }
 ```
-**NOTE: Some cards requires your ISO 3166 country code**  
-this can be done by specifying the country parameter before the network section in wpa_supplicant.conf: 
+**NOTE: Some cards requires your ISO 3166 country code.**  
+You can add this by specifying the `country` parameter before the `network` section in `wpa_supplicant.conf`: 
 ```
 country=XX
 network={
     ***example***
 }
 ```
-Where XX is any of [these officially assigned code elements](https://en.wikipedia.org/wiki/ISO_3166-1#Officially_assigned_code_elements).
+Where `XX` is any of [these officially assigned code elements](https://en.wikipedia.org/wiki/ISO_3166-1#Officially_assigned_code_elements).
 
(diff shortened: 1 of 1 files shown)


JamesH65:
Thanks - I thought we already said this somewhere, but this is a good addition. Pushing to CE.

configuration/wireless/wireless-cli.md
```diff
@@ -34,6 +34,16 @@ network={
     psk="testingPassword"
 }
 ```
+**NOTE: Some cards requires your ISO 3166 country code.**
```
XECDesign: Not sure I understand what "some cards requires.." means here.

Vanov: Me neither tbh, i just recognize that some of mine wont work without the country parameter set.

JamesH65: Does it need rephrasing? "Some USB wireless adapters"? Or something else?

JamesH65: Looking furthur down the page, there is a section on 3B+ and the requriements there for the country code to be set. I wonder if we simply add "for 3B+ and any USB wireless device using 5gig" or similar to that section rather than have this extra section? Or have people found its is necessary for other protocols as well?

XECDesign: I don't think it affects USB devices.

JamesH65:
Did we get to the bottom of what was required here?@vanov. Why do you think this change is needed?

Vanov:
Im using a RPI (Model B, no integrated wifi) with a TP-LINK TL-WN722N (V2 or V3, cannot remember). Im setting up raspbian lite with the wpa_supplicant.conf i created in boot partition. My problem is that unless i specify a country code, my card wont connect. 
The reason i suggest including this in the official documentation is because i had to reach out to other resources to figure this out.

JamesH65:
It's odd that that particular device needs the country code, doesn't appear to be 5GH band which is where is probably required. However, I can update the docs if others think its a good change given this information.

configuration/wireless/wireless-cli.md
lurch: In any case, "Some cards may require" is probably better than "Some cards requires" ? :wink:

JamesH65:
I'd be inclined to go with something like "Some USB wireless adapters may require a country code to be set before they will work correctly".

PR #1017, opened 2018-10-29
Add missing step for bridge access point setup

simonmeusel:
This PR add a missing step, which is mentioned in `Setting up a Raspberry Pi as an access point in a standalone network`, but not in `Using the Raspberry Pi as an access point to share an internet connection (bridge)`.

https://www.raspberrypi.org/documentation/configuration/wireless/access-point.md

### DIFF
--- a/configuration/wireless/access-point.md
+++ b/configuration/wireless/access-point.md
@@ -229,6 +229,18 @@ wpa_pairwise=TKIP
 rsn_pairwise=CCMP
 ```
 
+We now need to tell the system where to find this configuration file.
+
+```
+sudo nano /etc/default/hostapd
+```
+
+Find the line with #DAEMON_CONF, and replace it with this:
+
+```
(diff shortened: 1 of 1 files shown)


JamesH65:
Isn't this covered by the text "The access point setup is almost the same as that shown in the previous section. Follow the instructions above to set up the hostapd.conf file, but add bridge=br0 below the interface=wlan0 line, and remove or comment out the driver line. The passphrase must be between 8 and 64 characters long."?

simonmeusel:
From my understanding it says to configure the `hostapd.conf` file, but not the ` sudo nano /etc/default/hostapd` file. I think this way it's more clear.

But if you don't want to accept this PR, I don't care, just close it :)

JamesH65:
OK, so I actually fixed this in a different way by clarifying the original text.

PR #1020, opened 2018-11-04
Update cm-emmc-flashing.md

ghollingworth:
Added comment concerning the CM3 partition creation.  It needs to be done with the right command otherwise it can fail to boot correctly on a small percentage of CM3 devices.


JamesH65:
LGTM, pushing to copy edit.

PR #1021, opened 2018-11-14
Add info on how to enable ssh in headless mode.

jottr:
This is crucial information. Users shouldn't have to go to [stackoverflow](https://stackoverflow.com/questions/41318597/ssh-connection-refused-on-raspberry-pi) to find out about this.


JamesH65:
You don't need to go to stack overflow, it's all in section three here

https://www.raspberrypi.org/documentation/remote-access/ssh/README.md

jottr:
Interesting. It seems I overlooked it, since google offered me the `unix`-subsection first.

PR #1023, opened 2018-11-16
README.md: Correct baud rate is 115200 for UART

chckyn:
The correct baud rate for the UART serial connection is 115200. Tested on a Model 1 A+.


JamesH65:
@ghollingworth This docs change looks correct to me - can you confirm and I'll merge it.

ghollingworth:
Indeed slip of the keyboard!

PR #1030, opened 2018-11-27
Correct info on which Pi models support MSD boot

andrum99:
Update documentation/msd.md with information on which current Pi models support USB mass storage device booting.


JamesH65:
Not sure I like the table format, seems to take a lot of space when all you really need is "USB boot is available on the Raspberry Pi 3B, 3B+ and Raspberry Pi 2B v1.2  models only."

andrum99:
I concur - my original plan was to have a table which listed every single model, but I decided against that, then ended up with that version. I've updated the PR to include your wording instead.

PR #1032, opened 2018-11-27
Mention models of Pi that can do USB MSD boot

andrum99:
Only Pi 3 B, Pi 3 B+ and Pi 2 B model 1.2 can boot from USB mass storage device. We should probably mention that in the "Can I boot a Pi from a USB-attached hard drive instead of the SD card?" FAQ entry.


Doug-Wyman:
Only Pi 3 B, Pi 3 B+ and Pi 2 B model 1.2 can boot from USB mass
storage device. We should probably mention that in the "Can I boot a Pi
from a USB-attached hard drive instead of the SD card?" FAQ entry.

The new model A 3+ can also boot from USB

PR #1037, opened 2018-11-30
enable_uart=1 does NOT set core_freq=250

ak-ambi:
See: https://www.raspberrypi.org/forums/viewtopic.php?t=213499


pelwell:
I prefer the approach in #1038 .

JamesH65:
OK, closing, #1038 will sort this out.

PR #1039, opened 2018-12-05
Clarified overscan settings and their default values

hhromic:
According to the discussion in <https://github.com/raspberrypi/firmware/issues/1079>, better clarify the overscan settings. Changes:

* Added the default values for all overscan settings.
* Clarified that `disable_overscan` actually just sets the overscan edges to `0`, and other overscan settings can still be applied.
* Added a note that `overscan_scale` is generally not recommended.

### DIFF
--- a/configuration/config-txt/video.md
+++ b/configuration/config-txt/video.md
@@ -483,27 +483,39 @@ Setting `hdmi_ignore_hotplug` to `1` pretends that the HDMI hotplug signal is no
 
 ### disable_overscan
 
-Set `disable_overscan` to `1` to disable [overscan](../raspi-config.md#overscan).
+Set `disable_overscan` to `1` to disable the default values of [overscan](../raspi-config.md#overscan) the firmware sets. The default value of overscan for the left, right, top and bottom edges is `48` for HD CEA modes, `32` for SD CEA modes and `0` for DMT modes. The default value for `disable_overscan` is `0`.
+
+**NOTE:** Any further additional overscan options such as `overscan_scale` or overscan edges can still be applied after this option.
 
 ### overscan_left
 
-The `overscan_left` command specifies the number of pixels to skip on the left edge of the screen. Increase this value if the text flows off the left edge of the screen; decrease it if there is a black border between the left edge of the screen and the text.
+The `overscan_left` command specifies the number of pixels to add to the firmware default value of overscan on the left edge of the screen. The default value is `0`.
+
+Increase this value if the text flows off the left edge of the screen; decrease it if there is a black border between the left edge of the screen and the text.
 
 ### overscan_right
 
-The `overscan_right` command specifies the number of pixels to skip on the right edge of the screen.
+The `overscan_right` command specifies the number of pixels to add to the firmware default value of overscan on the right edge of the screen. The default value is `0`.
+
+Increase this value if the text flows off the right edge of the screen; decrease it if there is a black border between the right edge of the screen and the text.
 
 ### overscan_top
 
-The `overscan_top` command specifies the number of pixels to skip on the top edge of the screen.
+The `overscan_top` command specifies the number of pixels to add to the firmware default value of overscan on the top edge of the screen. The default value is `0`.
+
+Increase this value if the text flows off the top edge of the screen; decrease it if there is a black border between the top edge of the screen and the text.
 
 ### overscan_bottom
 
-The `overscan_bottom` command specifies the number of pixels to skip on the bottom edge of the screen.
+The `overscan_bottom` command specifies the number of pixels to add to the firmware default value of overscan on the bottom edge of the screen. The default value is `0`.
+
+Increase this value if the text flows off the bottom edge of the screen; decrease it if there is a black border between the bottom edge of the screen and the text.
 
 ### overscan_scale
 
-Set `overscan_scale` to `1` to force any non-framebuffer layers to conform to the overscan settings.
+Set `overscan_scale` to `1` to force any non-framebuffer layers to conform to the overscan settings. The default value is `0`.
+
+**NOTE:** This feature is generally not recommended because it can reduce image quality as all layers on the display will be scaled by the GPU. Disabling overscan on the display itself is the recommended option to avoid images being scaled twice (by the GPU and the display).
 
 ### framebuffer_width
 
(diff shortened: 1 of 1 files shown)


JamesH65:
This still seems vague to me, as to why there are two ways of 'disabling' overscan.

hhromic:
There aren't two ways to disable overscan, perhaps my wording is not the best, sorry.
The main message is that `disable_overscan` "disables" overscan by setting the four edges to `0` in one operation (as clarified to me on the referenced issue). This is why you can still use other overscan settings after setting `disable_overscan`.

How about rephrasing to this?
> This option disables overscan by setting `overscan_left`, `overscan_right`, `overscan_top` and `overscan_bottom` together to `0` in a single operation. Therefore, any further additional overscan setting such as `overscan_scale` can still be applied.

lurch:
It's not clear from the linked issue, but are you still able to use `disable_overscan` to simultaneously set all 4 borders to 0, **and** then still specify just one value (e.g. `overscan_right`) to apply a border to just that side? @popcornmix

hhromic:
Yes. Take for instance the following configuration with exaggerated values for illustration:
```
disable_overscan=1
overscan_left=100
overscan_right=100
overscan_scale=1
```
It produces an image with thick left and right borders and no top/bottom borders. Also applications are scaled to this overscan settings too. The `disable_overscan` is behaving exactly as @popcornmix mentioned (remove defaults).

configuration/config-txt/video.md
```diff
@@ -483,27 +483,31 @@ Setting `hdmi_ignore_hotplug` to `1` pretends that the HDMI hotplug signal is no
 
 ### disable_overscan
 
-Set `disable_overscan` to `1` to disable [overscan](../raspi-config.md#overscan).
+Set `disable_overscan` to `1` to disable [overscan](../raspi-config.md#overscan). The default value is `0`.
+
+Disabling overscan is accomplished by setting `overscan_left`, `overscan_right`, `overscan_top` and `overscan_bottom` together to `0`. Any further additional overscan setting such as `overscan_scale` can still be applied.
```
popcornmix: I think this should say:
Set `disable_overscan` to `1` to disable the default values of overscan the firmware sets.
The default values are `48` for HD CEA modes, `32` for SD CEA modes and `0` for DMT modes.

configuration/config-txt/video.md
```diff
@@ -483,27 +483,31 @@ Setting `hdmi_ignore_hotplug` to `1` pretends that the HDMI hotplug signal is no
...
 
 ### overscan_left
 
-The `overscan_left` command specifies the number of pixels to skip on the left edge of the screen. Increase this value if the text flows off the left edge of the screen; decrease it if there is a black border between the left edge of the screen and the text.
+The `overscan_left` command specifies the number of pixels to skip on the left edge of the screen. Increase this value if the text flows off the left edge of the screen; decrease it if there is a black border between the left edge of the screen and the text. The default value is `48` for HD CEA modes, `32` for SD CEA modes and `0` for DMT modes.
```
popcornmix: These can now be removed as the information is in the disable_overscan setting.

configuration/config-txt/video.md
```diff
@@ -483,27 +483,31 @@ Setting `hdmi_ignore_hotplug` to `1` pretends that the HDMI hotplug signal is no
...
 
-The `overscan_bottom` command specifies the number of pixels to skip on the bottom edge of the screen.
+The `overscan_bottom` command specifies the number of pixels to skip on the bottom edge of the screen. The default value is `48` for HD CEA modes, `32` for SD CEA modes and `0` for DMT modes.
 
 ### overscan_scale
 
-Set `overscan_scale` to `1` to force any non-framebuffer layers to conform to the overscan settings.
+Set `overscan_scale` to `1` to force any non-framebuffer layers to conform to the overscan settings. The default value is `0`.
+
+Note: this setting is generally not recommended because it can reduce performance as the scaling is done by the GPU.
```
popcornmix: Performance isn't affected but quality is as all layers on the display will be scaled. Disabling overscan on the TV is the recommended options to avoid images being scaled twice (by Pi and TV).

configuration/config-txt/video.md
hhromic: I'm not sure that is precise enough. What will be the values if you *disable the default values*?
I understand these values are set to (or left as?) `0`, hence why I detailed precisely what it does.

Also mentioning the default values of other options is not appropiate here IMHO.
Likewise in the other options of the documentation, each option precisely defines what is its default value independently and self-contained. This enables users to quickly consult the documentation as a reference guide, without needing to jump back and forth other sections.

configuration/config-txt/video.md
hhromic: I strongly think the defaults should be kept in each option as per any other option the documentation.

popcornmix: No, the default for overscan_left is not 48. It is always zero.
These values get added to the defaults which are set internally by the firmware when ``disable_overscan=0``

hhromic: Ahh I see, so the firmware actually does something like this (very pseudo-code)?
```
overscan_* = 0;
if (disable_overscan == 0) {
  overscan_* = 48, 32, 0; /* according to mode */
}
```

lurch: Ahhh, so there's "internal default" overscan values (determined either by the display-mode or disable_overscan), which the `overscan_left` etc. values then get added to?
I guess that implies that if the screen-mode is SD CEA, and disable_overscan isn't set, the minimum value for e.g. `overscan_right` is -32 ?

popcornmix: @lurch yes

hhromic: Mmm, @popcornmix, so to make things less confusing (for me), is this then what actually happens in the firmware (again very pseudo-code)?
```
internal_overscan_* = 0;
if (disable_overscan == 0) {
  internal_overscan = 48, 32, 0;  /* according to screen-mode */
}
internal_overscan_* += overscan_*;
```

hhromic:
Update the PR with the comments from @popcornmix. Hope is good now.

configuration/config-txt/video.md
popcornmix: Yes

popcornmix:
Change looks okay to me.

hhromic:
Great, thanks for all your clarifictaions, really made understanding overscan settings better.
@JamesH65  let me know if something else is needed for this PR :)

JamesH65:
Looks OK to me, have pushed it to copy editors. No idea when they will do it, probably all on holiday!

hhromic:
Looking fwd! Thanks and happy new year.

PR #1045, opened 2018-12-18
Original boot cmdline doesn't work anymore. Fixed the docs.

dagelf:
It worked as soon as I removed init=


JamesH65:
Bit confused, I cannot find an `init=` in the docs or in the cmdline.txt of a standard Pi. Where is the original cmdline you are talking about?

XECDesign:
It's in cmdline.txt before first boot. If it prevents netbooting, it should be fixed within the script.

XECDesign:
Yes, that would be me.

JamesH65:
@XECDesign  Thoughts?

XECDesign:
No objections. It will be a while before we release a new image, and it doesn't make sense to waste peoples' time with undocumented bugs.

PR #1047, opened 2018-12-21
Update conditional.md

JamesH65:
Updates the conditonal model filters section to use a table rather than long winded description. Should be easier to keep updated this way.

### DIFF
--- a/configuration/config-txt/conditional.md
+++ b/configuration/config-txt/conditional.md
@@ -18,10 +18,10 @@ It is usually a good idea to add an `[all]` filter at the end of groups of filte
 
 The conditional model filters are applied according to the following table.
 
-| Filter | Applicable Model |
+| Filter | Applicable Model(s) |
 |--------|------------------|
 | [pi1] | Model A, Model B, Compute Module |
-| [pi2] | Model 2B |
+| [pi2] | Model 2B (2836 or 2837 based) |
 | [pi3] | Model 3B, Compute Module 3 |
 | [pi3+]| Model 3A+, Model 3B+ |
 | [pi0] | Zero |
--- a/configuration/config-txt/conditional.md
+++ b/configuration/config-txt/conditional.md
@@ -18,11 +18,11 @@ It is usually a good idea to add an `[all]` filter at the end of groups of filte
 
 The conditional model filters are applied according to the following table.
 
-| Filter | Applicable Model(s) |
+| Filter | Applicable model(s) |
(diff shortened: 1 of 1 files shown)


JamesH65:
Yes, the Pi2 is the Pi2, irrespective of the SoC on it.

configuration/config-txt/conditional.md
```diff
@@ -14,17 +14,18 @@ This is the most basic filter. It resets all previously set filters and allows a
...
-Any settings below a `[pi2]` filter will only be applied to Pi 2 hardware. The `[pi3]` filter is applicable to Pi 3 and Compute Module 3 hardware, while `[pi0]` covers Pi Zero and Pi Zero W.
+The conditional model filters are applied according to the following table.
 
-    [pi1]
-    [pi2]
-    [pi3]
-    [pi3+]
-    [pi0]
-    [pi0w]
+| Filter | Applicable Model |
```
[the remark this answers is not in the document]
JamesH65: will change.

configuration/config-txt/conditional.md
```diff
@@ -14,17 +14,18 @@ This is the most basic filter. It resets all previously set filters and allows a
...
-    [pi0]
-    [pi0w]
+| Filter | Applicable Model(s) |
+|--------|------------------|
+| [pi1] | Model A, Model B, Compute Module |
+| [pi2] | Model 2B (2836 or 2837 based) |
+| [pi3] | Model 3B, Compute Module 3 |
+| [pi3+]| Model 3A+, Model 3B+ |
+| [pi0] | Zero |
+| [pi0w]| Zero W, Zero WH |
```
lurch: I've just spotted the section below which says
> It is important to note that the Raspberry Pi Zero W will see the contents of [pi0w] AND [pi0]. Likewise, a Raspberry Pi 3B Plus sees [pi3+] AND [pi3].

I wonder if that means the table above ought to say:
```
| [pi3] | Model 3B, Compute Module 3, Model 3A+, Model 3B+ |
| [pi3+]| Model 3A+, Model 3B+ |
| [pi0] | Zero, Zero W, Zero WH |
| [pi0w]| Zero W, Zero WH |
```
?

pelwell: Yes - that looks correct.

JamesH65: OK, will update as per Andrew's suggestion.

PR #1052, opened 2018-12-28
Fix typos :)

drmercer:
"Home directly" should say "home directory." It also seems there was a missing word in the last clause of that sentence.


JamesH65:
Thanks!

PR #1057, opened 2019-01-04
Fix broken link to Avahi Wikipedia Page

aloisklink:
The markdown renderer at <https://www.raspberrypi.org/documentation/remote-access/ip-address.md> seems to have a bug in it, since it thinks the markdown:

```md
[Avahi](https://en.wikipedia.org/wiki/Avahi_(software))
```

is a link to <https://en.wikipedia.org/wiki/Avahi_(software>, missing the final `)` and placing it into the text, ie:

```html
<a href="https://en.wikipedia.org/wiki/Avahi_(software">Avahi</a>)
```

By percent encoding the parentheses with `%28` and `%29`, this will be avoided.

Other potential solutions are discussed here <https://meta.stackexchange.com/a/13509> but URL encoding is probably the most likely to work.

### DIFF
--- a/remote-access/ip-address.md
+++ b/remote-access/ip-address.md
@@ -20,7 +20,7 @@ In a web browser navigate to your router's IP address e.g. `http://192.168.1.1`,
 
 ### Resolving `raspberrypi.local` with mDNS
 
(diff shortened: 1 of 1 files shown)


JamesH65:
Thanks!

Issue #61, opened 2014-05-26
The documentation/Troubleshooting section is inadequately scoped


P33M:
There are several stub pages in documentation/troubleshooting that require filling out.

At first when @bennuttall created the troubleshooting/usb page I was tempted just to cross-link to the "article" I already wrote on the USB hardware and its differences to EHCI, but then the obverse argument became apparent: users will hit (or be directed to) the troubleshooting guide without any understanding of the underlying issues, nor do they wish to know about why OTG is different from EHCI.

The way I see it, the Pi lacks tier 1 technical support. Well, it doesn't actually: it has a very well staffed and managed user forum but this is an energy barrier to people that just want their Pi to work.

I propose scoping the troubleshooting section of the documentation tree to an apparent symptom/recommended solution based approach. The top page should be "what is wrong with your Pi?" with the top four or five links directly linking to the most common causes of problems. Inexperienced users are typically terrible at diagnosing the root issue with unfamiliar hardware: a simple "my Pi doesn't do X" -> "page full of solutions for X" should nab 80% of nonfunctional cases.

ebenupton:
Close for inactivity.

PR #1061, opened 2019-01-10
James h65 patch led blinkies

JamesH65:
Added a page to describe the LED blink errors/warnings.  Linked from configuration page, as we don't seem to have any sort of 'troubleshooting' area.


lurch:
> we don't seem to have any sort of 'troubleshooting' area.

see #61 :wink:

JamesH65:
Note to self, this should only go live once the commit reordering the error codes has beenmerged in to the firmware.

PR #1064, opened 2019-01-13
The Embest CM3 Exists!

jriwanek:
I have two of them, and they're missing from the list


pelwell:
That's better - thanks.

PR #1068, opened 2019-01-17
cmdline-txt.md: add documentation

andrum99:
Improve documentation of:

- plymouth.ignore_serial_console (not currently explained)
- dwc_org.lpm_enable
- dwc_otg.speed

### DIFF
--- a/configuration/cmdline-txt.md
+++ b/configuration/cmdline-txt.md
@@ -25,7 +25,7 @@ There are many kernel command line parameters, some of which are defined by the 
 
  - splash: tells the boot to use a splash screen via the Plymouth module.
  - plymouth.ignore_serial_console - tells Plymouth not to output a boot splash on any serial console which may be configured.
- - dwc_otg.lpm_enable: turns off Link Power Management (LPM) in the dwc_otg driver. The dwc_otg driver is the driver for the USB controller built into the Raspberry Pi
+ - dwc_otg.lpm_enable: turns off Link Power Management (LPM) in the dwc_otg driver. The dwc_otg driver is the driver for the USB controller built into the Raspberry Pi.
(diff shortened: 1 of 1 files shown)


configuration/cmdline-txt.md
```diff
@@ -24,9 +24,9 @@ There are many kernel command line parameters, some of which are defined by the
 #### Other Entries (not exhaustive)
 
  - splash: tells the boot to use a splash screen via the Plymouth module.
- - plymouth.ignore_serial_console
- - dwc_otg.lpm_enable: turns off LPM in the dwc_otg (On The Go) driver.
- - dwc_otg.speed: sets the speed of the USB property. `dwc_otg.speed=1` will set it to USBv1.0 speed.
+ - plymouth.ignore_serial_console - tells Plymouth not to output a boot splash on any serial console which may be configured.
+ - dwc_otg.lpm_enable: turns off Link Power Management (LPM) in the dwc_otg driver. The dwc_otg driver is the driver for the USB controller built into the Raspberry Pi.
+ - dwc_otg.speed: sets the speed of the USB controller. `dwc_otg.speed=1` will set it to full speed (USB 1.0).
```
lurch: While this is technically correct (as USB2.0 is High Speed), I wonder if people not familiar with USB protocols might mistakenly think that they **want** to set their USB ports to "full speed", not realising that it would actually slow down their USB ports?

JamesH65: Good point, ought to say that its should not be changed unless having problems with the USB device, and probably emphasise that full speed is a slower speed.

andrum99: Added extra info per jamesh65's comments.

PR #1069, opened 2019-01-17
boot_folder.md: improve docs

andrum99:
Improve intro paragraph
Improve Device Tree files 
Add section explaining what the overlays sub-directory is
Make capitalisation of headings consistent

### DIFF
--- a/configuration/boot_folder.md
+++ b/configuration/boot_folder.md
@@ -1,6 +1,6 @@
 # The boot folder
 
-In a basic Raspbian install, the boot files are stored on the first partition of the SD card, which is formatted with the FAT filesystem. This means that it can be read on Windows, Mac and Linux devices.
+In a basic Raspbian install, the boot files are stored on the first partition of the SD card, which is formatted with the FAT file system. This means that it can be read on Windows, macOS, and Linux devices.
 
 When the Raspberry Pi is powered on, it loads various files from the boot partition/folder in order to start up the various processors, then it boots the Linux kernel.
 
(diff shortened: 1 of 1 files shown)


configuration/boot_folder.md
```diff
@@ -1,10 +1,12 @@
-# The Boot folder
+# The Boot Folder
 
-In a basic Raspbian install, the boot files are stored on the first partition of the SD card, which is VFAT-formatted. This means that it can be read on both Windows and Linux devices. 
+In a basic Raspbian install, the boot files are stored on the first partition of the SD card, which is formatted with the FAT filesystem. This means that it can be read on Windows, Mac and Linux devices.
+
+Once Linux has booted, the boot partition is mounted as /boot.
```
lurch: I wonder if this might be better as something like "Once Raspbian has booted on the Raspberry Pi, the boot partition is mounted as `/boot`." to avoid any ambiguity? (given that the previous paragraph talks about reading the SD card on Linux)

andrum99: I've moved that sentence to after the next paragraph.

configuration/boot_folder.md
```diff
@@ -31,6 +33,10 @@ Contains many configuration parameters for setting up the Pi. See [the `config.t
...
 
-### Device Tree files
+### Device tree files
+
+There are various device tree blob files, which have the extension `.dtb`. These contain the hardware definitions of the various models of Raspberry Pi, and are used on boot to set up the kernel according to which Pi model is detected. More [details here](device-tree.md).
+
+## Device tree overlays
 
-There are various Device Tree blob files, `\*.dtb`. These contain the hardware definitions of the various Pi models, and are used on boot to set up the kernel according to which Pi model is detected. More [details here](device-tree.md).
+The `overlays` sub-folder contains device tree overlays which are used to configure various hardware devices which may be attached to the system, for example the official touchscreen display, and third party sound boards. These overlays are selected using entries in config.txt - see [Device Trees, overlays and parameters, part 2](device-tree.md#part2)
```
lurch: Backticks on `config.txt` would be nice :)

configuration/boot_folder.md
```diff
@@ -33,10 +33,10 @@ Contains many configuration parameters for setting up the Pi. See [the `config.t
 
 Some text-based housekeeping information containing the date and git commit ID of the distribution.
 
-### Device Tree Files
+### Device tree files
```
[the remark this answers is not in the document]
andrum99: There's no reason why device tree needs to be capitalised, other than if you're going to then shorten it to DT. See https://en.wikipedia.org/wiki/Device_tree

configuration/boot_folder.md
andrum99: done

configuration/boot_folder.md
andrum99: done

configuration/boot_folder.md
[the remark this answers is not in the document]
andrum99: Device in this case is only capitalised because it is the first word of the heading, which is how the rest of the documentation does it - i.e. first word capitalised, subsequent words not.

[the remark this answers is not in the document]
pelwell: I tend to use the same capitalisation on the D and the T - see the heading further down the Wikipedia page.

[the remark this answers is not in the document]
andrum99: We also seem to have devicetree and DeviceTree, according to devicetree.org. I'll set it back to Device Tree per pelwell.

PR #1075, opened 2019-01-26
Add some troubleshooting to camera.md

IamPhytan:
Troubleshoot by enabling I2C or/and by upgrading firmware.

### DIFF
--- a/configuration/camera.md
+++ b/configuration/camera.md
@@ -35,13 +35,13 @@ The display should show a five-second preview from the camera and then take a pi
 
 If the camera is not working, you may need to enable I2C using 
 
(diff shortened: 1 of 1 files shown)


JamesH65:
No, this isn't right. We do not recommend the use of rpi-firmware unless really really needed, and since the firmware to use the camera has been in the standard firmware for 6 years, there is no need for rpi-update to get it. Also, the comment on I2C  is also not necessary. You enable the camera using the enable camera option in raspi-config, that all, you do not need to enable the I2C.

JamesH65:
Closing as a Won't Fix.

PR #1084, opened 2019-02-14
Renaming etcher to balenaEtcher

gmkey:
Etcher has been renamed to balenaEtcher and it's website changed to www.balena.io/etcher/

### DIFF
--- a/installation/installing-images/README.md
+++ b/installation/installing-images/README.md
@@ -22,9 +22,9 @@ If you're not using Etcher (see below), you'll need to unzip `.zip` downloads to
 
 You will need to use an image writing tool to install the image you have downloaded on your SD card.
 
-**Etcher** is a graphical SD card writing tool that works on Mac OS, Linux and Windows, and is the easiest option for most users. Etcher also supports writing images directly from the zip file, without any unzipping required. To write your image with Etcher:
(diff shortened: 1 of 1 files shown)


bennuttall:
Are you saying this as a member of the public or are you a Balena employee?

It doesn't seem to me that they have renamed the tool "balenaEtcher" just that they sometimes refer to it with the company name first.

They still refer to "Etcher Pro" and there's a sentence that says "So we built Etcher, an SD card flasher app" and there are links to "Etcher for Windows", etc.

JamesH65:
Also, the word Etcher is used on more than one page in our docs, so if we need to change we need to change on all pages that use it.

gmkey:
First: I'm a member of the public. 
Second: I've installed the app on my Mac and although it can still be found as Etcher, the name presented to me is balenaEtcher. (balenaEtcher.app)

If need be, I'm more than happy to create change requests for these.

gmkey:
https://www.balena.io/blog/resin-io-changes-name-to-balena-releases-open-source-edition/

bennuttall:
I don't think we should change it, unless someone from Balena agrees that's correct usage. @lurch do you know of someone who would know for sure?

gmkey:
Their statement seems pretty obvious to me. But then again, you might be better of at the source!

gmkey:
closing this one in favour of #1085 , which contains proposals for all  documents

PR #1085, opened 2019-02-14
renamed Etcher to balenaEtcher and etcher.io to www.balena.com/etcher/

gmkey:
made the modifications for all affected pages rather than one. closed #1084, which - in the event you approve the change - contained only one affected page


JamesH65:
There appear to be a number of whitespace related changes unrelated tothe balena change. I'd prefer, if possible, that they supplied as a separate PR, since that makes it easier to detemine relevent changes. Still waiting for confirmation from Balena themselves on the correct naming scheme.

gmkey:
Oh dear, that seems to be the setting on my text editor (Atom). Not sure how to revert these? (not that github prolific for that matter. Please advise

JamesH65:
Hmm, not sure of your procedure, but I'd revert the whitespace changes, then use git commit --amend to update the PR. Or, make another PR that just has the whitespace changes in, send that in, then rebase this PR with that one. Or something.

lurch:
Ping @thundron and @zvin (and see also #1084, where this was also discussed)

JamesH65:
Part of the problem is that balenaEtcher is such an awkward (and terrible) name. Everyone is still just going to call it Etcher (a great name) whatever its actual name is, and changing the docs might make it more confusing rather than less.

bennuttall:
FWIW etcher.io was a better URL too

lurch:
@JamesH65 See also "GNU/Linux" and "Microsoft Windows" :wink:   (which everyone calls just Linux and Windows)

gmkey:
I agree with most your comments and I do feel that some marketing intern at etcher has probably forgot to read branding 101. 
At the same time most of the target Community are inexperienced, the name could set them of on the wrong foot. 
IMHO waiting for a statement from etcher / baloney (err balena ;-) is the sensible thing to do.

gmkey:
PS “GNU/Linux” is actually a name given to the Linux project by the FSF. AFAIK the Linux Project at kernel.org talk only about Linux.

lurch:
Ping also @jviotti and @alexandrosm

JamesH65:
Going to merge this with the whitespace changes as well - bit confusing, but they need to be done anyway.

PR #1086, opened 2019-02-16
Changed 'dist-upgrade' to 'upgrade' in two spots

exaudio:
There were two spots where 'dist-upgrade' was specified when the correct command was 'upgrade'.

### DIFF
--- a/raspbian/updating.md
+++ b/raspbian/updating.md
@@ -13,7 +13,7 @@ sudo apt-get update
 Next, **upgrade** all your installed packages to their latest versions with the command:
 
 ```bash
-sudo apt-get dist-upgrade
+sudo apt-get upgrade
 ```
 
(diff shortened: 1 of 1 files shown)


bennuttall:
This undoes an intention change by @XECDesign in e12279fce4f282cf1e17138a28c7ce831faeeb5a. Serge, can you explain your reasoning?

XECDesign:
Sure. Dist-upgrade does what upgrade does, but also picks up dependency changes. It's all in the man page. Sometimes we make changes to dependencies, so dist-upgrade should be used.

lurch:
This is _almost_ a "documentation repo FAQ" :wink: 
https://github.com/raspberrypi/documentation/search?q=dist-upgrade&type=Issues

XECDesign:
Maybe `sudo apt full-ugprade` would clear up some confusion, but I'm not 100% sure it's the same as `sudo apt-get dist-upgrade`. The description in the documentation is a little different and possibly incomplete. After a few minutes of googling, I couldn't find any definitive answer whether they're equivalent commands. I think it would require a bit of testing.

PR #1090, opened 2019-02-23
Add installing images instructions for Chrome OS

srdjanrosic:
Chromebooks tend to be popular in education, add instructions on how to prepare a raspbian SD card (alongside windows/mac/linux).

### DIFF
--- a/installation/installing-images/README.md
+++ b/installation/installing-images/README.md
@@ -35,4 +35,4 @@ For more advanced control of this process, see our system-specific guides:
 - [Linux](linux.md)
 - [Mac OS](mac.md)
 - [Windows](windows.md)
-- [ChromeOS](chromeos.md)
+- [Chrome OS](chromeos.md)
--- a/installation/installing-images/README.md
+++ b/installation/installing-images/README.md
@@ -35,3 +35,4 @@ For more advanced control of this process, see our system-specific guides:
 - [Linux](linux.md)
 - [Mac OS](mac.md)
 - [Windows](windows.md)
+- [ChromeOS](chromeos.md)
--- a/installation/installing-images/chromeos.md
+++ b/installation/installing-images/chromeos.md
@@ -1,16 +1,16 @@
-# Installing operating system images using ChomeOS
+# Installing operating system images using Chrome OS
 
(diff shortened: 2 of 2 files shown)


andrum99:
In my view Chrome OS is too niche to warrant documentation on the official Raspberry Pi website.

I don't see any evidence of chromebooks being popular in education. I don't think there is any point adding these instructions to the documentation as anyone using a chromebook will have access to another suitable operating system that is documented. Also, all you need to install an OS image on ChromeOS is the dd command, and optionally unzip if your image is compressed. The ChromeOS shell has enough to image a drive with an OS image without downloading additional software. If this does get added to the documentation, please add command line instructions as well.

srdjanrosic:
> I don't see any evidence of chromebooks being popular in education.

from (seemingly very US centric) [wikipedia article](https://en.wikipedia.org/wiki/Chromebook#Education_market):

> ... By March 2018 Chromebooks made up 60% of computers used in schools ...

---

> Also, all you need to install an OS image on ChromeOS is the dd command

There's no `dd` in crosh.

To get to a regular bash shell to be able to run `dd`, typically, you need to put the device in developer mode, which wipes the TPM/EC on the device and is otherwise inconvenient for organizationally procured devices.

andrum99:
I withdraw my objection, since it appears there is indeed a large installed base of Chrome OS systems in use within the Raspberry Pi's primary target market (i.e. education). If this does get added, a direct link to the app to use directly within the documentation, as you've got, would be a good idea.

Is there a way to get NOOBS onto an SD card using Chrome OS?

srdjanrosic:
I think yes.

I've never tried NOOBS personally, but Chrome OS built-in Files app will let you right click on the SD card and format it as FAT32. You can then extract NOOBS .zip archive onto it.

andrum99:
I think it would be good to have both options in the docs, although you may wish to wait and see if this gets accepted.

JamesH65:
I think this is a good addition. I cannot, though, check that the instructions actually work! If anyone out there who isn't the author can check they work that would be most excellent. I will pass on to the copy editors in the meantime. There is a bit of a delay at the moment in copy editing.

PR #1092, opened 2019-03-02
Update security.md

thecolorjay:
I was going using the documentation myself and noticed some of the commands didn't have the intended effect, you'll see why.

### DIFF
--- a/configuration/security.md
+++ b/configuration/security.md
@@ -40,23 +40,29 @@ You will be prompted to create a password for the new user.
 
 The new user will have a home directory at `/home/alice/`.
 
-To add them to the `sudo` group to give them `sudo` permissions:
+To add them to the `sudo` group to give them `sudo` permissions as well as all of the other necessary permissions:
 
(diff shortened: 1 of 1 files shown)


andrum99:
+1 to removing the names of the Queen's offspring from the example usernames. The official Pi docs should be politically neutral, I would suggest.

JamesH65:
Royal Family are apolitical. So fair game.

helenlynn:
@JamesH65 I can't edit this file. I have two very minor edits I'd make, both rather nitpicking, neither currently affecting sense.

JamesH65:
The original author has deleted their github account, so I think that is why you cannot edit. Might need to merge and edit afterward.

lurch:
I guess the sudoers stuff also has small overlaps with #1195

JamesH65:
OK, since user has deleted their github account, going to merge this without CE as that seem to be problematical. I've read it through and it's not insane.

PR #1048, opened 2018-12-23
Update building.md

duchampdev:
list cross-compiling dependencies based on Ubuntu 18.10 "minimal installation"

PR #1100, opened 2019-03-14
Add new kernel build dependencies

pelwell:
Building current kernels needs a few more tools to be installed.


lurch:
Should https://github.com/raspberrypi/documentation/blob/master/linux/kernel/building.md#local-building and https://github.com/raspberrypi/documentation/blob/master/linux/kernel/building.md#build-sources be showing the same list of dependencies? (see also #1048 )

PR #1106, opened 2019-03-21
Update samba.md

mikerr:
Remove defaults & fix conflicting options (public vs valid users)


JamesH65:
Going to take a punt on this and assume that the change is correct! If anyone disagrees I can revert.

PR #1113, opened 2019-03-27
Update building.md

igormartsekha:
Wrong mount names

### DIFF
--- a/linux/kernel/building.md
+++ b/linux/kernel/building.md
@@ -169,8 +169,8 @@ Mount these first, adjusting the partition numbers for NOOBS cards:
 mkdir mnt
 mkdir mnt/fat32
 mkdir mnt/ext4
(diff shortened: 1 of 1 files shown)


XECDesign:
The original numbers are fine. It looks like you're using a noobs card, in which case those numbers aren't guaranteed either.

JamesH65:
@XECDesign NOOBS not my area, any thoughts?

XECDesign:
The instructions assume a normal image and already point out noobs partitions will need to be adjusted, right? Still, 6 and 7 are not guaranteed to always be right, varying with the number of operating systems installed.

lurch:
I wonder if changing "adjusting the partition numbers for NOOBS cards" to "adjusting the partition numbers for NOOBS cards (as necessary)" might clear this up? :man_shrugging:

PR #1116, opened 2019-04-05
Use `ssh-copy-id` to copy the public SSH key to the server

ftheile:


ftheile:
I've seen this was recommended before, sorry...

bennuttall:
Can you please open an issue suggesting this, and pointing out where it should be changed? The original suggestion was 5 years ago, I think it's time to revisit. Please mention that I recommended you do so. Thanks :+1:

PR #1120, opened 2019-04-14
--datestamp should be --datetime

pconwell:
--datestamp is not a valid option for raspistill. The correct option is --datetime.

There is a --timestamp option, so potentially it would be better (i.e. more consistent) to change the code for raspistill to --datestamp, but for now the correct (valid) option is --datetime and not --datestamp, so the documentation should be updated to match the code.

### DIFF
--- a/raspbian/applications/camera.md
+++ b/raspbian/applications/camera.md
@@ -438,7 +438,7 @@ The specific value is the time between shots in milliseconds. Note that you shou
 Specifies the first frame number in the timelapse. Useful if you have already saved a number of frames, and want to start again at the next frame.
 
 ```
---datestamp,   -dt
+--datetime,   -dt
 ```
 
 Instead of a simple frame number, the timelapse file names will use a date/time value of the format `aabbccddee`, where `aa` is the month, `bb` is the day of the month, `cc` is the hour, `dd` is the minute, and  `ee` is the second.


lurch:
Yep, confirmed by https://github.com/raspberrypi/userland/blob/master/host_applications/linux/apps/raspicam/RaspiStill.c#L216

lurch:
@JamesH65 https://github.com/raspberrypi/documentation/blob/master/raspbian/applications/camera.md says "as of 28 November 2018" and https://github.com/raspberrypi/userland/blob/master/host_applications/linux/apps/raspicam/README.md says "as of February 19th 2018".

Does it make sense to have the same info duplicated (and edited separately) in two separate places? Should the version in userland be updated to say something like "Please refer to https://www.raspberrypi.org/documentation/raspbian/applications/camera.md " ?

pconwell:
I honestly didn't realize there were two different readme files/locations. I think it would make much more sense to only have one location. The raspicam code base only exists in one location, right? As in, there is no conceivable reason to need two different sets of documentation? In other words, there are not two versions / code bases of raspicam floating around, right?

JamesH65:
There's only one source code repo, this is a mismatch between the readme in the source, and the documentation. I think the best thing is to simply remove the README in the source and point to the documentation instead. It a hangover from when there was no documentation repo, and since I wasn;t working at Pi at the time I wrote raspistill I did the README as documentation.

JamesH65:
OK, have merged this, and updated the README in the source to point to the camera docs rather than duplicate the information.

PR #1122, opened 2019-04-17
Improve configuration/external-storage.md

andrum99:
Add 'nofail' option to fstab entry in the docs, so if user unplugs their external storage the Pi will still boot.

### DIFF
--- a/configuration/external-storage.md
+++ b/configuration/external-storage.md
@@ -86,7 +86,7 @@ Before you unmount your HDD, ensure that there are no programs accessing the HDD
     ```
    where `PIHDD` is the mount point name.
 3. Manually close all the programs that are using the mount point.  
-4. Unmount the HDD. Run the following command:
+4. Unmount the HDD:
 
     ```
     sudo umount /mnt/PIHDD
--- a/configuration/external-storage.md
+++ b/configuration/external-storage.md
@@ -28,24 +28,24 @@ You can mount your HDD at a specific folder location. It is conventional to do t
     sudo apt-get update
     sudo apt-get install ntfs-3g
     ```
-4. Run the following command to get the location of the disk partition:
+6. Run the following command to get the location of the disk partition:
 
     ```
     sudo blkid
     ```
     For example, `/dev/sda1`.
-5. Create a target folder to be the mount point of the HDD. 
+7. Create a target folder to be the mount point of the HDD. 
    The mount point name used in this case is `PIHDD`. You can specify a name of your choice:
 
     ```
     sudo mkdir /mnt/PIHDD
     ```
-6. Mount the HDD at the mount point you created:
+8. Mount the HDD at the mount point you created:
 
     ```
     sudo mount /dev/sda1 /mnt/PIHDD
     ```
-7. Verify that the HDD is mounted successfully by listing the contents:
+9. Verify that the HDD is mounted successfully by listing the contents:
 
     ```
     ls /mnt/PIHDD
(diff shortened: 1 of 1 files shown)


JamesH65:
I know very little about this (old brain is full), so would welcome comments from others!

XECDesign:
I've seen a few people run into trouble without this on the forum, so it's probably a good idea.

JamesH65:
Found this statement on ArchWiki, "The nofail option is best combined with the x-systemd.device-timeout option. This is because the default device timeout is 90 seconds, so a disconnected external device with only nofail will make your boot take 90 seconds longer, unless you reconfigure the timeout as shown. Make sure not to set the timeout to 0, as this translates to infinite timeout."

G-Hopper:
Good idea, we just had a few reeeally long minutes because of the missing nofail option.

andrum99:
I didn't know about the x-systemd.device-timeout option - I've added it to my setup 😀 Do you want me to add it to this pull request?

andrum99:
Just noticed the page doesn't explain the need to put the filesystem the user's drive uses (e.g. NTFS) in the fstab entry - oops. Adding this here so I remember to address that 😜

andrum99:
I've made some additional changes. This PR now fixes the following problems:
- add nofail option to fstab entry
- remove exfat from fstab entry and explain need to specify correct filesystem type

It also includes the following changes to make the page easier to understand, and more easy to follow:
- move section about the exfat driver so it is inline within the instructions for mounting, rather than in its own section
- reword some sentences
- simplify wording of some steps
- remove unnecessary step: deleting the mount point

andrum99:
I've added some explanation of the fact that the Pi can be booted without the HDD attached, that that will take 90 seconds longer, and how to use the x-systemd.device-timeout option to change the timeout, and in passing what timeout means

andrum99:
OK, that's this one ready for review 😀

configuration/external-storage.md
```diff
@@ -22,6 +22,12 @@ You can mount your HDD at a specific folder location. It is conventional to do t
     sudo apt-get update
     sudo apt-get install exfat-fuse
     ```
+5. If your HDD uses an NTFS file system, you will have read-only access to the HDD. If you want to write to the HDD, you can install the ntfs-3g driver:
```
lurch: For the benefit of future editors, it might be nice to re-number the subsequent bullet points? :wink:

JamesH65: I don't thint these are markup bullet numbers, otherwise you just use 1. for all of them and it renumbers automagically.

JamesH65:
Thanks for the updates, I'll push to copy edit, but we are massively delayed on those at the moment, so no idea when it will be done.

configuration/external-storage.md
```diff
@@ -1,99 +1,110 @@
-# External Storage configuration
+# External storage configuration
 You can connect your external hard disk drive (HDD) to any of the USB ports on the Raspberry Pi, and mount the file system to access the data stored on your HDD.
```
lurch: Is it worth clarifying that this is e.g. "external hard disk drive or USB flash drive", since I believe from the Raspbian-side of things they work the same?

configuration/external-storage.md
```diff
@@ -1,99 +1,110 @@
-# External Storage configuration
+# External storage configuration
 You can connect your external hard disk drive (HDD) to any of the USB ports on the Raspberry Pi, and mount the file system to access the data stored on your HDD. 
 
-By default, Raspberry Pi automatically mounts some of the popular file systems such as FAT, NTFS, and HFS+ at `/media/pi/<HARD-DRIVE-LABEL>` location. 
+By default, Raspberry Pi automatically mounts some of the popular file systems such as FAT, NTFS, and HFS+ at `/media/pi/<HARD-DRIVE-LABEL>` location.
```
lurch: One for the copy-editors, but I dunno if this should be "Raspberry Pi" or "Raspbian" ?

configuration/external-storage.md
```diff
@@ -1,99 +1,110 @@
...

+4. Add the following line in the `fstab` file:
 
     ```
-    UUID=5C24-1453 /mnt/PIHDD exfat defaults,auto,umask=000,users,rw 0 0
+    UUID=5C24-1453 /mnt/PIHDD FSTYPE defaults,auto,umask=000,users,rw,nofail 0 0
     ```
+   Replace FSTYPE with the type of your file system, which you found in step 2 of 'Mounting an HDD' above.
+
+Now that you have set an entry in `fstab`, you can start up your Raspberry Pi with or without the HDD attached. Before you disconnect the HDD you must either shut down the Pi, or manually unmount the HDD using the steps in 'Unmounting an HDD' below.
```
lurch: disconnect -> unplug ?

configuration/external-storage.md
andrum99: I've added a bit at the start to specify hard disk, SSD and USB flash drive, then replaced HDD with 'storage device' (or 'device' for subsequent uses in the same paragraph to keep verbage down).

configuration/external-storage.md
andrum99: fixed

configuration/external-storage.md
andrum99: I will leave that for the copy editors, as I'm not sure what the convention is. It's kind of a non-issue anyway.

andrum99:
Thanks James.

configuration/external-storage.md
burtyb: If it says "Raspbian" it should probably say "Raspbian with desktop" or something as Lite doesn't mess with it.

andrum99: I don't want to make it too complicated. Raspbian with desktop is the default anyway - folk can always ask on the forums if they wonder why it's not automounted on lite.

PR #1125, opened 2019-04-23
Needed a Troubleshooting section

daveyostcom:
Point to a great summary on StackExchange of all the things that have to be right.


JamesH65:
I've started a troubleshooter on this branch, still has gaps. https://github.com/raspberrypi/documentation/tree/JamesH65-patch-troubleshooting

JamesH65:
Since we are working on an internal troubleshooting system, and also I don't like linking to unpredictable offsite pages, will close this. Thanks for the prompt to sort this out!

PR #1126, opened 2019-04-25
USB host/device boot - various improvements

andrum99:
The pages which describe the various USB host boot modes are inconsistent in what they say about which models of Pi can do which boot mode. Standardise all appropriate pages on the same wording (or very similar, given some describe a "sub-mode" of USB host boot i.e. MSD / net boot), and fix one instance of where the wording is slightly wrong.


hardware/raspberrypi/bootmodes/usb.md
```diff
@@ -7,7 +7,7 @@ There are two separate boot modes for USB:
   * [USB mass storage boot](msd.md)
   * [Network boot](net.md)
 
-Note that network boot is only possible on Raspberry Pi models that have a built-in wired Ethernet interface.
+**USB host boot is available on the Raspberry Pi 3B, 3B+, 3A+ and Raspberry Pi 2B v1.2 models only.**
```
lurch: Not sure that this needs to be mentioned here, given that it's also mentioned in host.md? :man_shrugging:

andrum99: It's already mentioned on this page, I'm just changing the wording.

andrum99: I personally think it should be mentioned on both pages, but I'm happy to change it. Possibly one for the copy editors since it is kind of style related.

lurch: It's just that this page mentions both "USB device boot" and "USB host boot" so I wondered why you were specifically only adding a note about "USB host boot"? :man_shrugging:  (and I like to avoid duplication where possible, but that might just be a personal thing)

I agree it's probably a copy-editors' decision.

andrum99:
I was leaving dealing with device boot to another PR. I also hadn't realised we had this inconsistency between how this page documents device versus host boot. It would therefore make sense to include fixing up the docs for device boot in this PR so this question of consistency can be addressed directly.

andrum99:
@ghollingworth Is the list of which devices can do USB device mode correct? Also, there is an issue around precedence of the OTP bits for device and host boot mode - can you possibly explain that so I can document it?

hardware/raspberrypi/bootmodes/usb.md
andrum99: I agree that we probably don't need the warning about which Pi's do USB host boot in that page. I have removed that, and instead added '(only available on certain models)' to the end of the first sentence. I think that is simpler, but still retains a warning that all Pi's don't do the USB boot modes. If folk want to know more, it should be obvious that they just click on the relevant sub-page.

hardware/raspberrypi/bootmodes/net.md
```diff
@@ -1,5 +1,7 @@
 # Network booting
 
+**Network booting is available on the Raspberry Pi 3B, 3B+, 3A+ and Raspberry Pi 2B v1.2 models only.**
```
lurch: The 3A+ doesn't have a built-in ethernet  adaptor... :wink:

andrum99: Sorry - I'm getting confused between device boot and host boot I think.

andrum99: Copy and paste error, combined with the fact I failed to remember the fact you've just pointed out. Fixed.

hardware/raspberrypi/bootmodes/host.md
```diff
@@ -1,6 +1,6 @@
 # USB host boot mode
 
-**USB host boot is available on the Raspberry Pi 3B, 3B+, 3A+ and Raspberry Pi 2B v1.2 models only.**
+**USB host boot is available on the Raspberry Pi 3B, 3B+ and Raspberry Pi 2B v1.2 models only.**
```
lurch: I've not tested it myself, but I believe the 3A+ should be able to do host-MSD boot, it just can't do host-net boot.

andrum99: Oops - I've edited the wrong file

andrum99: fixed

hardware/raspberrypi/bootmodes/host.md
```diff
@@ -1,4 +1,6 @@
-# USB Host boot mode
+# USB host boot mode
+
+**USB host boot is available on the Raspberry Pi 3B, 3B+, and Raspberry Pi 2B v1.2 models only. The 3A+ supports USB host boot as well, but only for mass storage device boot, not network boot.*
```
lurch: Perhaps this would sound better as "USB host boot is available on the Raspberry Pi 3B, 3B+, 3A+ and Raspberry Pi 2B v1.2 models only. The 3A+ only supports mass storage boot, not network boot." ?

andrum99: I agree - that's probably what the copy editors would have changed it to 👍.

andrum99:
I've also combined the pages describing the BCM2837A0 and B0. Apart from the fact that they are basically the same chip, the B0 is now also being used on the Pi 2 in place of the original BCM2837.

lurch:
> I've also combined the pages describing the BCM2837A0 and B0.

IMHO this would be better if done in a separate PR.

andrum99:
> > I've also combined the pages describing the BCM2837A0 and B0.
> 
> IMHO this would be better if done in a separate PR.

I think you're right. I will see what James says - I'm happy to back it out.

andrum99:
Backed out merging of BCM2837 and BCM2837B0 pages

andrum99:
OK, that's this one ready for review 😀

hardware/raspberrypi/bootmodes/bootflow.md
```diff
@@ -1,6 +1,8 @@
...
+**The following boot sequence applies to the BCM2837-based models of Raspberry Pi only. On all other models, the Pi will try [SD card boot](sdcard.md), following by [USB device mode boot](device.md).**
+
+The flow of boot begins with reading the OTP to determine which boot modes are enabled. By default, this is SD card boot followed by USB device boot. Subsequently, the boot ROM checks to see if the GPIO boot mode OTP bits have been programmed — one to enable GPIO host boot mode, and one to select the bank of GPIOs it uses to disable boot modes (low = GPIOs 22-26, high = GPIOs 39-43). This makes it possible to use hardware attached to the GPIO connector to choose between different boot modes.
```
lurch: I dunno why you've changed "GPIO boot mode" to "GPIO host boot mode" ?

andrum99: Neither do I. It would help if we had some documentation for what the 5 GPIOs actually do when GPIO boot mode is enabled. Thank you for your attentiveness - you're asking all the right questions 👍

@ghollingworth Can you possible explain what each of the 5 GPIOs does?

hardware/raspberrypi/bootmodes/usb.md
```diff
@@ -7,6 +7,6 @@ There are two separate boot modes for USB (only available on certain models):
...
   * [Network boot](net.md)
 
-The choice between the two boot modes is made by the firmware at boot time when it reads the OTP bits. There are two bits to control USB boot: the first enables device boot and is enabled by default on all Raspberry Pi computers. The second bit enables USB host boot; if this bit is also set, then the processor reads the OTGID pin to decide whether to boot as a host (driven to zero as on the Raspberry Pi Model B) or as a device (left floating). The Pi Zero has access to this pin through the OTGID pin on the USB connector, and the Compute Module has access to this pin on the edge connector.
+The choice between the two boot modes is made by the firmware at boot time when it reads the OTP bits. There are two bits to control USB boot: the first enables USB device boot, and the second enables USB host boot. If the USB host boot mode bit is set, then the processor reads the OTGID pin to decide whether to boot as a host (driven to zero as on the Raspberry Pi Model B) or as a device (left floating). The Pi Zero has access to this pin through the OTGID pin on the USB connector, and the Compute Module has access to this pin on the edge connector.
```
lurch: Given that the USB host boot mode has to be manually enabled on all Pis without a BCM2837B0, I think it might be worth leaving in the text clarifying that USB device boot is enabled by default (i.e. _doesn't_ need to be manually enabled).
AFAIK USB device boot **is** enabled on all Pis, it's just that on Pi's with a built-in USB hub (i.e. all the B/B+ models) the downstream hub forces the SoC's USB port into host mode, and prevents it from being able to switch into device mode.  But I'm no USB expert... :wink: 

There's also a little bit of duplication between the text here and the more-detailed info in the bootflow.md page, but that's probably not worth worrying about?

hardware/raspberrypi/bootmodes/bootflow.md
lurch: > Thank you for your attentiveness

Yeah, many people have told me I have a great eye for detail  :eye: :eyes: :wink: 

> if we had some documentation for what the 5 GPIOs actually do

That's explained further down in the bootflow.md page? @ghollingworth wrote much of the bootloader code and these bootloader docs, and he's better at the former than the latter :joy:

hardware/raspberrypi/bootmodes/usb.md
andrum99: I've added a bit back so it now reads 

> the first enabled USB device boot which is enabled by default

In other words, the net effect is to remove the bit that says "on all Raspberry Pi's". If folk want to desolver the LAN7500 / 9500 chip to get device boot working, the page still provides enough information to tell them that this can be done. It just no longer suggests that device mode works on all models of Pi (because it doesn't).

andrum99:
This one should be ready to go to CE now.

burtyb:
The  3A+ shouldn't be on the "USB device mode" page as the OTG_ID pin is tied to ground.

andrum99:
> The 3A+ shouldn't be on the "USB device mode" page as the OTG_ID pin is tied to ground.

Thanks. I'm awaiting confirmation of the device list for USB device boot mode from the folks at Pi towers.

andrum99:
OK, so this is awaiting one more piece of information: confirmation of which Raspberry Pis can do USB device mode boot @JamesH65 @ghollingworth

JamesH65:
I've no idea, but when you think you have completed a PR (not just this one) can you post asking me to push it to copy edit. Otherwise, I have no idea whether the work has been completed.

andrum99:
Will do - thanks

JamesH65:
Is this ready to go to copy edit?

andrum99:
> Is this ready to go to copy edit?

Not quite - need confirmation from @ghollingworth that the list of devices that can use USB device mode boot is correct, namely:

- Compute Module
- Compute Module 3 and 3+
- Pi Zero, Zero W and WH
- Pi A and 3A+

ghollingworth:
Basically anything that doesn't have a hub in the way!

burtyb:
> Basically anything that doesn't have a hub in the way!

But not the 3A+ ( https://www.raspberrypi.org/forums/viewtopic.php?f=63&t=227044#p1392654 ).

ghollingworth:
More recent versions of 3A+ will work...  It's only the first few thousand units that had the USB host boot bit set

lurch:
@ghollingworth Ahhh, so does that mean that "recent" 3A+ boards function like the 3B, in that they won't USB host-boot (i.e. boot off a USB flash drive) until the relevant OTP bit is explicitly set? (which presumably also then disables USB device-boot, like on "early" 3A+ boards?)

andrum99:
Thanks Gordon - this one is now ready to go to CE @JamesH65

hardware/raspberrypi/bootmodes/device.md
```diff
@@ -1,13 +1,15 @@
...
+The following devices can boot using USB device boot mode:
 
-* Pi CM
-* Pi CM3
+* Pi Compute Module 
+* Pi Compute Module 3
 * Pi Zero
 * Pi Zero W
+* Pi A
+* Pi 3A+
```
lurch: Also the Pi A+ ?

andrum99: Ooops - forgot about that one - fixed

burtyb:
> More recent versions of 3A+ will work... It's only the first few thousand units that had the USB host boot bit set

That's good to know but it's a shame the post was just closed and no one bothered to update it with the current status.

andrum99:
> > More recent versions of 3A+ will work... It's only the first few thousand units that had the USB host boot bit set
> 
> That's good to know but it's a shame the post was just closed and no one bothered to update it with the current status.

It would certainly have been useful if Gordon (@ghollingworth) had mentioned that nugget of information on the forum thread before locking it. Don't know if it is possible to unlock it now and do that - paging @ghollingworth 😉

Issue #793, opened 2018-01-02
bootflow.md: please clarify how to disable boot modes via GPIO


rd235:
from bootflow.md:
*Uses program_gpio_bootmode to disable some modes by reading GPIOs 22-26 or 39-43 to see if the default values do not equal the default pull to '0'. If it is low, it will disable that boot mode for each of SD1, SD2, NAND, SPI, USB. If the value read is a '1', then that boot mode is enabled (note this cannot enable boot modes that have not already been enabled in the OTP). The default pull resistance is around 50K ohm, so a smaller pull up of 5K should suffice to enable the boot mode but still allow the GPIO to be operational without consuming too much power.*
This seems to mean that all boot mode are *disabled* by default: in fact, all the pins 22-26 or 39-41 are down at startup time.
(see pag 102 of https://www.raspberrypi.org/documentation/hardware/raspberrypi/bcm2835/BCM2835-ARM-Peripherals.pdf -- these values should be the same for BCM2837)
So, from your boot flow description at least one 5K resistor should be needed to *enable* the desired boot mode.
On the contrary there are on the net several videos showing how to boot a RPi3 from MSD or network, and no one mentions about the need of resistors.
Please rephrase the sentence above to clarify when pull up resistors are needed and which is the behavior with and without the pullup.
Thank you.

JamesH65:
Ping @ghollingworth Is this one for you?

ghollingworth:
The above is only valid when program_gpio_bootmode is set...  Once this bit has been set then the boot modes will be disabled, you'll then require a pull up to re-enable your specific boot mode

Gordon

JamesH65:
Do we need to clarify the docs?

ghollingworth:
No I don't think so, it says this at the top of the document.  Although James, maybe just have a read through and make sure you agree...  Obviously being the person to write the data does tend to give one a different viewpoint!

rd235:
There are two OTP flags: _program_usb_boot_mode_ and _program_gpio_bootmode_. Somewhere I would say that _program_gpio_bootmode_ is *not* needed to boot a RPI from a MSD or from the network. (am i misundertanding the specifications?)
When program_gpio_bootmode is UNSET, are all the devices enabled (SD1, SD2, NAND, SPI, USB) or just SD1 and USB (provided _program_usb_boot_mode_ is 1)?

ghollingworth:
[...]

PR #1131, opened 2019-04-30
Document GPIO boot mode on its own page

andrum99:
Currently, there is no specific page which documents GPIO boot mode. This PR proposes to add one. This will clarify that this is a specific boot mode, enabled by a specific GPIO. It will also document exactly which GPIO activates which type of boot (SD, MSD, Ethernet etc) - the current docs are not ideal in this regard.

### DIFF
--- a/hardware/raspberrypi/bootmodes/gpio.md
+++ b/hardware/raspberrypi/bootmodes/gpio.md
@@ -15,11 +15,11 @@ Once GPIO boot mode is enabled, the Raspberry Pi will no longer boot. You must p
 Once enabled, the GPIOs are allocated to boot modes as follows:
 
 |Bank 1|Bank 2|boot type|
-|:----:|:---:|:---- ---:|
+|:----:|:---:|:--------:|
 |22    |39   |SD0       |
 |23    |40   |SD1       |
-|24    |41   |NAND      |
-|25    |42   |SPI       |
+|24    |41   |NAND (no Linux support at present)    |
+|25    |42   |SPI (no Linux support at present)    |
 |26    |43   |USB       |
 
 The various boot modes are attempted in the numerical order of the GPIO pins, i.e. SD0, then SD1, then NAND and so on.
--- a/hardware/raspberrypi/bootmodes/gpio.md
+++ b/hardware/raspberrypi/bootmodes/gpio.md
@@ -1,1 +1,23 @@
 # GPIO boot mode
+
+The Raspberry Pi can be configured to allow the boot mode to be selected at run time using hardware attached to the GPIO connector: this is GPIO boot mode. This is done by setting a bit in the OTP memory of the SoC. Once the bit is set, it permanently allocates 5 GPIOs to allow this selected to be made. Once the OTP bit is set it cannot be undone, so you should think carefully about enabling this, since those 5 GPIO lines will then be permanently unavailable for any other use.
+
+To enable GPIO boot mode, add the following line to the config.txt file:
+
+```
+program_gpio_bootmode=n
+```
+
+And reboot the Pi once to program the OTP with this setting. Where n is the bank of GPIOs which you wish to use. Bank 1 is GPIOs 22-26, bank 2 is GPIOs 39-43. Unless you have a compute module, you must use bank 1: the GPIOs in bank 2 are only available on the compute module.
+
+Once enabled, the GPIOs are mapped as follows:
+
+|Bank 1|Bank2|boot type|
+|:----:|:---:|:-------:|
+|22    |39   |SD1      |
+|23    |40   |SD2      |
+|24    |41   |NAND     |
+|25    |42   |SPI      |
+|26    |43   |USB      |
+
+SD1 is the <xxx> SD card interface. SD2 is the <yyy> SD card interface. USB enables both USB device mode and USB host mode boot.
--- a/hardware/raspberrypi/bootmodes/gpio.md
+++ b/hardware/raspberrypi/bootmodes/gpio.md
@@ -2,7 +2,7 @@
 
 The Raspberry Pi can be configured to allow the boot mode to be selected at power on using hardware attached to the GPIO connector: this is GPIO boot mode. This is done by setting bits in the OTP memory of the SoC. Once the bits are set, they permanently allocate 5 GPIOs to allow this selection to be made. Once the OTP bits are set they cannot be unset so you should think carefully about enabling this, since those 5 GPIO lines will then be unavailable for any other use.
 
-To enable GPIO boot mode, add the following line to the [config.txt](configuration/config-txt/README.md) file:
+To enable GPIO boot mode, add the following line to the [config.txt](../../../configurationconfig-txt/README.md) file:
 
 ```
 program_gpio_bootmode=n
@@ -28,4 +28,4 @@ SD1 is the Broadcom SD card / MMC interface, which is used by default for the bu
 
 USB in the table selects USB device boot mode and USB host boot mode, which are only available on certain models of Raspberry Pi.
 
-The default pull resistance on the GPIO lines is 50K ohm, as documented on page 102 of the [BCM2835 ARM peripherals datasheet](hardware/raspberrypi/bcm2835/BCM2835-ARM-Peripherals.pdf). A pull resistance of 5K ohm is recommended to pull a GPIO line up: this will allow the GPIO to function but not consume too much power.
+The default pull resistance on the GPIO lines is 50K ohm, as documented on page 102 of the [BCM2835 ARM peripherals datasheet](../../hardware/raspberrypi/bcm2835/BCM2835-ARM-Peripherals.pdf). A pull resistance of 5K ohm is recommended to pull a GPIO line up: this will allow the GPIO to function but not consume too much power.
--- a/hardware/raspberrypi/bootmodes/gpio.md
+++ b/hardware/raspberrypi/bootmodes/gpio.md
@@ -26,7 +26,7 @@ The various boot modes are attempted in the numerical order of the GPIO pins, i.
 
 SD0 is the Broadcom SD card / MMC interface. When the boot ROM within the SoC runs, it always connects SD0 to the built-in SD card slot, or on the Compute Module, to the eMMC device. SD1 is the Arasan SD card / MMC interface which is also capable of SDIO. All Raspberry Pi models with built-in wifi use SD1 to connect to the wifi chip via SDIO.
 
-It is possible to reassign both SD0 and SD1 to different uses after the Pi has booted, but the boot ROM will always use the assignments noted above at boot time .
+It is possible to reassign both SD0 and SD1 to different uses after the Pi has booted, but the boot ROM will always use the assignments noted above at boot time.
 
 USB in the table selects USB device boot mode and USB host boot mode, which are only available on certain models of Raspberry Pi. In order to use a USB boot mode, it must by enabled in the OTP memory. For more information, see [USB device boot](device.md) and [USB host boot](host.md).
 
--- a/hardware/raspberrypi/bootmodes/gpio.md
+++ b/hardware/raspberrypi/bootmodes/gpio.md
@@ -1,6 +1,8 @@
 # GPIO boot mode
 
-The Raspberry Pi can be configured to allow the boot mode to be selected at power on using hardware attached to the GPIO connector: this is GPIO boot mode. This is done by setting bits in the OTP memory of the SoC. Once the bits are set, they permanently allocate 5 GPIOs to allow this selection to be made. Once the OTP bits are set they cannot be unset so you should think carefully about enabling this, since those 5 GPIO lines will then be unavailable for any other use.
+**GPIO boot mode is only available on the Raspberry Pi 3B and 3B+.**
+
+The Raspberry Pi can be configured to allow the boot mode to be selected at power on using hardware attached to the GPIO connector: this is GPIO boot mode. This is done by setting bits in the OTP memory of the SoC. Once the bits are set, they permanently allocate 5 GPIOs to allow this selection to be made. Once the OTP bits are set they cannot be unset so you should think carefully about enabling this, since those 5 GPIO lines will always control booting. Although you can use the GPIOs for some other function once the Pi has booted, you must set them up so that they enable the desired boot modes when the Pi boots.
 
 To enable GPIO boot mode, add the following line to the [config.txt](../../../configuration/config-txt/README.md) file:
 
@@ -12,7 +14,8 @@ Where n is the bank of GPIOs which you wish to use. Then reboot the Pi once to p
 
 Once GPIO boot mode is enabled, the Raspberry Pi will no longer boot. You must pull up at least one boot mode GPIO pin in order for the Pi to boot.
 
-Once enabled, the GPIOs are allocated to boot modes as follows:
+## GPIO boot mode pin assignments
+### Raspberry Pi 3B
 
 |Bank 1|Bank 2|boot type|
 |:----:|:---:|:--------:|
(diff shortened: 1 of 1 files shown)


andrum99:
This is a follow on from #1126

lurch:
> GPIO activates which type of boot (SD, MSD, Ethernet etc)

Nitpick: the GPIOs allow you to select to enable "SD1, SD2, NAND, SPI, USB" and USB covers both MSD and Ethernet, so you can't use GPIO to select between them :stuck_out_tongue_winking_eye: 

And since SD2, NAND and SPI boot aren't supported, I guess it effectively only allows you to select between SD and USB boot.

andrum99:
I would have discovered that when I read Gordon's existing docs. Which I will do tomorrow.

ghollingworth:
Depends if you have control over the OTG_ID pin on the Pi compute module, if you do then you can choose between host and device mode.

andrum99:
@ghollingworth Do you mean OTG_ID is exposed on the compute module's edge connector?

ghollingworth:
Yes it is...  For device mode it should be left open circuit, for host mode it should be pulled / driven low.  On the CMIO board there is a resistor R12 which pulls this low permanently, I would suggest removing this and then using a jumper on the GPIO connector (the USB_ID pin comes out on pin 168 on the edge connector which is J6 pin 55 on the CMIO3 board)

lurch:
@ghollingworth Out of curiosity: if OTG_ID is permanently pulled low on the CMIO board, how does the usbboot flashing mode work? Doesn't that require the CM to be in device mode?

andrum99:
Thanks for the extra information @ghollingworth - I will add that to the new page. I pinged you on a couple of other documentation PRs - don't know if you saw those 😉

andrum99:
@ghollingworth Can you confirm 3 things:

1. What is the current state of SPI and NAND boot?

2. Currently bootflow.md says this: "The primary SD card boot mode is, as standard, set to be GPIOs 49-53". Does this mean that by default the primary SD card interface is on GPIOs 49-53?

3. Which SD card interface is SD1, and which is SD2?

lurch:
https://elinux.org/RPi_BCM2835_GPIOs

I suspect 'SD1' being discussed in bootflow.md is 'SD0' in that GPIOs table, and 'SD2' in bootflow.md is 'SD1' in that table.

hardware/raspberrypi/bootmodes/gpio.md
```diff
@@ -0,0 +1,27 @@
+# GPIO boot mode
+
+The Raspberry Pi can be configured to allow the boot mode to be selected at run time using hardware attached to the GPIO connector: this is GPIO boot mode. This is done by setting a bit in the OTP memory of the SoC. Once the bit is set, it permanently allocates 5 GPIOs to allow this selected to be made. Once the OTP bit is set it cannot be undone, so you should think carefully about enabling this, since those 5 GPIO lines will then be permanently unavailable for any other use.
```
lurch: Perhaps "at power-on time" or "at bootup time", instead of "at run time"?

lurch: "this selected" -> "this selection"

hardware/raspberrypi/bootmodes/gpio.md
```diff
@@ -0,0 +1,27 @@
...
+
+To enable GPIO boot mode, add the following line to the config.txt file:
+
+```
+program_gpio_bootmode=n
+```
+
+And reboot the Pi once to program the OTP with this setting. Where n is the bank of GPIOs which you wish to use. Bank 1 is GPIOs 22-26, bank 2 is GPIOs 39-43. Unless you have a compute module, you must use bank 1: the GPIOs in bank 2 are only available on the compute module.
```
lurch: Might be better with the first two sentences swapped.

lurch: And I think Compute Module ought to be capitalised?

hardware/raspberrypi/bootmodes/gpio.md
```diff
@@ -0,0 +1,27 @@
...
+
+Once GPIO boot mode is enabled, the Raspberry Pi will no longer boot. You must pull up at least one boot mode GPIO pin in order for the Pi to boot.
```
lurch: Probably also worth explaining the priority-order, if multiple bootmodes are selected?

andrum99: @ghollingworth - is the boot order in order of GPIO number?

andrum99:
> https://elinux.org/RPi_BCM2835_GPIOs
> 
> I suspect 'SD1' being discussed in bootflow.md is 'SD0' in that GPIOs table, and 'SD2' in bootflow.md is 'SD1' in that table.

That still doesn't explain which hardware block they are.

hardware/raspberrypi/bootmodes/gpio.md
```diff
@@ -1,18 +1,18 @@
 # GPIO boot mode
 
-The Raspberry Pi can be configured to allow the boot mode to be selected at run time using hardware attached to the GPIO connector: this is GPIO boot mode. This is done by setting a bit in the OTP memory of the SoC. Once the bit is set, it permanently allocates 5 GPIOs to allow this selected to be made. Once the OTP bit is set it cannot be undone, so you should think carefully about enabling this, since those 5 GPIO lines will then be permanently unavailable for any other use.
+The Raspberry Pi can be configured to allow the boot mode to be selected at power on using hardware attached to the GPIO connector: this is GPIO boot mode. This is done by setting bits in the OTP memory of the SoC. Once the bits are set, they permanently allocates 5 GPIOs to allow this selection to be made. Once the OTP bits are set they cannot be unset so you should think carefully about enabling this, since those 5 GPIO lines will then be permanently unavailable for any other use.
```
lurch: allocates -> allocate

lurch:
> That still doesn't explain which hardware block they are.

I'm not sure what you mean?

andrum99:
> > That still doesn't explain which hardware block they are.
> 
> I'm not sure what you mean?

There are two different implementations of an SD card interface present in the chip, and they come from two different vendors. It would be useful to know which one is SD1 and which is SD2. (Also, they should probably be called SD0 and SD1 if that's how the rest of the docs refer to them).

lurch:
> There are two different implementations of an SD card interface present in the chip, and they come from two different vendors.

Hmmm, I wasn't aware of that! I did a bit of googling and eventually found https://github.com/raspberrypi/linux/issues/1536 where @pelwell says "The alt3 function on pins 22-27 is SD1 a.k.a. MMC. You want SD0 a.k.a. SDHOST which is alt0"

pelwell:
SD0 is the Broadcom in-house SD/MMC interface - known as SDHOST - which is now used by default on current images to drive the SD card or (on the Compute Module) eMMC. SD1 is the standard Arasan MMC block, but without ADMA/bus mastering support. This block is SDIO-capable, which is why it is used to connect the WiFi module. (SDHOST can almost do SDIO, but some over-aggressive clock gating prevents it from working.)

andrum99:
Technical points that need to be confirmed:

1. What is the current state of SPI and NAND boot?

2. Currently bootflow.md says this: "The primary SD card boot mode is, as standard, set to be GPIOs 49-53". What does this actually mean? I suspect it is not actually necessary to have this on that page, since it appears to be simply saying which SD/MMC interface is the default.

3. Is boot order in numerical order of the GPIO pins?

4. Which way round is the bank selection bit? In other words, if you enable GPIO boot mode on one bank, selecting which bank will give you the option to select the opposite bank in the future?

5. It would also be useful to know if there is a way to select only USB host mode, or only USB device mode using the GPIOs.

ghollingworth:
1) SPI and NAND booting are not supported in software.
2) The primary SD card boot (using SDHOST) is selected in the OTP to boot from GPIOs 49-53
3) Yes they are in the same order
4) In the OTP there are two bits, bit 19 enables GPIO_BOOTMODE bit 20 selects the bank
  0 == bank 0 
  1 == bank 1
  The OTP bits are by default 0 when unprogrammed and 1 when programmed

Just looking into GPIO usb control...  This is a little more complex

ghollingworth:
OK,

GPIO enables:

20 or 37 - SDHOST boot enable
21 or 38 - ARASAN boot enable
22 or 39 - NAND boot enable (unsupported in software for any further reading)
23 or 40 - SPI boot enable (unsupported in software for any further reading)
24 or 41 - USB device mode enable
25 or 42 - USB host mass storage enable
26 or 43 - USB host ethernet enable

So each of these bits only has the ability to disable a boot source already enabled in OTP (currently the only enableable modes are SDHOST boot, USB device, USB mass storage and USB ethernet.

SDHOST and USB device are enabled by default, USB host is enabled on some devices (Pi 3B+)

So for example you can use these bits to switch between USB device and USB host boot modes (it will therefore ignore the OTGID pin.)  Or you can use it to disable the USB mass storage mode and instead boot only from Ethernet even if a mass storage device is plugged in

andrum99:
Does this means the boot rom will not actually attempt to boot from SPI and NAND, or just that there is currently no support in our build of linux? Also you mention enabling boot modes in OTP - is this something we should mention at all, given it is not documented anywhere?

pelwell:
Careful, @ghollingworth - the GPIO assignments changed with the addition of new boot modes (2837B0?). Instead of adding the new modes on the end, everything else was moved along by 2.

andrum99:
Are GPIOs 49 to 53 wired to the SD card slot, or are these GPIOs compute module only?

pelwell:
GPIOs _48_ to 53 are wired to SD or eMMC on all models of Pi.

andrum99:
For the OTP bits, do you mean that the program_host_bootmode bit must be set before you can use host boot mode via GPIO boot mode? I take it we don't want to document any other OTP boot mode enablement bits at the mo?

lurch:
> GPIOs 48 to 53 are wired to SD or eMMC on all models of Pi.

...or to the edge-connector on the CM3-Lite modules? :wink: (where it then connects to the microSD slot on the CMIO v3 board)

andrum99:
> > GPIOs 48 to 53 are wired to SD or eMMC on all models of Pi.
> 
> ...or to the edge-connector on the CM3-Lite modules? 😉 (where it then connects to the microSD slot on the CMIO v3 board)

Split CM into separate sentence, awaiting details of where SD0 is connected to when the boot ROM is running.

andrum99:
@pelwell @ghollingworth This PR just needs confirmation of the GPIO assignments, and which Pis have different GPIO enables as it seems from what has been said that newer Pis have more of them.

hardware/raspberrypi/bootmodes/gpio.md
```diff
@@ -24,7 +24,7 @@ Once enabled, the GPIOs are allocated to boot modes as follows:
 
 The various boot modes are attempted in the numerical order of the GPIO pins, i.e. SD0, then SD1, then NAND and so on.
 
-SD0 is the Broadcom SD card / MMC interface. When the boot ROM within the SoC runs, it always connects SD0 to the built-in SD card slot. On Compute Modules with an eMMC device, SD0 is connected to that; on the Compute Module Lite SD0 is available on the edge connector and connects to the SD card slot in the CMIO carrier board. SD1 is the Arasan SD card / MMC interface which is also capable of SDIO. All Raspberry Pi models with built-in wifi use SD1 to connect to the wifi chip via SDIO.
+SD0 is the Broadcom SD card / MMC interface. When the boot ROM within the SoC runs, it always connects SD0 to the built-in microSD card slot. On Compute Modules with an eMMC device, SD0 is connected to that; on the Compute Module Lite SD0 is available on the edge connector and connects to the microSD card slot in the CMIO carrier board. SD1 is the Arasan SD card / MMC interface which is also capable of SDIO. All Raspberry Pi models with built-in wifi use SD1 to connect to the wifi chip via SDIO.
```
lurch: Perhaps the first "microSD" here should be "SD / microSD"? (because the original Pis had full-size SD slots)

lurch:
@andrum99 The documentation you've got now seems to imply that it's possible to boot from either SD0 _or_ SD1, but from what I understand it's only possible to boot from SD0? Gordon earlier said "currently the only enableable modes are SDHOST boot, USB device, USB mass storage and USB ethernet".

hardware/raspberrypi/bootmodes/gpio.md
andrum99: That was deliberate - I was assuming GPIO boot mode is only available on the 2837. But that has not been confirmed - paging @ghollingworth 😉

andrum99:
> @andrum99 The documentation you've got now seems to imply that it's possible to boot from either SD0 _or_ SD1, but from what I understand it's only possible to boot from SD0? Gordon earlier said "currently the only enableable modes are SDHOST boot, USB device, USB mass storage and USB ethernet".

As does the existing documentation. It's not entirely clear in what way these modes are "enabled" and the ones which aren't available are "disabled". It could be that the boot ROM will try all of the modes listed (or else, why bother documenting them?), but that there is currently no support in Linux.

hardware/raspberrypi/bootmodes/gpio.md
lurch: Ahhh, I hadn't considered that. But even _without_ GPIO boot mode, "When the boot ROM within the SoC runs, it always connects SD0 to the built-in SD card slot." is still true for the original Raspberry Pi models :wink:

andrum99: If GPIO boot mode is available on the old Pi's then I will change the wording, otherwise it is correct as-is.

hardware/raspberrypi/bootmodes/gpio.md
```diff
@@ -0,0 +1,33 @@
+# GPIO boot mode
+
+The Raspberry Pi can be configured to allow the boot mode to be selected at power on using hardware attached to the GPIO connector: this is GPIO boot mode. This is done by setting bits in the OTP memory of the SoC. Once the bits are set, they permanently allocate 5 GPIOs to allow this selection to be made. Once the OTP bits are set they cannot be unset so you should think carefully about enabling this, since those 5 GPIO lines will then be unavailable for any other use.
```
ghollingworth: The GPIO boot modes do not permanently allocate the GPIOs, you can still use the pins afterwards for input or output.  To enable a mode it just requires a pull up to 3v3 strong enough to override the internal pull down (we suggest 5k).

hardware/raspberrypi/bootmodes/gpio.md
ghollingworth: Yes it is...

hardware/raspberrypi/bootmodes/gpio.md
```diff
@@ -0,0 +1,33 @@
...
+Once enabled, the GPIOs are allocated to boot modes as follows:
+
+|Bank 1|Bank 2|boot type|
+|:----:|:---:|:--------:|
+|22    |39   |SD0       |
+|23    |40   |SD1       |
+|24    |41   |NAND (no Linux support at present)    |
+|25    |42   |SPI (no Linux support at present)    |
+|26    |43   |USB       |
+
```
ghollingworth: This is correct for Pi 3 but the following was added for Pi 3+

bank1   bank2   type
20         37         SD0
21         38         SD1
22         39         NAND
23         40         SPI
24         41         USB device
25         42         USB host - MSD
26         43         USB host - Ethernet

hardware/raspberrypi/bootmodes/gpio.md
ghollingworth: GPIO boot mode is only available on 3B and 3B+

andrum99: Thanks @ghollingworth. So just to confirm - the 2837 checks to see which model if Pi it is running, and only allows GPIO boot mode if it is in a 3B or 3B+, so on the 3A+ and the 2837-based compute modules setting program_gpio_bootmode=n does nothing?

pelwell: No - the ROM knows nothing about the board it is fitted to. I think what @ghollingworth meant is that GPIO boot mode is new for 2837, so not available on earlier Pis that don't have one.

lurch: I wouldn't have thought so - otherwise why mention "the GPIOs in bank 2 are only available on the Compute Module" ? :man_shrugging: 
I wonder if Gordon meant "GPIO boot mode is only available on the BCM2837A0 and BCM2837B0" ?

andrum99: Thanks - I've added 3A+ and CM3 and 3+. There's also the 2837-based Pi 2B just to confuse things, but I'm not sure that's really worth documenting.

hardware/raspberrypi/bootmodes/gpio.md
```diff
@@ -1,6 +1,6 @@
 # GPIO boot mode
 
-**GPIO boot mode is only available on the Raspberry Pi 3B and 3B+.**
+**GPIO boot mode is only available on the Raspberry Pi 3A+, 3B, 3B+, Computer Module 3 and 3+**
```
lurch: "Computer Module" -> "Compute Module" :slightly_smiling_face:

hardware/raspberrypi/bootmodes/gpio.md
andrum99: Just need @ghollingworth to confirm that GPIO boot mode is only available on **all** 2837-based Pi's, then this PR is ready for CE. 😀

lurch:
@andrum99 I don't think it adds anything that you've not already covered, but I've just noticed there's some documentation about SD0 and SD1 in section 9.1.4 of https://www.raspberrypi.org/documentation/hardware/computemodule/datasheets/rpi_DATA_CM3plus_1p0.pdf

lurch:
I guess this would also fix #793

andrum99:
> I guess this would also fix #793

I guess it would to. Pinging @JamesH65 as this one is ready to go, but I hadn't pinged him before to tell him - my bad 😮

JamesH65:
Pushed to copy edit.

andrum99:
Just to clarify - I've changed GPIO pins TO lines, since the GPIO pins are not necessarily in the same order as the GPIO lines on the chip. The GPIO lines on the chip are tested in numerical order to select the desired boot modes. This order is not necessarily the same as the order of those GPIOs on the connector.

pelwell:
Can we get this merged ASAP?

JamesH65:
Cannot wait for CE, so merging.

andrum99:
Note that this PR on its own does not create ANY links to the new page documenting GPIO boot mode. I have created a new PR, #1194, to add a suitable link from the main bootmodes documentation page. Pinging @JamesH65.

PR #1153, opened 2019-05-23
Add codec key OTP register numbers

ptesarik:
After setting decode_MPG2 and decode_WVC1 _and_ checking whether the
codec is enabled, these register contain a copy of the keys.


JamesH65:
Thanks.

PR #1161, opened 2019-06-07
raspberrypi/bootmodes/netboot_server_easy.md: Banish rpi-update

andrum99:
We don't need the branch=NEXT firmware and kernel for netboot any more. rpi-update is now considered toxic for mere mortals, so purge it from this page of the docs.


lurch:
The "expand filesystem" section should probably be removed too, as Raspbian now does that automatically.

andrum99:
Good catch. I've also changed references to 'apt-get' to the more modern 'apt' command.

PR #1160, opened 2019-06-07
linux/kernel/updating.md: Stronger warning about not using rpi-update

andrum99:
Various web pages recommend using rpi-update. Let's not make this any worse by doing the same in the official docs.

PR #1162, opened 2019-06-07
glossary/README.md: insert big warning about rpi-update usage

andrum99:


lurch:
What was wrong with the previous version of the warning?

andrum99:
> What was wrong with the previous version of the warning?

See forum discussion here:

https://www.raspberrypi.org/forums/viewtopic.php?f=28&t=241977

Basically someone complained that we could do with a stronger wording on rpi-update, and that the official Pi docs should not be recommending using it, hence this and other PRs. I think the wording I have proposed here is more like what we need, with perhaps some polish from CE.

andrum99:
Actually my warning isn't really any stronger than the old warning, it just says not to use it unless advised by a Raspberry Pi engineer earlier in the paragraph. I've also added the exact same warning to the pages in other PRs, namely #1160, #1163 and #1165.

I've also created associated PRs to get rid of recommendations to use rpi-update from the official docs - see PRs #1164 and #1161.

JamesH65:
I think the original is fine in this case.

PR #1169, opened 2019-06-13
SDXC formatting: Mac Disk Utility no longer allows FAT32

parkr:
Instead, you need to use the diskutil utility:

> $ sudo diskutil eraseDisk FAT32 SDCARD MBRFormat /dev/disk2

### DIFF
--- a/installation/sdxc_formatting.md
+++ b/installation/sdxc_formatting.md
@@ -6,7 +6,7 @@ The Raspberry Pi's bootloader, built into the GPU and non-updateable, only has s
 
 ## Linux and Mac OS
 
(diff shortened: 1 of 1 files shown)


andrum99:
Not sure why the mac instructions are lumped together with Linux. It would make sense to pit mac in a separate section under its own heading

andrum99:
After actually reading the section, it looks like it makes sense to leave Mac and Linux together in the one section.

parkr:
@JamesH65 @andrum99 Anything else you need from me?

JamesH65:
No, its flagged up to our copy editors, so once they have checked it over, it will be merged.

parkr:
@helenlynn What kind of technical check are you looking for?

JamesH65:
That's down to me. Once a CE has been done, I double check that there has been no technical inaccuracies introduced. Hardly ever happens, but just to be on the safe side.

PR #1170, opened 2019-06-18
Update headless.md

pomali:
Added snippet with whole wpa_supplicant.conf since documentation for wireless-cli mentions only changes (you don't have base to start from) and there are required parts like country in base you can't see in cli help.

Added tip from my experience where I try to connect RPi with older dongles to 5GHz network and debug for half an hour why it isn't working.


andrum99:
Your last sentence, jf it gets included at all, should probably be more like "Don't forget that not all wifi adapters support the 5GHz band", since almost all older adapters are 2.4GHz only.

PR #1186, opened 2019-06-26
Update samba.md to fix some minor issues

stickyPiston:


lurch:
Ping @bennuttall and @grega - the "Wikipedia page" link on https://github.com/raspberrypi/documentation/blob/master/remote-access/samba.md works correctly, but on https://www.raspberrypi.org/documentation/remote-access/samba.md the URL gets converted from `https://en.wikipedia.org/wiki/Samba_(software)` to `https://en.wikipedia.org/wiki/Samba_(software` which is a broken link. The closing-bracket gets displayed in the text, so I guess the md->html process isn't working correctly when a URL ends with a `)` ?
```
[Wikipedia page](https://en.wikipedia.org/wiki/Samba_(software))
```

cc @JamesH65

PR #1187, opened 2019-06-27
configuration/hdmi-config.md: fix wording of new Pi4 paragraph

andrum99:
Note that this may still be slightly wrong, as it is not clear whether the Pi can drive two displays simultaneously at resolutions above 1080p but below 4K _**AT 60Hz**_, and if so, what the limit actually is for simultaneous 60Hz across 2 displays.


andrum99:
You might want to pass this to CE, since some of the sentences are a bit convoluted and there may well be better ways of saying what they state.

JamesH65:
That paragraph is getting really clunky. Not sure what to do with it. Might have chat with the writers in the comms team - should be in tomorrow.

andrum99:
> That paragraph is getting really clunky. Not sure what to do with it. Might have chat with the writers in the comms team - should be in tomorrow.

It is clunky, but is at least closer to the truth now 😛

lurch:
I wonder if changing the bit in the middle to something like "At 4K resolution you can only output 60Hz to a single screen, the maximum dual-display 4K frequency is 30Hz." might make it a little less clunky? :man_shrugging: 
And IMHO the last sentence is missing a comma between "4Kp60" and "depending".

andrum99:
> I wonder if changing the bit in the middle to something like "At 4K resolution you can only output 60Hz to a single screen, the maximum dual-display 4K frequency is 30Hz." might make it a little less clunky? 🤷‍♂
> And IMHO the last sentence is missing a comma between "4Kp60" and "depending".

Not sure - with the sentence that way round the explanation of hdmi_enable_4k=1 is then in the wrong place. I will leave exactly where to put the commas to CE or other suitably qualified persons 😁

andrum99:
OK, I think that's as good as I can get it. I will leave it in the very capable hands of @JamesH65 to decide what, if any, further work is required 😁

JamesH65:
I've merged as is, seems good enough to me, if anyone complains we can fix it up later.

PR #1189, opened 2019-06-28
Rewrite SD create on Mac:  Use etcher or 1 example using dd

DG12:
Previous multiple edits have made these instructions convoluted. 

With the availability of Etcher and Apple's System Integrity Protection restrictions and the variations with internal /external SD drive and new/unformattted/previously used cards and with the risk of overwriting the running system

Is it a better idea to just let etcher be THE way?


JamesH65:
I'd like to get the opinion of other Mac users as whether this is the best approach. @spl237 Any thoughts?

spl237:
Just use Etcher. It Just Works. We shouldn't be encouraging people to mess about with dd when there is no need at all to do so.

lurch:
Perhaps the instructions could do with editing to clarify that all the `diskutil` and `dd` steps are only needed if you're **not** using Etcher?

lurch:
@DG12 I've just noticed that even after these edits, the `sudo diskutil eject ...` bit appears twice, which looks weird :confused: 
Also, there seems to be a stray ` from diskutil>` ??

JamesH65:
@lurch Had corrected the issues reported, but should we merge this given the issues with Etcher on the latest MacOS - or accept and simply add the warning about Catalina (?)?

JamesH65:
@lurch Can you give this a quick proof read? I've fixed it up so it will merge cleanly just want it checked to make sure it's still OK.

lurch:
Just made a small edit. I can't check the accuracy of the commands as I don't have a Mac, but overall it LGTM.
As I mentioned earlier (and @DG12 gave :-1: to), I'm not sure if it's clear that when using balenaEtcher, you _don't_ need to follow the "Determine SD device", "Copy the image", etc. steps? :man_shrugging:

JamesH65:
I think I am inclined to merge this, then fix up any issues that crop up later. So will do that now.

PR #1206, opened 2019-07-19
balena.com → balena.io (very different)

Xenthys:
The balena.com website is not related to the balenaEtcher tool. It's also misconfigured and ends up redirecting somewhere else.


lurch:
Oops, looks like we all overlooked this mistake in #1085 :slightly_frowning_face: 

Thanks for catching it so promptly.

PR #1210, opened 2019-07-23
configuration/config-txt/video.md: clarify hdmi_blanking option

andrum99:
The wording was not ideal. I would propose running this through copy edit to apply more polish.

Also, is there anything specific to the Pi 4 that we need to mention here?

### DIFF
--- a/configuration/config-txt/video.md
+++ b/configuration/config-txt/video.md
@@ -116,7 +116,7 @@ The `hdmi_pixel_encoding` command forces the pixel encoding mode. By default, it
 
 ### hdmi_blanking
 
-The `hdmi_blanking` command controls what happens when the operating system asks for the display to be put into standby mode, using DPMS, to save power. If this option is not set or set to 0, the HDMI output is blanked but not switched off. In order to mimic the behaviour of other computers, you can set the HDMI output to switch off as well by setting this option to 1. In response to the HDMI signal being switched off, the attached display will go into a low power standby mode.
+The `hdmi_blanking` command controls what happens when the operating system asks for the display to be put into standby mode, using DPMS, to save power. If this option is not set or set to 0, the HDMI output is blanked but not switched off. In order to mimic the behaviour of other computers, you can set the HDMI output to switch off as well by setting this option to 1: the attached display will go into a low power standby mode.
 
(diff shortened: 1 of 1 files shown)


generaleramon:
If i understand correctly the `hdmi_blanking` option is not working as it should on the Pi 4
https://github.com/raspberrypi/linux/issues/3050
https://www.raspberrypi.org/forums/viewtopic.php?f=28&t=244796#p1504694

andrum99:
> If i understand correctly the `hdmi_blanking` option is not working as it should on the Pi 4
> [raspberrypi/linux#3050](https://github.com/raspberrypi/linux/issues/3050)
> https://www.raspberrypi.org/forums/viewtopic.php?f=28&t=244796#p1504694

That bug has someone saying hdmi_blanking is not yet implemented on Pi 4. Is this true, and if so, is that something we should mention in the docs?

mehrvarz:
> That bug has someone saying hdmi_blanking is not yet implemented on Pi 4. Is this true, and if so, is that something we should mention in the docs?

This not a PI4 specific bug. It's also not really a hdmi issues, as I point out here: https://github.com/raspberrypi/linux/issues/3050#issuecomment-514609300

andrum99:
> > That bug has someone saying hdmi_blanking is not yet implemented on Pi 4. Is this true, and if so, is that something we should mention in the docs?
> 
> This not a PI4 specific bug. It's also not really a hdmi issues, as I point out here: [raspberrypi/linux#3050 (comment)](https://github.com/raspberrypi/linux/issues/3050#issuecomment-514609300)

I was not referring to the bug, but to one of the comments in it - @timg stated that hdmi_blanking is not yet implemented on Pi 4.

andrum99:
@JamesH65 Can you possibly confirm with timg (who I cannot contact via github) if hdmi_blanking is implemented yet on the Pi 4, and also comment on whether I should mention that in this page of the docs?

JamesH65:
AFAIK, it is NOT yet implemented. Probably should be mentioned.

andrum99:
Just had a user on the forums wondering why hdmi_blanking=1 wasn't working on their Pi 4, so have added mention that this is not implemented yet.

JamesH65:
thanks.

PR #1212, opened 2019-07-25
hardware/raspberrypi/booteeprom.md: minor fixes

andrum99:
- Grammar fixes
- Make spaces after periods consistent - some had double spaces, some single spaces. Double spaces is an older standard which is not used any more, so standardise on single spaces.
- Re-word two parts to make it more obvious what they mean


JamesH65:
Thanks.

PR #1213, opened 2019-07-25
configuration/hdmi-config.md: mention Ethernet over HDMI is not supported

andrum99:


JamesH65:
I don't think this is necessary. There are lots of things we don't support, but we don't have a paragraph for every one. Whilst we do have a list of what we do support.

And ethernet over HDMI is as rare as rocking horse poo.

andrum99:
OK. Do you think we need a list of things we don't support? I'm trying to head off questions about whether things that are not supported are or are not supported, if you see what I mean.

JamesH65:
Not sure where a list would go, and how detailed it would need to be. There's already the basic ones on the product pages. Might be useful, perhaps add an issue?

I'll close this one for now.

PR #1218, opened 2019-07-27
Update video.md: long overdue update to sdtv_mode

andrum99:
The page currently says composite is output on the yellow RCA jack. This is wrong - original Pi's had an RCA socket, not jack, and they were not all yellow. Also, most Pi models output composite via the TRRS socket. Finally, Zero and CM don't have a connector for composite.

### DIFF
--- a/configuration/config-txt/video.md
+++ b/configuration/config-txt/video.md
@@ -4,7 +4,7 @@
 
 ### sdtv_mode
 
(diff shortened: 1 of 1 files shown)


andrum99:
Note that is may actually be best just to remove reference to where composite is output from, rather than say all of what I have added. Let me know if this is your preferred option and I will amend my PR.

JamesH65:
LGTM

lurch:
See also https://github.com/raspberrypi/documentation/blob/master/setup/monitor-connection.md :wink:

andrum99:
> See also https://github.com/raspberrypi/documentation/blob/master/setup/monitor-connection.md

I'm not seeing anything that needs updated on that page.

andrum99:
Ah - I've spotted it - no mention of Pi Zero. I'll get right onto it. (Don't think it needs instructions for CM - not the target audience for that page).

lurch:
Sorry, I wasn't implying that anything needed to be updated; I was just pointing out that there **is** already documentation which correctly mentions the different composite output options on different Pi models.

PR #1222, opened 2019-07-30
Update omxplayer.md

jonwitts:
Add info regarding Raspberry Pi 4 HDMI display options as per discussion here: https://www.raspberrypi.org/forums/viewtopic.php?f=63&t=244589


JamesH65:
Thanks.

Issue #1225, opened 2019-07-31
Etcher requires zenity on Linux 


alexlemaire:
## Issue :boom: 

Was unable to flash the latest [Raspbian Buster Lite image](https://www.raspberrypi.org/downloads/raspbian/) on my SD card with [balenaEtcher](https://www.balena.io/etcher/).

_Following the installation documentation [here](https://www.raspberrypi.org/documentation/installation/installing-images/README.md)._

## Steps to reproduce :computer: 

1. Download and extract balenaEtcher appImage on a Linux machine
2. Download the latest Raspbian Buster Lite image from the official website
3. Start balenaEtcher
4. Select the Raspbian Buster Lite image archive and your SD device and try to "Flash!"
5. Enter your password in the SUDO prompt

This should then throw an error where something unexpected seem to happen when trying to flash.

## Fix :wrench: 

After some Googling ([see here](https://linuxhint.com/install_etcher_linux/)), I realized that the system needs to have _zenity_ installed to be able to flash an SD card via balenaEtcher.

## Update request :question: 

Would be nice to **include a note about that for Linux users** to avoid troubles for them.

I would gladly fork and open a pull request to handle this if this issue is approved! :+1:

JamesH65:
Feel free to provide a pull request. I would suggest saying that zenity _might_ need to be installed if not present

lurch:
I think this might count as an Etcher problem, rather than a Raspberry Pi documentation issue? Try creating an issue [here](https://github.com/balena-io/etcher/issues).

alexlemaire:
Alright, I created a pull request here with the update but will also open an issue on Etcher side so they can get that fixed.
In the meantime I think it may be worth having the note inside of Raspberry Pi documentation just to avoid problems for users. This can be removed as soon as Etcher team found a workaround.

JamesH65:
Change merged. #1230

PR #1230, opened 2019-08-01
Added note regarding zenity package requirement on Linux distributions

alexlemaire:
This pull request follows the issue #1225 that I opened yesterday.

### DIFF
--- a/installation/installing-images/README.md
+++ b/installation/installing-images/README.md
@@ -32,7 +32,7 @@ You will need to use an image writing tool to install the image you have downloa
 - Select the SD card you wish to write your image to.
 - Review your selections and click 'Flash!' to begin writing data to the SD card.
 
-**Note**: for Linux users, _zenity_ might need to be installed on your machine for _balenaEtcher_ to be able to write the image on your SD card.
+**Note**: for Linux users, `zenity` might need to be installed on your machine for `balenaEtcher` to be able to write the image on your SD card.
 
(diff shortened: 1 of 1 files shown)


installation/installing-images/README.md
```diff
@@ -32,6 +32,8 @@ You will need to use an image writing tool to install the image you have downloa
 - Select the SD card you wish to write your image to.
 - Review your selections and click 'Flash!' to begin writing data to the SD card.
 
+**Note**: for Linux users, _zenity_ might need to be installed on your machine for _balenaEtcher_ to be able to write the image on your SD card.
```
lurch: I suspect you wanted to use backticks here
```
`
```
instead of underscores?
```
_
```
i.e. `zenity` and not _zenity_ :slightly_smiling_face:

alexlemaire: The underscores were intended here. I wanted to style it the same way as _balenaEtcher_ was already highlighted in the document.
But if you guys prefer using backticks for this I can go with it no problem! :)

lurch: ping @JamesH65

JamesH65: Hmm, we usually use backticks for command line text or similar. Not sure about references to packages.  Will have a think.

JamesH65: I've pushed to copy edit - they can decide!

PR #1231, opened 2019-08-01
Fix broken command and link

ollyhayes:
A heading had been changed so the link to it was broken.

Also fixed where the wrong command was given for upgrading:
`sudo dist-upgrade` -> `sudo apt dist-upgrade`

### DIFF
--- a/configuration/wireless/access-point.md
+++ b/configuration/wireless/access-point.md
@@ -7,7 +7,7 @@ The Raspberry Pi can be used as a wireless access point, running a standalone ne
 
 Note that this documentation was tested on a Raspberry Pi 3, and it is possible that some USB dongles may need slight changes to their settings. If you are having trouble with a USB wireless dongle, please check the forums.
 
-To add a Raspberry Pi-based access point to an existing network, see [this section](#user-content-using-the-raspberry-pi-as-an-access-point-to-share-an-internet-connection-bridge).
+To add a Raspberry Pi-based access point to an existing network, see [this section](#internet-sharing).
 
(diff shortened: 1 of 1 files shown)


lurch:
Links like that work on GitHub, but don't work when the documentation is viewed on https://www.raspberrypi.org/documentation/

For links that work in both places, have a look at where e.g. `<a name="hostapd-config"></a>` is used in https://raw.githubusercontent.com/raspberrypi/documentation/master/configuration/wireless/access-point.md
Which you can see working at https://www.raspberrypi.org/documentation/configuration/wireless/access-point.md#hostapd-config

XECDesign:
There's already a note at the very top of the page that links to the update page. Or, are we going to standardise on reproducing those instructions in full on every page they may be relevant? IMHO, it makes more sense to remove.

lurch:
Good point - perhaps the `dist-upgrade` isn't needed now, given that all the software in Raspbian is now newer than when this guide was first written?

Also (maybe it's just me), but I think https://www.raspberrypi.org/documentation/configuration/wireless/access-point.md looks a bit weird. Should the "Before proceeding ..." text be moved _below_ the "Setting up a ..." header? @helenlynn

XECDesign:
It would need its own header if any. The page is a bit odd because it has two sets of instructions and that snippet would need to apply to both, which is why I added it outside of the two sections, but I'll admit it does look wrong.

lurch:
> The page is a bit odd because it has two sets of instructions and that snippet would need to apply to both

#1232 is a potential fix for that @XECDesign

ollyhayes:
Ah cool, I didn't realise you could do links like that, updated.

I was thinking about removing the bit about updating as well so I made that change as well, let me know what you think.

JamesH65:
Thanks all.

Issue #1234, opened 2019-08-01
Possible incorrect cmdline.txt flag


OutsourcedGuru:
> plymouth.ignore_serial_console: tells Plymouth not to output a boot splash on any serial console which may be configured.

Having done plenty of research, I would suggest that this flag is plural rather than singular. And furthermore, there appears to be some confusion over the use of underscores versus hyphens in this flag.

This seemed to work as expected with Stretch using the singular version with underscores. After the upgrade to Buster it seems like undervoltage warnings from `dmesg` were clobbering the custom splash screen.

By changing my `/boot/cmdline.txt` to the plural version and appending `fbcon=map:2` it's now behaving again with Buster.

I reference the [main.c file](https://gitlab.freedesktop.org/plymouth/plymouth/blob/master/src/main.c#L2259) in the source code of plymouth. It appears to be in plural and possess hyphens rather than underscores.

What's difficult is that there are many competing suggestions for setting up plymouth on Raspbian. Perhaps someone can look into this and as necessary, update the documentation here.

andrum99:
Per https://gitlab.freedesktop.org/plymouth/plymouth/commit/6e50233b031ed301813de7f075ae0d606eae010c

`plymouth.ignore-serial-consoles` has always been the correct name of this option. The error in our docs comes from when that page was originally written - it seems nobody noticed until now, so good catch.

What does `fbcon=map:2` do? I've read the relevant section of https://www.kernel.org/doc/Documentation/fb/fbcon.txt and I don't really understand what it is saying.

OutsourcedGuru:
Bringing that section here:
```
fbcon=map:<0123>

        This is an interesting option. It tells which driver gets mapped to
        which console. The value '0123' is a sequence that gets repeated until
        the total length is 64 which is the number of consoles available. In
        the above example, it is expanded to 012301230123... and the mapping
        will be:

		tty | 1 2 3 4 5 6 7 8 9 ...
		fb  | 0 1 2 3 0 1 2 3 0 ...

		('cat /proc/fb' should tell you what the fb numbers are)

	One side effect that may be useful is using a map value that exceeds
	the number of loaded fb drivers. For example, if only one driver is
	available, fb0, adding fbcon=map:1 tells fbcon not to take over the
	console.

	Later on, when you want to map the console the to the framebuffer
	device, you can use the con2fbmap utility.
```

[...]

PR #1235, opened 2019-08-03
cmdline-txt.md: fix plymouth.ignore-serial-consoles

andrum99:
Per #1234 the documentation of this kernel boot option is wrong:

- option is mis-spelled
- description of what it does is wrong

This PR fixes both


andrum99:
Note: I have just tested this, and it does actually work as described in this PR.

JamesH65:
Thanks.

PR #1236, opened 2019-08-05
Fix typo in linux.md

bewuethr:


JamesH65:
Had to read that twice to spot the change - thanks!

PR #1239, opened 2019-08-12
Update README.md

GeorgBisseling:
Ask for clarification about who is actually reading config.txt.

### DIFF
--- a/configuration/config-txt/README.md
+++ b/configuration/config-txt/README.md
@@ -14,7 +14,7 @@ Note that there are a few config settings that cannot be retrieved using `vcgenc
 
 ## File format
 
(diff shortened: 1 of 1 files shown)


JamesH65:
Why does this need to be more specific? It's a section on the config .txt content, not the boot process.

https://www.raspberrypi.org/documentation/configuration/boot_folder.md
https://www.raspberrypi.org/documentation/hardware/raspberrypi/bootmodes/bootflow.md

lurch:
Indeed, IMHO this is a "it doesn't matter" detail. And in fact I believe some settings are read by `bootcode.bin` and others are read by `start.elf` - but again this shouldn't matter to the end user.

andrum99:
We've got 2 "no"s for this one - suggest it be closed.

PR #1243, opened 2019-08-13
configuration/config-txt/memory.md: Max for gpu_mem should be 512

andrum99:
Per timg236 on the forums (see https://www.raspberrypi.org/forums/viewtopic.php?f=28&t=247820#p1518230):

> There's no guarantee that anything > 512MB will work and is completely untested.

So fix the docs to specify 512 as the max for gpu_mem, and add a warning that it is possible to set gpu_mem greater than this but this is likely to cause breakage.

### DIFF
--- a/configuration/config-txt/memory.md
+++ b/configuration/config-txt/memory.md
@@ -2,7 +2,9 @@
 
 ## gpu_mem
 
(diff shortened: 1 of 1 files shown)


popcornmix:
I'm okay with this change.

andrum99:
I'm wondering if we also need a lower recommended max for the 256 and 512MB boards.

popcornmix:
Only if there is evidence the current values don't work.

JamesH65:
There's already a PR in progress that affects this section. Probably best to add this to that.

JamesH65:
See https://github.com/raspberrypi/documentation/pull/1240, I've made a similar change in that.

PR #1253, opened 2019-08-18
attempted fix for multi-level list bug

### LINKED ISSUE #1251: Multi-level lists in docs subtly broken on raspberrypi.org
(This is one I discovered a while ago but I'm just getting round to adding it as a separate bug).

Documentation pages which contain multi-level lists are currently subtly broken on the official Pi website (raspberrypi.org). The breakage applies only to list items where the indent level is a multiple of 3. In other words indent levels 3, 6, 9 etc are affected. For those indent levels, any list item will be rendered as if it belongs to one below the actual indent level. So level 3 list items are rendered as level 2 list items, level 6 list items as level 5 items, etc.

Compare and contrast the following pages:

https://www.raspberrypi.org/documentation/hardware/raspberrypi/bootmodes/host.md
https://github.com/raspberrypi/documentation/blob/master/hardware/raspberrypi/bootmodes/host.md

Rendered on github, the page shows the correct layout of multi-level lists. On raspberrypi.org it is broken as described above.

Interestingly, github.com seems to know about the possible breakage: if you go into edit mode it  renders 3rd, 6th etc level list items in red - see https://github.com/raspberrypi/documentation/edit/master/hardware/raspberrypi/bootmodes/host.md

andrum99:
Attempt to fix #1251 by making all list indents on this page 4 spaces instead of 2.

### DIFF
--- a/hardware/raspberrypi/bootmodes/host.md
+++ b/hardware/raspberrypi/bootmodes/host.md
@@ -6,19 +6,19 @@ The USB host boot mode follows this sequence:
 
 * Enable the USB port and wait for D+ line to be pulled high indicating a USB 2.0 device (we only support USB2.0)
 * If the device is a hub:
-  * Enable power to all downstream ports of the hub
-  * For each port, loop for a maximum of two seconds (or five seconds if `program_usb_boot_timeout=1` has been set)
-    * Release from reset and wait for D+ to be driven high to indicate that a device is connected
-    * If a device is detected:
-      * Send "Get Device Descriptor"
-        * If VID == SMSC && PID == 9500
-          * Add device to Ethernet device list
-        * If class interface == mass storage class
-          * Add device to mass storage device list
+    * Enable power to all downstream ports of the hub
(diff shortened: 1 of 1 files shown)


lurch:
Pushing this directly to master to see if it fixes #1251 on https://www.raspberrypi.org/documentation/hardware/raspberrypi/bootmodes/host.md
If it doesn't fix things I'll revert it.

andrum99:
Nope - still broken in the same way it was before. so it seems https://github.com/Python-Markdown/markdown/issues/3 is not the culprit.

lurch:
AFAIK raspberrypi.org doesn't auto-update, but is done as a scheduled-task or something. I'll check in a few hours...

andrum99:
I've just checked https://www.raspberrypi.org/documentation/hardware/raspberrypi/bootmodes/host.md and it is actually now rendering correctly. I jumped the gun slightly there - sorry.

Looking again at this page, it could do with different list indicator character for each indent level - at the moment it is slightly confusing how raspberrypi.org renders it all with red hyphens - the various levels of indent would be much more obvious if each indent level used a different character, ike github.com uses.

andrum99:
See #1254

PR #1254, opened 2019-08-19
test different list item indicator characters

andrum99:
At present, raspberrypi.org uses a red hyphen to indicate all unordered list items. Markdown supports using different characters in the source: *, - and +. Test how raspberrypi.org renders these by applying them to different levels of the multi-level lists on hardware/raspberrypi/bootmodes/host.md.

The motivation for doing this is that it makes it much more obvious at which indent level each list item is, if the indent level above and below it uses a different indicator character.

### DIFF
--- a/hardware/raspberrypi/bootmodes/host.md
+++ b/hardware/raspberrypi/bootmodes/host.md
@@ -6,19 +6,19 @@ The USB host boot mode follows this sequence:
 
 * Enable the USB port and wait for D+ line to be pulled high indicating a USB 2.0 device (we only support USB2.0)
 * If the device is a hub:
-    * Enable power to all downstream ports of the hub
-    * For each port, loop for a maximum of two seconds (or five seconds if `program_usb_boot_timeout=1` has been set)
-        * Release from reset and wait for D+ to be driven high to indicate that a device is connected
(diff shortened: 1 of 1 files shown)


andrum99:
This is a follow-on to #1253.

lurch:
Everything I've found (e.g. https://github.github.com/gfm/#list-items ) suggests that `*`, `-` and `+` will all get treated equivalently, so I'm not going to merge this. AFAIK the different list-indicator-characters are a CSS-level thing rather than a HTML-level thing.

I won't be merging this, but I'll attempt to find out more about the MarkDown -> HTML conversion process on raspberrypi.org ...

lurch:
> I won't be merging this, but I'll attempt to find out more about the MarkDown -> HTML conversion process on raspberrypi.org ...

... which may take a while, because lots of people are on holiday at the moment :wink: :sun_with_face:

andrum99:
> > I won't be merging this, but I'll attempt to find out more about the MarkDown -> HTML conversion process on raspberrypi.org ...
> 
> ... which may take a while, because lots of people are on holiday at the moment 😉 🌞

Lucky them. I'm off in September when it's cheaper 😎 #scorchio

Issue #1108, opened 2019-03-22
Missing documentation for Revision a22083


matteobogo:
Hello,
Recently i have received a new Raspberry Pi 3 Model B v1.3 with Revision Code: a22083.
I didn't find any info about it in the official documentation.
![WP_20190322_19_04_39_Rich](https://user-images.githubusercontent.com/8475090/54843909-cfd37e80-4cd5-11e9-8045-f77ac261116f.jpg)
![WP_20190322_19_12_56_Rich](https://user-images.githubusercontent.com/8475090/54845289-4a51cd80-4cd9-11e9-94d4-6a2ddfa9698c.jpg)
![WP_20190322_19_10_36_Rich (2)](https://user-images.githubusercontent.com/8475090/54844249-b54dd500-4cd6-11e9-93b2-eed53d5a30bc.jpg)

bennuttall:
Thanks, it's a Pi 3B manufactured by Embest. I've added it to the table: https://github.com/raspberrypi/documentation/blob/master/hardware/raspberrypi/revision-codes/README.md

PR #1260, opened 2019-08-28
Raspberry pi code a22083 is 3B+ Model

arbabseyfola:


lurch:
That's not what #1108 says? :confused:

arbabseyfola:
> That's not what #1108 says? 😕
Why board revision is 1.3? What is the difference between 1.2 and 1.31?

lurch:
> Why board revision is 1.3? What is the difference between 1.2 and 1.31?

AFAIK Raspberry Pi doesn't provide that level of detail about their hardware.

JamesH65:
Closing, that revision code is definitely a 3B, not a Plus.

PR #1261, opened 2019-08-29
Modules install command fix

rkoptev:
Running modules installation command produces "make: arm-linux-gnueabihf-gcc: Command not found" error. This is caused by running this command from root via sudo, where the path to the toolchain is not available. Passing PATH variable to the command fixes this problem.

Also here is related discussion on this topic: https://www.raspberrypi.org/forums/viewtopic.php?t=209621

### DIFF
--- a/linux/kernel/building.md
+++ b/linux/kernel/building.md
@@ -194,7 +194,7 @@ sudo mount /dev/sdb7 mnt/ext4
 Next, install the modules:
 
 ```bash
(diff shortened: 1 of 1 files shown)


lurch:
Is this the same issue as #872 ?

JamesH65:
@popcornmix @pelwell This looks like the correct fix to me, any comments?

pelwell:
Looks OK to me.

JamesH65:
Thanks @pelwell for review, and thanks @rkoptev,

PR #1267, opened 2019-09-05
Update access-point.md

nick-ivanov:
'reload dnsmasq' causes an error on RPi Zero W (Raspbian Buster). enable+start works great.

### DIFF
--- a/configuration/wireless/access-point.md
+++ b/configuration/wireless/access-point.md
@@ -73,7 +73,8 @@ There are many more options for dnsmasq; see the [dnsmasq documentation](http://
 
 Reload `dnsmasq` to use the updated configuration:
 ```
-sudo systemctl reload dnsmasq
+sudo systemctl enable dnsmasq
+sudo systemctl start dnsmasq
 ```
 
 <a name="hostapd-config"></a>


XECDesign:
The steps to enable and start it are already further up. Running those commands again will have no effect.

lurch:
Presumably #1268 makes this redundant?

JamesH65:
I would think so, but does this fix need to be in the PR (or is it already?)

XECDesign:
This isn't a fix, it would break the previous instructions. The new instructions in the separate PR ask the user to reboot instead of reloading the configuration.

lurch:
ping @epoch1970

epoch1970:
> ping @epoch1970

Sorry I was unavoidably detained... Reloading network and services is a lot of gyrations for nothing much in the end since we don't display what an OK runtime state should look like, and how to fix deviations.
 
So I just went for the acid test: reboot and see if the service comes up. If it doesn't, [read again and] call the forums...

benshaker:
For what it's worth, the steps to enable and start `dnsmasq` are not listed anywhere else on this document. I spent a good bit of time hunting down an error tonight, and I wouldn't have had such trouble if I knew to enable this service earlier on.

JamesH65:
@epoch1970 Is this still required in the newly merged docs?

epoch1970:
No it isn't.

JamesH65:
Closing, no longer needed.

PR #1268, opened 2019-09-05
Access point doc refactoring

epoch1970:
- Split routed and bridged Access Point cases

- General refactoring

- Frenglish beware!

Tested w/ Pi3B / Raspbian Buster Lite.

### DIFF
--- a/configuration/wireless/access-point-bridged.md
+++ b/configuration/wireless/access-point-bridged.md
@@ -49,13 +49,13 @@ sudo systemctl unmask hostapd
 sudo systemctl enable hostapd
 ```
 
-This completes software installation. We will configure the access point software later on.
+Software installation is complete. We will configure the access point software later on.
 
 ## Configure bridge networking
 
 ### Create a bridge device and populate the bridge (systemd-networkd)
 
-Add a bridge network device called br0 in the system by creating the following file:
+Add a bridge network device named `br0` by creating the following file:
 
 ```
 sudo nano /etc/systemd/network/bridge-br0.netdev
@@ -65,7 +65,7 @@ Name=br0
 Kind=bridge
 ```
 
-In order to bridge the ethernet network with the wireless network, add the built-in ethernet interface (eth0) as bridge member with the following file:
+In order to bridge the ethernet network with the wireless network, first add the built-in ethernet interface (`eth0`) as bridge member with the following file:
 
 ```
 sudo nano /etc/systemd/network/br0-member-eth0.network
@@ -77,7 +77,7 @@ Name=eth0
 Bridge=br0
 ```
 
-*Note* The access point software will add the wireless interface wlan0 to the bridge when the service starts. There is no need to create a file for the wireless interface.
+**Note:** The access point software will add the wireless interface `wlan0` to the bridge when the service starts. There is no need to create a file for it. This situation is particular to WiFi network interfaces.
 
 ### Configure DHCP networking (dhcpcd)
 
--- a/configuration/wireless/access-point-bridged.md
+++ b/configuration/wireless/access-point-bridged.md
@@ -57,7 +57,7 @@ Software installation is complete. We will configure the access point software l
 <a name="bridging"></a>
 ## Setup the network bridge
 
-### Create a bridge device and populate the bridge (systemd-networkd)
+### Create a bridge device and populate the bridge
 
 Add a bridge network device named `br0` by creating the following file:
 
--- a/configuration/wireless/access-point-bridged.md
+++ b/configuration/wireless/access-point-bridged.md
@@ -57,6 +57,8 @@ Software installation is complete. We will configure the access point software l
 <a name="bridging"></a>
 ## Setup the network bridge
 
+A bridge network device in the Raspberry Pi will join the ethernet and wireless networks using its built-in interfaces.
+
 ### Create a bridge device and populate the bridge
 
 Add a bridge network device named `br0` by creating the following file:
--- a/configuration/wireless/access-point-bridged.md
+++ b/configuration/wireless/access-point-bridged.md
@@ -3,7 +3,7 @@
 
 The Raspberry Pi can be used as a bridged wireless access point within an existing ethernet network. This will extend the network to wireless computers and devices.
 
-If you wish to setup a stand-alone wireless network, consider instead setting up a [routed access point.](./access-point-routed.md)
+If you wish to create a wireless network that could function stand-alone, consider instead setting up a [routed access point.](./access-point-routed.md)
 
 ```
                                          +- RPi -------+
--- a/configuration/wireless/access-point-bridged.md
+++ b/configuration/wireless/access-point-bridged.md
@@ -35,6 +35,7 @@ Note that this documentation was tested on a Raspberry Pi 3, and it is possible 
 * Ensure the Raspbian OS on your Raspberry Pi is [up to date](../../raspbian/updating.md) and reboot if packages were installed in the process.
 * Have a wireless client (laptop, smartphone, ...) ready to test your new access point.
 
+<a name="hostapd-install"></a>
 ## Install the access point software (hostapd)
 
 In order to work as a bridged access point, the Raspberry Pi needs to have the `hostapd` access point software package installed:
@@ -51,7 +52,8 @@ sudo systemctl enable hostapd
 
 Software installation is complete. We will configure the access point software later on.
 
-## Configure bridge networking
+<a name="bridging"></a>
+## Setup the network bridge
 
 ### Create a bridge device and populate the bridge (systemd-networkd)
 
@@ -97,6 +99,7 @@ sudo nano /etc/dhcpcd.conf
 
 Add `denyinterfaces wlan0 eth0` near the beginning of the file (above the first `interface` line, if any). Interface `br0` will be configured as per defaults, no specific entry is necessary. Save the file to complete the IP configuration of the machine.
 
+<a name="hostapd-config"></a>
 ## Configure the access point software (hostapd)
 
 Create the `hostapd` configuration file, located at `/etc/hostapd/hostapd.conf`, to add the various parameters for your wireless network. 
@@ -105,15 +108,15 @@ Create the `hostapd` configuration file, located at `/etc/hostapd/hostapd.conf`,
 sudo nano /etc/hostapd/hostapd.conf
 ```
 
-Add the information below to the configuration file. This configuration assumes we are using channel 7, with a network name of NameOfNetwork, and a password AardvarkBadgerHedgehog. Note that the name and password should **not** have quotes around them. The passphrase should be between 8 and 64 characters in length.
+Add the information below to the configuration file. This configuration assumes we are using channel 7, with a network name of `NameOfNetwork`, and a password `AardvarkBadgerHedgehog`. Note that the name and password should **not** have quotes around them. The passphrase should be between 8 and 64 characters in length.
 
-To use the 5 GHz band, you can change the operations mode from hw_mode=g to hw_mode=a. Possible values for hw_mode are:
+To use the 5 GHz band, you can change the operations mode from `hw_mode=g` to `hw_mode=a`. Possible values for `hw_mode` are:
  - a = IEEE 802.11a (5 GHz)
  - b = IEEE 802.11b (2.4 GHz)
  - g = IEEE 802.11g (2.4 GHz)
  - ad = IEEE 802.11ad (60 GHz)
 
-Note the lines `interface=wlan0` and `bridge=br0`: this directs hostapd to add the `wlan0` interface as bridge member to `br0` when the access point starts, completing the bridge.
+Note the lines `interface=wlan0` and `bridge=br0`: this directs `hostapd` to add the `wlan0` interface as bridge member to `br0` when the access point starts, completing the bridge.
 
 ```
 interface=wlan0
@@ -132,7 +135,6 @@ wpa_key_mgmt=WPA-PSK
 wpa_pairwise=TKIP
 rsn_pairwise=CCMP
 ```
-
 ## Run your new wireless access point
 
 It is now time restart your Raspberry Pi and verify the wireless access point becomes automatically available.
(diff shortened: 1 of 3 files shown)


JamesH65:
This looks like a really good change, but I do not at the moment have time to check that it all works! Anyone out there willing to work through these instructions and confirm all is well?

Only comment I have is that it would be good to use proper graphics for the current text drawn diagrams. I'll see if our in house team is up for that, but in the meantime, I can push this to copy edit, on the assumption that it all works! Thanks!

XECDesign:
I'll take a look, since I touched it last.

configuration/wireless/access-point-bridged.md
```diff
@@ -0,0 +1,157 @@
...
+                 +-------------+     |
+                                     |   +- PC#1 ------+
+                                     +---+ 192.168.1.4 |
+                                         +-------------+
+```
+
+This can be done using the inbuilt wireless features of the Raspberry Pi 4, Raspberry Pi 3 or Raspberry Pi Zero W, or by using a suitable USB wireless dongle that supports access point mode.
+It is possible that some USB dongles may need slight changes to their settings. If you are having trouble with a USB wireless dongle, please check the forums.
+
+This documentation was tested on a Raspberry Pi 3B running a factory installation of Raspbian Buster Lite (Jul. 2019).
```
lurch: IMHO 'Jul.' should be expanded to 'July', and perhaps "a fresh installation" makes more sense than "a factory installation" ?

configuration/wireless/access-point-bridged.md
```diff
@@ -0,0 +1,157 @@
...
+<a name="ap-config"></a>
+## Configure the access point software
+
+Create the `hostapd` configuration file, located at `/etc/hostapd/hostapd.conf`, to add the various parameters for your wireless network. 
+
+```
+sudo nano /etc/hostapd/hostapd.conf
+```
+
+Add the information below to the configuration file. This configuration assumes we are using channel 7, with a network name of `NameOfNetwork`, and a password `AardvarkBadgerHedgehog`. Note that the name and password should **not** have quotes around them. The passphrase should be between 8 and 64 characters in length.
```
lurch: I've not followed this setup myself, but is it worth suggesting that the `NameOfNetwork` is different to any existing wireless network names?

configuration/wireless/access-point-bridged.md
```diff
@@ -0,0 +1,157 @@
...
+## Run your new wireless access point
+
+It is now time restart your Raspberry Pi and verify the wireless access point becomes automatically available.
+
+```
+sudo systemctl reboot
+```
+Once your Raspberry Pi has restarted, search for WiFi networks with your wireless client. The network SSID you specified in file `/etc/hostapd/hostapd.conf` should now be present, and it should be accessible with the specified password.
+
+If your wireless client has access to the local network and the Internet, congratulations on your new access point!
```
lurch: "congratulations on setting up your new access point!" ?

configuration/wireless/access-point-bridged.md
```diff
@@ -0,0 +1,157 @@
...
+
+If you encounter difficulties, contact the forums for assistance. Please refer to this page in your message.
```
lurch: Add a link to the forums?

configuration/wireless/access-point-routed.md
```diff
@@ -0,0 +1,190 @@
+# Setting up a Raspberry Pi as a Routed Wireless Access Point
+
+The Raspberry Pi can be used as a wireless access point routing to an existing ethernet network. This will create a new wireless network entirely managed by the Raspberry Pi.
```
lurch: I'm not sure if the average reader will understand "a wireless access point routing to an existing ethernet network" - perhaps it could be clarified slightly? :man_shrugging: 

Although the IP-address diagram below definitely helps.

configuration/wireless/access-point-routed.md
```diff
@@ -0,0 +1,190 @@
...
+
+```
+                                         +- RPi -------+
+                                     +---+ 10.10.0.2   |          +- Laptop ----+
+                                     |   |     WiFi AP +-)))  (((-+ WiFi STA    |
+                                     |   | 192.168.4.1 |          | 192.168.4.2 |
+                                     |   +-------------+          +-------------+
+                 +- Router ----+     |
+                 | Firewall    |     |   +- PC#2 ------+
+(Internet)---WAN-+ DHCP server +-LAN-+---+ 10.10.0.3   |
```
lurch: For consistency with the diagram on the other page, maybe it's worth switching the Ethernet network IPs to 192.168.1.x and then have 10.10.0.x on the WiFi network?

configuration/wireless/access-point-routed.md
```diff
@@ -0,0 +1,190 @@
...
+
+In order to provide network management services (DNS, DHCP) to wireless clients, the Raspberry Pi needs to have the `dnsmasq` software package installed:
+
+```
+sudo apt install dnsmasq
+```
+Finally, install `netfilter-persistent` and its plugin `iptables-persistent`. This utilty helps saving firewall rules and restores them when the Raspberry Pi boots:
+
+```
+sudo DEBIAN_FRONTEND=noninteractive apt install -y netfilter-persistent iptables-persistent
```
lurch: Is `DEBIAN_FRONTEND=noninteractive` really necessary? (if it is, maybe it's worth explaining why?)

configuration/wireless/access-point-routed.md
epoch1970: I've been struggling with that... The current README.md file uses this expression: "[A guide to setting the Raspberry Pi up] as an access point, and as an internet sharing device"

On retrospect I think I should have used that.
"The Raspberry Pi can be used as a wireless access point, and as an internet sharing device. This will create ..."?

configuration/wireless/access-point-routed.md
```diff
@@ -0,0 +1,190 @@
...
+```
+interface wlan0
+    static ip_address=192.168.4.1/24
+    nohook wpa_supplicant
+```
+
+### Enable routing and IP masquerading
+
+This section configures the Raspberry Pi to let wireless clients access computers on the main network, and from there the Internet.
+*NOTE:* If you wish to block wireless clients from accessing the main network and the Internet, skip this section.
```
lurch: Is there any _easy_ way of setting things up to allow the wireless clients (connected to the Pi) to access the internet **without** allowing them to access other computers on the ethernet network? E.g. setting up a 'guest' wireless network.

If it's not easy, then ignore this question as I don't want to make these instructions more complicated than they already are! :slightly_smiling_face:

configuration/wireless/access-point-routed.md
epoch1970: Boring explanation: The install process interactively asks if you want to save the current iptables state. "noninteractive" + option "-y" skips the question and answers yes, making copy-paste possible.

To me it is a pure technicality, I wouldn't mind not explaining the unusual options used here.

configuration/wireless/access-point-bridged.md
XECDesign: I'd rather it just said the latest Raspbian image. Otherwise it's an extra place that needs to be updated each time an image comes out.

configuration/wireless/access-point-routed.md
```diff
@@ -0,0 +1,190 @@
...
+
+```
+interface wlan0
+    static ip_address=192.168.4.1/24
+    nohook wpa_supplicant
+```
+
+### Enable routing and IP masquerading
+
+This section configures the Raspberry Pi to let wireless clients access computers on the main network, and from there the Internet.
```
lurch: To avoid any ambiguity, maybe it's worth specifically referring to the "ethernet network" (or "wired network"), rather than the "main network"?

configuration/wireless/access-point-routed.md
epoch1970: Sorry, I didn't see that. I used 10.10.0.0 instead of 192.168.1.0 to make the contrast more apparent between the 2 nets. (192.168.4.1 and 192.168.1.1 differ less visually)

Could have been a bad idea...

configuration/wireless/access-point-routed.md
```diff
@@ -0,0 +1,190 @@
...
+This process is configured by adding a single firewall rule in the Raspberry Pi:
+
+```
+sudo iptables -t nat -A POSTROUTING -o eth0 -j MASQUERADE
+```
+Now save the current firewall rules for IPv4 (including the rule above) and IPv6 to be loaded at boot by the `netfilter-persistent` service:
+```
+sudo netfilter-persistent save
+```
+Filtering rules are saved to directory `/etc/iptables/`. If in the future you change the configuration of your firewall, make sure to save the configuration to files before rebooting.
```
lurch: Does this sentence still work if "to files" is removed?

configuration/wireless/access-point-bridged.md
lurch: Probably still worth specifying "Raspbian Buster" though?

configuration/wireless/access-point-routed.md
epoch1970: Hmm. I think yes, by adding an additional filtering rule that would block traffic destined to the local LAN. (something like "... --dest 10.10.0.0/24 -j DROP") 
But I am not exactly the iptables guru.

lurch:
IMHO `configuration/wireless/access-point.md` shouldn't be deleted (in case external pages are linking to it), but instead just have it provide links to the two new pages you've created.

I agree with James, these changes look great! :tada:

configuration/wireless/access-point-routed.md
epoch1970: It works better :)

configuration/wireless/access-point-routed.md
lurch: I'm not sure that works either, since both pages talk about "wireless access point" and both pages end up sharing the internet to the wireless clients?
Would phrasing like "extending an existing network" and "creating a new sub-network" work? :man_shrugging:

configuration/wireless/access-point-bridged.md
epoch1970: I think so, yes. The "netfilter-persistent" thing is likely to blow in the face of people using older releases.

XECDesign:
Just went through the router config. Works well, clear and easy to follow.

One issue is that rfkill will be enabled out of the box on 5GHz pis. If the user skips the first run wizard and never sets the country, they'll run into issues. It may be worth adding `sudo rfkill unblock wifi` somewhere in the instructions just in case.

XECDesign:
Bridged config works as well. Same caveat about `rfkill` needed.

JamesH65:
One this is CE'd and merged I'll ask the comms team to produce better graphics for the diagrams currently drawn using text, but no need to wait for it.

epoch1970:
> If the user ... never sets the country, they'll run into issues. It may be worth adding `sudo rfkill unblock wifi` somewhere in the instructions just in case.

I understand setting the regulatory domain is i) the correct thing to do, ii) is persistent, while `rfkill unblock` is a runtime setting?

If so, then I suggest adding a bullet to the **Before you start** section:

-  Make sure to run `raspi-config` and verify the WiFi Country setting matches the country you're in.

(or something like that)

XECDesign:
> I understand setting the regulatory domain is i) the correct thing to do, ii) is persistent, while `rfkill unblock` is a runtime setting?
> 
> If so, I suggest adding a bullet to the **Before you start** section:
> 
>     * Make sure to run `raspi-config` and verify the WiFi Country setting matches the country you're in.

systemd has a service that makes rfkill persistent.

`raspi-config` only sets the country in `wpa_supplicant.conf` which isn't used in this case.

`hostapd.conf` has its own way of setting the country code:
```
# Country code (ISO/IEC 3166-1). Used to set regulatory domain.
# Set as needed to indicate country in which device is operating.
# This can limit available channels and transmit power.
# These two octets are used as the first two octets of the Country String
# (dot11CountryString)
#country_code=US
```

epoch1970:
Guys, just to be sure. I don't know what "copy-edit" entails, is anyone waiting on me?
If not, all is fine, no need to answer. Cheers.

JamesH65:
We have a copy editing team who checks all contributions to make sure they adhere to our styles, are grammatically correct etc. Unfortunately they are horribly backlogged.

helenlynn:
@JamesH65 Since this is 2.7K ww of new technical content I'm exceedingly tempted to ask a handy freelance technical editor to take a look, but would it make sense to resolve the conflicts first?

JamesH65:
Replaced by #https://github.com/raspberrypi/documentation/pull/new/epoch1970-ap-refactor
which is a simple merge (no changes from OP's contribution)

configuration/wireless/access-point-bridged.md
epoch1970: > IMHO 'Jul.' should be expanded to 'July', and perhaps "a fresh installation" makes more sense than "a factory installation" ?

Yes. fresh install is better

configuration/wireless/access-point-bridged.md
epoch1970: Sort answer: no I don't think it is useful.
I haven't touched the wireless setup part of the original doc, mostly.
I do not think there is a requirement for the SSID to be unique, in case of multiple identical network names (with identical password) a client should be able to roam with more or less success, depending on client, signals strengths...
Of course if people start connecting to the wrong AP they might get confused by what they see. But the number of users of this tutorial having multiple APs is probably low anyway.

configuration/wireless/access-point-bridged.md
epoch1970: Is it obviously frenglish? Anyway do as you see best

configuration/wireless/access-point-bridged.md
epoch1970: I leave that to you. I think it would be a good idea.

configuration/wireless/access-point-routed.md
epoch1970: "extending an existing network" works for me for bridged
 "creating a new sub-network" not as much for routed, it sounds too much as "net/supernet/subnet" which have a definite meaning.
What about "secondary network" for routed?

configuration/wireless/access-point-routed.md
epoch1970: Nitpicking a bit, nothing says that there isn't a wireless bridge in the main network.
Also see above https://github.com/raspberrypi/documentation/pull/1268#discussion_r346414657 my proposal for using "main/secondary networks".
But yes, if the sentence is unclear, by all means amend it.

JamesH65:
@helenlynn Because I had to merge this offline and produce a new PR, you'll (or whosoever) will need to cross reference @epoch1970 comments to the new PR at #1333.

epoch1970:
So this definitely needs fixing in the doc.
I don't have a 5GHz capable Pi, so I'm not sure of the most elegant way of doing it.
I leave this to you.

(Quoting XECDesign, above)
> It may be worth adding sudo rfkill unblock wifi somewhere in the instructions just in case.
...
> `hostapd.conf` has its own way of setting the country code:
> 
> ```
> # Country code (ISO/IEC 3166-1). Used to set regulatory domain.
> # Set as needed to indicate country in which device is operating.
> # This can limit available channels and transmit power.
> # These two octets are used as the first two octets of the Country String
> # (dot11CountryString)
> #country_code=US
> ```

configuration/wireless/access-point-bridged.md
lurch: > Of course if people start connecting to the wrong AP they might get confused by what they see.

I'm imagining that most people following this tutorial will probably be home users with an existing wireless network being served from their broadband router.
But maybe I'm just being over-cautious and it's obvious that the network-name they choose for their RPi AP should be different from their router's AP? :man_shrugging:

helenlynn:
@JamesH65 and @epoch1970 Noted; we will refer to comments here when editing the new PR #1333.

PR #1051, opened 2018-12-28
removed umask=000 because it fails with `Unrecognized mount option "u…

brunoamaral:
…mask=000" or missing value`

Using Raspbian 9 (stretch)

PR #1271, opened 2019-09-06
external-storage.md: use umask=000 only for FAT and NTFS

andrum99:
Per discussion on #1051, update instructions to use umask=000 only when mounting FAT or NTFS filesystem.

Since instructions don't currently deal with trying to mount an HFS or HFS+ filesystem, don't bother mentioning that. If folk need this, it can be added in a future PR.

### DIFF
--- a/configuration/external-storage.md
+++ b/configuration/external-storage.md
@@ -68,9 +68,11 @@ You can modify the `fstab` file to define the location where the storage device 
 4. Add the following line in the `fstab` file:
 
     ```
-    UUID=5C24-1453 /mnt/mydisk FSTYPE defaults,auto,umask=000,users,rw,nofail 0 0
+    UUID=5C24-1453 /mnt/mydisk FSTYPE defaults,auto,users,rw,nofail 0 0
     ```
    Replace FSTYPE with the type of your file system, which you found in step 2 of 'Mounting a storage device' above.
+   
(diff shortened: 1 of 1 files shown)


andrum99:
Ready for review @JamesH65

configuration/external-storage.md
```diff
@@ -68,9 +68,11 @@ You can modify the `fstab` file to define the location where the storage device
 4. Add the following line in the `fstab` file:
 
     ```
-    UUID=5C24-1453 /mnt/mydisk FSTYPE defaults,auto,umask=000,users,rw,nofail 0 0
+    UUID=5C24-1453 /mnt/mydisk FSTYPE defaults,auto,users,rw,nofail 0 0
     ```
    Replace FSTYPE with the type of your file system, which you found in step 2 of 'Mounting a storage device' above.
+   
+5. If the filesystem type is FAT or NTFS, add the option `umask=000` immediately after `nofail` - this will allow all users full read/write access to every file on the storage device.
```
JamesH65: I think having the complete line with the umask bit inserted here would make this clearer - at the moment do you need a comma after nofail for example?

JamesH65:
Thanks all!

configuration/external-storage.md
andrum99: Oops - forgot about the comma. The current wording matches how the page already suggests the `x-systemd.device-timeout` option for something else.

PR #1276, opened 2019-09-11
Replace releases notes with a link

timg236:
Link to release notes in the rpi-eeprom Github because these are updated as soon as a beta image is published.


timg236:
@JamesH65 @XECDesign

Issue #1228, opened 2019-07-31
Problem with URLs ending with brackets


lurch:
As I mentioned [here](https://github.com/raspberrypi/documentation/pull/1186#issuecomment-512287987) and [here](https://github.com/raspberrypi/documentation/pull/1208#issuecomment-513811950) there's a problem with https://www.raspberrypi.org/documentation/ displaying URLs that end with a `)` character.
Opening an issue so that this doesn't get forgotten about...

andrum99:
According to https://en.wikipedia.org/wiki/Percent-encoding, parentheses are reserved characters in RFC3986 - Uniform Resource Identifier (URI): Generic Syntax. This means they are meant to be escaped. The mapping is:

( -> %28
) -> %29

lurch:
Fixed, see #1278

Thanks @andrum99 :+1:

PR #1278, opened 2019-09-12
The wikipedia link to the ZIP64 is broken

SibeeshVenu:
Updated the ZIP64 Wikipedia link


lurch:
Yup, #1228

JamesH65:
So I think the ( and ) should be converted to %28 and %29 respectively? Are we sure the transfer from github to the website copes with that?

lurch:
I've got no idea, but I guess there's only one way to find out... :confused:

JamesH65:
Currently finding out!

JamesH65:
And it worked. 

So no need for this PR now, but thanks to @SibeeshVenu for pointing out the problem.

PR #1281, opened 2019-09-16
How linux filesystem works

weg2g:
More info on Linux file system management


JamesH65:
We generally don't like links off site for things like this - it's a maintenance burden and we can never be sure of the quality at the end of the link.

JamesH65:
Rejecting as unnecessary, we don't like links offsite, and destination website too commercially oriented.

PR #1282, opened 2019-09-16
Resources for Linux Kernel

weg2g:
Adding more resource for Linux Kernel


JamesH65:
As with comment on you rother PR, on the whole we don't like links offsite for things like this. I'll discuss with the comms team.

lurch:
I guess there's also the level-of-detail to consider - IMHO stuff about "Bottom Half Handling Data Structures" certainly seems _well_ beyond the scope of the beginner-oriented nature of the Raspberry Pi Documentation? :man_shrugging:

JamesH65:
I hadn't looked at the content, but that is a fair point.

JamesH65:
Not happy with simply adding links to random outside sites over which we have zero control, so rejecting this PR.

PR #1283, opened 2019-09-16
Add more resource info

weg2g:
More info for each linux filesystem

### DIFF
--- a/linux/filesystem/whole-filesystem.md
+++ b/linux/filesystem/whole-filesystem.md
@@ -2,9 +2,9 @@
 
 A list of directories is shown below. For more information, please refer to the Linux [Filesystem Hierarchy Standard](https://wiki.debian.org/FilesystemHierarchyStandard).
 
-- bin
+- [bin](https://coding-bootcamps.com/linux/filesystem/bin.html)
 
-- boot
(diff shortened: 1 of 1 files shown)


XECDesign:
Why not link to the source instead (http://refspecs.linuxfoundation.org/FHS_3.0/fhs/index.html)? The Debian wiki link on there already opens up a few other resources. If it's of value, why not include all the relevant information on the page itself, without going to an external source?

This feels a bit like SEO.

lurch:
> This feels a bit like SEO.

Yeah, especially when it says "we offer private custom tutoring classes..." at the bottom of the page :confused:

weg2g:
I just checked Linux foundation docs and mine is more comprehensive than theirs. 

I can not include all of them as it includes long (20 pages). Also, we spent months to preparing docs, so just we like to maintain our copywright over our contents. 

Anyway, if you think it is SEO, go ahead and reject my changes. 

Also, "we offer private custom tutoring classes" is a part of web page template not SEO

JamesH65:
As before, we are not keen on links offsite for things like this. Will discuss with comms.

JamesH65:
Decided this is not a suitable change for our documentation, too 'commercial', and too detailed for the intended audience.

PR #1284, opened 2019-09-18
Update audio-config.md

javaspaces:
I just went through this process in order to get audio working on my Raspberry Pi 4. On my system, "advanced options" was Option 7 (not option 8). Also, the "audio" sub-option was "A4" on my system.  Finally, audio did not play through the 3.5mm jack until I rebooted my Pi 4.

### DIFF
--- a/configuration/audio-config.md
+++ b/configuration/audio-config.md
@@ -34,11 +34,17 @@ This will open the configuration screen:
 
 ![raspi-config screen](images/raspi-config.png)
 
-Select Option 8 `Advanced Options` and press `Enter`, then select Option A6: `Audio` and press `Enter`:
+Select `Advanced Options` (here shown as Option 7, but yours may be different) and press `Enter`.
+
+Now select the Option named, `Audio` (here shown as A6, but yours may be different) and press `Enter`:
 
 ![Audio configuration screen](images/raspi-config-audio.png)
 
 Now you are presented with the two modes explained above as an alternative to the default `Auto` option. Select a mode, press `Enter` and press the right arrow key to exit the options list, then select `Finish` to exit the configuration tool.
+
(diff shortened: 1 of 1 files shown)


JamesH65:
Thanks.

lurch:
@spl237 Does you GUI wrapper around raspi-config also cover audio configuration? Is it worth mentioning that in these instructions too? (or should we just stick to the command-line tool for simplicity?)

spl237:
Audio config is all managed by right-clicking the speaker icon on the taskbar - the raspi-config controls are only in there for people using the command line. The audio switcher in the speaker icon is a lot more useful, as it will also connect to Bluetooth and USB devices, and (in the version I am working on as I type) will also set input devices. I'm not sure anyone is still using the raspi-config version any more; certainly not if they are running the desktop.

JamesH65:
The whole page needs a revamp, but I've merged this PR as its better than before, if not what we really need in the long run.

javaspaces:
@spl237 Here's a comment on the overall user experience, and not the documentation...

It would be really great (and more intuitive from my perspective) if all audio controls and options were merged into a single "applet" (for lack of a better term). I tried to change my audio settings on the Buster Desktop using LEFT-CLICK and was disappointed that I couldn't switch my audio output from there. I didn't even know that RIGHT-CLICK gave a different view.  

Additionally (for consistency), the network, Bluetooth, and clock icons work as expected in the top-right panel. Perhaps the audio icon user experience could be made consistent with them (in other words, LEFT-CLICK to change all options for the item, and RIGHT CLICK for panel settings)

spl237:
The audio switcher was one of the first things I added - it's now worked like that for five years. I think that changing that now would add more confusion than it removes...

PR #1289, opened 2019-09-20
Update boot.md

chryana:
Added comment to mention the default kernel booted on the Pi 4 is now kernel7l.img.

### DIFF
--- a/configuration/config-txt/boot.md
+++ b/configuration/config-txt/boot.md
@@ -35,7 +35,7 @@ Set the `disable_commandline_tags` command to `1` to stop `start.elf` from filli
 
 ## kernel
 
-`kernel` is the alternative filename on the boot partition to use when loading the kernel. The default value on the Pi 1, Pi Zero, and Compute Module is `kernel.img`, and on the Pi 2, Pi 3, and Compute Module 3 it is `kernel7.img`. If `kernel8.img` is present on the Pi 3 or Compute Module 3, it will be loaded in preference and entered in 64-bit mode. **NOTE**: This must be an uncompressed kernel image file.
+`kernel` is the alternative filename on the boot partition to use when loading the kernel. The default value on the Pi 1, Pi Zero, and Compute Module is `kernel.img`; on the Pi 2, Pi 3 and Compute Module 3 it is `kernel7.img`; on the Pi 4, it is `kernel7l.img`. If `kernel8.img` is present on the Pi 3 or Compute Module 3, it will be loaded in preference and entered in 64-bit mode. **NOTE**: This must be an uncompressed kernel image file.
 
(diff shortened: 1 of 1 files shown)


configuration/config-txt/boot.md
```diff
@@ -35,7 +35,7 @@ Set the `disable_commandline_tags` command to `1` to stop `start.elf` from filli
 
 ## kernel
 
-`kernel` is the alternative filename on the boot partition to use when loading the kernel. The default value on the Pi 1, Pi Zero, and Compute Module is `kernel.img`, and on the Pi 2, Pi 3, and Compute Module 3 it is `kernel7.img`. If `kernel8.img` is present on the Pi 3 or Compute Module 3, it will be loaded in preference and entered in 64-bit mode. **NOTE**: This must be an uncompressed kernel image file.
+`kernel` is the alternative filename on the boot partition to use when loading the kernel. The default value on the Pi 1, Pi Zero, and Compute Module is `kernel.img`; on the Pi 2, Pi 3 and Compute Module 3 it is `kernel7.img`; on the Pi 4, it is `kernel7l.img`. If `kernel8.img` is present on the Pi 3 or Compute Module 3, it will be loaded in preference and entered in 64-bit mode. **NOTE**: This must be an uncompressed kernel image file.
```
lurch: @pelwell Does the Pi4 also use the same `kernel8.img` filename for 64-bit mode?

pelwell: Yes it does - the trial 64-bit builds included in new firmwares will work on on Pi 4 and Pi 3 (and with a very up-to-date firmware, Pi 2 rev 1.2).

Howeve, this paragraph is no longer correct - the `arm_64bit` property controls whether or not to start in 64-bit mode, including selecting `kernel8.img` unless there is an explicit `kernel=...`, and the default value is currently 0 on all platforms so `kernel8.img` will _not_ be loaded in preference.

lurch: @JamesH65 Currently `arm_control` and `arm_64bit` are documented in configuration/config-txt/misc.md - I guess it probably makes sense to move them to configuration/config-txt/boot.md ?

pelwell: And we should deprecate arm_control - arm_64bit is where it's at.

JamesH65: OK, will sort that out.

JamesH65:
OK, new PR to replace this one here https://github.com/raspberrypi/documentation/pull/1290

JamesH65:
New PR has overtaken this one, but thanks to @chryana for the heads up.

chryana:
You're welcome. Thanks for updating the documentation so quickly.

PR #1300, opened 2019-10-02
Updated app list for raspbian

cntrlANC:


raspbian/applications/README.md
```diff
@@ -4,3 +4,7 @@
     - Information on the demonstration camera software
 - [OMXPlayer](omxplayer.md)
     - Information on the OMXPlayer command line media player
+- [Minecraft for pi]
```
lurch: @JamesH65 Perhaps this is supposed to be linking to `usage/minecraft/README.md` ?

raspbian/applications/README.md
```diff
@@ -4,3 +4,7 @@
...
+    - 3D block based building platform with its own simple 
+      line and command coding and also teaches electronic
+      circuitry.
```
lurch: @cntrlANC Just out of curiosity, how does Minecraft teach electronic circuitry?

raspbian/applications/README.md
JamesH65: Have added link and removed the electronic circuit bit, which is a feature, but don't need to reference it here I don't think.

raspbian/applications/README.md
XECDesign: Probably referring to redstone, but I don't think that works in the pi edition.

cntrlANC: yes.

cntrlANC: I can confirm that there is support for redstone as it is a more polished version. Minecraft pi edition also has support for extended programming languages.

lurch:
Hmmm, on seconds thoughts, how do we determine what gets listed on https://github.com/raspberrypi/documentation/blob/master/raspbian/applications/README.md and what gets listed on https://github.com/raspberrypi/documentation/blob/master/usage/README.md ?
@JamesH65 @bennuttall 

_Perhaps_ it makes sense to move the pages under `raspbian/applications/` to instead be under `usage/` (with appropriate 'redirect pages' in place, to prevent external-links breaking) and then remove the "Applications" section from https://github.com/raspberrypi/documentation/blob/master/raspbian/README.md ?

bennuttall:
> Hmmm, on second thought, how do we determine what gets listed on...

Just whatever makes sense

PR #1308, opened 2019-10-16
Update updating.md

JamesH65:
dist-upgrade explanation

### DIFF
--- a/raspbian/updating.md
+++ b/raspbian/updating.md
@@ -26,7 +26,7 @@ Next, **upgrade** all your installed packages to their latest versions with the 
 sudo apt full-upgrade
 ```
 
-Note that `dist-upgrade` is used in preference to a simple `upgrade`, as it also picks up any dependency changes that may have been made. 
+Note that `full-upgrade` is used in preference to a simple `upgrade`, as it also picks up any dependency changes that may have been made. 
 
 Generally speaking, doing this regularly will keep your installation up to date for the particular major Raspbian release you are using (e.g. Stretch). It will not update from one major release to another, for example, Stretch to Buster.
 
--- a/raspbian/updating.md
+++ b/raspbian/updating.md
@@ -26,6 +26,8 @@ Next, **upgrade** all your installed packages to their latest versions with the 
 sudo apt-get dist-upgrade
 ```
 
+Note that `dist-upgrade` is used in preference to a simple `upgrade`, as it also picks up any dependency changes that may have been made. 
+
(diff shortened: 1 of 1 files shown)


JamesH65:
I wonder if we should also change apt-get to apt at this point?

XECDesign:
> I wonder if we should also change apt-get to apt at this point?

Yup

JamesH65:
Oh, goodo, not too many then!! I'll grab a full copy of the tree and do a global replace in a separate PR.

JamesH65:
See #1310

raspbian/updating.md
```diff
@@ -36,7 +38,7 @@ The kernel and firmware are installed as a Debian package, and so will also get
 
 ### Running out of space
 
-When running `sudo apt-get dist-upgrade`, it will show how much data will be downloaded and how much space it will take up on the SD card. It's worth checking with `df -h` that you have enough free disk space, as unfortunately `apt` will not do this for you. Also be aware that downloaded package files (`.deb` files) are kept in `/var/cache/apt/archives`. You can remove these in order to free up space with `sudo apt-get clean`.
+When running `sudo apt dist-upgrade`, it will show how much data will be downloaded and how much space it will take up on the SD card. It's worth checking with `df -h` that you have enough free disk space, as unfortunately `apt` will not do this for you. Also be aware that downloaded package files (`.deb` files) are kept in `/var/cache/apt/archives`. You can remove these in order to free up space with `sudo apt clean`.
```
XECDesign: One difference in behaviour between `apt` and `apt-get` is that `apt` will remove the downloaded archives afterwards. So the last two sentences are not necessary.

lurch:
Should this be modified to use `full-upgrade` too, @XECDesign ?

XECDesign:
I'm guessing that since this PR will conflict with #1310, it should probably all just be done there.

raspbian/updating.md
```diff
@@ -23,7 +23,7 @@ sudo apt update
 Next, **upgrade** all your installed packages to their latest versions with the following command:
 
 ```bash
-sudo apt dist-upgrade
+sudo apt full-upgrade
 ```
 
 Note that `dist-upgrade` is used in preference to a simple `upgrade`, as it also picks up any dependency changes that may have been made.
```
lurch: I guess this line should have been changed to `full-upgrade` too? :eyes:

JamesH65: No worries, I only spotted it on a proof read before merging!

PR #1315, opened 2019-10-26
Further hardening

thecolorjay:
Working progress...


JamesH65:
Happy with much of this, but not happy about the TOR/Onionising stuff. That would need to be removed before the PR could be accepted. It's not Raspberry Pi's job to document TOR, and it could be misconstrued.

thecolorjay:
> Happy with much of this, but not happy about the TOR/Onionising stuff. That would need to be removed before the PR could be accepted. It's not Raspberry Pi's job to document TOR, and it could be misconstrued.

Sweet. Okay, I'll remove Tor.

configuration/security.md
```diff
@@ -92,6 +92,28 @@ alice ALL=(ALL) PASSWD: ALL
...
+sudo passwd -l root
+```
+
+To enable root login, use:
+
+```bash
+sudo passwd -u root
+```
+
+To enable root login again, use:
```
lurch: I don't understand why there's "To enable root login" and "To enable root login again" ??

configuration/security.md
```diff
@@ -218,20 +296,33 @@ Limit login attempts on ssh port using tcp: this denies connection if an IP addr
...
+Allow incoming requests to port 22 from IP address 192.168.2.1
+
+```bash
+sudo ufw allow in proto tcp from 192.168.2.1 to any port 22
+```
+
+Limit incoming requests to port 22 from IP address 192.168.2.1
+
+```bash
+sudo ufw limit in proto tcp from 192.168.2.1 to any port 22
```
lurch: I wonder if it'd make sense to use different example IP addresses (rather than using `192.168.2.1` each time) for the "Deny", "Allow" and "Limit" examples? :man_shrugging:

configuration/security.md
```diff
@@ -218,20 +296,33 @@ Limit login attempts on ssh port using tcp: this denies connection if an IP addr
...
+sudo ufw deny in proto tcp from 192.168.2.1 to any port 22
 ```
 
+Allow incoming requests to port 22 from IP address 192.168.2.1
+
+```bash
+sudo ufw allow in proto tcp from 192.168.2.1 to any port 22
+```
+
+Limit incoming requests to port 22 from IP address 192.168.2.1
```
lurch: Would "Limit incoming requests from IP address 192.168.2.1 to port 22 only" make this easier to read?

lurch:
ping @mythic-beasties in case Pete wants to have a quick look at this.

mythic-beasties:
I would avoid dropping ICMP messages, especially IPv6 messages as they are likely to cause operational issues (time-exceeded and parameter problem spring to mind). c.f. 

https://tools.ietf.org/html/rfc4890

See also

https://security.stackexchange.com/questions/22711/is-it-a-bad-idea-for-a-firewall-to-block-icmp

In general I'd recommend to leave ICMP alone unless you know what and why you're filtering. The security benefit is minimal but your probability of degrading your connectivity and/or disabling your debugging tools for real problems is pretty high.

configuration/security.md
thecolorjay: By locking it, you're disabling it entirely. By enabling it again after unlocking it, you're setting the password.

configuration/security.md
thecolorjay: Possibly, although I don't think it matters, it's just an example.

configuration/security.md
thecolorjay: Good point, I'll change that.

thecolorjay:
> I would avoid dropping ICMP messages, especially IPv6 messages as they are likely to cause operational issues (time-exceeded and parameter problem spring to mind). c.f.
> 
> https://tools.ietf.org/html/rfc4890
> 
> See also
> 
> https://security.stackexchange.com/questions/22711/is-it-a-bad-idea-for-a-firewall-to-block-icmp
> 
> In general I'd recommend to leave ICMP alone unless you know what and why you're filtering. The security benefit is minimal but your probability of degrading your connectivity and/or disabling your debugging tools for real problems is pretty high.

Alright, thanks for letting me know. I think it's up to the administrator though, at least we're providing them with the information. [https://wiki.archlinux.org/index.php/Uncomplicated_Firewall#Disable_remote_ping](https://wiki.archlinux.org/index.php/Uncomplicated_Firewall#Disable_remote_ping)

thecolorjay:
> Happy with much of this, but not happy about the TOR/Onionising stuff. That would need to be removed before the PR could be accepted. It's not Raspberry Pi's job to document TOR, and it could be misconstrued.

By that logic though, nobody would be writing documentation for Tor. I'm pretty sure Tor Project encourages everyone to write documentation for Tor, we're a community after all, where's your free software spirit?

JamesH65:
If Mythic seem to think ICMP stuff is likely to cause problems, then I am inclined to remove it - we cannot afford for an everyday user to break their systems simply by following a set of instructions on the site and this has the capability of doing that. So either needs a big warning around it "experienced admins only as this may cause problems unless you really know what you are doing" or not to be there at all. 

As for TOR, you could supply that in a separate PR, and I can pass it by legal and the boss. I'm not convinced though, and neither were the people I passed it past this morning, but worth a try.

thecolorjay:
> If Mythic seem to think ICMP stuff is likely to cause problems, then I am inclined to remove it - we cannot afford for an everyday user to break their systems simply by following a set of instructions on the site and this has the capability of doing that. So either needs a big warning around it "experienced admins only as this may cause problems unless you really know what you are doing" or not to be there at all.
> 
> As for TOR, you could supply that in a separate PR, and I can pass it by legal and the boss. I'm not convinced though, and neither were the people I passed it past this morning, but worth a try.

Okay, I'll add a warning.

configuration/security.md
lurch: Sorry, I'm still confused - is there now a difference between "locking" & "disabling" and "enabling" & "unlocking" ?

I guess what I was asking originally (but perhaps not very clearly), was: should the "To enable root login" and "To enable root login again" sections be merged into one?

configuration/security.md
```diff
@@ -158,17 +180,53 @@ UsePAM no
...
+
+# ok icmp codes for INPUT
+-A ufw6-before-input -p icmpv6 --icmpv6-type destination-unreachable -j DROP
+-A ufw6-before-input -p icmpv6 --icmpv6-type source-quench -j DROP
+-A ufw6-before-input -p icmpv6 --icmpv6-type time-exceeded -j DROP
+-A ufw6-before-input -p icmpv6 --icmpv6-type parameter-problem -j DROP
+-A ufw6-before-input -p icmpv6 --icmpv6-type echo-request -j DROP
+```
+
+To reset the firewall, use:
```
lurch: Is it worth a bit more explanation about what "reset the firewall" means / does ?

configuration/security.md
```diff
@@ -218,20 +296,33 @@ Limit login attempts on ssh port using tcp: this denies connection if an IP addr
...
 
-Deny access to port 30 from IP address 192.168.2.1
+Deny incoming requests from IP address 192.168.2.1 to port 22 only
 
 ```bash
-sudo ufw deny from 192.168.2.1 port 30
+sudo ufw deny in proto tcp from 192.168.2.1 to any port 22
 ```
 
+Allow incoming requests from IP address 192.168.2.1 to port 22 only
```
lurch: What's the difference between "Allow incoming requests from ..." and "Limit incoming requests from ..." ?

(sorry for being picky, but I think if we're advising (potentially) novice users on how to "harden" their system, we need to be clear exactly what we're doing in order that they don't accidentally enable / disable something they didn't mean to)

configuration/security.md
```diff
@@ -218,20 +296,33 @@ Limit login attempts on ssh port using tcp: this denies connection if an IP addr
 sudo ufw limit ssh/tcp
 ```
 
-Deny access to port 30 from IP address 192.168.2.1
+Deny incoming requests from IP address 192.168.2.1 to port 22 only
```
lurch: I've never set up a firewall myself (so apologies of this is a naive question), but if you're blocking a specific IP address, isn't it more likely that you'd block access to all ports, rather than just a specific port? :man_shrugging:

configuration/security.md
thecolorjay: > Sorry, I'm still confused - is there now a difference between "locking" & "disabling" and "enabling" & "unlocking" ?
> 
> I guess what I was asking originally (but perhaps not very clearly), was: should the "To enable root login" and "To enable root login again" sections be merged into one?

No, they do two different things. Once you lock and unlock the account, you've also gotta use passwd to reset the password.

configuration/security.md
thecolorjay: The difference is that by limiting incoming requests you're adding brute-force protection by only allowing 6 requests to be made from that IP address every 30 seconds. I can add some explanation for this.

JamesH65: Everything needs to be thoroughly explained - this is a beginners guide so everything has to be crystal clear in its explanations and its implications.

configuration/security.md
thecolorjay: Yes, it's actually redundant if you're blocking all incoming requests by default already but this is just an example of what's possible. This is why the other rules for allowing and limiting are included as well.

configuration/security.md
thecolorjay: > Is it worth a bit more explanation about what "reset the firewall" means / does ?

Yes, I can provide some explanation for that. You're resetting the firewall back to it's defaults and disabling it. I'll be sure to clarify this.

configuration/security.md
thecolorjay: Fair enough, I've just added a fair bit of explanation and I'll be looking through it to see if I should add anything else.

configuration/security.md
lurch: > Once you lock and unlock the account, you've also gotta use `$ sudo passwd root` to reset the password.

I think that's exactly what I was suggesting with:
> I guess what I was asking originally (but perhaps not very clearly), was: should the "To enable root login" and "To enable root login again" sections be merged into one?

:question: :confused:

lurch: > Once you lock and unlock the account, you've also gotta use `$ sudo passwd root` to reset the password.

I think that's exactly what I was suggesting with:
> I guess what I was asking originally (but perhaps not very clearly), was: should the "To enable root login" and "To enable root login again" sections be merged into one?

:question: :confused: 

(your current instructions don't make it clear that after unlocking / enabling a previously-locked root account, you **also** need to change / reset the root password)

configuration/security.md
```diff
@@ -180,6 +217,24 @@ To disable the firewall, and disable start up on boot, use:
 sudo ufw disable
 ```
 
+To disable logging, use:
```
lurch: Is it worth explaining why a user might want to / not want to disable logging?

configuration/security.md
```diff
@@ -208,30 +263,43 @@ The status command lists all current settings for the firewall:
 
 ```bash
 sudo ufw status
+sudo ufw status verbose
+sudo ufw status numbered
 ```
 
 The rules can be quite complicated, allowing specific IP addresses to be blocked, specifying in which direction traffic is allowed, or limiting the number of attempts to connect, for example to help defeat a Denial of Service (DoS) attack. You can also specify the device rules are to be applied to (e.g. eth0, wlan0). Please refer to the `ufw` man page (`man ufw`) for full details, but here are some examples of more sophisticated commands.
 
 Limit login attempts on ssh port using tcp: this denies connection if an IP address has attempted to connect six or more times in the last 30 seconds:
```
lurch: It seems the example following this text has disappeared?

configuration/security.md
lurch: Extra explanatory text looks good :+1:

configuration/security.md
thecolorjay: Okay, I added clarification.

configuration/security.md
thecolorjay: > Is it worth explaining why a user might want to / not want to disable logging?

Good point.

configuration/security.md
thecolorjay: What example are you referring to? I've checked and everything I've added is still there as far as I can see.

configuration/security.md
lurch: The new text suggests _when_ you might want to disable logging, but it still doesn't suggest _why_ :wink:

configuration/security.md
lurch: It seems you've accidentally removed the `sudo ufw limit ssh/tcp` line?

configuration/security.md
```diff
@@ -92,6 +92,30 @@ alice ALL=(ALL) PASSWD: ALL
 
 Now save the file.
 
+
+## Disable or enable root login
+
+To disable root login, use:
```
lurch: @spl237 @XECDesign  Doesn't Raspbian already come without a root login? If so, should this section be removed?

configuration/security.md
thecolorjay: Alright, changing it now.

configuration/security.md
thecolorjay: Oh, I got rid of that because I didn't see it as necessary since there are already examples for that. Should I put it back?

lurch: That's up to you - but if you leave it out then I think you need to remove the "Limit login attempts on ssh port using tcp: ..." bit too?

configuration/security.md
lurch: "you may wish to disable logging as you'll be denying quite a lot of requests"

Still not clear (to me) _why_ this means you'd want to disable logging? Does logging slow down ufw? Or does it use a lot of diskspace? Or something else?

JamesH65:
Did you mean to close this - thought we were getting it in pretty good shape?

thecolorjay:
> Did you mean to close this - thought we were getting it in pretty good shape?

Yes I did, I'll take what I have elsewhere. And I noticed you updated what you guys already have with something you took from my pull request...

JamesH65:
Can you be more specific? I've not taken anything from anything today.  We have merged a few PR's that have been waiting for copy editing for some weeks, is that what you mean?

JamesH65:
@lurch Is it worth keeping this PR? Got some useful stuff in it.

PR #1318, opened 2019-10-29
Hints for use with IPv6 added

thomasschaeferm:
There are big differences in handling ipv4 and ipv6 for remote access in residential environments(dsl,cable,fibre, maybe mobile too).
In ipv4 it was very common to use address and port forwarding. The router was the only device with a public address. So the router got a packet and translates it (address and port) to a private ip on the lan. Also the router was (mostly) responsible for getting a (nice) dns name.

With ipv6 something has changed. Depending on your isp (ds-lite(carrier grade nat) vs. dualstack) you don't get a public ipv4 address anymore.
Instead of one public ipv4 address you get a whole net with ipv6 addresses.
Your computers get one or more addresses from that net. They all are public, but usually protected by the routers firewall against unwanted incoming packets.

To use these addresses for remote access we need to do at least two things:

open the routers firewall or a particularly service/port for the device on the lan (no nat anymore)
create one dns name for that device (not for the routers ip address anymore)

opening/configuring the firewall has to be done on the router
dns updates can be managed by the device itself or by the router.

Since privacy extensions (randomizing ip addresses) are in conflict with server use, I recommend to disable privacy extensions for that use case

My examples:

web server: http://tschaefer.dynv6.net
remote desktop: xfreerdp +glyph-cache /v:raspberrypi4.zts9jt8obs8kionr.myfritz.net

ssh: both

### DIFF
--- a/remote-access/access-over-Internet/README.md
+++ b/remote-access/access-over-Internet/README.md
@@ -6,9 +6,30 @@ You can connect to your Raspberry Pi over the internet from another computer or 
 
 One method is to set up port forwarding on your router. To do this, you must change the configuration of your router to forward all inbound traffic from the internet on a specific port to the local IP address of your Raspberry Pi. Most routers have this feature available. However, every router is different so you will need to consult your router's user manual for instructions. The settings can be tricky if your Pi is behind a firewall or if there is more than one router. One disadvantage of port forwarding is that it exposes a network port on your private LAN to the public internet. This is a known security vulnerability and must be managed carefully.
 
-## Port releasing (IPv6, e.g. DS-lite)
+## Introduction for IPv6 newcomers
+
+There are big differences in handling IPv4 and IPv6 for remote access in residential environments(dsl,cable,fibre, maybe mobile too).
+With IPv4 it was and is still very common to use the public address of the router in combination with port forwarding. The router is the only device with a public address. So the router gets a packet and translates it (address and port) to a private ip on the lan. Also the router is (mostly) responsible for getting a (nice) DNS name.
+With IPv6 some things have changed. Depending on your isp (different types of ds-lite and dualstack) you don't get a public ipv4 address anymore.
+Instead of one public IPv4 address you get a whole network with IPv6 addresses (prefix).
+Your computers get one or more addresses from that network. They all are public, but usually protected by the routers firewall against unwanted inbound traffic.
+To use these addresses for remote access we need to do at least two things. 
+Open the routers firewall or a particularly service/port for the device on the lan (there is no nat anymore). 
+Create one DNS name for that device (not for the routers ip address anymore).
+Opening/configuring the firewall has to be done on the router.
+DNS updates can be managed by the device itself(similar to IPv4 updaters) or by the router. 
+Since privacy extensions (randomizing ip addresses) are in conflict with server use, it is recommend to disable privacy extensions for that use case.
+It is possible to use IPv4 port forwarding and IPv6 Port opening at the same time. It depends on your ISP.
+In case of ds-lite you have only the choice to use IPv6 directly or to use some cloud relays/mappers.
+
+
+## Port opening (IPv6, e.g. DS-lite)
 
 Another method is to release the IP address (also called "exposed host") or ports of that IP address on your router. To do this, you must change the firewall configuration to allow inbound traffic to a specific port/IP. Sometimes it is called "interface id" or just "host part" of the IPv6 address. To get a stable interface id you should change `slaac private` to `slaac hwaddr` in `/etc/dhcpcd.conf` . Afterwards you can use DNS-services like dynv6.com or myfritz.net. Security considerations are the same as for IPv4 mentioned above. Additionally you may lose a little bit privacy while using the `hwaddr`.
+
+This description was tested with dynv6.com service(DNS-updater installed on the PI) and with myfritz service by avm (DNS-updater integrated in the fritzbox). It should work also with other IPv6 capable dyndns providers.
+
+
 
 
 ## Alternatives
--- a/remote-access/access-over-Internet/README.md
+++ b/remote-access/access-over-Internet/README.md
@@ -15,11 +15,11 @@ Instead of one public IPv4 address you get a whole network with IPv6 addresses (
 Your computers get one or more addresses from that network. They all are public, but usually protected by the routers firewall against unwanted inbound traffic.
 To use these addresses for remote access we need to do at least two things. 
 
-Open the routers firewall or a particularly service/port for the device on the lan (there is no nat anymore). 
+Open the routers firewall or a particularly service/port for the device on the lan (there is no nat anymore).
+
 Create one DNS name for that device (not for the routers ip address anymore).
 
 Opening/configuring the firewall has to be done on the router. Maybe on your Pi too if you are using iptables/nft.
-
 DNS updates can be managed by the device itself(similar to IPv4 updaters) or by the router. 
 Since privacy extensions (randomizing ip addresses) are in conflict with server use, it is recommend to disable privacy extensions for that use case.
 It is possible to use IPv4 port forwarding and IPv6 Port opening at the same time. It depends on your ISP.
--- a/remote-access/access-over-Internet/README.md
+++ b/remote-access/access-over-Internet/README.md
@@ -8,7 +8,7 @@ One method is to set up port forwarding on your router. To do this, you must cha
 
 ## Port releasing (IPv6, e.g. DS-lite)
 
-Another method is to release the IP address (also called "exposed host") or ports of that IP address on your router. To do this, you must change the firewall configuration to allow inbound traffic to a specific port/IP. Sometimes it is called "interface id" or just "host part" of the IPv6 address. To get a stable interface id you should change `slaac private` to `slaac hwaddr` in `/etc/dhcpcd.conf` . Afterwards you can use DNS-services like dynv6.com or myfritz.net. Security considerations are the same as for IPv4 mentioned above. Addtionally you may lose a little bit privacy while using the `hwaddr`.
+Another method is to release the IP address (also called "exposed host") or ports of that IP address on your router. To do this, you must change the firewall configuration to allow inbound traffic to a specific port/IP. Sometimes it is called "interface id" or just "host part" of the IPv6 address. To get a stable interface id you should change `slaac private` to `slaac hwaddr` in `/etc/dhcpcd.conf` . Afterwards you can use DNS-services like dynv6.com or myfritz.net. Security considerations are the same as for IPv4 mentioned above. Additionally you may lose a little bit privacy while using the `hwaddr`.
 
 
 ## Alternatives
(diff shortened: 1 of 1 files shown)


lurch:
@JamesH65 Any idea how widespread IPv6 networks are? Do you think they're common enough to warrant inclusion in the RPi documentation?

thomasschaeferm:
There statistics by google, akamai, facebook and apnic.
https://www.google.de/ipv6/statistics.html
https://www.facebook.com/ipv6/?tab=ipv6_country
https://www.akamai.com/de/de/resources/our-thinking/state-of-the-internet-report/state-of-the-internet-ipv6-adoption-visualization.jsp
https://stats.labs.apnic.net/ipv6

At least in Germany they are very common. Also in Germany DS-lite by Vodafone Cable (no public IPv4 addresses anymore !) is very common.
So for millions of users IPv6 is the only possibility for remote access. My proposal is just about some general router/dns hints. The rpi itself works with IPv6 like a charm.

JamesH65:
I like the idea, but reading it I'm still not sure what is actually happening. Can you add some stuff, won't need much, to explain specifically what this is doing?

thomasschaeferm:
I added a description.

JamesH65:
I am not seeing an new commit on the PR? Did you push it?

thomasschaeferm:
I did not change the pull request itself. I think it should not be bigger. 
I added a motivation and a description to the first/initial comment of this conversation.
The initial description of the pull request was empty before.

JamesH65:
AH, I see. I think some or all of that description does need to be in the PR text, or people will not actually know what they are doing and why, which will make for problems with support later. The more information provided the better.

remote-access/access-over-Internet/README.md
```diff
@@ -2,10 +2,39 @@
...
+## Introduction for IPv6 newcomers
+
+There are big differences in handling IPv4 and IPv6 for remote access in residential environments(dsl,cable,fibre, maybe mobile too).
+With IPv4 it was and is still very common to use the public address of the router in combination with port forwarding. The router is the only device with a public address. So the router gets a packet and translates it (address and port) to a private IP address on the lan. Also the router is (mostly) responsible for getting a (nice) DNS name.
+With IPv6 some things have changed. Depending on your isp (different types of ds-lite and dualstack) you don't get a public IPv4 address anymore.
+Instead of one public IPv4 address you get a whole network with IPv6 addresses (prefix).
+Your computers get one or more addresses from that network. They all are public, but usually protected by the routers firewall against unwanted inbound traffic.
+To use these addresses for remote access we need to do at least two things. 
+
+* Open the routers firewall or a particularly service/port for the device on the lan (there is no nat anymore).
```
lurch: "(there is no nat anymore)" seems to imply that NAT (which I guess should be capitalised?) has already been defined / explained previously in the document? :wink:

remote-access/access-over-Internet/README.md
```diff
@@ -2,10 +2,39 @@
...
 
+## Introduction for IPv6 newcomers
+
+There are big differences in handling IPv4 and IPv6 for remote access in residential environments(dsl,cable,fibre, maybe mobile too).
+With IPv4 it was and is still very common to use the public address of the router in combination with port forwarding. The router is the only device with a public address. So the router gets a packet and translates it (address and port) to a private IP address on the lan. Also the router is (mostly) responsible for getting a (nice) DNS name.
+With IPv6 some things have changed. Depending on your isp (different types of ds-lite and dualstack) you don't get a public IPv4 address anymore.
+Instead of one public IPv4 address you get a whole network with IPv6 addresses (prefix).
```
lurch: What does "(prefix)" mean here?

remote-access/access-over-Internet/README.md
```diff
@@ -2,10 +2,39 @@
...
 
+## Introduction for IPv6 newcomers
+
+There are big differences in handling IPv4 and IPv6 for remote access in residential environments(dsl,cable,fibre, maybe mobile too).
+With IPv4 it was and is still very common to use the public address of the router in combination with port forwarding. The router is the only device with a public address. So the router gets a packet and translates it (address and port) to a private IP address on the lan. Also the router is (mostly) responsible for getting a (nice) DNS name.
+With IPv6 some things have changed. Depending on your isp (different types of ds-lite and dualstack) you don't get a public IPv4 address anymore.
```
lurch: What's ds-lite? When I google for it I only get results about Nintendo :confused:

lurch:
@thomasschaeferm Starting to look much better now, thanks :+1:

remote-access/access-over-Internet/README.md
thomasschaeferm: My raspberry pi 2 has
`2001:a61:475:6601:ba27:ebff:feb6:f293`
My raspberry pi 4 has
`2001:a61:475:6601:dea6:32ff:fe23:6be1`
They share the same prefix `2001:a61:475:6601::/64` I got that prefix from my ISP ( in reality it is /56, but some providers give more, some only /64)

remote-access/access-over-Internet/README.md
thomasschaeferm: That' also a problem with the regional/language settings of google. I mean the technical term, defined in https://tools.ietf.org/html/rfc6333 .
It is not the only one, there are different implementations e.g. from greece
https://blog.apnic.net/2018/07/04/a-large-scale-lw4o6-deployment-for-greek-broadband-users/ 
They have on common, that IPv6 is usable in both directions, IPv4 only as client

remote-access/access-over-Internet/README.md
lurch: Ahh, in that case maybe it's worth expanding "(prefix)." to something like "(with a common prefix)." ?

remote-access/access-over-Internet/README.md
lurch: Perhaps it's worth mentioning the phrase "Dual-Stack Lite" at least once, so people have more context?

JamesH65:
Everyone happy with this? Reads OK to me.

remote-access/access-over-Internet/README.md
```diff
@@ -2,10 +2,39 @@
...
 You can connect to your Raspberry Pi over the internet from another computer or a mobile device. There are a number of ways to do this, which we cover below.
 
-## Port forwarding
+## Port forwarding (IPv4)
 
 One method is to set up port forwarding on your router. To do this, you must change the configuration of your router to forward all inbound traffic from the internet on a specific port to the local IP address of your Raspberry Pi. Most routers have this feature available. However, every router is different so you will need to consult your router's user manual for instructions. The settings can be tricky if your Pi is behind a firewall or if there is more than one router. One disadvantage of port forwarding is that it exposes a network port on your private LAN to the public internet. This is a known security vulnerability and must be managed carefully.
 
+## Introduction for IPv6 newcomers
+
+There are big differences in handling IPv4 and IPv6 for remote access in residential environments(dsl,cable,fibre, maybe mobile too).
```
lurch: spacing: "environments(dsl,cable,fibre," -> "environments (dsl, cable, fibre,"

remote-access/access-over-Internet/README.md
```diff
@@ -2,10 +2,39 @@
...
+With IPv4 it was and is still very common to use the public address of the router in combination with port forwarding. The router is the only device with a public address. So the router gets a packet and translates it (address and port) to a private IP address on the lan. Also the router is (mostly) responsible for getting a (nice) DNS name.
+With IPv6 some things have changed. Depending on your isp (different types of dualstack-lite and dualstack) you don't get a public IPv4 address anymore.
+Instead of one public IPv4 address you get a whole network with IPv6 addresses. Your devices share a common public prefix, usually with a length of 64bit or less.
```
lurch: "a whole network with IPv6 addresses" -> "a whole network of IPv6 addresses" ?

remote-access/access-over-Internet/README.md
```diff
@@ -2,10 +2,39 @@
...
+Your computers assign themself one or more addresses from that network(SLAAC) or they get it assigned(DHCPv6). They all are public, but usually protected by the routers firewall against unwanted inbound traffic.
```
lurch: Space before brackets please :slightly_smiling_face:

remote-access/access-over-Internet/README.md
```diff
@@ -2,10 +2,39 @@
...
+To use these addresses for remote access we need to do at least two things. 
+
+* Open the routers firewall or a particularly service/port for the device on the lan. There is no NAT anymore. (NAT = network address/port translation, where the IPv4 port forwarding is based on).
```
lurch: routers -> router's
particularly -> particular

lurch:
I wonder if @mythic-beasties might want to have a quick look at this, since IIRC he's _very_ familiar with IPv6 ?

mythic-beasties:
I suspect this is going to be pretty hard to make generic as life is now a lot more complicated. You've got a variety of network types, 

Traditional v4 + NAT (most fixed line)
Your router has a public v4 address which every device behind it uses. You may be able to configure it to forward to a private v4 address on your Pi. Your public address may be static, may be sticky dynamic (i.e. basically never changes) or change every reconnect.

Traditional v4 with CGNAT (lots of mobile, some fixed line)
Your router has a private v4 address, the public address is somewhere in the network. Inbound traffic is prohibited.

Traditional v4 + NAT + v6 (some fixed line)
See above, but you have a static or dynamic v6 allocation. In addition to port forwarding, you can also just open up the router to allow inbound v6 connections. By default your Pi will use privacy addresses so it's v6 address will continuously change, you'll want it to have a static or hardware defined address, and you may also need external DNS if your v6 allocation keeps changing.

Traditional v4 with CGNAT + v6 (lots of mobile, newer ISPs)
The only way to have inbound traffic is via v6 as above.

v6 only with NAT64
Nobody is brave enough to do this as a residential ISP, but the Mythic Beasts PiCloud does this. IPv4 just isn't supported at all, all v4 connections get translated to v6 ones.

There's a whole bunch of implementation details about how modern networks are being rolled out, depending on if they've gone all in for IPv4 only and CGNAT (e.g. Vodafone's new 1Gbps fibre setup), or a pure v6 network with the IPv4 being provided as a compatibility layer on the top (Sky Italia).

The trend is clear though, v4 inbound is only likely to work on an expensive business only line, v6 inbound is increasingly likely to work on comon connections, if you're allowed to enable inbound connections.

However, I think general documentation from Pi Towers  can only be starting point due to the variability of what's offered.

thomasschaeferm:
Thank you for the long comment. Despite the variety of network types we should provide a path where the user can start from.

lurch:
Thanks Pete :heart:

JamesH65:
Whereabouts are we with this?

JamesH65:
Sorry about the delay. This PR seems OK to me, it gives more information than before, and is a starting point. I wonder if Mythic-beasties comments should make it in to the docs somehow as well.

aallan:

thomasschaeferm:
I think it is still relevant, except you cancel the whole topic about "Access your Raspberry Pi over the internet".
Of course my original version was shorter. It was suggested to make a short introduction to IPv6. And then we did it.
(I do it that way with all the Raspberry PIs I have.)

PR #1320, opened 2019-10-30
find IP addresses

thomasschaeferm:
find IP addresses


lurch:
This PR seems to include some (unfinished?) `xrdp` changes too?

thomasschaeferm:
Sorry, I made to things at once. xrdp-doc is indeed unfinished (but it works nice by the way).
 I am not very familiar with git/github. I try to separate the finished part from the unfinished part.
Or you just copy my proposal.

PR #1322, opened 2019-10-31
Update ip-address.md

thomasschaeferm:
finding the ipv6 with ping

### DIFF
--- a/remote-access/ip-address.md
+++ b/remote-access/ip-address.md
@@ -103,7 +103,7 @@ ping -c 2 -I 2001:db8:494:9d01:ba27:ebff:feb6:f293  ff02::1%eth0
      
 `ff02::1` is a well known multicast address for all nodes on the link, so it behaves like a local broadcast, usually it is defined in `/etc/hosts` so you can also use the name (`ip6-allnodes` or `ipv6-allnodes`) instead of the literal address
 
-some newer systems expect the interface ID behind the multicast address    
+Some newer systems expect the interface ID behind the multicast address.   
 
 ```
 ping -c 2 -I 2001:db8:494:9d01:ba27:ebff:feb6:f293 ip6-allnodes
@@ -124,7 +124,7 @@ rtt min/avg/max/mdev = 0.480/1.283/2.623/0.735 ms
 This should result in replies from all the nodes on your (W)LAN link, with associated DNS names.
 
 Exclude your own IP( here `2001:db8:494:9d01:ba27:ebff:feb6:f293` ), 
-then check the others by trying to connect them via SSH
+then check the others by trying to connect them via SSH.
 
 ```
 ssh pi@2001:db8:494:9d01:dea6:32ff:fe23:6be1
--- a/remote-access/ip-address.md
+++ b/remote-access/ip-address.md
@@ -89,7 +89,7 @@ by `hostname -I`
 The example shows two IP addresses. The first one is a so called unique local unicast address(`fc00::/7`). The second one is the global unicast address(`2000::/3`). It is also possible to see only one them depending on your network (router) configuration. Both addresses are valid for reaching the Raspberry Pi within your LAN. The address out of `2000::/3` is accessible world wide, provided your router's firewall is opened.
 
 
-Now use one of IPs from the first step for ping all local nodes:
+Now use one of IPs from the first step to ping all local nodes:
 
 ```
 ping -c 2 -I 2001:db8:494:9d01:ba27:ebff:feb6:f293  ff02::1
--- a/remote-access/ip-address.md
+++ b/remote-access/ip-address.md
@@ -92,8 +92,8 @@ The example shows two IP addresses. The first one is a so called unique local un
 Now use one of IPs from the first step for ping all local nodes:
 
 ```
-ping -c 2 -I 2001:db8:494:9d01:ba27:ebff:feb6:f293  ff02::1`
-ping -c 2 -I 2001:db8:494:9d01:ba27:ebff:feb6:f293  ff02::1%eth0`
+ping -c 2 -I 2001:db8:494:9d01:ba27:ebff:feb6:f293  ff02::1
+ping -c 2 -I 2001:db8:494:9d01:ba27:ebff:feb6:f293  ff02::1%eth0
 ```
 `-c 2` stands for sending two echo requests
 
--- a/remote-access/ip-address.md
+++ b/remote-access/ip-address.md
@@ -121,10 +121,10 @@ PING ip6-allnodes(ip6-allnodes (ff02::1)) from 2001:db8:494:9d01:ba27:ebff:feb6:
 rtt min/avg/max/mdev = 0.480/1.283/2.623/0.735 ms
 
 ```
-Now, you got the answers from all your nodes on the link (your (W)LAN), with a little bit luck - also with dns names
+This should result in replies from all the nodes on your (W)LAN link, with associated DNS names.
 
 Exclude your own IP( here `2001:db8:494:9d01:ba27:ebff:feb6:f293` ), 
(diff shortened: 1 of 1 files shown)


remote-access/ip-address.md
```diff
@@ -79,6 +79,41 @@ Nmap done: 256 IP addresses (4 hosts up) scanned in 2.41 seconds
...
+64 bytes from 2001:db8:494:9d01:dea6:32ff:fe23:6be1: icmp_seq=2 ttl=64 time=0.994 ms (DUP!)
+64 bytes from 2001:db8:494:9d01:da37:beff:fefd:f09d: icmp_seq=2 ttl=255 time=0.997 ms (DUP!)
+64 bytes from 2001:db8:494:9d01:464e:6dff:fe72:8a08: icmp_seq=2 ttl=64 time=1.69 ms (DUP!)
+64 bytes from 2001:db8:494:9d01:ba27:ebff:feb6:f293: icmp_seq=3 ttl=64 time=0.564 ms
+
+--- ff02::1 ping statistics ---
+3 packets transmitted, 3 received, +6 duplicates, 0% packet loss, time 2003ms
+rtt min/avg/max/mdev = 0.454/1.173/2.518/0.599 ms
+```
+exclude your your own IP, check the others
```
lurch: Could you add a bit more explanation here about what this means?

lurch:
Thanks, looking better now. It's still not clear what "check the others" means though - are you effectively saying "See which IPv6 addresses respond to an SSH login attempt" ?

remote-access/ip-address.md
```diff
@@ -79,6 +79,52 @@ Nmap done: 256 IP addresses (4 hosts up) scanned in 2.41 seconds
 
 Here you can see a device with hostname `raspberrypi` has IP address `192.168.1.8`.
 
+### Getting IPv6 addresses by pinging from a second device (linux device or raspberry pi)
+
+find your own IP by `hostname -I`
```
lurch: Might be worth clarifying that this is the "own IP" of the "second device" ? :man_shrugging:

thomasschaeferm:
"See which IPv6 addresses respond to an SSH login attempt"

That is, what I am trying to say. There easier ways with fritz.box routers, but the manual should be more general.

remote-access/ip-address.md
thomasschaeferm: your are right, I try to follow the nmap example

remote-access/ip-address.md
```diff
@@ -79,6 +79,54 @@ Nmap done: 256 IP addresses (4 hosts up) scanned in 2.41 seconds
 
 Here you can see a device with hostname `raspberrypi` has IP address `192.168.1.8`.
 
+### Getting IPv6 addresses by pinging from a second device (linux device or raspberry pi)
+
+First find your own IP address, in other words the one of the computer you're using to find your Pi's IP address
+by `hostname -I`
+
+`fd00::ba27:ebff:feb6:f293 2001:db8:494:9d01:ba27:ebff:feb6:f293`
```
lurch: I've just noticed that there's two IP addresses here. :eyes: I've never used IPv6 myself, but how does a user know which IP address to use with the `ping` command below?

thomasschaeferm: I will explain it.

aallan:
In light of #1911 is this PR still relevant? If there is supporting third-party documentation outside of this repo, and there hasn't been much recent progress I'd be inclined to close it?

thomasschaeferm:
I think it is still relevant.

remote-access/ip-address.md
```diff
@@ -79,6 +79,68 @@ Nmap done: 256 IP addresses (4 hosts up) scanned in 2.41 seconds
...
+
+First find your own IP address(es), in other words the one of the computer you're using to find your Pi's IP address
+by `hostname -I`
+
+`fd00::ba27:ebff:feb6:f293 2001:db8:494:9d01:ba27:ebff:feb6:f293`
+
+The example shows two IP addresses. The first one is a so called unique local unicast address(`fc00::/7`). The second one is the global unicast address(`2000::/3`). It is also possible to see only one them depending on your network (router) configuration. Both addresses are valid for reaching the Raspberry Pi within your LAN. The address out of `2000::/3` is accessible world wide, provided your router's firewall is opened.
+
+
+Now use one of IPs from the first step for ping all local nodes:
```
JamesH65, line 92: "to ping" ?

remote-access/ip-address.md
```diff
@@ -79,6 +79,68 @@ Nmap done: 256 IP addresses (4 hosts up) scanned in 2.41 seconds
...
+```
+`-c 2` stands for sending two echo requests
+
+`-I` with the IP address, it sets the interface and the source address of the echo request,
+     it is necessary to choose the interface's IP address, 
+     `eth0` isn't sufficient - the answer would be the local link address(`fe80::/10`), we need the global or local unicast address
+     
+`ff02::1` is a well known multicast address for all nodes on the link, so it behaves like a local broadcast, usually it is defined in `/etc/hosts` so you can also use the name (`ip6-allnodes` or `ipv6-allnodes`) instead of the literal address
+
+some newer OS expect the interface ID here behind the multicast address
```
JamesH65, line 106: Capital S

remote-access/ip-address.md
```diff
@@ -79,6 +79,68 @@ Nmap done: 256 IP addresses (4 hosts up) scanned in 2.41 seconds
...
+64 bytes from fusion.fritz.box (2001:db8:494:9d01:1e6f:65ff:fec9:8746): icmp_seq=1 ttl=255 time=2.12 ms (DUP!)
+64 bytes from fritz.box (2001:db8:494:9d01:464e:6dff:fe72:8a08): icmp_seq=1 ttl=64 time=2.62 ms (DUP!)
+64 bytes from raspberrypi.fritz.box (2001:db8:494:9d01:ba27:ebff:feb6:f293): icmp_seq=2 ttl=64 time=0.480 ms
+
+--- ip6-allnodes ping statistics ---
+2 packets transmitted, 2 received, +5 duplicates, 0% packet loss, time 1001ms
+rtt min/avg/max/mdev = 0.480/1.283/2.623/0.735 ms
+
+```
+Now, you got the answers from all your nodes on the link (your (W)LAN), with a little bit luck - also with dns names
```
JamesH65, line 124: Rephrase, something like "This should result in replies from all the nodes on your (W)LAN link, with associated DNS names" ???

remote-access/ip-address.md
```diff
@@ -79,6 +79,68 @@ Nmap done: 256 IP addresses (4 hosts up) scanned in 2.41 seconds
...
+
+Exclude your own IP( here `2001:db8:494:9d01:ba27:ebff:feb6:f293` ), 
+check the others by trying to connect them via ssh
```
JamesH65, line 127: "then check"
"SSH"

remote-access/ip-address.md
thomasschaeferm: "to ping" you are right.

remote-access/ip-address.md
thomasschaeferm: It sounds better. Sorry again for my bad English.

remote-access/ip-address.md
thomasschaeferm: Your proposal is better.

remote-access/ip-address.md
thomasschaeferm: May be it is better to avoid the abbreviation of OS. (https://en.wikipedia.org/wiki/Operating_system ).
should be replaced by linux operating system or just systems

aallan:
Have your points been addressed by the subsequent changes @JamesH65? Ready to merge?

JamesH65:
LGTM

PR #1330, opened 2019-11-09
Removed "sh -c" so it follows the whole document pattern

evandropomatti:
Just to avoid confusion I suggest to remove the "sh -c", it seems not required.

### DIFF
--- a/configuration/wireless/access-point.md
+++ b/configuration/wireless/access-point.md
@@ -156,7 +156,7 @@ sudo iptables -t nat -A  POSTROUTING -o eth0 -j MASQUERADE
 Save the iptables rule.
 
 ```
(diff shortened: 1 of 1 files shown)


lurch:
It's being done that way because `/etc/` isn't writable by normal users.

```bash
sudo sh -c "iptables-save > /etc/iptables.ipv4.nat"
```
runs the `iptables-save` command as `root`, and also writes to the `/etc/iptables.ipv4.nat` file as `root`.

```bash
sudo iptables-save > /etc/iptables.ipv4.nat
```
would only run the `iptables-save` command as `root`, and try to write to the `/etc/iptables.ipv4.nat` file as a normal user, which wouldn't work.

JamesH65:
Closing this issue as questions answered/issue resolved.

PR #1333, opened 2019-11-14
Epoch1970: Access Point refactoring

JamesH65:
This is a merge from the original OP's branch, but still needs copy editing, so added as a new PR.


lurch:
Note to the copy-editor(s) - I made a bunch of comments (not all of which the OP replied to) on #1268 which you may or may not find useful.

epoch1970:
> Note to the copy-editor(s) - I made a bunch of comments (not all of which the OP replied to) on #1268 which you may or may not find useful.

I tried replying to most.

JamesH65:
@helenlynn It would be great if we could get this one through CE, it's been hanging around for a while now. Is this the one being dealt with offsite?

JamesH65:
@mwainwright Is the CE on this now complete? You can flag it up as needing a final tech check using the Labels option on the right bar, prior to being merged. (Normal process is Get PR, I tech check it, then flag as Ready For CE; once CE is complete, the editor flags as ready for tech check to ensure the CE hasn't broken anything, then it is merged)

configuration/wireless/access-point-bridged.md
```diff
@@ -0,0 +1,163 @@
...
+# Setting up a Raspberry Pi as a bridged wireless access point
+
+The Raspberry Pi can be used as a bridged wireless access point within an existing Ethernet network. This will extend the network to wireless computers and devices.
+
+If you wish to create a standalone wireless network, consider instead setting up a [routed access point.](./access-point-routed.md)
+
+```
+                                         +- RPi -------+
+                                     +---+ Bridge      |
+                                     |   | WiFi AP     +-)))
```
helenlynn: @JamesH65 review use of "WiFi" please

configuration/wireless/access-point-bridged.md
```diff
@@ -0,0 +1,163 @@
...
+                                     |   | 192.168.1.2 |         +- Laptop ----+
+                                     |   +-------------+     (((-+ WiFi STA    |
```
helenlynn: @JamesH65 review use of "WiFi"

configuration/wireless/access-point-routed.md
```diff
@@ -0,0 +1,193 @@
+# Setting up a Raspberry Pi as a routed wireless access point
+
+A Raspberry Pi within an Ethernet network can be used as a wireless access point, creating a secondary network. The resulting new wireless network is entirely managed by the Raspberry Pi.
+
+If you wish to extend an existing Ethernet network to wireless clients, consider instead setting up a [bridged access point.](./access-point-bridged.md)
+
+```
+                                         +- RPi -------+
+                                     +---+ 10.10.0.2   |          +- Laptop ----+
+                                     |   |     WiFi AP +-)))  (((-+ WiFi STA    |
```
helenlynn: @JamesH65 review use of "WiFi"

configuration/wireless/access-point-routed.md
```diff
@@ -0,0 +1,193 @@
...
+```
+sudo nano /etc/sysctl.d/routed-ap.conf
+```
+File contents:
+```
+# https://www.raspberrypi.org/documentation/configuration/wireless/access-point-routed.md
+# Enable IPv4 routing
+net.ipv4.ip_forward=1
+```
+Enabling routing will allow hosts from network `192.168.4.0/24` to reach the LAN and the main router towards the internet. In order to allow traffic between clients on this foreign wireless network and the internet without changing the configuration of the main router, the Raspberry Pi can substitute the IP address of wireless clients with its own IP address on the LAN using a "masquerade" firewall rule.
```
helenlynn: @JamesH65 Neither available copy-editor understands "foreign" here :) Please can you confirm that it's correct?

helenlynn:
@JamesH65 as discussed, @mwainwright's changes have been waiting on me; I was away by the time he finished CE. This is ready for a tech check now; please note comments tagging you, all very limited in scope.

A broader query relating to one of the changed files, [access-point-routed.md](https://github.com/raspberrypi/documentation/blob/aaf97e8c5c85e1aeb33bdd5e4119e6330a2e2f51/configuration/wireless/access-point-routed.md), is whether and how to implement the suggestion/discussion (in the conversation on the closed PR https://github.com/raspberrypi/documentation/pull/1268) about adding something about rfkill, and similarly about possibly changing the diagram. That needs a technical person to decide and suggest what copy to add and where.

configuration/wireless/access-point-bridged.md
```diff
@@ -0,0 +1,163 @@
...
+                 | Firewall    |     |   +- PC#2 ------+         +-------------+
+(Internet)---WAN-+ DHCP server +-LAN-+---+ 192.168.1.3 |
+                 | 192.168.1.1 |     |   +-------------+
+                 +-------------+     |
+                                     |   +- PC#1 ------+
+                                     +---+ 192.168.1.4 |
+                                         +-------------+
+```
+
+A bridged wireless access point can be created using the inbuilt wireless features of the Raspberry Pi 4, Raspberry Pi 3 or Raspberry Pi Zero W, or by using a suitable USB wireless dongle that supports access point mode.
```
andrum99: Product names are usually expanded out to the actual product, i.e. Raspberry Pi 4B, Raspberry Pi 3B and Raspberry Pi 3B+. Also, Raspberry Pi Zero WH needs to be added.

configuration/wireless/access-point-routed.md
andrum99: That sounds like it should be on its own page - I would expect it to be called something like "Advanced - routing and IP masquerading. Normally wireless access points just bridge onto the local ethernet network.

andrum99: That needs reworded. "Foreign" here assumes that "native" is the wired LAN segment, and the user requires routing to be set up in order for things to work, since they have some setup that won't route traffic unless it comes from a specific IP address range.

JamesH65: I cannot comment one way or the other, as this is out of my level of knowledge.

JamesH65: @epoch1970 Have you got time to clarify this?

configuration/wireless/access-point-bridged.md
andrum99: That actually needs to be expanded to the actual product names, since the Raspberry Pi 3A+ does not have a wireless interface.

configuration/wireless/access-point-routed.md
epoch1970: Routers route or drop traffic whether they have a relevant route for it, or not. Firewalls filter traffic that isn't dropped.
Between a source and a destination host (and back again), each router along the path makes a _fatal_ decision (pass/drop) based on the _local information_ it has.

Say a wifi client 192.168.4.99 queries Google's DNS 8.8.8.8, and receives a response; The path includes 2 local routers:  Pi > ISP router > (internet routers) > ISP router > Pi. Any router can break the chain.
Internet routers automatically exchange routes, they know their stuff. Local routers require local administration...
 - The Pi router knows all networks involved, it will let traffic pass. 
 - The ISP router does not know about network 192.168.4.0/24, it _will_ drop traffic. The simplest and most portable solution is to hide the hosts in this network under the identity of the Pi, using the well-known LAN address 10.10.0.2.

Tutorials all choose this double-NAT solution. None (I think) try to explain the reason for  masquerading. So let me try again:

Enabling routing will allow hosts from network `192.168.4.0/24` to reach the LAN and the main router towards the internet. 
In order to ensure that the main router will allow traffic between wireless clients and the internet, the Raspberry Pi can substitute their IP address with its own IP address on the LAN using a "masquerade" firewall rule.

Hope that's more clear.

configuration/wireless/access-point-routed.md
```diff
@@ -32,7 +32,7 @@ This documentation was tested on a Raspberry Pi 3B running a fresh installation
 * Ensure the Raspbian OS on your Raspberry Pi is [up to date](../../raspbian/updating.md) and reboot if packages were installed in the process.
 * Take note of the IP configuration of the Ethernet network the Raspberry Pi is connected to: 
     * In this document, we assume IP network `10.10.0.0/24` is configured on the Ethernet LAN, and the Raspberry Pi is to manage IP network `192.168.4.0/24` for wireless clients.
```
lurch: "is to manage" -> "is going to manage" ?

configuration/wireless/access-point-routed.md
lurch: @epoch1970 I'm not sure that an explanation of masquerading / routing is what was being asked for?

To answer @helenlynn 's original question, I think "foreign wireless network" here is referring to the wireless network that's being managed exclusively by the RPi, rather than the wireless network being managed by the user's "main router". I wonder if "new wireless network" would be a suitable replacement? :man_shrugging:

configuration/wireless/access-point-bridged.md
```diff
@@ -0,0 +1,163 @@
+
+# Setting up a Raspberry Pi as a bridged wireless access point
+
+The Raspberry Pi can be used as a bridged wireless access point within an existing Ethernet network. This will extend the network to wireless computers and devices.
+
+If you wish to create a standalone wireless network, consider instead setting up a [routed access point.](./access-point-routed.md)
```
lurch: Only a minor nitpick, but maybe the full-stop at the end of this sentence should be moved to after the `)`, rather than being before the `]` ?

configuration/wireless/access-point-bridged.md
lurch: IIRC in all the documentation I've seen, "Raspberry Pi Zero W" is assumed to include "Raspberry Pi Zero WH"?

Also, is Raspberry Pi Zero W even a sensible suggestion for this, since although it has a built-in wireless interface, it doesn't have a built-in ethernet interface?

configuration/wireless/access-point-routed.md
epoch1970: Given the question and previous remarks, I think the long reply might be on point.
As I said, in the previous tutorial, you can read "Add a masquerade for outbound traffic on eth0:". You can leave it at that if you prefer. Proven and native.

configuration/wireless/access-point-bridged.md
lurch: Looks like "WiFi STA" is an abbreviation for "WiFi Station", but I'm not sure how commonly-used that is?  "wireless client" would probably work equally as well?
(And similarly, "WiFi AP" would probably work as "wireless access point")

configuration/wireless/access-point-bridged.md
```diff
@@ -0,0 +1,163 @@
...
+Add the following line near the beginning of the file (above the first `interface xxx` line, if any):
+```
+denyinterfaces wlan0 eth0
+```
+Interface `br0` will be configured in accordance with the defaults via DHCP; no specific entry is necessary. Save the file to complete the IP configuration of the machine.
+
+<a name="ap-config"></a>
+## Configure the access point software
+
+Create the `hostapd` configuration file, located at `/etc/hostapd/hostapd.conf`, to add the various parameters for your wireless network.
```
lurch: Perhaps change "your wireless network" to "your new wireless network" ?

configuration/wireless/access-point-bridged.md
```diff
@@ -0,0 +1,163 @@
...
+wpa_pairwise=TKIP
+rsn_pairwise=CCMP
+```
+Note the lines `interface=wlan0` and `bridge=br0`: these direct `hostapd` to add the `wlan0` interface as a bridge member to `br0` when the access point starts, completing the bridge between Ethernet and wireless.
+
+To use the 5 GHz band, you can change the operations mode from `hw_mode=g` to `hw_mode=a`. Possible values for `hw_mode` are:
+ - a = IEEE 802.11a (5 GHz)
+ - b = IEEE 802.11b (2.4 GHz)
+ - g = IEEE 802.11g (2.4 GHz)
+ - ad = IEEE 802.11ad (60 GHz)
```
lurch: I'm sure I suggested elsewhere that there doesn't seem much sense in listing `ad` mode, since the Pi doesn't support 60 GHz wifi?

configuration/wireless/access-point-bridged.md
```diff
@@ -0,0 +1,163 @@
...
+## Run your new wireless access point
+
+Now restart your Raspberry Pi and verify that the wireless access point becomes automatically available.
+
+```
+sudo systemctl reboot
+```
+Once your Raspberry Pi has restarted, search for wireless networks with your wireless client. The network SSID you specified in file `/etc/hostapd/hostapd.conf` should now be present, and it should be accessible with the specified password.
+
+If your wireless client has access to the local network and the internet, congratulations on your new access point!
```
lurch: "congratulations on your new access point!" -> "congratulations on setting up your new access point!" ?

configuration/wireless/access-point-routed.md
```diff
@@ -0,0 +1,193 @@
...
+
+<a name="routing"></a>
+## Set up the network router
+
+The Raspberry Pi will run and manage a standalone wireless network. It will also route between the wireless and Ethernet networks, providing internet access to wireless clients. If you prefer, you can choose to skip the routing by skipping the section "Enable routing and IP masquerading" below, and run the wireless network in complete isolation. 
+
+### Define the wireless interface IP configuration
+
+In this document, we assume IP network `10.10.0.0/24` is configured on an Ethernet LAN, and the Raspberry Pi will manage IP network `192.168.4.0/24` for wireless clients.
+**Note:** Please select another IP network for wireless, e.g. `192.168.10.0/24`, if IP network `192.168.4.0/24` is already in use by your Ethernet LAN.
```
lurch: This sentence seems to be duplicated in the "Before you start" section?

configuration/wireless/access-point-routed.md
```diff
@@ -0,0 +1,193 @@
...
+```
+Now save the current firewall rules for IPv4 (including the rule above) and IPv6 to be loaded at boot by the `netfilter-persistent` service:
+```
+sudo netfilter-persistent save
+```
+Filtering rules are saved to the directory `/etc/iptables/`. If in the future you change the configuration of your firewall, make sure to save the configuration before rebooting.
+
+### Configure the DHCP and DNS services for the wireless network
+
+The DHCP and DNS services are provided by `dnsmasq`. The default configuration file serves as a template for all possible configuration options, when we only need a few. It is easier to start from an empty file.
```
lurch: when -> whereas?

configuration/wireless/access-point-routed.md
```diff
@@ -0,0 +1,193 @@
...
+dhcp-range=192.168.4.2,192.168.4.20,255.255.255.0,24h
+                # Pool of IP addresses served via DHCP
+domain=wlan     # Local wireless DNS domain
+address=/gw.wlan/192.168.4.1
+                # Alias for this router
+```
+
+The Raspberry Pi will deliver IP addresses between `192.168.4.2` and `192.168.4.20`, with a lease time of 24 hours, to wireless DHCP clients. You should be able to reach the Raspberry Pi under the name `gw.wlan` from wireless clients.
+
+There are many more options for `dnsmasq`; see the default configuration file or the [online documentation](http://www.thekelleys.org.uk/dnsmasq/doc.html) for details.
```
lurch: "... default configuration file or the ..." -> "... default configuration file (`/etc/dnsmasq.conf.orig`) or the ..." for clarity?

lurch:
Many of the comments I've just made about `access-point-bridged.md` also apply to `access-point-routed.md`

configuration/wireless/access-point-bridged.md
lurch: Also, does this need to mention which models of Pi support 2.4 GHz WiFi and which support 5 GHz WiFi?

andrum99: > Also, does this need to mention which models of Pi support 2.4 GHz WiFi and which support 5 GHz WiFi?

That's probably a good idea, or a link to a page that does.

configuration/wireless/access-point-bridged.md
andrum99: As Helen Lyn has suggested - 'wifi' needs to be change to 'wireless' everywhere it occurs.

JamesH65: I think in the case of ASCII art, WLAN would be sufficient and the same number of letters.

configuration/wireless/access-point-routed.md
JamesH65: Removed.

configuration/wireless/access-point-routed.md
JamesH65: I've made all other changes requested but am not sure what to do here, any suggestions?

helenlynn: @lurch 's suggestion of "new wireless network" is possibly suitable if it is technically correct, which I suspect that it is, but I'm not in a position to be sure.

epoch1970: No that suggestion was not correct. The fact the new network is wireless is irrelevant. What counts is that the new IP network (192.168.4.0) is different from the one managed by the main router (10.0.0.0). See my answer above for details.

configuration/wireless/access-point-bridged.md
lurch: > since the Raspberry Pi 3A+ does not have a wireless interface.

Err, yes it does :wink: https://www.raspberrypi.org/products/raspberry-pi-3-model-a-plus/

lurch:
On the off-chance that there are any external resources (e.g. user-written tutorials?) pointing to `access-point.md`, should we re-instate it with links to the two new pages?

configuration/wireless/access-point-bridged.md
andrum99: > > since the Raspberry Pi 3A+ does not have a wireless interface.
> 
> Err, yes it does 😉 https://www.raspberrypi.org/products/raspberry-pi-3-model-a-plus/

You're quite right. I must have been thinking of the original A - sorry!

configuration/wireless/access-point-routed.md
andrum99: I think the solution is to include some discussion about what "routed" means in this context, at the start of the page. You then just need to call the subnet something like "the routed subnet" consistently throughout the page (rather than foreign).

Also, a minor nit: it is common practice when using /24 networks to put routers at the 'top' of the subnet, starting at .254 downwards, leaving the low-numbered addresses (.1, .2 etc) for the "normal" hosts.

lurch: > it is common practice when using /24 networks to put routers at the 'top' of the subnet, starting at .254 downwards, leaving the low-numbered addresses (.1, .2 etc) for the "normal" hosts.

Oh? Most of the home routers I've seen default to something like 192.168.0.1 or 192.168.1.1

andrum99: > > it is common practice when using /24 networks to put routers at the 'top' of the subnet, starting at .254 downwards, leaving the low-numbered addresses (.1, .2 etc) for the "normal" hosts.
> 
> Oh? Most of the home routers I've seen default to something like 192.168.0.1 or 192.168.1.1

While it is certainly true that most home routers give themselves addresses ending in .1, that's not really how it's meant to be done. BT manage to get it right though. Like I said, it's a minor nit, which I am picking 😋

epoch1970:
#1423 needs to be reopened. Or otherwise note the howto is confirmed broken. Something changed w/ systemd or dhcpcd and the instructions no longer work.
This is a blocking issue.
I am not very keen on wasting my life making systemd-networkd work personally.

configuration/wireless/access-point-bridged.md
```diff
@@ -0,0 +1,162 @@
...
+```
+sudo nano /etc/dhcpcd.conf
+```
+
+Add the following line near the beginning of the file (above the first `interface xxx` line, if any):
+```
+denyinterfaces wlan0 eth0
+```
+Interface `br0` will be configured in accordance with the defaults via DHCP; no specific entry is necessary. Save the file to complete the IP configuration of the machine.
+
```
epoch1970: Go to the end of the file and add the following:

```
interface br0

```
With this line, interface `br0` will be configured in accordance with the defaults via DHCP. Save the file to complete the IP configuration of the machine.

epoch1970: What's above is a proposed fix, as per #1423.
The routed ap howto already asked users to add something at the beginning of file dhcpcd.conf, and something else at the end of that same file. Now this seems necessary for the bridged ap as well.

fikrikarim:
What a nice PR.

I found this PR after trying to set up an AP as a bridge on 3B+. I follow the guide on this PR, but somehow I got incorrect password everytime I tried to connect via Wi-Fi using my laptop or handphone. The problem is exactly similar [to this one](https://raspberrypi.stackexchange.com/questions/108820/using-a-pi-as-a-wireless-access-point-with-bridge-incorrect-password?newreg=1f25b7fb1b40415aa3f4145c5ba2c7e1).

In the end I tried reflashing my PI, and follow the guide that I got from one of the [answer there](https://raspberrypi.stackexchange.com/questions/88214/setting-up-a-raspberry-pi-as-an-access-point-the-easy-way/88234#88234). Using that answer now everything is working so far. 

I'm not sure what my problem was. Just want to share my experience here. Anyway this PR is needed because I think the one on the master branch is pretty old and need update.

JamesH65:
This has been hanging around for quite a while now - I am inclined to merge it (once rebased) since it's better than the current pages, even if not quite perfect.

JamesH65:
@epoch1970. Really sorry  about the delay here. If you can rebase it to current tree then I'll give a last once over and merge it.

epoch1970:
No problem. However, you're speaking to a dummy: I don't understand what rebasing entails. What do you want me to do?

JamesH65:
There are apparently conflicts which prevent it from merging, so something has changed on our repothat means you PR won't apply. Rebasing really just means fixing it up so it will reapply without conflicts. ie rebasing you change on top on the new top of tree docs. On the command line I would "git pull --rebase upstream master" or similar then fix the conflicts, but not sure how you do that in github...Might be easier if I do it here. Will try!

JamesH65:
I think that has done it....

epoch1970:
Great!

PR #1338, opened 2019-11-17
Update samba.md to allow wirtes

neilh10:
As written it didn't work for me on the latest rasberrian/RaspberryPi3.   Without the user share chmod it forbids access, and then with the specified samba uses nobody

### DIFF
--- a/remote-access/samba.md
+++ b/remote-access/samba.md
@@ -80,8 +80,13 @@ Firstly, create a folder to share. This example creates a folder called `shared`
 ```bash
 cd ~
 mkdir shared
+chmod 0740 share
 ```
+Now we need to tell Samba there is a 'pi'user when accessing that folder. When asked enter the passwd of pi - ether default or hopefully your own security password
 
+```bash
+sudo smbpasswd -a pi
+```
 Now we need to tell Samba to share this folder, using the Samba configuration file.
 
 ```bash


remote-access/samba.md
```diff
@@ -80,8 +80,13 @@ Firstly, create a folder to share. This example creates a folder called `shared`
 ```bash
 cd ~
 mkdir shared
+chmod 0740 share
```
lurch: Was this supposed to be `shared` rather than `share` ?

remote-access/samba.md
```diff
@@ -80,8 +80,13 @@ Firstly, create a folder to share. This example creates a folder called `shared`
...
 ```
+Now we need to tell Samba there is a 'pi'user when accessing that folder. When asked enter the passwd of pi - ether default or hopefully your own security password
```
lurch: ether -> either ?

JamesH65: Rewritten that line, but still not entirely happy with it.

lurch: AFAICT the password used by [smbpasswd](https://www.samba.org/samba/docs/current/man-html/smbpasswd.8.html) is actually different to the password used to login to Raspbian. I _think_ the OP was trying to say something like "you can use the same password for both your regular user account and your samba account, but it's more secure if you choose a separate password for samba".

Of course I suspect that in practice 90% of users _will_ use the same password for both :wink:

JamesH65: Feel free to change as you see fit. Although I wrote the original page, I am not particular familiar with this stuff.

neilh10:
Thanks for looking at it. I've just noticed this conversation.

andrum99:
This PR makes the shared directory read/write by all users, including guest. I don't think this was the intention of the instructions as they were originally written. I read it as showing how to configure the share so that only the `pi` user can connect to it, however it was missing the step where you add the Samba password using `smbpasswd`. You should not need the `chmod` part if you do the `smbpasswd` correctly, then use the correct username and password to connect to the Samba share.

aallan:

PR #1341, opened 2019-11-23
Improve description of vcgencmd get_throttled output

andrum99:
Recently a forum user asked for help decoding the values returned by vcgencmd get_throttled. (https://www.raspberrypi.org/forums/viewtopic.php?f=63&t=257569). This flagged up that the explanation of how to do this in the docs could be improved.

Explain how to read off the individual bit values from the hex value returned by the command.

### DIFF
--- a/raspbian/applications/vcgencmd.md
+++ b/raspbian/applications/vcgencmd.md
@@ -45,7 +45,7 @@ Returns the throttled state of the system. This is a bit pattern - a bit being s
 
 A value of zero indicates that none of the above conditions is true.
 
-To find the if one of these bits has been set, convert the value returned to binary, then number each bit along the top. You can then see which bits are set. For example:
+To find if one of these bits has been set, convert the value returned to binary, then number each bit along the top. You can then see which bits are set. For example:
 
 ``0x50000 = 0101 0000 0000 0000 0000``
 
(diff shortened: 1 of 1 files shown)


andrum99:
Ready for review.

lurch:
@JamesH65 Maybe it's also worth clarifying when the "has occurred" bits get cleared to 0? Is it every time the Pi reboots / powers-off ?

JamesH65:
@6by9 quickly looked at the code, and it appears that it will also be reset if you add a an extra parameter to get_throtled, so `vcgencmd get_throttled 1` will also reset the "has occurred" bits.

raspbian/applications/vcgencmd.md
```diff
@@ -15,7 +15,7 @@ Some of the more useful  commands are described below.
 
 #### vcos
 
-The `vcos` cammand has a number of sub commands
+The `vcos` cammand has a number of sub commands.
```
JamesH65: Cammand?  yes, I know that's what the original said as well!!!

raspbian/applications/vcgencmd.md
```diff
@@ -68,7 +81,7 @@ This returns the current frequency of the specified clock. The options are:
 | hdmi | HDMI |
 | dpi | Display Peripheral Interface |
 
-e.g. `vcgencmd measure_clock arm`
+e.g. `vcgencmd measure_clock arm`.
```
JamesH65: Not sure we should have a period here - people might think they need to type it in.

raspbian/applications/vcgencmd.md
andrum99: That's the difference between one of my PRs and one of @lurch's - my eyes see only some of the mistakes 😁

PR #1342, opened 2019-11-25
James h65 patch overlay renames

JamesH65:


lurch:
@pelwell It says "ttyS0 on a Pi 3, ttyAMA0 otherwise" - doesn't this also now apply to other Pi models too? Is it worth correcting this text?

PR #1344, opened 2019-11-26
Update README.md

eitanchud:
Added UpSwift.io as free remote shell access over the internet solution for RaspberryPi boards.

### DIFF
--- a/remote-access/access-over-Internet/README.md
+++ b/remote-access/access-over-Internet/README.md
@@ -10,6 +10,7 @@ One method is to set up port forwarding on your router. To do this, you must cha
 
 Rather than using port forwarding, there are a number of alternative third-party services available. These provide varying levels of functionality - see their websites for more details.
 
+- [UpSwift.io](https://upswift.io)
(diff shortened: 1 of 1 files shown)


JamesH65:
Looks like a pay for service to me (for over three devices), couldn't even get pricing without signing up, so I don't think  we will merge this.

JamesH65:
Actually, given the other options are also pay under certain circumstances, if you can make you price list more obvious this could probably go through - because I needed to register before seeing pricing, it make it less clear to our users which to chose.

JamesH65:
Also, please put your link at the end of the list rather than the beginning. No need to 'jump the queue', those others were here first.

eitanchud:
Hi James, as you mentioned, the pricing isn't relevant for that case, because we provide our platform completely free and full-featured for Makers and for Prototyping. 

Regarding the link - You are completely right, I will move it to the end of the list, there was no intention to 'jump the queue'.

JamesH65:
Merged from #1345

PR #1350, opened 2019-11-29
User security

thecolorjay:
Disabling the `pi` user and disabling autologin of the `pi` user.

### DIFF
--- a/configuration/security.md
+++ b/configuration/security.md
@@ -92,7 +92,7 @@ and change the `pi` entry (or whichever usernames have superuser rights) to:
 autologin-user=alice
 ```
 
-This can also be done by using raspi-config, but you'll have to install it.
+This can also be done by using raspi-config which is recommended as it will disable autologin for the console as well.
 
 ## Make `sudo` require a password
 
--- a/configuration/security.md
+++ b/configuration/security.md
@@ -60,7 +60,7 @@ Once you have confirmed that the new account is working, you can delete the `pi`
 sudo pkill -u pi
 ```
 
-Please note that with the current Raspbian distribution, there are some aspects that require the `pi` user to be present. If you are unsure whether you will be affected by this, then leave the `pi` user in place. Work is being done to reduce the dependency on the `pi` user.
+Please note that with the current Raspbian distribution, there are some aspects that require the `pi` user to be present. If you are unsure whether you will be affected by this, then either leave the `pi` user in place or disable it. Work is being done to reduce the dependency on the `pi` user.
 
 To delete the `pi` user, type the following:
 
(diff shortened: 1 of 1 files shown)


JamesH65:
@XECDesign Does this look sane to you?

XECDesign:
Seems coherent, yes. I'd recommend using raspi-config instead since that will also handle console autologin. Otherwise this needs an extra step added to adjust the getty override as well.

JamesH65:
@thecolorjay As @XECDesign has suggested, raspi-config is a better option for turning off auto login, so the instructions should be amended to use that instead. Otherwise, all good.

thecolorjay:
> @thecolorjay As @XECDesign has suggested, raspi-config is a better option for turning off auto login, so the instructions should be amended to use that instead. Otherwise, all good.

That's actually not installed by default, they'd have to install it first. But, I'll point that out.

XECDesign:
Where is `raspi-config` not installed by default?

thecolorjay:
> Where is `raspi-config` not installed by default?

On Raspbian, it's not preinstalled.

JamesH65:
> > Where is `raspi-config` not installed by default?
> 
> On Raspbian, it's not preinstalled.

Yes it is, and has been almost since the very first Pi.

thecolorjay:
> > > Where is `raspi-config` not installed by default?
> > 
> > 
> > On Raspbian, it's not preinstalled.
> 
> Yes it is, and has been almost since the very first Pi.

Well I had to install it because it didn't exist, and I'm using the latest version of Raspbian. Try it yourself.

XECDesign:
The core issue here is that the instructions are incomplete. Following them will disable lightdm autologin, but will do nothing about console autologin. Whereas using raspi-config instead will do both.

XECDesign:
No need to try it ourselves, it's one of the first things people run and if it wasn't there, we'd be the first to notice.

thecolorjay:
> The core issue here is that the instructions are incomplete. Following them will disable lightdm autologin, but will do nothing about console autologin. Whereas using raspi-config instead will do both.

Okay I'll just get rid of that all together

thecolorjay:
> The core issue here is that the instructions are incomplete. Following them will disable lightdm autologin, but will do nothing about console autologin. Whereas using raspi-config instead will do both.

Okay, it's done. That good?

thecolorjay:
I didn't mean to close the pull request, that was a mistake.

thecolorjay:
> No need to try it ourselves, it's one of the first things people run and if it wasn't there, we'd be the first to notice.

That may be, but I'm also a user of Raspbian and it's not there for me. As I pointed out, I had to install it. Weird, but that's true.

thecolorjay:
You know, this is a fucking waste of my time. Fuck your upper management!

JamesH65:
I'll pass on your message. Just a FYI though, I spend a lot of time working on this documentation. It's not a waste of time.

JamesH65:
Despite over the top comments from OP, this is worth keeping. But I don't seem to be able to reopen it. Will just copy over when I get the chance.

thecolorjay:
> Despite over the top comments from OP, this is worth keeping. But I don't seem to be able to reopen it. Will just copy over when I get the chance.

I've already deleted the pull request, I don't see the point in trying to contribute anything when the smallest change is rejected.

XECDesign:
It wasn't rejected.

JamesH65:
This is documentation that is read by millions of people. It has a process to go through before it goes live - its checked for technical accuracy, checked for completeness, checked its useful enough to go in, then it's checked by copy editors to ensure its grammatically correct and adheres to our publishing standards. Then it is technically checked again to make sure copy editors haven't broken anything. Then its merged. We were still going through that process, at no point was it rejected.

PR #1352, opened 2019-12-02
Update Mac image installation instructions

bshaw:
- Provides additional commands required when using the command line and provides some output examples.
- Cleanup some markdown syntax


helenlynn:
@JamesH65 the table in lines 15-37 appears to me to be misaligned, unless the current alignment holds some meaning with which I am not familiar. Has it? If not, I will line up columns under headings.

JamesH65:
I think the format has gone a bit haywire. Not familiar with Mac's, but I thin lining up is probably the way to go.

helenlynn:
Now aligned and ready for tech check etc.

JamesH65:
rebased and merged.

PR #1354, opened 2019-12-03
building.md: mention make

proger:
You also need make to run the build (testing on a fresh ubuntu vm).


JamesH65:
Odd that it is not installed by default. But if there are no objections, happy to merge this.

XECDesign:
Just confirming that `make` is missing from Raspbian lite, and is not guaranteed to be installed on other systems when cross-compiling.

JamesH65:
@proger Thanks!

Issue #1018, opened 2018-10-30
Failed to start hostapd.service: Unit hostapd.service is masked.


ijash:
`sudo systemctl start hostapd `returned `Failed to start hostapd.service: Unit hostapd.service is masked.`

JamesH65:
I have no idea why you have posted this issue. Do you have a specific part of the documentation that is showing this error?

ijash:
> I have no idea why you have posted this issue. Do you have a specific part of the documentation that is showing this error?
sorry for not pointing the documentation.

i have followed https://www.raspberrypi.org/documentation/configuration/wireless/access-point.md
and get those error during `sudo systemctl start`.

i am using RPi3 B

ijash:
i'll close this topic, as i long forgot the steps to reproduce.

beta-tester:
i have the same issue.

today, i took the actual rasbian image [2018-11-13-raspbian-stretch.zip](https://downloads.raspberrypi.org/raspbian/images/raspbian-2018-11-15/2018-11-13-raspbian-stretch.zip)
and flashed it to an SD card for my RPi3B+.

i enabled SSH by adding a `ssh` file to boot partition befor very first boot.

at reboot (after automatic resizing/growing filesystem pass) i could shortly see a message with something like that the WiFi was disables because of unsetted country code or something similar.

from SSH i ran `sudo raspi-config` to setup **I4: WiFi country code** to DE for Germany.

after `sudo apt update && sudo apt upgrade -y` i installed packages `sudo apt install -y dnsmasq hostapd` to be able to setup Wifi AccessPoint i tried to setup my RPi3B+ folowed the steps for [Access-Point (NAT)](https://www.raspberrypi.org/documentation/configuration/wireless/access-point.md)

but when i try to start/enable hostapd with `sudo systemctl start hostapd` i get an error message
`Failed to start hostapd.service: Unit hostapd.service is masked`.

how can i enable WiFi again? how can i unmask hostapd?

XECDesign:
@beta-tester try unmasking, enabling and starting the service:
https://github.com/raspberrypi/documentation/pull/1097/files#diff-bab12433eaae7aea98b21a2978c8ba52R223

beta-tester:
yes, thank you this is working.
```
sudo systemctl unmask hostapd
sudo systemctl enable hostapd
sudo systemctl start hostapd
```

scudrunner:
[...]

PR #1359, opened 2019-12-11
About (re)starting dnsmasq service

ftinetti:
The line
sudo systemctl reload dnsmasq
generates an error, because the dnsmasq is not running (at least it happens in the last raspbian: Raspbian Buster Lite, Minimal image based on Debian Buster, Version: September 2019)

### DIFF
--- a/configuration/wireless/access-point.md
+++ b/configuration/wireless/access-point.md
@@ -71,7 +71,7 @@ So for `wlan0`, we are going to provide IP addresses between 192.168.4.2 and 192
 
 There are many more options for dnsmasq; see the [dnsmasq documentation](http://www.thekelleys.org.uk/dnsmasq/doc.html) for more details.
 
-Start `dnsmasq` (it was stopped), it will use the updated configuration:
+Start `dnsmasq` (it was stopped), it will now use the updated configuration:
 ```
 sudo systemctl start dnsmasq
 ```
--- a/configuration/wireless/access-point.md
+++ b/configuration/wireless/access-point.md
@@ -71,9 +71,9 @@ So for `wlan0`, we are going to provide IP addresses between 192.168.4.2 and 192
 
 There are many more options for dnsmasq; see the [dnsmasq documentation](http://www.thekelleys.org.uk/dnsmasq/doc.html) for more details.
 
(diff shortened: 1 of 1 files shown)


configuration/wireless/access-point.md
```diff
@@ -71,9 +71,9 @@ So for `wlan0`, we are going to provide IP addresses between 192.168.4.2 and 192
 
 There are many more options for dnsmasq; see the [dnsmasq documentation](http://www.thekelleys.org.uk/dnsmasq/doc.html) for more details.
 
-Reload `dnsmasq` to use the updated configuration:
+Start `dnsmasq` (it was stopped), it will use the updated configuration:
```
andrum99: It is enough to say "Start dnsmasq:" - I don't think you need the rest - I would keep it simple.

lurch:
Duplicate of #1018 ? (which I think there's already a pending PR to fix it?)

ftinetti:
> 
> 
> Duplicate of #1018 ? (which I think there's already a pending PR to fix it?)

I'm quite sure it is.

I think the original problem in that thread is what I've reported as
"sudo systemctl reload dnsmasq
generates an error, because the dnsmasq is not running "

However, I do not fully understand all the discussion later in that thread.

Besides, the problem is solved (at least in a clean current raspbian, as I've reported
"raspbian: Raspbian Buster Lite, Minimal image based on Debian Buster, Version: September 2019"
replacing the original line

sudo systemctl reload dnsmasq

by the one I've used and propose to include in the documentation 

sudo systemctl start dnsmasq

BR,

Fernando.

JamesH65:
Is this PR still relevant given changes in this area that have been merged? (I think)

ftinetti:
Idk, I just reported the problem I found as well as the fix I tried and worked out

andrum99:
This change is still relevant, and is actually the correct fix. At the moment the page tells you to 'stop' the dnsmasq service, then the next instruction to do with that service is to 'reload' it. You can only 'reload' a service that is running, so the fix in this PR is correct.

JamesH65:
Cool, thanks, will sort it tomorrow.

PR #1360, opened 2019-12-12
Update msd.md

Paul-Ver:
Updated the file to properly include 3B+ and 4B added background info from other sources on raspberrypi.org
I'm specifically curious if the part on Raspberry Pi 3B+, the info included in brackets "(" ")", which are test notes are relevant for this page. But my feeling is that they are important for reproducibility.

- Made it more clear that Pi 3B+ supports Mass Storage boot out of the box (Tested myself)
- Changed the text on setting up into a step-by-step guide with comments on how it was tested in my case.
- Split the article per version of Raspberry Pi, to keep it more readable. (In the original article 3B and 3B+ weren't as well separated)
- Added a reference to the 'Special bootcode.bin boot mode', which can be an interesting alternative for Pi 1/2 users.
- Added known issues.
- Added list of known working devices
- Added extra info; reference to Blog post of developer during Beta of USB Mass storage boot.

### DIFF
--- a/hardware/raspberrypi/bootmodes/msd.md
+++ b/hardware/raspberrypi/bootmodes/msd.md
@@ -47,7 +47,7 @@ You can now boot from an USB Mass Storage device in the same way as booting from
 
 The Raspberry Pi 3B+ and CM3+ support USB Mass Storage boot out of the box. The settings specific to the previous versions of Raspberry Pi do not have to be executed.
 
-This is verified using the following steps, this is the same (procedure)[../../../installation/installing-images/] as for SD cards.
+This is verified using the following steps, this is the same [procedure](../../../installation/installing-images/) as for SD cards.
 
 On Windows you may have to initialise a (clean) harddrive before it's visible and has a drive letter (which may be required to write the image).
 
--- a/hardware/raspberrypi/bootmodes/msd.md
+++ b/hardware/raspberrypi/bootmodes/msd.md
@@ -69,4 +69,4 @@ The Raspberry Pi 4's bootcode is stored in [EEPROM](../booteeprom.md) and can be
 
 - The default timeout for checking bootable USB devices is 2 seconds. Some flash drives and rotational harddrives power up too slowly. It’s possible to extend this timeout to five seconds (add a new file `timeout` to the SD card), but there are devices that fail to respond within this period as well.
 - Some flash drives have a very specific protocol requirement that is not handled by the bootcode and may thus be incompatible.
-- 3.5" HDD's commonly require 12V as well as 5V and may draw too much current for the Pi's USB connection. Use an externally powered 3.5" HDD in this case.
+- Lack of power can be an issue, as some devices draw too much current from the Raspberry Pi's USB connection and fail to start, or prevent the Raspberry Pi from booting. An external power supply for the device will be required in these cases.
--- a/hardware/raspberrypi/bootmodes/msd.md
+++ b/hardware/raspberrypi/bootmodes/msd.md
@@ -4,17 +4,17 @@
 
 This tutorial explains how to boot your Raspberry Pi from a USB mass storage device such as a flash drive or a USB hard disk. Note that this feature does not work with all USB mass storage devices.
 
-See the [documentation](README.md) for the boot sequence and alternative boot modes (Network, USB device, GPIO or SD boot).
+See the [bootmodes documentation](README.md) for the boot sequence and alternative boot modes (Network, USB device, GPIO or SD boot).
 
 Note that "USB Mass Storage Boot" is different from "USB Device Boot Mode". The [device boot mode](device.md) allows a Raspberry Pi connected to a computer to boot using files on that computer.
 
-[Pi 2B v1.2 up to 3](#raspberry-pi-2b-v12-3a-3b-cm3)
+[Pi 2B v1.2 up to 3](#pi2to3)
 
-[Pi 3B+ CM3+](#raspberry-pi-3b-cm3)
+[Pi 3B+ CM3+](#pi3plus)
 
-[Unsupported devices](#unsupported-devices)
+[Unsupported devices](#unsupported)
 
-<a name="Raspberry Pi 2B v1.2, 3A+, 3B, CM3"></a>
+<a name="pi2to3"></a>
 ## Raspberry Pi 2B v1.2, 3A+, 3B, CM3
 
 On the Raspberry Pi 2B v1.2, 3A+, 3B, CM3, first USB [host boot mode](host.md) should be enabled. This is to allow Mass Storage Boot / Network boot (Network boot not supported on 3A+).
@@ -50,33 +50,33 @@ If you wish, you can remove the `program_usb_boot_mode` line from config.txt, so
 
 You can now boot from an USB Mass Storage device in the same way as booting from SD, see the following paragraph for Raspberry Pi 3B+, CM3+.
 
-<a name="Raspberry Pi 3B+, CM3+"></a>
+<a name="pi3plus"></a>
 ## Raspberry Pi 3B+, CM3+
 
 The Raspberry Pi 3B+ and CM3+ support USB Mass Storage boot out of the box. The settings specific to the previous versions of Raspberry Pi do not have to be executed.
 
-This is the same [procedure](../../../installation/installing-images/installation/installing-images/) as for SD cards.
-
-On Windows you may have to initialise a (clean) harddrive before it's visible and has a drive letter (which may be required to write the image).
+This is the same [procedure](../../../installation/installing-images) as for SD cards.
 
 After preparing the storage device, connect the drive to the Raspberry Pi and power up the Pi (be aware of the extra USB power usage from external drive).
 After five to ten seconds, the Raspberry Pi should begin booting and show the rainbow splash screen on an attached display. Make sure that you do not have an SD card inserted in the Pi, since if you do, it will boot from that first.
 
-<a name="unsupported devices"></a>
+<a name="unsupported"></a>
 ## Unsupported devices
 
 Including Raspberry Pi 1, 2 (V1.1), Compute Module, Zero.
 
-The boot code for USB is stored in the BCM2837(B0) device only. The Pi 1 A/B, Pi 2 B (v1.1), and Pi Zero will all require SD cards as they are based on the BCM2835 and BCM2836. This boot code for USB is stored in ROM (except Pi 4B) which by definition cannot be changed.
+The boot code for USB Mass Storage boot is stored in the BCM2837(B0)'s ROM only. Earlier versions such as the BCM2836 and BCM2835 do not have this version of bootcode with USB Mass Storage boot.
+
+Since up to the Pi 4B, the boot code is written in ROM (Read only memory) it cannot be updated. For this reason the Pi 1 A/B, Pi 2 B (v1.1), and Pi Zero will all require SD cards.
 
 An alternative is to use the 'special bootcode.bin-only boot mode' as described [here](README.md). This still requires/boots from an SD-card, but `bootcode.bin` is the only file read from the SD-Card.
 
 ### Raspberry Pi 4
 
-Regarding the Raspberry Pi 4, the bootcode is stored in [EEPROM](../booteeprom.md) and can be updated. Support for mass storage boot will be added in a future update.
+The Raspberry Pi 4's bootcode is stored in [EEPROM](../booteeprom.md) and can be updated. Support for mass storage boot will be added in a future update.
 
 ## Known issues
 
 - The default timeout for checking bootable USB devices is 2 seconds. Some flash drives and rotational harddrives power up too slowly. It’s possible to extend this timeout to five seconds (add a new file `timeout` to the SD card), but there are devices that fail to respond within this period as well.
-- Some flash drives have a very specific protocol requirement that we don’t handle; as a result of this, we can’t talk to these drives correctly.
+- Some flash drives have a very specific protocol requirement that is not handled by the bootcode and may thus be incompatible.
 - 3.5" HDD's commonly require 12V as well as 5V and may draw too much current for the Pi's USB connection. Use an externally powered 3.5" HDD in this case.
--- a/hardware/raspberrypi/bootmodes/msd.md
+++ b/hardware/raspberrypi/bootmodes/msd.md
@@ -67,6 +67,6 @@ The Raspberry Pi 4's bootcode is stored in [EEPROM](../booteeprom.md) and can be
 
 ## Known issues
 
-- The default timeout for checking bootable devices is 2 seconds. Some flash drives and rotational harddrives power up too slowly. It’s possible to extend this timeout to five seconds (add a new file `timeout` to the SD card), but there are devices that fail to respond within this period as well.
+- The default timeout for checking bootable USB devices is 2 seconds. Some flash drives and rotational harddrives power up too slowly. It’s possible to extend this timeout to five seconds (add a new file `timeout` to the SD card), but there are devices that fail to respond within this period as well.
 - Some flash drives have a very specific protocol requirement that is not handled by the bootcode and may thus be incompatible.
 - 3.5" HDD's commonly require 12V as well as 5V and may draw too much current for the Pi's USB connection. Use an externally powered 3.5" HDD in this case.
--- a/hardware/raspberrypi/bootmodes/msd.md
+++ b/hardware/raspberrypi/bootmodes/msd.md
@@ -66,7 +66,7 @@ Including Raspberry Pi 1, 2 (V1.1), Compute Module, Zero
 
 The boot code for USB is stored in the BCM2837 device only, so the Pi 1 A/B, Pi 2 B (v1.1), and Pi Zero will all require SD cards as they are based on the BCM2835 and BCM2836. This boot code is stored in ROM (except Pi 4B) which by definition cannot be changed.
 
-Regarding the Raspberry Pi 4, the bootcode is stored in [EEPROM](../booteeprom.md) and can be updated. Support for mass storage boot will be added in a future update. In the meanwhile you can use the alternative below. 
+Regarding the Raspberry Pi 4, the bootcode is stored in [EEPROM](../booteeprom.md) and can be updated. Support for mass storage boot will be added in a future update. In the meantime you can use the alternative below. 
 
 An alternative is to use the 'special bootcode.bin-only boot mode' as described [here](./). This still requires/boots from an SD-card, but `bootcode.bin` is the only file read from the SD-Card.
 
--- a/hardware/raspberrypi/bootmodes/msd.md
+++ b/hardware/raspberrypi/bootmodes/msd.md
@@ -4,16 +4,17 @@
 
 This tutorial explains how to boot your Raspberry Pi from a USB mass storage device such as a flash drive or a USB hard disk. Note that this feature does not work with all USB mass storage devices.
 
-See the [documentation](./) for the boot sequence and alternative boot modes (Network, USB device, GPIO or SD boot).
+See the [documentation](README.md) for the boot sequence and alternative boot modes (Network, USB device, GPIO or SD boot).
 
-Note that "USB Mass Storage Boot" is different from "USB Device Boot Mode", the device boot mode allows a Raspberry Pi connected to a computer to boot using files on that computer.
+Note that "USB Mass Storage Boot" is different from "USB Device Boot Mode". The [device boot mode](device.md) allows a Raspberry Pi connected to a computer to boot using files on that computer.
 
-[Pi 2B v1.2 up to 3](./#raspberry-pi-2b-v12-3a-3b-cm3)
+[Pi 2B v1.2 up to 3](#raspberry-pi-2b-v12-3a-3b-cm3)
 
 [Pi 3B+ CM3+](#raspberry-pi-3b-cm3)
 
 [Unsupported devices](#unsupported-devices)
 
+<a name="Raspberry Pi 2B v1.2, 3A+, 3B, CM3"></a>
 ## Raspberry Pi 2B v1.2, 3A+, 3B, CM3
 
 On the Raspberry Pi 2B v1.2, 3A+, 3B, CM3, first USB [host boot mode](host.md) should be enabled. This is to allow Mass Storage Boot / Network boot (Network boot not supported on 3A+).
@@ -49,17 +50,19 @@ If you wish, you can remove the `program_usb_boot_mode` line from config.txt, so
 
 You can now boot from an USB Mass Storage device in the same way as booting from SD, see the following paragraph for Raspberry Pi 3B+, CM3+.
 
+<a name="Raspberry Pi 3B+, CM3+"></a>
 ## Raspberry Pi 3B+, CM3+
 
 The Raspberry Pi 3B+ and CM3+ support USB Mass Storage boot out of the box. The settings specific to the previous versions of Raspberry Pi do not have to be executed.
 
-This is verified using the following steps, this is the same [procedure](../../../installation/installing-images/) as for SD cards.
+This is the same [procedure](../../../installation/installing-images/installation/installing-images/) as for SD cards.
 
 On Windows you may have to initialise a (clean) harddrive before it's visible and has a drive letter (which may be required to write the image).
 
-After preparing the storage device, connect the drive to the Raspberry Pi and power up the Pi (mind extra USB power usage from external drive).
+After preparing the storage device, connect the drive to the Raspberry Pi and power up the Pi (be aware of the extra USB power usage from external drive).
 After five to ten seconds, the Raspberry Pi should begin booting and show the rainbow splash screen on an attached display. Make sure that you do not have an SD card inserted in the Pi, since if you do, it will boot from that first.
 
+<a name="unsupported devices"></a>
 ## Unsupported devices
 
 Including Raspberry Pi 1, 2 (V1.1), Compute Module, Zero
@@ -68,10 +71,10 @@ The boot code for USB is stored in the BCM2837 device only, so the Pi 1 A/B, Pi 
 
 Regarding the Raspberry Pi 4, the bootcode is stored in [EEPROM](../booteeprom.md) and can be updated. Support for mass storage boot will be added in a future update. In the meantime you can use the alternative below. 
 
-An alternative is to use the 'special bootcode.bin-only boot mode' as described [here](./). This still requires/boots from an SD-card, but `bootcode.bin` is the only file read from the SD-Card.
+An alternative is to use the 'special bootcode.bin-only boot mode' as described [here](README.md). This still requires/boots from an SD-card, but `bootcode.bin` is the only file read from the SD-Card.
 
 ## Known issues
 
 - The default timeout for checking bootable USB devices is 2 seconds. Some flash drives and rotational harddrives power up too slowly. It’s possible to extend this timeout to five seconds (add a new file `timeout` to the SD card), but there are devices that fail to respond within this period as well.
 - Some flash drives have a very specific protocol requirement that we don’t handle; as a result of this, we can’t talk to these drives correctly.
-- 3.5" HDD's commonly require 12V as well as 5V and may draw too much current for the Pi's USB connections.
+- 3.5" HDD's commonly require 12V as well as 5V and may draw too much current for the Pi's USB connection. Use an externally powered 3.5" HDD in this case.
--- a/hardware/raspberrypi/bootmodes/msd.md
+++ b/hardware/raspberrypi/bootmodes/msd.md
@@ -84,16 +84,6 @@ https://www.raspberrypi.org/blog/pi-3-booting-part-i-usb-mass-storage-boot/
 
 ## Known issues
 
-- The default timeout for checking bootable USB devices is 2 seconds. Some flash drives and rotational harddrives power up too slowly. It’s possible to extend this timeout to five seconds (add a new file 'timeout'to the SD card), but there are devices that fail to respond within this period as well, such as the Verbatim PinStripe 64GB. 
-- Some flash drives have a very specific protocol requirement that we don’t handle; as a result of this, we can’t talk to these drives correctly. An example of such a drive would be the Kingston Data Traveller 100 G3 32G.
-- 3.5" HDD's commonly require 12V as well as 5V and commonly draw too much current for the Pi's USB connections.
-
-## Known working devices
-
-| Device Name                                       | Reported by |
-|---|---|
-| Sandisk Cruzer Fit 16GB                           | [Henry Budden](http://www.raspberrypitutorials.yolasite.com/) |
-| Sandisk Cruzer Blade 16Gb                         | [Henry Budden](http://www.raspberrypitutorials.yolasite.com/) |
-| Samsung 32GB USB 3.0 drive                        | [Henry Budden](http://www.raspberrypitutorials.yolasite.com/) |
-| MeCo 16GB USB 3.0                                 | [Henry Budden](http://www.raspberrypitutorials.yolasite.com/) |
-| Kingston A400 240GB (external 2.0 USB case MA6116) Raspbian Buster Lite | Paul-Ver | 
+- The default timeout for checking bootable USB devices is 2 seconds. Some flash drives and rotational harddrives power up too slowly. It’s possible to extend this timeout to five seconds (add a new file 'timeout'to the SD card), but there are devices that fail to respond within this period as well.
+- Some flash drives have a very specific protocol requirement that we don’t handle; as a result of this, we can’t talk to these drives correctly.
+- 3.5" HDD's commonly require 12V as well as 5V and may draw too much current for the Pi's USB connections.
--- a/hardware/raspberrypi/bootmodes/msd.md
+++ b/hardware/raspberrypi/bootmodes/msd.md
@@ -68,7 +68,7 @@ Including Raspberry Pi 1, 2 (V1.1), Compute Module, Zero
 
 The boot code for USB is stored in the BCM2837 device only, so the Pi 1 A/B, Pi 2 B (v1.1), and Pi Zero will all require SD cards as they are based on the BCM2835 and BCM2836. This boot code is stored in ROM (except Pi 4B) which by definition cannot be changed.
 
-An alternative is to use the 'special bootcode.bin-only boot mode' as described [here](./). This still requires/boots from an SD-card but allows to run on an USB Device.
+An alternative is to use the 'special bootcode.bin-only boot mode' as described [here](./). This still requires/boots from an SD-card, but bootcode.bin is the only file read from the SD-Card.
 
 
 ## Extra info
--- a/hardware/raspberrypi/bootmodes/msd.md
+++ b/hardware/raspberrypi/bootmodes/msd.md
@@ -4,7 +4,7 @@
 
 This tutorial explains how to boot your Raspberry Pi from a USB mass storage device such as a flash drive or a USB hard disk. Note that this feature does not work with all USB mass storage devices.
 
-See the [documentation](https://www.raspberrypi.org/documentation/hardware/raspberrypi/bootmodes/) for the boot sequence and alternative boot modes (Network, USB device, GPIO or SD boot).
+See the [documentation](./) for the boot sequence and alternative boot modes (Network, USB device, GPIO or SD boot).
 
 Note that "USB Mass Storage Boot" is different from "USB Device Boot Mode", the device boot mode allows a Raspberry Pi connected to a computer to boot using files on that computer.
 
@@ -13,11 +13,11 @@ Note that "USB Mass Storage Boot" is different from "USB Device Boot Mode", the 
 **These models are not supported for USB Mass Storage Boot.**
 The boot code for USB is stored in the BCM2837 device only, so the Pi 1 A/B, Pi 2 B (v1.1), and Pi Zero will all require SD cards as they are based on the BCM2835 and BCM2836. This boot code is stored in ROM (except Pi 4B) which by definition cannot be changed.
 
-An alternative is to use the 'special bootcode.bin-only boot mode' as described [here](https://www.raspberrypi.org/documentation/hardware/raspberrypi/bootmodes/). This still requires/boots from an SD-card but allows to run on an USB Device.
+An alternative is to use the 'special bootcode.bin-only boot mode' as described [here](./). This still requires/boots from an SD-card but allows to run on an USB Device.
 
 ## Raspberry Pi 2B v1.2, 3A+, 3B
 
-On the Raspberry Pi 2B v1.2, 3A+, 3B, first USB [host boot mode](https://www.raspberrypi.org/documentation/hardware/raspberrypi/bootmodes/host.md) should be enabled. This is to allow Mass Storage Boot / Network boot (Network boot not supported on 3A+).
+On the Raspberry Pi 2B v1.2, 3A+, 3B, first USB [host boot mode](host.md) should be enabled. This is to allow Mass Storage Boot / Network boot (Network boot not supported on 3A+).
 
 To enable USB host boot mode, the Raspberry Pi needs to be booted from an SD card with a special option to set the USB host boot mode bit in the OTP (one-time programmable) memory. 
 
@@ -54,7 +54,7 @@ You can now boot from an USB Mass Storage device in the same way as booting from
 
 The Raspberry Pi 3B+ supports USB Mass Storage boot out of the box. The settings specific to the previous versions of Raspberry Pi do not have to be executed.
 
-This is verified using the following steps, this is the same (procedure)[https://www.raspberrypi.org/documentation/installation/installing-images/] as for SD cards though many alternative ways are functional:
+This is verified using the following steps, this is the same (procedure)[../../../installation/installing-images/] as for SD cards though many alternative ways are functional:
 
 1. Download Raspbian Buster Lite
 
--- a/hardware/raspberrypi/bootmodes/msd.md
+++ b/hardware/raspberrypi/bootmodes/msd.md
@@ -8,16 +8,9 @@ See the [documentation](./) for the boot sequence and alternative boot modes (Ne
 
 Note that "USB Mass Storage Boot" is different from "USB Device Boot Mode", the device boot mode allows a Raspberry Pi connected to a computer to boot using files on that computer.
 
-## Raspberry Pi 1, 2 (V1.1), Compute Module, Zero
+## Raspberry Pi 2B v1.2, 3A+, 3B, CM3
 
-**These models are not supported for USB Mass Storage Boot.**
-The boot code for USB is stored in the BCM2837 device only, so the Pi 1 A/B, Pi 2 B (v1.1), and Pi Zero will all require SD cards as they are based on the BCM2835 and BCM2836. This boot code is stored in ROM (except Pi 4B) which by definition cannot be changed.
-
-An alternative is to use the 'special bootcode.bin-only boot mode' as described [here](./). This still requires/boots from an SD-card but allows to run on an USB Device.
-
-## Raspberry Pi 2B v1.2, 3A+, 3B
-
-On the Raspberry Pi 2B v1.2, 3A+, 3B, first USB [host boot mode](host.md) should be enabled. This is to allow Mass Storage Boot / Network boot (Network boot not supported on 3A+).
+On the Raspberry Pi 2B v1.2, 3A+, 3B, CM3, first USB [host boot mode](host.md) should be enabled. This is to allow Mass Storage Boot / Network boot (Network boot not supported on 3A+).
 
 To enable USB host boot mode, the Raspberry Pi needs to be booted from an SD card with a special option to set the USB host boot mode bit in the OTP (one-time programmable) memory. 
 
@@ -50,7 +43,7 @@ If you wish, you can remove the `program_usb_boot_mode` line from config.txt, so
 
 You can now boot from an USB Mass Storage device in the same way as booting from SD, see the following paragraph for Raspberry Pi 3B+.
 
-## Raspberry Pi 3B+
+## Raspberry Pi 3B+, CM3+
 
 The Raspberry Pi 3B+ supports USB Mass Storage boot out of the box. The settings specific to the previous versions of Raspberry Pi do not have to be executed.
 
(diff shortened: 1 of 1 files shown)


hardware/raspberrypi/bootmodes/msd.md
```diff
@@ -4,7 +4,20 @@
 
 This tutorial explains how to boot your Raspberry Pi from a USB mass storage device such as a flash drive or a USB hard disk. Note that this feature does not work with all USB mass storage devices.
 
-## Program USB host boot mode (Raspberry Pi 3A+ and 3B only)
+See the [documentation](https://www.raspberrypi.org/documentation/hardware/raspberrypi/bootmodes/) for the boot sequence and alternative boot modes (Network, USB device, GPIO or SD boot).
```
lurch: Please make these relative links, rather than absolute URLs.

hardware/raspberrypi/bootmodes/msd.md
```diff
@@ -4,7 +4,20 @@
...
+
+Note that "USB Mass Storage Boot" is different from "USB Device Boot Mode", the device boot mode allows a Raspberry Pi connected to a computer to boot using files on that computer.
+
+## Raspberry Pi 1, 2 (V1.1), Compute Module, Zero
+
+**These models are not supported for USB Mass Storage Boot.**
+The boot code for USB is stored in the BCM2837 device only, so the Pi 1 A/B, Pi 2 B (v1.1), and Pi Zero will all require SD cards as they are based on the BCM2835 and BCM2836. This boot code is stored in ROM (except Pi 4B) which by definition cannot be changed.
```
lurch: Also the Pi1 A+/B+, the Pi Zero W and the CM1. Rather than listing the models that **don't** support MSD boot without an SD card, it might be easier to list only those models that **do**? (and then say something like "all other models of Raspberry Pi will need to use the 'special bootcode.bin-only boot mode'") :man_shrugging:

hardware/raspberrypi/bootmodes/msd.md
```diff
@@ -4,7 +4,20 @@
...
+
+An alternative is to use the 'special bootcode.bin-only boot mode' as described [here](https://www.raspberrypi.org/documentation/hardware/raspberrypi/bootmodes/). This still requires/boots from an SD-card but allows to run on an USB Device.
```
lurch: I think "run on an USB Device" is a bit vague? The benefit of this 'special bootcode.bin-only boot mode' is that `bootcode.bin` is the _only_ file read from the SD-card.

JamesH65:
I don't think we need the Known working devices section - it will immediately be out of date, and a nightmare to keep valid. The previous section has ones that don't work, which should be sufficient.

Maybe something along the lines of "We cannot guarantee all devices will work due to varying blah blah blah" as a catch all?

hardware/raspberrypi/bootmodes/msd.md
```diff
@@ -4,7 +4,20 @@
...
+
+## Raspberry Pi 2B v1.2, 3A+, 3B
+
+On the Raspberry Pi 2B v1.2, 3A+, 3B, first USB [host boot mode](https://www.raspberrypi.org/documentation/hardware/raspberrypi/bootmodes/host.md) should be enabled. This is to allow Mass Storage Boot / Network boot (Network boot not supported on 3A+).
```
lurch: And also the CM3 ?

hardware/raspberrypi/bootmodes/msd.md
```diff
@@ -35,12 +48,52 @@ Check that the output `0x3020000a` is shown. If it is not, then the OTP bit has
...
+
+The Raspberry Pi 3B+ supports USB Mass Storage boot out of the box. The settings specific to the previous versions of Raspberry Pi do not have to be executed.
+
+This is verified using the following steps, this is the same (procedure)[https://www.raspberrypi.org/documentation/installation/installing-images/] as for SD cards though many alternative ways are functional:
+
+1. Download Raspbian Buster Lite
+
+2. Unzip the .zip file (tested with 7zip, but any other should be fine) to an .img file.
+
+3. Initialise the drive (tested with MBR scheme, no partition/clean disk).
```
lurch: This step is never necessary, as flashing the .img overwrites the MBR

hardware/raspberrypi/bootmodes/msd.md
```diff
@@ -35,12 +48,52 @@ Check that the output `0x3020000a` is shown. If it is not, then the OTP bit has
...
+
+4. Flash the .img file to the external drive (tested using Win32DiskImager on Windows 10)
```
lurch: ...or just use Etcher, which eliminates the need for step 2 :wink:

hardware/raspberrypi/bootmodes/msd.md
```diff
@@ -35,12 +48,52 @@ Check that the output `0x3020000a` is shown. If it is not, then the OTP bit has
...
+
+6. Connect the drive to the Raspberry Pi and power up the Pi (mind extra USB power usage from external drive).
+
+7. After five to ten seconds, the Raspberry Pi should begin booting and show the rainbow splash screen on an attached display. Make sure that you do not have an SD card inserted in the Pi, since if you do, it will boot from that first.
+
+## Raspberry Pi 4B
+
+Support for USB mass storage boot will be added to the Raspberry Pi 4B in a future software update.
+
+According to rpdom in a [forum post](https://www.raspberrypi.org/forums/viewtopic.php?t=243995#p1488036):
```
lurch: @JamesH65 will need to confirm, but I don't think we want links from the official documentation to arbitrary forum posts?

hardware/raspberrypi/bootmodes/msd.md
```diff
@@ -35,12 +48,52 @@ Check that the output `0x3020000a` is shown. If it is not, then the OTP bit has
...
+
+The Pi4 uses a different boot loader to the earlier models. It is stored in eeprom on the board instead of part in the chip and part on the SD card. An update and instructions on how to apply it will be issued when the USB and network boot is ready.
+
+## Extra info
+
+https://www.raspberrypi.org/blog/pi-3-booting-part-i-usb-mass-storage-boot/
```
lurch: Ironically, this blog post link was removed by an earlier PR!

hardware/raspberrypi/bootmodes/msd.md
```diff
@@ -35,12 +48,52 @@ Check that the output `0x3020000a` is shown. If it is not, then the OTP bit has
...
 
-Copy a Raspbian or Raspbian Lite installation image directly onto your USB device, in the same way that you would for an SD card. To perform this step, follow the instructions [here](../../../installation/installing-images/README.md), remembering to select the drive that corresponds to your USB mass storage device. Make sure that you use a recent version of Raspbian/Raspbian Lite.
+## Known issues
 
-Once you have finished imaging your USB mass storage device, remove it from your computer and insert it into your Raspberry Pi.
+- The default timeout for checking bootable USB devices is 2 seconds. Some flash drives and rotational harddrives power up too slowly. It’s possible to extend this timeout to five seconds (add a new file 'timeout'to the SD card), but there are devices that fail to respond within this period as well, such as the Verbatim PinStripe 64GB.
```
lurch: Missing space in `'timeout'to` (and IMHO timeout should be in backticks rather than single quotes)

hardware/raspberrypi/bootmodes/msd.md
JamesH65: Correct, either copy data in to the docs, or don't use it. There is no guarantee that forum post will continue to exist and we should not be adding users names either.

hardware/raspberrypi/bootmodes/msd.md
```diff
@@ -35,12 +48,52 @@ Check that the output `0x3020000a` is shown. If it is not, then the OTP bit has
...
+## Known issues
 
-Once you have finished imaging your USB mass storage device, remove it from your computer and insert it into your Raspberry Pi.
+- The default timeout for checking bootable USB devices is 2 seconds. Some flash drives and rotational harddrives power up too slowly. It’s possible to extend this timeout to five seconds (add a new file 'timeout'to the SD card), but there are devices that fail to respond within this period as well, such as the Verbatim PinStripe 64GB. 
+- Some flash drives have a very specific protocol requirement that we don’t handle; as a result of this, we can’t talk to these drives correctly. An example of such a drive would be the Kingston Data Traveller 100 G3 32G.
+- 3.5" HDD's commonly require 12V as well as 5V and commonly draw too much current for the Pi's USB connections.
 
-## Boot your Raspberry Pi from the USB mass storage device
+## Known working devices
```
lurch: It'll be impossible to keep lists of "known working devices" or "known not-working devices" up-to-date, so these shouldn't be included in the documentation repo. (same applies to the earlier mentions of "Verbatim PinStripe 64GB" and "")Kingston Data Traveller 100 G3 32G

hardware/raspberrypi/bootmodes/msd.md
```diff
@@ -35,12 +48,52 @@ Check that the output `0x3020000a` is shown. If it is not, then the OTP bit has
 
 If you wish, you can remove the `program_usb_boot_mode` line from config.txt, so that if you put the SD card into another Raspberry Pi, it won't program USB host boot mode. Make sure there is no blank line at the end of config.txt.
 
-## Prepare the USB mass storage device
+You can now boot from an USB Mass Storage device in the same way as booting from SD, see the following paragraph for Raspberry Pi 3B+.
+
+## Raspberry Pi 3B+
+
+The Raspberry Pi 3B+ supports USB Mass Storage boot out of the box. The settings specific to the previous versions of Raspberry Pi do not have to be executed.
```
lurch: ...and the CM3+ ?

lurch:
ping also @andrum99 as he made a lot of edits in this area recently.

hardware/raspberrypi/bootmodes/msd.md
andrum99: Not needed - the way this page is found (excluding external links) is via the boot modes page.

If an external link is used, the page includes "breadcrumbs" which the user can use to go up the tree to the boot modes page.

lurch: The reason I prefer relative links, is that it makes the documentation pages equally readable on https://github.com/raspberrypi/documentation/ and https://www.raspberrypi.org/documentation/

hardware/raspberrypi/bootmodes/msd.md
andrum99: FIrst paragraph - not needed, since again this page will usually be found via the boot modes page, which includes a link to device mode boot which explains what device mode boot does.

The rest is not needed and will only add confusion - the first paragraph of the page already specifies the models which the page applies to.

hardware/raspberrypi/bootmodes/msd.md
andrum99: Again, probably not needed, since the main boot modes page includes information about the bootcode.bin only method of booting.

hardware/raspberrypi/bootmodes/msd.md
andrum99: Again, probably not needed, since the usual method of getting to this page is via the boot modes page.

andrum99:
Apologies - I started adding comments about sections of the page that were not affected by this PR, although for some reason guthub has marked them in green, and @lurch has already commented on them. I've deleted my comments since they were therefore in error.

hardware/raspberrypi/bootmodes/msd.md
cleverca22: for the rpi4, the boot code is stored on both rom and spi flash

the rpi4 rom appears to only be able to run code from 2 places: `recovery.bin` on an SD card, or a blob in the SPI flash

and then that SPI flash would add the usb support (i think its currently missing)

hardware/raspberrypi/bootmodes/msd.md
Paul-Ver: It actually was required on Windows for my 'clean' harddrive to get assigned a drive letter. Before I was able to write to it using Win32DiskImager.

In fact; I'm not even sure if we should include these steps here, as it will lead to double documentation and it has two issues:
- If you update one; you would have to update the other.
- There's two places with the same info; so it may be harder to find in search results.
- It leads to confusion as to which is the "correct" solution.

hardware/raspberrypi/bootmodes/msd.md
Paul-Ver: I've changed it to: "This still requires/boots from an SD-card, but bootcode.bin is the only file read from the SD-Card."

However; this means that the SD-Card is not read from anymore after the boot process as well?

hardware/raspberrypi/bootmodes/msd.md
Paul-Ver: Cleverca22

Yes; it can boot from SPI EEPROM (I think SPI Flash is incorrect as that more describes an SD card). But it currently does not support USB boot. I've added a link to the Pi 4B EEPROM boot page; to not have it documented twice and have one place with the complete description.

See: https://www.raspberrypi.org/documentation/hardware/raspberrypi/booteeprom.md

hardware/raspberrypi/bootmodes/msd.md
Paul-Ver: Thank you; I've added the CM3+.

Initially I started this update to add my experiences on the RPI 3B+ working out of the box; which wasn't clearly reflected in this documentation (in my opinion). I have no experience with the CM3+ but given it's similarity in hardware; I can assume it will work?

hardware/raspberrypi/bootmodes/msd.md
Paul-Ver: True; It was an attempt at merging the info of the blogpost into this documentation.
Also it would have been interesting info to anyone that has issues or wants to avoid issues, but I agree such list is unmaintainable.

I've removed the list.

hardware/raspberrypi/bootmodes/msd.md
Paul-Ver: To be honest it was a very usefull resource for getting my setup to work and a dev-blog isn't just a random blogpost in my opinion.

However; I've merged as much of the (relevant) info of that blogpost in this article; so we could indeed remove the link. (And I did)

hardware/raspberrypi/bootmodes/msd.md
Paul-Ver: Agreed, it was an attempt to validate the info I was writing, but random forum posts that may/not continue to exist are a poort attempt.

I've edited it to reference to https://www.raspberrypi.org/documentation/hardware/raspberrypi/booteeprom.md (as a relative link (: )

hardware/raspberrypi/bootmodes/msd.md
andrum99: Yes, but the bootcode.bin file has to be the only file on the SD card for this to work. If any of the other boot files are present, then bootcode.bin will read those, and will try to complete the boot process from the SD card.

Paul-Ver:
I think the edits cover the comments above.
I've also changed the setup of the document a bit; so that the supported platforms come before the unsupported ones (but added an index to make it more obvious where to find info for the unsupported ones).

I haven't pressed the "resolve conversation" buttons as I'm not quite sure if that's up to me or up to you guys.

I think that in overal the document is more up-to-date and slightly more readable. Thank you guys very much for your input!

andrum99:
> I haven't pressed the "resolve conversation" buttons as I'm not quite sure if that's up to me or up to you guys.

If you have addressed the issues raised, then I would press the "resolve conversation" buttons. That way the reviewer(s) know you have addressed the issues. If they have further comments, they will soon pipe up 😋

hardware/raspberrypi/bootmodes/msd.md
lurch: :+1: for not having duplicated documentation.

hardware/raspberrypi/bootmodes/msd.md
```diff
@@ -4,7 +4,19 @@
 
 This tutorial explains how to boot your Raspberry Pi from a USB mass storage device such as a flash drive or a USB hard disk. Note that this feature does not work with all USB mass storage devices.
 
-## Program USB host boot mode (Raspberry Pi 3A+ and 3B only)
+See the [documentation](./) for the boot sequence and alternative boot modes (Network, USB device, GPIO or SD boot).
```
lurch: Could you change the `documentation` link-text to `bootmodes documentation` please?

@JamesH65 Should we prefer a link-url like `./README.md` or is `./` okay?

hardware/raspberrypi/bootmodes/msd.md
```diff
@@ -4,7 +4,19 @@
...
+
+Note that "USB Mass Storage Boot" is different from "USB Device Boot Mode", the device boot mode allows a Raspberry Pi connected to a computer to boot using files on that computer.
```
lurch: I wonder if the comma here should instead be a semi-colon?

hardware/raspberrypi/bootmodes/msd.md
```diff
@@ -4,7 +4,19 @@
...
+
+[Pi 2B v1.2 up to 3](./#raspberry-pi-2b-v12-3a-3b-cm3)
+
+[Pi 3B+ CM3+](#raspberry-pi-3b-cm3)
```
lurch: To get anchor-links like this working on raspberrypi.org you need to add explicit anchors - see e.g. https://raw.githubusercontent.com/raspberrypi/documentation/master/configuration/raspi-config.md for how this is done.

hardware/raspberrypi/bootmodes/msd.md
```diff
@@ -35,12 +47,31 @@ Check that the output `0x3020000a` is shown. If it is not, then the OTP bit has
...
 If you wish, you can remove the `program_usb_boot_mode` line from config.txt, so that if you put the SD card into another Raspberry Pi, it won't program USB host boot mode. Make sure there is no blank line at the end of config.txt.
 
-## Prepare the USB mass storage device
+You can now boot from an USB Mass Storage device in the same way as booting from SD, see the following paragraph for Raspberry Pi 3B+, CM3+.
+
+## Raspberry Pi 3B+, CM3+
+
+The Raspberry Pi 3B+ and CM3+ support USB Mass Storage boot out of the box. The settings specific to the previous versions of Raspberry Pi do not have to be executed.
+
+This is verified using the following steps, this is the same [procedure](../../../installation/installing-images/) as for SD cards.
```
lurch: I'm not sure that "This is verified" is relevant?

hardware/raspberrypi/bootmodes/msd.md
```diff
@@ -35,12 +47,31 @@ Check that the output `0x3020000a` is shown. If it is not, then the OTP bit has
...
+
+On Windows you may have to initialise a (clean) harddrive before it's visible and has a drive letter (which may be required to write the image).
+
+After preparing the storage device, connect the drive to the Raspberry Pi and power up the Pi (mind extra USB power usage from external drive).
```
lurch: "mind" -> "be aware of the" ?  :man_shrugging:

hardware/raspberrypi/bootmodes/msd.md
```diff
@@ -35,12 +47,31 @@ Check that the output `0x3020000a` is shown. If it is not, then the OTP bit has
...
+After five to ten seconds, the Raspberry Pi should begin booting and show the rainbow splash screen on an attached display. Make sure that you do not have an SD card inserted in the Pi, since if you do, it will boot from that first.
+
+## Unsupported devices
+
+Including Raspberry Pi 1, 2 (V1.1), Compute Module, Zero
+
+The boot code for USB is stored in the BCM2837 device only, so the Pi 1 A/B, Pi 2 B (v1.1), and Pi Zero will all require SD cards as they are based on the BCM2835 and BCM2836. This boot code is stored in ROM (except Pi 4B) which by definition cannot be changed.
```
lurch: I think these sentences need to be tweaked slightly, as it's not clear what "this boot code" in the second sentence is referring to?

hardware/raspberrypi/bootmodes/msd.md
```diff
@@ -35,12 +47,31 @@ Check that the output `0x3020000a` is shown. If it is not, then the OTP bit has
...
+An alternative is to use the 'special bootcode.bin-only boot mode' as described [here](./). This still requires/boots from an SD-card, but `bootcode.bin` is the only file read from the SD-Card.
 
-## Boot your Raspberry Pi from the USB mass storage device
+## Known issues
 
-Attach the USB mass storage device to your Raspberry Pi, and power up the Pi. After five to ten seconds, the Raspberry Pi should begin booting and show the rainbow splash screen on an attached display. Make sure that you do not have an SD card inserted in the Pi, since if you do, it will boot from that first.
+- The default timeout for checking bootable USB devices is 2 seconds. Some flash drives and rotational harddrives power up too slowly. It’s possible to extend this timeout to five seconds (add a new file `timeout` to the SD card), but there are devices that fail to respond within this period as well.
+- Some flash drives have a very specific protocol requirement that we don’t handle; as a result of this, we can’t talk to these drives correctly.
+- 3.5" HDD's commonly require 12V as well as 5V and may draw too much current for the Pi's USB connections.
```
lurch: Perhaps mention that a 3.5" HDD with its own power-supply should be okay?

hardware/raspberrypi/bootmodes/msd.md
```diff
@@ -35,12 +47,31 @@ Check that the output `0x3020000a` is shown. If it is not, then the OTP bit has
...
+After five to ten seconds, the Raspberry Pi should begin booting and show the rainbow splash screen on an attached display. Make sure that you do not have an SD card inserted in the Pi, since if you do, it will boot from that first.
+
+## Unsupported devices
+
+Including Raspberry Pi 1, 2 (V1.1), Compute Module, Zero
+
+The boot code for USB is stored in the BCM2837 device only, so the Pi 1 A/B, Pi 2 B (v1.1), and Pi Zero will all require SD cards as they are based on the BCM2835 and BCM2836. This boot code is stored in ROM (except Pi 4B) which by definition cannot be changed.
 
-Copy a Raspbian or Raspbian Lite installation image directly onto your USB device, in the same way that you would for an SD card. To perform this step, follow the instructions [here](../../../installation/installing-images/README.md), remembering to select the drive that corresponds to your USB mass storage device. Make sure that you use a recent version of Raspbian/Raspbian Lite.
+Regarding the Raspberry Pi 4, the bootcode is stored in [EEPROM](../booteeprom.md) and can be updated. Support for mass storage boot will be added in a future update. In the meantime you can use the alternative below.
```
lurch: @timg236 Does 'special bootcode.bin-only boot mode' allow USB-booting on the Pi 4 ?

timg236: The current Raspberry Pi 4 bootloader does not support USB mass storage boot. It's possible to have the root file system on USB because start.elf loads the kernel from /boot or the network and then the kernel can read the root file system itself.
That's normally the workaround that people refer to, so the OS is booting from from USB but the bootloader is not.

lurch: Right. But does the Pi4 need start.elf, cmdline.txt, etc. on the SD-card too, or does it (also) support having _only_ bootcode.bin on the SD-card?

timg236: Pi4 needs start.elf, fixup.dat, kernel, overlays and dtbs on the sd-card (or TFTP server)

bootcode.bin is not required because it's loaded from the EEPROM

lurch: Thanks @timg236 !
@Paul-Ver This means that the "In the meantime you can use the alternative below." text isn't relevant for the Pi 4.

hardware/raspberrypi/bootmodes/msd.md
Paul-Ver: Ah!
I tried `./readme.md` but that didn't work `./README.md` does work. I thought URL's weren't case sensitive; but ofc it is case sensitive when referring to a file.

I strongly agree with you on `./README.md` as it's more direct/obvious to where it leads.
In fact; I've changed them to `README.md` which better matches the way these references are/were done in the rest of the document(s).

hardware/raspberrypi/bootmodes/msd.md
Paul-Ver: Not relevant; it was a remainant of (my) old edit. (in an attempt to make it reproducable, but I agree it looks weird in official documentation)
Edited.

hardware/raspberrypi/bootmodes/msd.md
Paul-Ver: Agreed and edited.

hardware/raspberrypi/bootmodes/msd.md
Paul-Ver: Agreed and edited.

hardware/raspberrypi/bootmodes/msd.md
andrum99: >I thought URL's weren't case sensitive; but ofc it is case sensitive when referring to a file.

Generally URLs are case sensitive. The exception is Windows servers, where they are not case sensitive.

hardware/raspberrypi/bootmodes/msd.md
Paul-Ver: I've changed it to a `.` 
I'm not sure what fits best as I'm not a native English speaker, but I feel there is no use in concatenating both sentences.

hardware/raspberrypi/bootmodes/msd.md
Paul-Ver: I've added these links; but I'm not quite sure if it works on the webpage; it did on GitHub (but was working on GitHub before also, I believe).

hardware/raspberrypi/bootmodes/msd.md
Paul-Ver: To "the boot code" in the first sentence. I've made some small edits in my second commit, but I'm not quite sure how to properly rephrase it.

hardware/raspberrypi/bootmodes/msd.md
Paul-Ver: Ah okay; I wasn't quite sure if it'd be working; but couldn't make up why not.
We may want to add the steps to do the above to boot the OS from USB (with bootloader from USB), but I'm not quite sure on these steps.

For now I've put the Pi 4 in a sub in the "unsupported devices"

hardware/raspberrypi/bootmodes/msd.md
```diff
@@ -4,16 +4,17 @@
 
 This tutorial explains how to boot your Raspberry Pi from a USB mass storage device such as a flash drive or a USB hard disk. Note that this feature does not work with all USB mass storage devices.
 
-See the [documentation](./) for the boot sequence and alternative boot modes (Network, USB device, GPIO or SD boot).
+See the [documentation](README.md) for the boot sequence and alternative boot modes (Network, USB device, GPIO or SD boot).
```
lurch: I still think the link-text should be "bootmodes documentation" rather than just "documentation" :wink:

hardware/raspberrypi/bootmodes/msd.md
lurch: Github adds automatic-anchor-links to Markdown headings, which is helpful in general but not in this particular case :wink: 
To work on the website, the text in the anchor-link needs to match the anchor-name, so to get `[Pi 3B+ CM3+](#raspberry-pi-3b-cm3)` to work you'd need to use `<a name="raspberry-pi-3b-cm3">`. (But there's no need for the anchors to precisely match the name of the heading, so `[Pi 3B+ CM3+](#pi_3plus)` and `<a name="pi_3plus">` would also work)

I hope that all makes sense! :smiley:

hardware/raspberrypi/bootmodes/msd.md
```diff
@@ -35,12 +48,35 @@ Check that the output `0x3020000a` is shown. If it is not, then the OTP bit has
...
 
-## Prepare the USB mass storage device
+You can now boot from an USB Mass Storage device in the same way as booting from SD, see the following paragraph for Raspberry Pi 3B+, CM3+.
+
+<a name="Raspberry Pi 3B+, CM3+"></a>
+## Raspberry Pi 3B+, CM3+
+
+The Raspberry Pi 3B+ and CM3+ support USB Mass Storage boot out of the box. The settings specific to the previous versions of Raspberry Pi do not have to be executed.
+
+This is the same [procedure](../../../installation/installing-images/installation/installing-images/) as for SD cards.
```
lurch: Did something go wrong with the URL here?

hardware/raspberrypi/bootmodes/msd.md
```diff
@@ -35,12 +48,35 @@ Check that the output `0x3020000a` is shown. If it is not, then the OTP bit has
...
+
+On Windows you may have to initialise a (clean) harddrive before it's visible and has a drive letter (which may be required to write the image).
```
lurch: I'm not sure this sentence is necessary, as AFAIK it's dependent on what image-writing software you're using. (and also kinda conflicts with the "After preparing the storage device" sentence below)

hardware/raspberrypi/bootmodes/msd.md
```diff
@@ -35,12 +48,35 @@ Check that the output `0x3020000a` is shown. If it is not, then the OTP bit has
...
+
+After preparing the storage device, connect the drive to the Raspberry Pi and power up the Pi (be aware of the extra USB power usage from external drive).
+After five to ten seconds, the Raspberry Pi should begin booting and show the rainbow splash screen on an attached display. Make sure that you do not have an SD card inserted in the Pi, since if you do, it will boot from that first.
+
+<a name="unsupported devices"></a>
+## Unsupported devices
+
+Including Raspberry Pi 1, 2 (V1.1), Compute Module, Zero.
+
+The boot code for USB is stored in the BCM2837(B0) device only. The Pi 1 A/B, Pi 2 B (v1.1), and Pi Zero will all require SD cards as they are based on the BCM2835 and BCM2836. This boot code for USB is stored in ROM (except Pi 4B) which by definition cannot be changed.
```
lurch: Hmm, what about something like
"The boot code for USB Mass Storage is present in the (non-updateable) ROM of the BCM2837 devices only. The Pi 1 A/B/A+/B+, Pi 2 B (v1.1), and Pi Zero are based on the earlier BCM2835 and BCM2836 devices which lack this boot code."
:grey_question: 
Although even that will need tidying up a bit in the copy-edit stage.

hardware/raspberrypi/bootmodes/msd.md
```diff
@@ -35,12 +48,35 @@ Check that the output `0x3020000a` is shown. If it is not, then the OTP bit has
...
+
+An alternative is to use the 'special bootcode.bin-only boot mode' as described [here](README.md). This still requires/boots from an SD-card, but `bootcode.bin` is the only file read from the SD-Card.
 
-Copy a Raspbian or Raspbian Lite installation image directly onto your USB device, in the same way that you would for an SD card. To perform this step, follow the instructions [here](../../../installation/installing-images/README.md), remembering to select the drive that corresponds to your USB mass storage device. Make sure that you use a recent version of Raspbian/Raspbian Lite.
+### Raspberry Pi 4
 
-Once you have finished imaging your USB mass storage device, remove it from your computer and insert it into your Raspberry Pi.
+Regarding the Raspberry Pi 4, the bootcode is stored in [EEPROM](../booteeprom.md) and can be updated. Support for mass storage boot will be added in a future update.
```
lurch: I'd be tempted to change the first sentence to something like:
"The Raspberry Pi 4's bootcode is stored in [EEPROM](../booteeprom.md) and **can** be updated."

hardware/raspberrypi/bootmodes/msd.md
```diff
@@ -35,12 +48,35 @@ Check that the output `0x3020000a` is shown. If it is not, then the OTP bit has
...
 
-## Boot your Raspberry Pi from the USB mass storage device
+## Known issues
 
-Attach the USB mass storage device to your Raspberry Pi, and power up the Pi. After five to ten seconds, the Raspberry Pi should begin booting and show the rainbow splash screen on an attached display. Make sure that you do not have an SD card inserted in the Pi, since if you do, it will boot from that first.
+- The default timeout for checking bootable USB devices is 2 seconds. Some flash drives and rotational harddrives power up too slowly. It’s possible to extend this timeout to five seconds (add a new file `timeout` to the SD card), but there are devices that fail to respond within this period as well.
+- Some flash drives have a very specific protocol requirement that we don’t handle; as a result of this, we can’t talk to these drives correctly.
```
lurch: Perhaps rewrite this sentence to avoid using "we"? E.g.
"Some flash drives have a very specific protocol requirement that isn’t handled; as a result of this, the boot code can’t talk to these drives correctly."

Paul-Ver: Agreed, was copied from the blog and didn't edit it as it was more or less a quote, but I agree the documentation shouldn't include quotes of not neccesary.

Edited.

hardware/raspberrypi/bootmodes/msd.md
Paul-Ver: Agreed and edited.

hardware/raspberrypi/bootmodes/msd.md
Paul-Ver: I've just 'completely' rewritten that part based on your input.

I agree since it wasn't a very clear message, edited.

hardware/raspberrypi/bootmodes/msd.md
Paul-Ver: Agreed and edited.

I may; but there's enough resources (and simply too many ways) to do the installing. I feel we all have our different favorite tools for it; and it's not really up to me to decide the "officially supported procedure" for flashing :P

hardware/raspberrypi/bootmodes/msd.md
Paul-Ver: Looks like an oopsie, edited.

hardware/raspberrypi/bootmodes/msd.md
Paul-Ver: Agreed and edited.

hardware/raspberrypi/bootmodes/msd.md
Paul-Ver: I think it should work now, I was using the anchor names as GitHub is doing, but this edit should 'probably' also work?
Though I can't really test it unless if it's live as a webpage?
I have no clue how these .md pages are directly converted to HTML and hosted, I do think it's nice, but have no clue on how to reproduce this at home (w/o figuring it all out from scratch :p)

hardware/raspberrypi/bootmodes/msd.md
Paul-Ver: Outdated as this part has been rewritten.

hardware/raspberrypi/bootmodes/msd.md
```diff
@@ -35,12 +48,35 @@ Check that the output `0x3020000a` is shown. If it is not, then the OTP bit has
...
+
+<a name="unsupported"></a>
+## Unsupported devices
+
+Including Raspberry Pi 1, 2 (V1.1), Compute Module, Zero.
+
+The boot code for USB Mass Storage boot is stored in the BCM2837(B0)'s ROM only. Earlier versions such as the BCM2836 and BCM2835 do not have this version of bootcode with USB Mass Storage boot.
+
+Since up to the Pi 4B, the boot code is written in ROM (Read only memory) it cannot be updated. For this reason the Pi 1 A/B, Pi 2 B (v1.1), and Pi Zero will all require SD cards.
+
```
JamesH65: "since up to the Pi 4B" - sounds odd, and please use the entire (tedious I know) Pi name e.g. "From the Raspberry Pi 4B onwards" or something.

JamesH65:
Apart from one inline comment, pretty happy with this, so once the line is updated, I can merge unless others have any issues.

hardware/raspberrypi/bootmodes/msd.md
andrum99: I don't think you need the "unsupported devices" section, since the page already specifies exactly which devices MSD boot works with at the very top.

Paul-Ver: I agree on the "Unsupported devices", it's a bit of an odd heading for it; as well as it gives a lot of info about devices for which this page isn't really relevant. And the reasoning is also documented on the "Special bootcode.bin-only boot mode page".

I did move the reference to the 'special bootcode.bin-only boot mode' to the intro; so that any user can still find it's way to an alternative solution for his Raspberry Pi if it's not supported.

Also to still comply with  #1361 

By that the whole "Since up to" is gone as well :)

Paul-Ver: Also; I've changed the capitalisation from "USB Mass Storage Boot mode" to "USB mass storage boot". For some reason it doesn't feel better; but it does seem more correct.

andrum99: > Also; I've changed the capitalisation from "USB Mass Storage Boot mode" to "USB mass storage boot". For some reason it doesn't feel better; but it does seem more correct.

Without the capitalisation is correct. Originally there was a page that called it MSD boot, so that's probably how the capitalisation crept in.

Paul-Ver:
> Apart from one inline comment, pretty happy with this, so once the line is updated, I can merge unless others have any issues.

Personally (somewhat biased) I think it adds a lot of relevant info:
- Pi 3B+, CM3+ (that it will boot out of the box; was unclear/outdated in the original page)
- Adds reference to alternative bootmode for unsupported devices ( #1361 )
- Known issues; to aid troubleshooting

Though I'm not all too sure about the "Index links" to the different headers; and if we should include the link to "Pi 4" at the moment. In the bold text at the start: "Support for USB mass storage boot will be added to the Raspberry Pi 4B in a future software update.", it's actually already covered.

hardware/raspberrypi/bootmodes/msd.md
```diff
@@ -2,9 +2,22 @@
...
-## Program USB host boot mode (Raspberry Pi 3A+ and 3B only)
+See the [bootmodes documentation](README.md) for the boot sequence and alternative boot modes (Network, USB device, GPIO or SD boot).
+
+Note that "USB mass storage boot" is different from "USB device boot mode". The [device boot mode](device.md) allows a Raspberry Pi connected to a computer to boot using files on that computer.
+
+For devices that are not supported, an alternative is to use the 'special bootcode.bin-only boot mode' as described [here](README.md). This still requires/boots from an SD-card, but `bootcode.bin` is the only file read from the SD-Card.
+
+[Pi 2B v1.2 up to 3](#pi2to3)
+
+[Pi 3B+ CM3+](#pi3plus)
```
andrum99: I think these should be removed. It's not obvious they refer to stuff further down the page, and they're not really necessary anyway.

hardware/raspberrypi/bootmodes/msd.md
```diff
@@ -35,12 +48,25 @@ Check that the output `0x3020000a` is shown. If it is not, then the OTP bit has
...
+The Raspberry Pi 4's bootcode is stored in [EEPROM](../booteeprom.md) and can be updated. Support for mass storage boot will be added in a future update.
 
-## Boot your Raspberry Pi from the USB mass storage device
+## Known issues
 
-Attach the USB mass storage device to your Raspberry Pi, and power up the Pi. After five to ten seconds, the Raspberry Pi should begin booting and show the rainbow splash screen on an attached display. Make sure that you do not have an SD card inserted in the Pi, since if you do, it will boot from that first.
+- The default timeout for checking bootable USB devices is 2 seconds. Some flash drives and rotational harddrives power up too slowly. It’s possible to extend this timeout to five seconds (add a new file `timeout` to the SD card), but there are devices that fail to respond within this period as well.
+- Some flash drives have a very specific protocol requirement that is not handled by the bootcode and may thus be incompatible.
+- 3.5" HDD's commonly require 12V as well as 5V and may draw too much current for the Pi's USB connection. Use an externally powered 3.5" HDD in this case.
```
andrum99: There shouldn't be any 3.5" hard disks attempting to power themselves off the USB bus. I would remove this and add a remark that may issues are power related, so using a powered USB hub will help.

hardware/raspberrypi/bootmodes/msd.md
```diff
@@ -35,12 +48,25 @@ Check that the output `0x3020000a` is shown. If it is not, then the OTP bit has
...
+<a name="pi4"></a>
+## Raspberry Pi 4
 
-Once you have finished imaging your USB mass storage device, remove it from your computer and insert it into your Raspberry Pi.
+The Raspberry Pi 4's bootcode is stored in [EEPROM](../booteeprom.md) and can be updated. Support for mass storage boot will be added in a future update.
 
-## Boot your Raspberry Pi from the USB mass storage device
+## Known issues
 
-Attach the USB mass storage device to your Raspberry Pi, and power up the Pi. After five to ten seconds, the Raspberry Pi should begin booting and show the rainbow splash screen on an attached display. Make sure that you do not have an SD card inserted in the Pi, since if you do, it will boot from that first.
```
andrum99: Please add this paragraph back in somewhere.

lurch: It's been moved up to earlier in the page?

andrum99: My mistake - I saw it in red and assumed it had gone.

JamesH65:
Have adjusted according to comment above. I'd like to merge this now, unless there are objections.

hardware/raspberrypi/bootmodes/msd.md
```diff
@@ -69,4 +63,4 @@ The Raspberry Pi 4's bootcode is stored in [EEPROM](../booteeprom.md) and can be
 
 - The default timeout for checking bootable USB devices is 2 seconds. Some flash drives and rotational harddrives power up too slowly. It’s possible to extend this timeout to five seconds (add a new file `timeout` to the SD card), but there are devices that fail to respond within this period as well.
 - Some flash drives have a very specific protocol requirement that is not handled by the bootcode and may thus be incompatible.
-- 3.5" HDD's commonly require 12V as well as 5V and may draw too much current for the Pi's USB connection. Use an externally powered 3.5" HDD in this case.
+- Lack of power can be an issue, as some devices draw too much current from the Raspberry Pi's USB connection and fail to start, or prevent the Raspberry Pi from booting. An external power supply for the device will be required in these cases.
```
andrum99: I think you need to mention a powered USB hub here as well.

JamesH65: "An external power supply or powered USB hub will be..." ?

andrum99: > "An external power supply or powered USB hub will be..." ?

That would be OK. Usually devices either have their own PSU or use bus power, so I would go with something like:

"Lack of power can be an issue, so it is recommended to use a powered USB hub, particuarly if you are attaching more than one storage device to the Raspberry Pi. If your device has its own power supply then be sure to use it.

JamesH65:
80 comments on one single page change was quite enough, so now merged! Please send in a PR for any further changes.

hardware/raspberrypi/bootmodes/msd.md
```diff
@@ -35,12 +43,24 @@ Check that the output `0x3020000a` is shown. If it is not, then the OTP bit has
 
 If you wish, you can remove the `program_usb_boot_mode` line from config.txt, so that if you put the SD card into another Raspberry Pi, it won't program USB host boot mode. Make sure there is no blank line at the end of config.txt.
```
lurch: @JamesH65 Is the bit about "Make sure there is no blank line at the end of config.txt." true / relevant?

hardware/raspberrypi/bootmodes/msd.md
```diff
@@ -35,12 +43,24 @@ Check that the output `0x3020000a` is shown. If it is not, then the OTP bit has
...
 
-## Prepare the USB mass storage device
+You can now boot from an USB mass storage device in the same way as booting from SD, see the following paragraph for Raspberry Pi 3B+, CM3+.
+
+## Raspberry Pi 3B+, CM3+
+
+The Raspberry Pi 3B+ and CM3+ support USB mass storage boot out of the box. The settings specific to the previous versions of Raspberry Pi do not have to be executed.
+
+This is the same [procedure](../../../installation/installing-images) as for SD cards.
```
lurch: "This is the same procedure..." feels a bit out of context IMHO? Perhaps a slight tweak of previous text like "Copy a Raspbian or Raspbian Lite installation image onto your USB device using the same procedure..." would be better?

hardware/raspberrypi/bootmodes/msd.md
```diff
@@ -2,9 +2,17 @@
...
+
+Note that "USB mass storage boot" is different from "USB device boot mode". The [device boot mode](device.md) allows a Raspberry Pi connected to a computer to boot using files on that computer.
+
+For devices that are not supported, an alternative is to use the 'special bootcode.bin-only boot mode' as described [here](README.md). This still requires/boots from an SD-card, but `bootcode.bin` is the only file read from the SD-Card.
+
+## Raspberry Pi 2B v1.2, 3A+, 3B, CM3
+
+On the Raspberry Pi 2B v1.2, 3A+, 3B, CM3, first USB [host boot mode](host.md) should be enabled. This is to allow "mass storage boot" and ["network boot"](net.md) (Network boot not supported on 3A+).
 
 To enable USB host boot mode, the Raspberry Pi needs to be booted from an SD card with a special option to set the USB host boot mode bit in the OTP (one-time programmable) memory.
```
lurch: Is it worth making "OTP" here a link to https://github.com/raspberrypi/documentation/blob/master/hardware/raspberrypi/otpbits.md ?

hardware/raspberrypi/bootmodes/msd.md
```diff
@@ -35,12 +43,24 @@ Check that the output `0x3020000a` is shown. If it is not, then the OTP bit has
 
 If you wish, you can remove the `program_usb_boot_mode` line from config.txt, so that if you put the SD card into another Raspberry Pi, it won't program USB host boot mode. Make sure there is no blank line at the end of config.txt.
 
-## Prepare the USB mass storage device
+You can now boot from an USB mass storage device in the same way as booting from SD, see the following paragraph for Raspberry Pi 3B+, CM3+.
+
+## Raspberry Pi 3B+, CM3+
+
+The Raspberry Pi 3B+ and CM3+ support USB mass storage boot out of the box. The settings specific to the previous versions of Raspberry Pi do not have to be executed.
```
lurch: Just out of curiosity, is this because the "USB host boot mode" OTP bit is already set on the Pi 3B+, or is it because the bootcode in the BCM2837 doesn't check for this bit?
If it's the former, perhaps "The Raspberry Pi 3B+ and CM3+ support USB mass storage boot out of the box, because the host boot mode OTP bit is already set." would clarify this? (or perhaps that's unnecessary extra explanation?) :man_shrugging:

hardware/raspberrypi/bootmodes/msd.md
andrum99: You've just reviews a PR that has already merged, just to make you aware.

lurch: I realise that _now_, but I think that's because I was making comments at the same time that James was merging it :wink:

andrum99: > I realise that _now_, but I think that's because I was making comments at the same time that James was merging it 😉

Indeed

lurch:
Thanks for your contribution @Paul-Ver :+1:

PR #1365, opened 2019-12-25
Update windows.md

eitanchud:
New img flasher portable software for Windows. Has been built specifically for Raspberry Pi makers. Completely free :)

### DIFF
--- a/installation/installing-images/windows.md
+++ b/installation/installing-images/windows.md
@@ -1,6 +1,6 @@
 # Installing operating system images using Windows
 
-[balenaEtcher](https://www.balena.io/etcher/) is typically the easiest option for most users to write images to SD cards, so it is a good place to start. If you're looking for an alternative on Windows, you can use Win32DiskImager.
+[balenaEtcher](https://www.balena.io/etcher/) is typically the easiest option for most users to write images to SD cards, so it is a good place to start. If you're looking for an alternative on Windows, you can use Win32DiskImager or imgFlasher.
 
 ## balenaEtcher
 
@@ -19,6 +19,14 @@
 - In the device box, select the drive letter of the SD card. Be careful to select the correct drive: if you choose the wrong drive you could destroy the data on your computer's hard disk! If you are using an SD card slot in your computer, and can't see the drive in the Win32DiskImager window, try using an external SD adapter.
 - Click 'Write' and wait for the write to complete.
 - Exit the imager and eject the SD card.
+
+## Upswift imgFlasher
+
+- Download portable Windows version from [upswift.io](https://www.upswift.io/imgflasher/)
+- Run imgFlasher and choose an image or zip file
+- Choose SD card or USB drive
+- Click on 'Flash'
+- Wait until the flash is completed.
 
 ---
 


JamesH65:
When I tried to run this on WIndows, Defender brought up a big warning and prevented it from running. Presumably it needs to be signed to stop that happening. Maybe Win32DiskIMager does the same thing? Dunno, but it put me off from running it. Inclined to not merge this until the process is less scary.

eitanchud:
Hi James,
Thanks for your reply :)

I have checked the issue with Windows Defender, and it seems that you are correct, and a warning pop-up appears. The imgFlasher has been built using python3 and merged into 1 exe file using Pyinstaller. From what I have read, when using Pyinstaller, the Windows Defender automatically suspects that exe file. The only option I found to eliminate this warning is as soon as this exe will have enough users, Windows Defender will recognize it as legit and known software.

Purchasing a certificate that signs the software isn't going to help since the Defender still recognizes the Pyinstaller exe as a suspicious one.

I understand the problem here, therefore I have published the source code of the imgFlasher.exe.
You can check it at this repository: https://github.com/eitanchud/imgFlasher

andrum99:
Given there is already a working solution on Windows, this PR should probably be closed.

eitanchud:
Hi, although there are already 2 solutions, Etcher/win32diskimager, there is no reason to not add another open-source solution that has its own benefits and has been built and tested specifically for Raspberry Pis. 
Same as any other topic that has tools and they are all mentioned here at this documentation repository.

andrum99:
We should not be recommending solutions which cause security warnings on end-users' systems.

JamesH65:
If Windows Defender did not bring up this warning, I probably would be happy to link to it, but the warning is off putting, and most users will not have the skills/equipment to check the source code and build themselves.

lurch:
> add another open-source solution that has its own benefits

I don't use Windows so haven't tried this new imgFlasher myself, but what benefits does it offer over Etcher or Win32DiskImager?  (I don't mean to sound rude, I'm just curious)

eitanchud:
The main benefit of imgFlasher is its size. It’s one third in size of Etcher.

**James,** 
I have contacted Microsoft and uploaded imgFlasher.exe for them to review it and remove it from the Defender detection.
They have just answered that they have removed the detection. I guess the next defender malware update should be tomorrow, and after it is published, it should not pop that warning anymore.

Here is their answer:

defender microsoft answer

andrum99:
> The main benefit of imgFlasher is its size. It’s one third in size of Etcher.
> 
> **James,**
> I have contacted Microsoft and uploaded imgFlasher.exe for them to review it and remove it from the Defender detection.
> They have just answered that they have removed the detection. I guess the next defender malware update should be tomorrow, and after it is published, it should not pop that warning anymore.
> 
> Here is their answer:
> 
> defender microsoft answer

I didn't know you could do that! 😁 BalenaEtcher is huge as it has to download a private copy of some VM it uses to run, and there are privacy concerns since it "phones home" every time it runs. I would recommend we wait a while (like 2 weeks ,say) for folk's Windows Defender definitions to update, then include imgflasher in the docs.

lurch:
BalenaEtcher doesn't use a "private VM", it's huge because it runs on the Electron framework, and because it uses a huge stack of javascript libraries. The Etcher developers are currently trimming down that list https://github.com/balena-io/etcher/pull/3026 although there have been several complaints about how big it is e.g. https://github.com/balena-io/etcher/issues/2857 :wink:  :elephant:

eitanchud:
Alright, I've been in conversation with Microsoft regarding this in the past few weeks. I've been told that imgFlasher is added to the allowed programs in Windows Defender. It should stop alerting on that gradually in the following weeks depending on Defender updates downloaded by the Windows users.

I hope that now you could accept that merge request. Thanks!

eitanchud:
Hey :) many people told me that the application is working fine now, besides that, the code is completely open-source as I published here at GitHub and Microsoft has approved imgFlasher as a safe application.
I would be very happy if you could merge it :)

JamesH65:
OK, have merged this, but if I get any reports of problems it's likely to be removed without warning.

PR #1368, opened 2020-01-02
Edit the correct fstab file

peternewman:
Untested, but I think from reading it the current code is targeting the wrong one, or assuming it's run from the Pi that was booted via the network rather than the Pi serving the images over the network.

I may be entirely wrong @chunky but I assume this is what you intended?


JamesH65:
I have no idea whether this is correct or not. Needs input from the original author.

lurch:
ping @liamfraser

JamesH65:
I'm assuming this is correct. If not i am sure some will will let us know.

peternewman:
I think this also closes #1380 @JamesH65 ?

Issue #1372, opened 2020-01-05
Documentation refers to Pi 4 HDMI port 0 without specifying which that is.


HughR:
[HDMI Config](https://www.raspberrypi.org/documentation/configuration/hdmi-config.md) says that HDMI port 0 must be used for UltraHD at 60Hz.  But it isn't clear which of the two ports is 0.

I read the [schematic](https://www.raspberrypi.org/documentation/hardware/raspberrypi/schematics/rpi_SCH_4b_4p0_reduced.pdf) to find port 0 is labelled "j6".  This is silk-screened onto the circuit board.

I think that the clearest description is "the port closest to the USB C power connector".  This should be added to the HDMI Config page.

andrum99:
> I read the [schematic](https://www.raspberrypi.org/documentation/hardware/raspberrypi/schematics/rpi_SCH_4b_4p0_reduced.pdf) to find port 0 is labelled "j6". This is silk-screened onto the circuit board.

It's actually labelled HDMI0 on the silk screen. I've opened #1374 to address this.

JamesH65:
Fixed merged. Thanks all.

PR #1374, opened 2020-01-06
configuration/hdmi-config.md - clarify which port HDMI 0 is

andrum99:
See #1372


configuration/hdmi-config.md
```diff
@@ -2,7 +2,7 @@
 
 In the vast majority of cases, simply plugging your HDMI-equipped monitor into the Raspberry Pi using a standard HDMI cable will automatically lead to the Pi using the best resolution the monitor supports. On the Raspberry Pi 4 there are two micro HDMI ports, so you will need either one or two micro-HDMI-to-full-size-HDMI leads or adapters, depending on how many displays you wish to attach. You should connect any HDMI leads before turning on the Raspberry Pi.
```
lurch: Not directly related to this PR, but is it worth squeezing a sentence in here about the Raspberry Pi Zero using a mini HDMI port?  WDYT @JamesH65 ?

JamesH65: Good point - we call out the Pi4, we should also call out the Zero.

HughR:
Thanks for this change.

I don't see a label "HDMI 0" on my board.  How about s/labelled/known as/ ?

HughR:
> Thanks for this change.
> 
> I don't see a label "HDMI 0" on my board. How about s/labelled/known as/ ?

Oops.  Now I do see the label.  Faint and sideways, but there.

JamesH65:
> Thanks for this change.
> 
> I don't see a label "HDMI 0" on my board. How about s/labelled/known as/ ?

Odd, my board has HDMI0 silk screened next to the socket. They all should.

JamesH65:
Ah, OK, crossed in post. Glad you found it, was about to ask for a photo as they all should be clearly marked!

configuration/hdmi-config.md
andrum99: I would go the other way - if someone is buying a Pi it should be obvious if there is an adapter needed, from whatever webpage they bought it from. That way the docs stay simple.

andrum99: Although maybe it was me that put the bit about the adaptor for the Pi 4 in to start with lol.

lurch: Already added (and merged) in #1376 :wink:

andrum99: I know.

JamesH65: https://www.youtube.com/watch?v=4LaYn0541ro

PR #1382, opened 2020-01-08
Update building.md

psmdream:
The branch rpi-4.18.y doesn't have bcm2711_defconfig in arch/arm64/configs.  rpi-4.19.y has it.
It can make RP4 users confused.
Need to change from rpi-4.18.y to rpi-4.19.y.

### DIFF
--- a/linux/kernel/building.md
+++ b/linux/kernel/building.md
@@ -27,7 +27,7 @@ The `git clone` command above will download the current active branch (the one w
 To download a different branch (again with no history), use the `--branch` option:
 
 ```bash
-git clone --depth=1 --branch rpi-4.18.y https://github.com/raspberrypi/linux
+git clone --depth=1 --branch rpi-4.19.y https://github.com/raspberrypi/linux
 ```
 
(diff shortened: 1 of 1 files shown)


lurch:
ping @pelwell

lurch:
Hmm, this change may be warranted even _despite_ the RPi4, since https://github.com/raspberrypi/linux now defaults to the `rpi-4.19.y` branch, and https://www.raspberrypi.org/downloads/raspbian/ says Raspbian is now shipping with a 4.19 kernel.

pelwell:
This is just an example of how to nominate a specific branch - nowhere does it say that this is the branch to use. rpi-4.19.y is the default branch, so in a way there's little point also showing it as an alternate branch.

If anything, the specific branch name should be replaced with `<branch>`, with the next line saying "where `<branch>` is the name of the branch to download".

JamesH65:
Agree with @pelwell on this one, will update appropriately. Thanks to the OP for pointing out the issue.

PR #1384, opened 2020-01-08
Update net_tutorial.md

adduty:
Raspbian now supports NFSv4.1, which results in a much faster boot time. TCP is the standard protocol for NFSv4, and UDP is disabled by default for some NFS servers.

### DIFF
--- a/hardware/raspberrypi/bootmodes/net_tutorial.md
+++ b/hardware/raspberrypi/bootmodes/net_tutorial.md
@@ -239,14 +239,14 @@ sudo systemctl restart nfs-kernel-server
 Edit `/tftpboot/cmdline.txt` and from `root=` onwards, and replace it with:
 
 ```
-root=/dev/nfs nfsroot=10.42.0.211:/nfs/client1,vers=3 rw ip=dhcp rootwait elevator=deadline
+root=/dev/nfs nfsroot=10.42.0.211:/nfs/client1,vers=4.1,proto=tcp rw ip=dhcp rootwait elevator=deadline
 ```
 
 You should substitute the IP address here with the IP address you have noted down. Also remove any part of the command line starting with init=.
 
 Finally, edit `/nfs/client1/etc/fstab` and remove the `/dev/mmcblk0p1` and `p2` lines (only `proc` should be left). Then, add the boot partition back in:
 ```
-echo "10.42.0.211:/tftpboot /boot nfs defaults,vers=3 0 0" | sudo tee -a /etc/fstab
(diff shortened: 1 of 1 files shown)


andrum99:
Redhat and Fedora use TCP by default for NFS. Does debian do the same?

adduty:
I am not sure, but it seems that TCP is the preferred protocol for NFSv4. See section 'TCP for Transport' in this doc: https://www.snia.org/sites/default/files/SNIA_An_Overview_of_NFSv4-3_0.pdf.

andrum99:
If TCP is the preferred protocol for NFSv4 then why do you need to specify proto=tcp?

andrum99:
If you can put wireshark on your network to see what is going on traffic-wise without proto=TCP that would be helpful, or find out how this works. I'm inclined to say that the version is unnecessary as well, as we should probably just let Raspbian choose its own preferred version. I suspect it will choose NFSv4.1

adduty:
I have been monitoring the boot sequence from the TFTP/NFS server using tcpdump, and when vers=4.1 is specified without proto=tcp, I see no NFS traffic (sudo tcpdump port nfs or port rpcbind or port rpc or port sunrpc). If I do not specify a vers= or proto= option, the Pi defaults to using NFSv3 over UDP.

andrum99:
NFSv4.1 doesn't use the old NFS port (port 111), it uses port 2049 so you're not going to see any traffic if you're only looking at port 111, rpcbind, rpc and sunrpc. I suspect tcpdump is calling port 111 nfs.

So we definitely need to specify vers=4.1, but the jury is still out on whether proto=TCP is required.

adduty:
Yes, I know, but my tcpdump filter catches all NFSv3 and NFSv4 traffic (2049 will show from 'port nfs'). Using that same filter (sudo tcpdump port nfs or port rpcbind or port rpc or port sunrpc), I see the traffic when the Pi successfully boots using vers=4.1,proto=tcp.

andrum99:
Good stuff. So your original suggestion has been shown to be correct.

JamesH65:
So, we are happy with this change? But it does need to be rebased as their are conflicts. If @adduty can do that i can merge if everyone is happy.

adduty:
I think I have rebased, but I don't actually use git very much, so I'm not sure if I did it correctly.

andrum99:
> I think I have rebased, but I don't actually use git very much, so I'm not sure if I did it correctly.

Github says there are no conflicts, so it seems you did 👍

PR #1391, opened 2020-01-25
building.md: Add extra packages to be able to cross-compile the kernel

GeVanCo:
I started a fresh Linux Mint 19 XFCE installation and then wanted to build the Raspberry Pi kernel.
Following the documentation, I was missing two packages:
- libc6-dev (build error because sys/types.h was missing)
- libncurses5-dev (needed when you run make menuconfig...)

Adding those two missing packages solved the issue and I was able to build the kernel without problems.

### DIFF
--- a/configuration/wireless/access-point.md
+++ b/configuration/wireless/access-point.md
@@ -135,8 +135,8 @@ sudo systemctl start hostapd
 Do a quick check of their status to ensure they are active and running:
 
 ```
-sudo systemctl status hostapd
-sudo systemctl status dnsmasq
+systemctl status hostapd
+systemctl status dnsmasq
 ```
 
 ### Add routing and masquerade
--- a/configuration/wireless/access-point.md
+++ b/configuration/wireless/access-point.md
@@ -135,8 +135,8 @@ sudo systemctl start hostapd
 Do a quick check of their status to ensure they are active and running:
 
 ```
-systemctl status hostapd
-systemctl status dnsmasq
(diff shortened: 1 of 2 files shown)


JamesH65:
@pelwell @popcornmix Does this look sane to you guys? Seems OK to me as even if not required on some systems, it's not going to break anything.

pelwell:
It looks fine - libc6-dev is installed by default on Raspbian (non-Lite - I've not tested Lite), and libncurses5-dev pulls in an additional 6kB library (plus the Debian package overhead which is probably larger).

GeVanCo:
Thanks for the comments.  Just to be clear: I didn't build on a RPI but on a VM on Windows using Linux Mint 19 XFCE.  And for that, I needed those two extra libs.

pelwell:
Yes, I understood - I was explaining how small the impact would be on a Pi.

popcornmix:
Looks fine.

GeVanCo:
Added more changes to the documentation.  This time in the file `access-point.md`.  There's no need to be `sudo` when running `systemctl status <service>`.  Can be done without.

lurch:
> This time in the file access-point.md

Could you please make that change in a separate PR, rather than making multiple unrelated changes in the same PR? Thanks.

GeVanCo:
Advised to revert my latest commit since "unrelated".  Hope this went well, first time I'm doing such git action...

JamesH65:
Thanks.

Issue #1397, opened 2020-02-11
Rpi Wireless Access Point bridge setup documentation improvement


jpcornil-git:
Reference: https://github.com/raspberrypi/documentation/blob/master/configuration/wireless/access-point.md (bridge section)

- Remove the following section (this step is not required; it just creates a bridge using the command line but related configuration is not persistent nor a preliminary step for anything):

---
Add a new bridge, which in this case is called br0.

`sudo brctl addbr br0`

Connect the network ports. In this case, connect eth0 to the bridge br0.

`sudo brctl addif br0 eth0`

---

- Given the above you can also remove the bridge-utils installation and replace

---
`sudo apt install hostapd bridge-utils`

---
with:

---
`sudo apt install hostapd`

---

- In order to handle systems running/conflict with networking.service, I suggest to replace:

---
Finally, restart systemd-networkd:

`sudo systemctl restart systemd-networkd`

You can also use the brctl tool to verify that a bridge br0 has been created.

---
with:

---
Finally, (re)start systemd-networkd:

`sudo systemctl disable networking.service`
`sudo systemctl enable systemd-networkd`
`sudo systemctl restart systemd-networkd`

You can also use the bridge tool to verify that a bridge br0 has been created.

`bridge -d link`

---

JamesH65:
This would be much better submitted as a pull request.

jpcornil-git:
Indeed/agree, just created https://github.com/raspberrypi/documentation/pull/1398

lurch:
Probably better to wait until after #1333 has been merged?

jpcornil-git:
Sure, not a big deal to rework this on top of it

jpcornil-git:
Closing old issue (all fixed in https://www.raspberrypi.org/documentation/configuration/wireless/access-point-bridged.md)

PR #1398, opened 2020-02-12
Correction/improvement of the bridge section

jpcornil-git:
PR related to https://github.com/raspberrypi/documentation/issues/1397#

### DIFF
--- a/configuration/wireless/access-point.md
+++ b/configuration/wireless/access-point.md
@@ -184,7 +184,7 @@ One common use of the Raspberry Pi as an access point is to provide wireless con
 To do this, a 'bridge' needs to put in place between the wireless device and the Ethernet device on the access point Raspberry Pi. This bridge will pass all traffic between the two interfaces. Install the following packages to enable the access point setup and bridging.
 
 ```
-sudo apt install hostapd bridge-utils
+sudo apt install hostapd
 ```
 
 Since the configuration files are not ready yet, turn the new software off as follows:
@@ -199,18 +199,6 @@ sudo nano /etc/dhcpcd.conf
 ```
 
 Add `denyinterfaces wlan0` and `denyinterfaces eth0` to the end of the file (but above any other added `interface` lines) and save the file.
-
-Add a new bridge, which in this case is called `br0`.
-
-```
-sudo brctl addbr br0
-```
-
-Connect the network ports. In this case, connect `eth0` to the bridge `br0`.
-
-```
-sudo brctl addif br0 eth0
-```
 
 Now the interfaces file needs to be edited to adjust the various devices to work with bridging. To make this work with the newer systemd configuration options, you'll need to create a set of network configuration files.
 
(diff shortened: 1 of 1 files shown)


JamesH65:
@jpcornil-git If this PR is still relevant given today's merge of  #1333, can you redo it to match the changes please?

JamesH65:
@epoch1970 I guess this one can be closed as well???!!

epoch1970:
Absolutely. it is covered by #1333

jpcornil-git:
Much better, I was also suggesting to disable networking service before
enabling networkd (should only be one managing network)

sudo systemctl disable networking.service
sudo systemctl enable systemd-networkd

Cheers
jpc

epoch1970:
There is no _need_ to disable networking service, in order to get a working setup.
So I would leave it alone.

JamesH65:
Thanks all!

jpcornil-git:
I would recommend to either:
- mention that no network interfaces should be configured in
/etc/network/interfaces when systemd-networkd is used (avoid conflicting
dhcp clients, race conditions between both network mngt systems, ...), i.e.
only use /etc/dhcpcd.conf
- disable legacy networking.services (avoid issues related to mixed
configurations by construction)

Making that explicit will ring a bell to whoever has an issue after
applying something seen on internet based on legacy networking.services (or
even proactively prevent it)

Cheers,
jpc

Le ven. 24 avr. 2020 à 17:53, James Hughes <notifications@github.com> a
écrit :

epoch1970:
I think I’ll agree to disagree.

Yes ifupdown must die.
No it is not the job of this guide to right every wrong. Its job is to work, out of the box, with the advertised Raspbian version. I think it does that.
Otherwise we’ll need a paragraph about how not to shoot yourself in the foot by making resolv.conf immutable, having ip=1.2.3.4 in cmdline.txt and all kinds of folklore.

Cheers

jpcornil-git:
Agree about the 2nd paragraph :-)

just saying that:

sudo systemctl enable systemd-networkd

<https://github.com/raspberrypi/documentation/blob/master/configuration/wireless/access-point-bridged.md#define-the-bridge-device-ip-configuration>is
changing the box networking behavior/create a slipery slope and that:

sudo systemctl disable networking.service

<https://github.com/raspberrypi/documentation/blob/master/configuration/wireless/access-point-bridged.md#define-the-bridge-device-ip-configuration>(or
a note about possible conflicts) fixes potential related/vicious issues.

Feel free to continue to disagree as well/nothing wrong with that, I said
what I had to say/won't insist anymore :-)

Cheers,
jpc

epoch1970:
First of all I have no authority here, your vote is a good as mine.

I could follow you on the idea that "systemd-networkd enable" are words none should ever utter :) 
Unfortunately, as Raspbian is today, it is the only reasonable option because dhcpcd does not handle bridge creation. It can be done, I have a prototype script lying somewhere in the forums, it's not perfect and anyways the destiny of dhcpcd in Raspbian is better to be discussed between its author and RPT.

So ATM the least of 2 evils (ifupdown, systemd) is to enable networkd and have it create the bridge. Personally I hope dhcpcd will live on and prevail in future Raspbian releases. I think it's very good software and a great asset to Raspbian.

jpcornil-git:
Legacy-only networking can be used as well but a bit more tricky/scattered
to setup when a wifi ifce is involved, e.g.
https://wiki.debian.org/BridgeNetworkConnections#Bridging_with_a_wireless_NIC

Cheers
jpc

lurch:
ping @XECDesign in case he has anything to add to the discussion here? :shrug:

XECDesign:
Nothing useful, unfortunately. In a display of cognitive dissonance, I agree with both points of view.

Yes, you should avoid having 3 different network managers. Yes, if the user hasn't configured ifupdown it won't be used anyway. Yes, there will be people who followed random guides online from 2010 or have a cargo cult setup put together from random forum posts and there's no way this guide can undo all the damage.

I'd lean towards the notion that disabling ifupdown does no harm and is technically the right thing to do.

PR #1399, opened 2020-02-12
Update toolchain path for cross-compiling

### LINKED ISSUE #1388: Cross-compiling Linux 5.4 requires an updated GCC
https://github.com/raspberrypi/documentation/blob/master/linux/kernel/building.md
Cross compiling references
~/tools/arm-bcm2708/gcc-linaro-arm-linux-gnueabihf-raspbian/bin or ~/tools/arm-bcm2708/gcc-linaro-arm-linux-gnueabihf-raspbian-x64/bin

Those are gcc 4,8 and won't work for the 5.4 kernel.
It should now be ```~/tools/arm-bcm2708/arm-linux-gnueabihf/bin``` which symlinks to the latest toolchain (currently should be 4.9.3 in ~/tools/arm-bcm2708/arm-linux-gnueabihf/bin).

@pelwell please confirm I'm correct on these details.

6by9:
Fixes #1388

### DIFF
--- a/linux/kernel/building.md
+++ b/linux/kernel/building.md
@@ -101,7 +101,7 @@ source ~/.bashrc
 ```
 If you are on a 64-bit host system, you should use:
 ```bash
-echo PATH=\$PATH:~/tools/arm-bcm2708/gcc-linaro-arm-linux-gnueabihf-raspbian-x64/bin >> ~/.bashrc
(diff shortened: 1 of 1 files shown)


lurch:
Should we remove the section about 32-bit hosts, if the 32-bit toolchain isn't new enough, as per #1388 ? :man_shrugging:

JamesH65:
This was fixed by 1413

PR #1402, opened 2020-02-14
Update access-point.md

asap2004:
the line driver=nl80211 causes hostapd to crash. not needed anymore


lurch:
Probably better to wait until after #1333 has been merged?

JamesH65:
#1333 is now merged, so this PR now has a conflict. @epoch1970 Is this change still necessary or have I missed something?

epoch1970:
You can safely close this. That line is gone in both routed and bridged documents.
(and I doubt it caused crashing, at least w/ the built-in wifi interface)

PR #1403, opened 2020-02-19
Update cmdline-txt.md

JamesH65:
Added some detail on setting video stuff (e.g. rotate) via the command line.

### DIFF
--- a/configuration/cmdline-txt.md
+++ b/configuration/cmdline-txt.md
@@ -28,11 +28,20 @@ The firmware automatically adds a preferred resolution and overscan settings via
 
 ```video=HDMI-A-1:1920x1080M@60,margin_left=0,margin_right=0,margin_top=0,margin_bottom=0```
 
-This default entry can be modified by duplicating the entry above manually in /boot/cmdline.txt and making required changes to the margin parameters. In addition, it is possible to add rotation and reflect parameters as documented in the standard [Linux framebuffer documentation](https://github.com/raspberrypi/linux/blob/rpi-4.19.y/Documentation/fb/modedb.txt)
+This default entry can be modified by duplicating the entry above manually in /boot/cmdline.txt and making required changes to the margin parameters. In addition, it is possible to add rotation and reflect parameters as documented in the standard [Linux framebuffer documentation](https://github.com/raspberrypi/linux/blob/rpi-4.19.y/Documentation/fb/modedb.txt). By default the `margin_*` options are set from the `overscan` entries in config.txt, if present. The firmware can be prevented from making any KMS specific changes to the command line by adding `disable_fw_kms_setup=1` to `config.txt`
 
-e.g. 
+An example entry may be as follows:
 
 ```video=HDMI-A-1:1920x1080M@60,margin_left=0,margin_right=0,margin_top=0,margin_bottom=0,rotate=90,reflect_x```
+
+Possible options for the display type, the first part of the `video=` entry,  are as follows:
+
+| video option  | Display |
+|:---:|:---|
+| HDMI-A-1      | HDMI 1 (0 on silkscreen of Pi4B) |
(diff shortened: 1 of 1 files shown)


JamesH65:
I'll update the config.txt video page which talks about rotation to point to this page for those using (F)KMS, if people think that this is the right approach.

configuration/cmdline-txt.md
```diff
@@ -20,6 +20,17 @@ There are many kernel command line parameters, some of which are defined by the
  - elevator: specifies the I/O scheduler to use. `elevator=deadline` means the kernel imposes a deadline on all I/O operations to prevent request starvation.
  - quiet: sets the default kernel log level to `KERN_WARNING`, which suppresses all but very serious log messages during boot.
 
+#### Display Entries in FKMS and KMS modes
+
+The firmware already adds a preferred resolution and overscan settings via an entry such as:
```
lurch: already -> automatically ? :man_shrugging:

configuration/cmdline-txt.md
```diff
@@ -20,6 +20,17 @@ There are many kernel command line parameters, some of which are defined by the
...
+
+```video=HDMI-A-1:1920x1080M@60,margin_left=0,margin_right=0,margin_top=0,margin_bottom=0```
+
+This default entry can be modified by duplicating the line above manually in /boot/cmdline.txt and making required changes to the margin parameters. In addition, it is possible to add rotation and reflect parameters as documented in the standard [Linux framebuffer documentation](https://github.com/raspberrypi/linux/blob/rpi-4.19.y/Documentation/fb/modedb.txt)
```
lurch: Might be worth changing "line above" to "entry above", as people might think "line" means it has to be on a separate line in `cmdline.txt` ?

Might be worth adding some clarification on if / how this relates to the margin settings in `config.txt`? (are the config.txt entries only used when FKMS isn't being used? Does the firmware copy the margin settings from config.txt into the automatic cmdline.txt entry? etc.) ping @6by9 

Perhaps it's worth explaining that if `video=...` _isn't_ explicitly set in `cmdline.txt`, then the automatic firmware-supplied entry can be read in Raspbian by doing `cat /proc/cmdline`

Is it necessary to explain what the `HDMI-A-1` part means and/or what the available options here are?

configuration/cmdline-txt.md
```diff
@@ -4,7 +4,7 @@ The Linux kernel accepts a command line of parameters during boot. On the Raspbe
 ```
 sudo nano /boot/cmdline.txt
 ```
-Note that we have to use `sudo` to edit anything in the boot partition.
+Note that we have to use `sudo` to edit anything in the boot partition, and that all parameters in `cmdline.txt` must be on the same line (no carriage returns)
```
lurch: Missing full-stop at the end of the sentence. :stuck_out_tongue_winking_eye:

configuration/cmdline-txt.md
JamesH65: AIUI, those margin entries come from any config.txt entries for overscan. Have added an extra sentence in first paragraph to explain use of cat /proc/cmdline. Not sure about describing the HDMI-A bit. It is in the linked Linux docs.

6by9: Displays are HDMI-A-1, HDMI-A-2, DSI-1 (covers DPI as well), and Composite-1.
```HDMI-A``` is because it is an HDMI type A connector. Technically it's a type D on Pi4, and type C on Pi0, but electrically they are identical. Type B isn't, and is the alternative from https://github.com/raspberrypi/linux/blob/rpi-4.19.y/drivers/gpu/drm/drm_connector.c#L71.

Margin properties are generated by the firmware based on the config.txt overscan_ properties.
The firmware ```video=``` can be disabled via ```disable_fw_kms_setup=1```.

JamesH65: OK, have added a table and rejigged a few lines.

PR #1406, opened 2020-02-23
Specify which machine to issue command from

Sefton419:
This was the only point of confusion while following these instructions


JamesH65:
Thanks! Good idea.

Issue #1407, opened 2020-02-24
Kernel build instructions don't advice to change LOCALVERSION


olifre:
I walked through the instructions (`building.md`) just now and wonder if the instructions should advice to set / hint on setting:
```
CONFIG_LOCALVERSION="-v7l-MY_CUSTOM_KERNEL"
```
If this is not done, existing modules from a distro installation in `/lib/modules/` might be overwritten, at least when a release tag is used as base for the self-compiled kernel. This may come as a surprise for users compiling their kernel for the first time. 

It might also be useful to hint on how to "go back" to the distro kernel (is it just reinstallation of `raspberrypi-kernel/`)?

lurch:
ping @pelwell and @popcornmix

pelwell:
It sounds like a reasonable suggestion - do you feel like submitting a Pull Request using your recent experience?

olifre:
@pelwell Yes, can do, just wanted to clarify first whether the change is considered reasonable. 
I'll prepare a PR then soon-ish. 

About the "how to go back to the distro kernel", do you have experience on this already?

pelwell:
I'm not the best person to comment since I either build my own or use the rpi-update builds, but `sudo apt install --reinstall raspberrypi-bootloader` should work.

olifre:
@pelwell Thanks. I'll then submit a PR for the first point (`LOCALVERSION`) in the next minutes, and as soon as I rollback to distro kernel, will try `sudo apt install --reinstall raspberrypi-bootloader`. If that also works well, I'll do another PR with that one.

JamesH65:
Fixed by #1408

olifre:
@pelwell and others:
I'm not finally in the reverting procedure. I found that the following steps are needed to get back to the previous state:
```
sudo apt install --reinstall raspberrypi-bootloader
sudo apt install --reinstall raspberrypi-kernel
rm /boot/bcm2835-rpi-*
rm /boot/bcm2837-rpi-3-b*
rm /boot/bcm2836-rpi-2-b.dtb
rm -rf /lib/modules/*-${LOCALVERSION}
```
This seems to catch everything (obviously, `LOCALVERSION` is also needed here). 

Dou you believe it is worthwhile documenting at the same place? As part of the same PR?

lurch:
#1408 has been merged, so closing this.

PR #1408, opened 2020-02-24
building.md: Add hint to adjust CONFIG_LOCALVERSION (#1407).

olifre:
Adjusting LOCALVERSION ensures the custom kernel
can be identified from uname and distro modules
are not overwritten.

### DIFF
--- a/linux/kernel/building.md
+++ b/linux/kernel/building.md
@@ -64,7 +64,9 @@ KERNEL=kernel7l
 make bcm2711_defconfig
 ```
 
-After this is done, you will likely want to adjust the `LOCALVERSION` to ensure your new kernel does not get the same version string as the upstream kernel. This both clarifies you are running your own kernel in the output of `uname` and ensures existing modules in `/lib/modules` are not overwritten.
+#### Customising the Kernel version using LOCALVERSION
+
+In addition to your kernel configuration changes, you may wish to adjust the `LOCALVERSION` to ensure your new kernel does not receive the same version string as the upstream kernel. This both clarifies you are running your own kernel in the output of `uname` and ensures existing modules in `/lib/modules` are not overwritten.
 
 To do so, change the following line in `.config`:
 ```
--- a/linux/kernel/building.md
+++ b/linux/kernel/building.md
@@ -36,11 +36,14 @@ Refer to the [original GitHub repository](https://github.com/raspberrypi/linux) 
 
 ### Kernel configuration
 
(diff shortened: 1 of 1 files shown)


olifre:
@JamesH65 Thanks from my side, this extra headline clarifies it better that there's an extra paragraph :+1: .

lurch:
@pelwell Should we be recommending editing `.config` directly, rather than using `menuconfig`?  (Is this common practice? :man_shrugging: )

olifre:
@lurch My reasoning for describing how to edit the file was that using `menuconfig` requires installing `libncurses5-dev` which is described in `configuring.md`, so people only reading through `building.md` will not have it installed.

lurch:
So (if I've understood things correctly), I guess if you use both LOCALVERSION (which keeps the modules in a separate directory) and you _also_ use the suggestion from the bottom of this page about using a different filename for the kernel, then reverting to the "stock" kernel would simply be a matter of removing "kernel=kernel-myconfig.img" from `config.txt`? :slightly_smiling_face:

olifre:
You'll still be overwriting the device tree overlays, but that probably does not harm if the version you built is not too distant from the "stock" kernel. Leaving to others if that's a good idea ;-).

lurch:
@JamesH65 This is OK with me, if @pelwell is happy :slightly_smiling_face:

linux/kernel/building.md
```diff
@@ -64,7 +64,9 @@ KERNEL=kernel7l
 make bcm2711_defconfig
 ```
 
-After this is done, you will likely want to adjust the `LOCALVERSION` to ensure your new kernel does not get the same version string as the upstream kernel. This both clarifies you are running your own kernel in the output of `uname` and ensures existing modules in `/lib/modules` are not overwritten.
+#### Customising the Kernel version using LOCALVERSION
```
pelwell: This would make more sense with one fewer `#`, since it doesn't belong to the list formed by the items above.

lurch: But OTOH it probably *does* belong under the `### Kernel configuration` section? :man_shrugging: 
Maybe we need a new `#### Apply the default configuration` header, and then move the "Raspberry Pi 4" etc. headers another level down? How many levels of subheadings does https://www.raspberrypi.org/documentation/ support?

popcornmix:
No objections to this change.

linux/kernel/building.md
pelwell: (Editing comments doesn't trigger notifications...)
That looks OK to me.

JamesH65:
Thanks, sorry about the delay.

PR #1413, opened 2020-02-25
Update cross compile toolchain path

JamesH65:
Update cross compile toolchain path

### DIFF
--- a/linux/kernel/building.md
+++ b/linux/kernel/building.md
@@ -100,7 +100,7 @@ Updating the $PATH environment variable makes the system aware of file locations
 echo PATH=\$PATH:~/tools/arm-bcm2708/arm-linux-gnueabihf/bin >> ~/.bashrc
 source ~/.bashrc
 ```
-If you are using a 32 bit Operating system (for example, our Raspberry Pi Desktop for PC), then you may need to install an addition set of libraries.
+If you are using a 32-bit operating system (for example, our Raspberry Pi Desktop for PC), then you may need to install an additional set of libraries:
 
 `sudo apt install zlib1g-dev:amd64`
 
--- a/linux/kernel/building.md
+++ b/linux/kernel/building.md
@@ -94,16 +94,13 @@ Use the following command to download the toolchain to the home folder:
 git clone https://github.com/raspberrypi/tools ~/tools
 ```
 
-Updating the $PATH environment variable makes the system aware of file locations needed for cross-compilation. On a 32-bit host system you can update and reload it using:
+Updating the $PATH environment variable makes the system aware of file locations needed for cross-compilation. 
+
 ```bash
-echo PATH=\$PATH:~/tools/arm-bcm2708/gcc-linaro-arm-linux-gnueabihf-raspbian/bin >> ~/.bashrc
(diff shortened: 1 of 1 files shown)


JamesH65:
Fixes #1388

6by9:
OK, thanks for fixing up lurch's query on #1399.

The only niggle being that our x86 "Raspberry Pi Desktop" distro is 32 bit, and therefore can't cross compile a kernel. Do we care?

pelwell:
The commit message needs to explain the reason for the change, and particularly why we are removing reference to 32- and 64-bit systems.

lurch:
Do we need to explicitly mention that this will no longer work on a 32-bit host / that it's for 64-bit hosts only?

lurch:
Or, looking at things from another angle, is it possible to get an updated 32-bit cross-compile toolchain? :man_shrugging:

JamesH65:
Now it's been pointed out I am slightly concerned with only providing a 64bit tool chain when we only provide a 32bit OS (Desktop on x86). Any way round this?

JamesH65:
Apparently we have multiarch so this should still work on x86 Desktop.

XECDesign:
Seems to work as long as `zlib1g-dev:amd64` is installed - that drags in the other necessary dependencies as well.

JamesH65:
Thanks  @XECDesign, I think that I'll add that specifically to the docs, or was it obvious that is what needed installing?

XECDesign:
It wasn't obvious because libc6:amd64 isn't installed by default. When you try to run a binary without libc present, it will give you a misleading error about not being able to find the file you're trying to run. If you just install libc, then it will fall over later when trying to use zlib with a more sensible error message.

In general, I don't think multiarch is something many people have experience with, so it's best not to assume anything is obvious.

lurch:
> I don't think multiarch is something many people have experience with

I didn't even realise multiarch allowed you to run 64-bit binaries on a 32-bit host? I assumed it only allowed a 64-bit host to also run 32-bit binaries.

XECDesign:
The 32-bit on 64-bit scenario is just more common. As long as the kernel supports the architecture you want to run, it's fine.

Our Debian + RPD setup is a bit unusual because we try to support very old hardware (down to i686 without PAE, if I recall correctly), but if a 64-bit capable system is detected, it will use a 64-bit kernel. Initially it was done that way to be able to support systems running 64-bit UEFI, but it also makes multiarch work.

lurch:
Ahhh, neat. So I guess the 64bit cross-compilers won't work with multiarch on x86 Raspbian Desktop **only if** it has been installed onto a i686 machine? I guess that's pretty unlikely, and anybody running such old hardware probably isn't interested in cross-compiling anyway :wink:

XECDesign:
> So I guess the 64bit cross-compilers won't work with multiarch on x86 Raspbian Desktop **only if** it has been installed onto a i686 machine? I guess that's pretty unlikely, and anybody running such old hardware probably isn't interested in cross-compiling anyway

Exactly

JamesH65:
I've just made a small change to add the library need on Raspberry Pi desktop for PC. If everyone is happy with that I will merge this.

pelwell:
LGTM.

PR #1415, opened 2020-02-25
otpbits.md: revision number -> revision code

andrum99:
Minor nit - for consistency. I suspect 'revision number' was their original name, when they were purely sequential.

Also mention this register stores other things too - currently 3 features not related to the revision code. Not sure if this is really needed, but it improves the 'discoverability' of this information slightly.


lurch:
I guess this needs to be merged _after_ #1370 ?

andrum99:
> I guess this needs to be merged _after_ #1370 ?

Good point - yes, assuming the footnote is deemed a worthwhile addition, otherwise feel free to make this PR a one word change (number -> code) and push

JamesH65:
Thanks - sorry about the delay - only just noticed this one.

PR #1417, opened 2020-02-26
Update boot.md to add uart_2ndstage information

JamesH65:


configuration/config-txt/boot.md
```diff
@@ -99,6 +99,12 @@ If `disable_splash` is set to `1`, the rainbow splash screen will not be shown o
 
 On the Raspberry Pi 4B, if this value is set to `0` then the interrupts will be routed to the ARM cores using the legacy interrupt controller, rather than via the GIC-400. The default value is `1`.
 
+## uart_2ndstage
+
+Setting `uart_2ndstage=1` causes the second-stage loader (`bootcode.bin` on devices prior to the Raspberry Pi 4, or the boot code in the  EEPROM for Raspberry Pi 4 devices) and the main firmware (`start*.elf`) to output diagnostic information to UART0.
```
lurch: Do we need to specify the baudrate here? Or should people just assume that it uses the same 115200 default that the Linux serial console uses?

configuration/config-txt/boot.md
```diff
@@ -99,6 +99,12 @@ If `disable_splash` is set to `1`, the rainbow splash screen will not be shown o
 
 On the Raspberry Pi 4B, if this value is set to `0` then the interrupts will be routed to the ARM cores using the legacy interrupt controller, rather than via the GIC-400. The default value is `1`.
 
+## uart_2ndstage
+
+Setting `uart_2ndstage=1` causes the second-stage loader (`bootcode.bin` on devices prior to the Raspberry Pi 4, or the boot code in the  EEPROM for Raspberry Pi 4 devices) and the main firmware (`start*.elf`) to output diagnostic information to UART0. 
+
+Be aware that output is likely to interfere with Bluetooth operation unless it is disabled (`dtoverlay=disable-bt`) or switched to the other UART (`dtoverlay=miniuart-bt`), and if the UART is accessed simultaneously to output from Linux then data loss can occur leading to corrupted output.  This feature should only be required when trying to diagnose a early boot loading problem.
```
lurch: Worth adding a link to https://github.com/raspberrypi/documentation/blob/master/configuration/uart.md ?

> if the UART is accessed simultaneously to output from Linux

Hmm, wouldn't the bootloader have finished printing stuff to UART by the time Linux has loaded (and has access to the UART) ?

pelwell: > Hmm, wouldn't the bootloader have finished printing stuff to UART by the time Linux has loaded (and has access to the UART) ?

Not necessarily - the firmware is constantly running and could choose to output error messages at any time.

configuration/config-txt/boot.md
JamesH65: I think that is more detail than needed for this particular reference.

configuration/config-txt/boot.md
lurch: Ahhh. In that case should this emphasise that if `uart_2ndstage=1` is used, the Linux serial console should be disabled? (since I assume people using the UART to debug bootloader problems will probably also be using the UART to debug kernel problems?)

pelwell: No - a warning is probably sufficient.

JamesH65: I reckon anyone who uses this option, given the warnings present, will know what they are doing.

PR #1425, opened 2020-03-03
Installation from images

pgramme:
For user-friendliness, add instructions for booting and changing password

### DIFF
--- a/installation/installing-images/README.md
+++ b/installation/installing-images/README.md
@@ -48,4 +48,4 @@ You can now insert the SD card into the Raspberry Pi and power it up.
 
 For the official Raspbian OS, if you need to manually log in, the default user name is `pi`, with password `raspberry`. Remember the default keyboard layout is set to UK.
 
-You should change the default password straight away to ensure your Raspberry Pi is secure.
+You should change the default password straight away to ensure your Raspberry Pi is [secure](../../configuration/security.md).
--- a/installation/installing-images/README.md
+++ b/installation/installing-images/README.md
@@ -46,6 +46,6 @@ For more advanced control of this process, see our system-specific guides:
 
 You can now insert the SD card into the Raspberry Pi and power it up. 
 
-For the official Raspbian OS, if you need to enter it, the default user name is `pi`, with password `raspberry`. Remember the default keyboard layout is set to UK.
+For the official Raspbian OS, if you need to manually log in, the default user name is `pi`, with password `raspberry`. Remember the default keyboard layout is set to UK.
 
-You should change the default password straight away by running `passwd`, [as explained here](../../linux/usage/users.md)
+You should change the default password straight away to ensure your Raspberry Pi is secure.
--- a/installation/installing-images/README.md
+++ b/installation/installing-images/README.md
@@ -42,10 +42,10 @@ For more advanced control of this process, see our system-specific guides:
 - [Chrome OS](chromeos.md)
 
 
-## Boot your new Raspbian
+## Boot your new OS
 
-You can now insert the SD card into the Pi. The default user name is `pi`, with password `raspberry`.
+You can now insert the SD card into the Raspberry Pi and power it up. 
 
-**Note**: Remember the default keyboard layout is qwerty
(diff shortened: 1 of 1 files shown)


JamesH65:
@lurch Seems like a useful addition, might need some tiny tweaks to phrasing. Thoughts?

lurch:
This page is a generic "installing images", whereas the added instructions are Raspbian-specific.
This seems to touch on https://github.com/raspberrypi/documentation/blob/master/raspbian/README.md and https://github.com/raspberrypi/documentation/blob/master/linux/usage/users.md and https://github.com/raspberrypi/documentation/blob/master/configuration/security.md :question: 

BTW, IIRC the default keyboard in Raspbian won't be qwerty if you've installed Raspbian via NOOBS, and you selected your correct keyboard in the NOOBS interface? @XECDesign

JamesH65:
Could remove the Raspbian reference and replace with OS. Add a rider that the password etc relate to Raspbian?

JamesH65:
@lurch Made a few minor changes, any thoughts?

installation/installing-images/README.md
```diff
@@ -40,3 +40,12 @@ For more advanced control of this process, see our system-specific guides:
 - [Mac OS](mac.md)
 - [Windows](windows.md)
 - [Chrome OS](chromeos.md)
+
+
+## Boot your new OS
+
+You can now insert the SD card into the Raspberry Pi and power it up. 
+
+For the official Raspbian OS, if you need to enter it, the default user name is `pi`, with password `raspberry`. Remember the default keyboard layout is set to UK.
```
lurch, line 49: Perhaps "if you need to enter it" should be replaced with "if you need to manually login" ? (IIRC most new installs just boot straight to a logged-in desktop?)

installation/installing-images/README.md
```diff
@@ -40,3 +40,12 @@ For more advanced control of this process, see our system-specific guides:
...
+
+You should change the default password straight away by running `passwd`, [as explained here](../../linux/usage/users.md)
```
lurch, line 51: Isn't this already a suggested step in the "first run wizard"? ping @spl237

spl237: Yes, it is part of the wizard; no need to do it from the terminal.

JamesH65: What about Raspbian lite?

spl237: Lite doesn't have the wizard, but it would seem to me that trying to use the same documentation for installing the desktop version as for lite is not the greatest idea...

JamesH65: We've don't tend to distinguish, and TBH, there's very little documentation on the desktop in this repo at all.

spl237: In that case, probably best to just recommend people change the password without being specific as to how...

installation/installing-images/README.md
```diff
@@ -40,3 +40,12 @@ For more advanced control of this process, see our system-specific guides:
...
 - [Chrome OS](chromeos.md)
+
+
+## Boot your new OS
+
+You can now insert the SD card into the Raspberry Pi and power it up. 
+
+For the official Raspbian OS, if you need to manually log in, the default user name is `pi`, with password `raspberry`. Remember the default keyboard layout is set to UK.
+
+You should change the default password straight away to ensure your Raspberry Pi is secure.
```
andrum99, line 51: Ubuntu for the Pi forces the user to change the primary user password at first log on. Is that something Raspbian should be doing? (Not a question for the documentation, I know).

lurch: For the audience that Raspbian is aimed at, I suspect enforcing a mandatory password change might cause more problems than it "fixes"? See also Simon's earlier comments above...

lurch: Perhaps it's worth linking the word "secure" in the sentence above to https://github.com/raspberrypi/documentation/blob/master/configuration/security.md ? :man_shrugging:

JamesH65: Ah, good idea.

JamesH65:
Thanks to the original writer!

pgramme:
Thanks for your improvements (and sorry for the late reaction)

PR #1435, opened 2020-03-16
Use Raspberry Pi Imager for recovery

timg236:
Direct people towards the Raspberry Pi Image for creating recovery sd cards. I've created an internal ticket to update the text on the downloads page accordingly.

### DIFF
--- a/hardware/raspberrypi/booteeprom.md
+++ b/hardware/raspberrypi/booteeprom.md
@@ -19,7 +19,7 @@ To check that the bootloader is working correctly, turn off the power, unplug ev
 
 ## Recovery image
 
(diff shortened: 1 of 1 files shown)


timg236:
Thanks!

lurch:
Will the downloads page still be offering a link (even if indirect) to the recovery image, for the benefit of anyone unable to run the Raspberry Pi Imager tool?

timg236:
Yes, that's what I asked, although eventually I think all binaries will be moved to Github as releases and raspberrypi.org will just point to that because it's impractical to try to synchronise documentation updates with software releases.

JamesH65:
> Yes, that's what I asked, although eventually I think all binaries will be moved to Github as releases and raspberrypi.org will just point to that because it's impractical to try to synchronise documentation updates with software releases.

Really should have gone this way ages ago.

PR #1437, opened 2020-03-16
Major Device Tree update

pelwell:
This is an extensive revision of the DT documentation. It includes some new features:
* New parameter types (byte strings, inverted booleans, lookup tables)
* Literal assignments
* Exporting symbols

The rest of the text as been generally updated and improved.

### DIFF
--- a/configuration/device-tree.md
+++ b/configuration/device-tree.md
@@ -1,34 +1,10 @@
 # Device Trees, overlays, and parameters
 
-Raspberry Pi's latest kernels and firmware, including Raspbian and NOOBS releases, now use a Device Tree (DT) to manage some resource allocation and module loading by default. This was implemented to alleviate the problem of multiple drivers contending for system resources, and to allow HAT modules to be auto-configured.
+Raspberry Pi kernels and firmware use a Device Tree (DT) to describe the hardware present in the Pi. These Device Trees may include DT parameters that provide a degree of control over some onboard features. DT overlays allow optional external hardware to be described and configured, and they also support parameters for more control.
 
-The current implementation is not a pure Device Tree system – there is still board support code that creates some platform devices – but the external interfaces (I2C, I2S, SPI), and the audio devices that use them, must now be instantiated using a Device Tree Blob (DTB) passed to the kernel by the loader (`start.elf`).
+The firmware loader (`start.elf` and its variants) is responsible for loading the DTB (Device Tree Blob - a machine readable DT file). It chooses which one to load based on the board revision number, and makes certain modifications to further tailor it (memory size, Ethernet addresses etc.). This runtime customisation avoids the need for lots of DTBs with only minor differences.
 
-The main impact of using Device Tree is to change from **everything on**, relying on module blacklisting to manage contention, to **everything off unless requested by the DTB**. In order to continue to use external interfaces and the peripherals that attach to them, you will need to add some new settings to your `config.txt`. Full details are given in [Part 3](#part3), but these are a few examples:
-
-```
-# Uncomment some or all of these lines to enable the optional hardware interfaces
-#dtparam=i2c_arm=on
-#dtparam=i2s=on
-#dtparam=spi=on
-
-# Uncomment one of these lines to enable an audio interface
-#dtoverlay=hifiberry-amp
-#dtoverlay=hifiberry-dac
-#dtoverlay=hifiberry-dacplus
-#dtoverlay=hifiberry-digi
-#dtoverlay=iqaudio-dac
-#dtoverlay=iqaudio-dacplus
-#dtoverlay=audioinjector-wm8731-audio
-
-# Uncomment this to enable the lirc-rpi module
-#dtoverlay=lirc-rpi
-
-# Uncomment this to override the defaults for the lirc-rpi module
-#dtparam=gpio_out_pin=16
-#dtparam=gpio_in_pin=17
-#dtparam=gpio_in_pull=down
-```
+`config.txt` is scanned for user-provided parameters, along with any overlays and their parameters, which are then applied. The loader examines the result to learn (for example) which UART, if any, is to be used for the console. Finally it launches the kernel, passing a pointer to the merged DTB.
 
 <a name="part1"></a>
 ## Device Trees
@@ -173,9 +149,11 @@ The `status` property indicates whether a device is enabled or disabled. If the 
 
 A modern SoC (System on a Chip) is a very complicated device; a complete Device Tree could be hundreds of lines long. Taking that one step further and placing the SoC on a board with other components only makes matters worse. To keep that manageable, particularly if there are related devices that share components, it makes sense to put the common elements in `.dtsi` files, to be included from possibly multiple `.dts` files.
 
-But when a system like Raspberry Pi supports optional plug-in accessories, such as HATs, the problem grows. Ultimately, each possible configuration requires a Device Tree to describe it, but once you factor in different base hardware (models A, B, A+, and B+) and gadgets only requiring the use of a few GPIO pins that can coexist, the number of combinations starts to multiply rapidly.
+When a system like Raspberry Pi also supports optional plug-in accessories such as HATs, the problem grows. Ultimately, each possible configuration requires a Device Tree to describe it, but once you factor in all the different base models and the large number of available accessories, the number of combinations starts to multiply rapidly.
 
 What is needed is a way to describe these optional components using a partial Device Tree, and then to be able to build a complete tree by taking a base DT and adding a number of optional elements. You can do this, and these optional elements are called "overlays".
+
+Unless you want to learn how to write overlays for Raspberry Pis, you might prefer to skip on to [Part 3: Using Device Trees on Raspberry Pi)(#part3).
 
 <a name="part2.1"></a>
 ### 2.1: Fragments
@@ -188,7 +166,7 @@ A DT overlay comprises a number of fragments, each of which targets one node and
 /plugin/;
 
 / {
-    compatible = "brcm,bcm2708";
+    compatible = "brcm,bcm2835";
 
     fragment@0 {
         target = <&i2s>;
@@ -199,7 +177,7 @@ A DT overlay comprises a number of fragments, each of which targets one node and
 };
 ```
 
-The `compatible` string identifies this as being for BCM2708, which is the base architecture of the BCM2835 part. For the BCM2836 part you could use a compatible string of "brcm,bcm2709", but unless you are targeting features of the ARM CPUs, the two architectures should be equivalent, so sticking to "brcm,bcm2708" is reasonable. Then comes the first (and in this case only) fragment. Fragments are numbered sequentially from zero. Failure to adhere to this may cause some or all of your fragments to be missed.
+The `compatible` string identifies this as being for BCM2835, which is the base architecture for the Raspberry Pi SoCs; if the overlay makes use of features of a Pi 4 then `brcm,bcm2711` is the correct value to use, otherwise `brcm,bcm2835` can be used for all Pi overlays. Then comes the first (and in this case only) fragment. Fragments should be numbered sequentially from zero. Failure to adhere to this may cause some or all of your fragments to be missed.
 
 Each fragment consists of two parts: a `target` property, identifying the node to apply the overlay to; and the `__overlay__` itself, the body of which is added to the target node. The example above can be interpreted as if it were written like this:
 
@@ -332,10 +310,85 @@ Note that a property is assigned the value `false` by not defining it. Boolean p
 name = <&label>,"property?";
 ```
 
-where `label` and `property` are replaced by suitable values. Boolean parameters can cause properties to be created or deleted.
+where `label` and `property` are replaced by suitable values.
+
+Inverted booleans invert the input value before applying it in the same was as a regular boolean; they are declared similarly, but use `!` to indicate the inversion:
+
+```
+name = <&label>,"property!";
+```
+
+Boolean parameters can cause properties to be created or deleted, but they can't delete a property that already exists in the base DTB.
 
 <a name="part2.2.4"></a>
-#### 2.2.4 Overlay/fragment parameters
+#### 2.2.4: Byte string parameters
+
+Byte string properties are arbitrary sequences of bytes, e.g. MAC addresses. They accept strings of hexadecimal bytes, with or without colons between the bytes.
+
+```
+mac_address = <&ethernet0>,"local_mac_address[";
+```
+The `[` was chosen to match the DT syntax for declaring a byte string:
+```
+local_mac_address = [aa bb cc dd ee ff];
+```
+
+<a name="part2.2.5"></a>
+#### 2.2.5: Parameters with multiple targets
+
+There are some situations where it is convenient to be able to set the same value in multiple locations within the Device Tree. Rather than the ungainly approach of creating multiple parameters, it is possible to add multiple targets to a single parameter by concatenating them, like this:
+
+```
+    __overrides__ {
+        gpiopin = <&w1>,"gpios:4",
+                  <&w1_pins>,"brcm,pins:0";
+        ...
+    };
+```
+
+(example taken from the `w1-gpio` overlay)
+
+Note that it is even possible to target properties of different types with a single parameter. You could reasonably connect an "enable" parameter to a `status` string, cells containing zero or one, and a proper boolean property.
+
+<a name="part2.2.6"></a>
+#### 2.2.6: Literal assignments
+
+As seen in [2.2.5](#part2.2.5), the DT parameter mechanism allows multiple targets to be patched from the same parameter, but the utility is limited by the fact that the same value has to be written to all locations (except for format conversion and the negation available from inverted booleans). The addition of embedded literal assignments allows a parameter to write arbitrary values, regardless of the parameter value supplied by the user.
+
+Assignments appear at the end of a declaration, and are indicated by a `=`:
+
+```
+str_val  = <&target>,"strprop=value";              // 1
+int_val  = <&target>,"intprop:0=42                 // 2
+int_val2 = <&target>,"intprop:0=",<42>;            // 3
+bytes    = <&target>,"bytestr[=b8:27:eb:01:23:45"; // 4
+```
+Lines 1, 2 and 4 are fairly obvious, but line 3 is more interesting because the value appears as an integer (cell) value. The DT compiler evaluates integer expressions at compile time, which might be convenient (particularly if macro values are used), but the cell can also contain a reference to a label:
+```
+// Force an LED to use a GPIO on the internal GPIO controller.
+exp_led = <&led1>,"gpios:0=",<&gpio>,
+          <&led1>,"gpios:4";
+```
+When the overlay is applied, the label will be resolved against the base DTB in the usual way. Note that it is a good idea to split multi-part parameters over multiple lines like this to make them easier to read - something that becomes more necessary with the addition of cell value assignments like this.
+
+Bear in mind that parameters do nothing unless they are applied - a default value in a lookup table is ignored unless the parameter name is used without assigning a value.
+
+<a name="part2.2.7"></a>
+#### 2.2.7: Lookup tables
+
+Lookup tables allow parameter input values to be transformed before they are used. They act as associative arrays, rather like switch/case statements:
+```
+phonetic = <&node>,"letter{a=alpha,b=bravo,c=charlie,d,e,='tango uniform'}";
+bus      = <&fragment>,"target:0{0=",<&i2c0>,"1=",<&i2c1>,"}";
+```
+A key with no `=value` means to use the key as the value, an `=` with no key before it is the default value in the case of no match, and starting or ending the list with a comma (or an empty key=value pair anywhere) indicates that the unmatched input value should be used unaltered; otherwise, not finding a match is an error.
+
+Note that the comma separator within the table string after a cell integer value is implicit - adding one explicitly creates an empty pair (see above).
+
+N.B. As lookup tables operate on input values and literal assigments ignore them, it's not possible to combine the two - characters after the closing `}` in the lookup declaration are treated as an error.
+
+<a name="part2.2.8"></a>
+#### 2.2.8 Overlay/fragment parameters
 
 The DT parameter mechanism as described has a number of limitations, including the inability to change the name of a node and to write arbitrary values to arbitrary properties when a parameter is used. One way to overcome some of these limitations is to conditionally include or exclude certain fragments.
 
@@ -354,10 +407,29 @@ just_one    = <0>,"+1-2"; // Enable 1, disable 2
 conditional = <0>,"=3!4"; // Enable 3, disable 4 if value is true,
                           // otherwise disable 3, enable 4.
 ```
-The i2c-mux overlay uses this technique.
+The `i2c-rtc` overlay uses this technique.
 
-<a name="part2.2.5"></a>
-#### 2.2.5 Examples
+<a name="part2.2.9"></a>
+#### 2.2.9 Special properties
+
+A few property names, when targeted by a parameter, get special handling. One you may have noticed already - `status` - which will convert a boolean to either `okay` for true and `disabled` for false.
+
+Assigning to the `bootargs` property appends to it rather than overwriting it - this is how settings can be added to the kernel command line.
+
+The `reg` property is used to specify device addresses - the location of a memory-mapped hardware block, the address on an I2C bus, etc. The names of child nodes should be qualified with their addresses in hexadecimal, using `@` as a separator:
+```
+		bmp280@76 {
+			reg = <0x77>;
+			...
+		};
+```
+When assigning to the `reg` property, the address portion of the parent node name will be replaced with the assigned value. This can be used to prevent a node name clash when using the same overlay multiple times - a technique used by the `i2c-gpio` overlay.
+
+The `name` property is a pseudo-property - it shouldn't appear in a DT, but assigning to it causes the name of its parent node to be changed to the assigned value. Like the `reg` property, this can be used to give nodes unique names.
+
+
+<a name="part2.2.10"></a>
+#### 2.2.10 Examples
 
 Here are some examples of different types of properties, with parameters to modify them:
 
@@ -376,6 +448,8 @@ Here are some examples of different types of properties, with parameters to modi
 				u64s = /bits/ 64 < 0xaaaaa5a55a5a5555 0x0000111122223333>;
 				bool1; // Defaults to true
 				       // bool2 defaults to false
+				mac = [01 23 45 67 89 ab];
+				spi = <&spi0>;
 			};
 		};
 	};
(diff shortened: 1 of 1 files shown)


configuration/device-tree.md
```diff
@@ -1,44 +1,20 @@
 # Device Trees, overlays, and parameters
 
-Raspberry Pi's latest kernels and firmware, including Raspbian and NOOBS releases, now use a Device Tree (DT) to manage some resource allocation and module loading by default. This was implemented to alleviate the problem of multiple drivers contending for system resources, and to allow HAT modules to be auto-configured.
+Raspberry Pi kernels and firmware use a Device Tree (DT) to describe the hardware present in the Pi. These Device Trees may include DT parameters that provide a degree of control over some onboard features. DT overlays allow optional external hardware to be described and configured, and they too often support parameters for more control.
```
JamesH65: "too often" sounds weird.

configuration/device-tree.md
```diff
@@ -354,10 +407,29 @@ just_one    = <0>,"+1-2"; // Enable 1, disable 2
...
-<a name="part2.2.5"></a>
-#### 2.2.5 Examples
+#### 2.2.9 Special properties
+<a name="part2.2.9"></a>
+
+A few property names, when targeted by a parameter, get special handling. One you may have noticed already - `status` - which will convert a boolean to either `okay` for true and `disabled` for false.
+
+Assigning to the `bootargs` property appends to it rather than overwriting it - this is how settings can be added to the kernel command line.
+
+The `reg` property is used to specify device addresses - the address a memory-mapped hardware block appears, the address on an I2C bus, etc. The names of child nodes should be qualified with their addresses in hexadecimal, using `@` as a separator:
```
JamesH65: "the address a memory-mapped hardware block appears" doesn't read right.

configuration/device-tree.md
pelwell: "also" then.

configuration/device-tree.md
pelwell: I'm not convinced it's wrong, but I'll change it to "the address where...".

configuration/device-tree.md
```diff
@@ -766,28 +831,30 @@ to get:
                 spi@7e2150C0 {
 ```
 
-<a name="part4.3"></a>
+The <a href="https://github.com/raspberrypi/utils">Utils</a> repo includes another DT utility - `ovmerge`. Unlike `dtmerge`, `ovmerge` combines file and applies overlays in source form. Because the overlay is never compiled, labels are preserved and the result is usually more readable. It also has a number of other tricks, such as the ability to list the order of file inclusion.
```
JamesH65: Use [text](URL) syntax for offsite links?

configuration/device-tree.md
```diff
@@ -826,3 +891,7 @@ Specifies a subdirectory/prefix from which to load overlays - defaults to "overl
 
 Further ports can be controlled by the DT, for more details see [section 3](#part3).
 
+### 4.7 Further help
+<a name="part4.7"></a>
+
+If you've read through this document and not found the answer to a Device Tree problem, there is help available. The author can usually be found on Raspberry Pi forums, particularly the <a href="https://www.raspberrypi.org/forums/viewforum.php?f=107">Device Tree</a> forum.
```
JamesH65: [test](url) format?

configuration/device-tree.md
JamesH65: Ah, I needed to use `[test](url)`

configuration/device-tree.md
JamesH65: `[text](url)` markdown

configuration/device-tree.md
pelwell: Yes, I got sucked in by the `<a` tags.

pelwell:
Thanks for the feedback so far. @lurch, @6by9 - can you cast an eye over it?

configuration/device-tree.md
```diff
@@ -199,7 +177,7 @@ A DT overlay comprises a number of fragments, each of which targets one node and
 };
 ```
 
-The `compatible` string identifies this as being for BCM2708, which is the base architecture of the BCM2835 part. For the BCM2836 part you could use a compatible string of "brcm,bcm2709", but unless you are targeting features of the ARM CPUs, the two architectures should be equivalent, so sticking to "brcm,bcm2708" is reasonable. Then comes the first (and in this case only) fragment. Fragments are numbered sequentially from zero. Failure to adhere to this may cause some or all of your fragments to be missed.
+The `compatible` string identifies this as being for BCM2835, which is the base architecture for the Raspberry Pi SoCs; if the overlay makes use of features of a Pi 4 then `brcm,bcm2711` is the correct value to use, otherwise `brcm,bcm2835` can be used for all Pi overlays. Then comes the first (and in this case only) fragment. Fragments should be numbered sequentially from zero. Failure to adhere to this may cause some or all of your fragments to be missed.
```
6by9: Is the compatible string in the overlay matched on now by the firmware/libdft? That'd be very useful in merging the vc4-kms-v3d and vc4-kms-v3d-pi4 overlays.

configuration/device-tree.md
```diff
@@ -405,56 +479,58 @@ Here are some examples of different types of properties, with parameters to modi
...
 
-Note that DT and ATAGs are mutually exclusive. As a result, passing a DT blob to a kernel that doesn't understand it causes a boot failure. To guard against this, the loader checks kernel images for DT-compatibility, which is marked by a trailer added by the `mkknlimg` utility; this can be found in the `scripts` directory of a recent kernel source tree. Any kernel without a trailer is assumed to be non-DT-capable.
+Note that DT and ATAGs are mutually exclusive, and passing a DT blob to a kernel that doesn't understand it causes a boot failure. The firmware will always try to load the DT and pass it to the kernel, since all kernels since rpi-4.4.y will not function without a DTB. You can override this by adding `device_tree=` in config.txt, forcing the use of ATAGs, which can be useful for simple "bare-metal" kernels.
```
6by9: "doesn't understand it **will cause** a boot failure"?

configuration/device-tree.md
```diff
@@ -532,7 +596,7 @@ i2c_baudrate/i2c_arm_baudrate
 i2c_vc_baudrate
 ```
 
-These are aliases for `i2c0`, `i2c1`, `i2c0_baudrate`, and `i2c1_baudrate`. It is recommended that you only use `i2c_vc` and `i2c_vc_baudrate` if you really need to - for example, if you are programming a HAT EEPROM. Enabling `i2c_vc` can stop the Pi Camera being detected.
+These are aliases for `i2c0`, `i2c1`, `i2c0_baudrate`, and `i2c1_baudrate`. It is recommended that you only use `i2c_vc` and `i2c_vc_baudrate` if you really need to - for example, if you are programming a HAT EEPROM (which is better done using a software I2C bus using the `i2c-gpio` overlay). Enabling `i2c_vc` can stop the Pi Camera being detected.
```
6by9: Enabling `i2c_vc` can stop the Pi Camera or 7" DSI display functioning correctly.

configuration/device-tree.md
```diff
@@ -547,24 +611,24 @@ fragment@0 {
 
 Any overlays using the numeric variants will be modified to use the new aliases.
 
-<a name="part3.4"></a>
 ### 3.4: HATs and Device Tree
+<a name="part3.4"></a>
 
-A Raspberry Pi HAT is an add-on board for a "Plus"-shaped (A+, B+ or Pi 2 B) Raspberry Pi with an embedded EEPROM. The EEPROM includes any DT overlay required to enable the board, and this overlay can also expose parameters.
+A Raspberry Pi HAT is an add-on board for a "Plus"-shaped (A+, B+, Pi 2, Pi 3 and Pi 4) Raspberry Pi, with an embedded EEPROM. The EEPROM includes any DT overlay required to enable the board (or the name of an overlay to load from the filing system), and this overlay can also expose parameters.
```
6by9: Did the pHAT for the Zero ever get standardised? I'm assuming that'll have the EEPROM on as well.

configuration/device-tree.md
```diff
@@ -766,28 +831,30 @@ to get:
...
 ```
 device_tree=my-pi.dtb
 ```
 
-<a name="part4.4"></a>
 ### 4.4: Disabling Device Tree usage
+<a name="part4.4"></a>
 
-Since the switch to the 4.4 kernel and the use of more upstream drivers, Device Tree usage is required in Pi kernels. The method of disabling DT usage is to add:
+Since the switch to the 4.4 kernel and the use of more upstream drivers, Device Tree usage is required in Pi Linux kernels. However, for bare metal and other OSs, the method of disabling DT usage is to add:
```
6by9: required by Raspberry Pi Linux kernels.

configuration/device-tree.md
pelwell: Not at present, no. How were you thinking it could work?

configuration/device-tree.md
pelwell: That sounds better.

configuration/device-tree.md
lurch: IIRC it was called a uHAT in the TV-tuner addon? I think pHAT is a Pimoroni thing?
_...searches..._
https://www.raspberrypi.org/blog/raspberry-pi-tv-hat/

configuration/device-tree.md
pelwell: If we must.

configuration/device-tree.md
pelwell: Yes - it's just a mechanical difference. I'll change it to "add-on board with an embedded EEPROM designed for a Raspberry Pi with a 40-pin header".

configuration/device-tree.md
pelwell: Will amend.

configuration/device-tree.md
```diff
@@ -1,44 +1,20 @@
...
+The firmware loader (`start.elf` and its variants) is responsible for loading the DTB (Device Tree Blob - a machine readable DT file). It chooses which one to load based on the board revision number, and makes certain modifications to further tailor it (memory size, Ethernet addresses etc.). This runtime customisation avoids the need for lots of DTBs with only minor differences.
 
-The main impact of using Device Tree is to change from **everything on**, relying on module blacklisting to manage contention, to **everything off unless requested by the DTB**. In order to continue to use external interfaces and the peripherals that attach to them, you will need to add some new settings to your `config.txt`. Full details are given in [Part 3](#part3), but these are a few examples:
+Any user-provided parameters are applied, along with any overlays and their parameters. The loader then examines the result to learn (for example) which UART, if any, is to be used for the console. Finally it launches the kernel, passing a pointer to the merged DTB.
```
lurch: Is it worth mentioning here where these user-provided parameters are read from?

configuration/device-tree.md
```diff
@@ -1,44 +1,20 @@
...
-
-# Uncomment this to override the defaults for the lirc-rpi module
-#dtparam=gpio_out_pin=16
-#dtparam=gpio_in_pin=17
-#dtparam=gpio_in_pull=down
-```
-
-<a name="part1"></a>
 ## Device Trees
+<a name="part1"></a>
```
lurch: It's probably better to put the anchors _above_ the headings, so any links jumping to that anchor will also display the heading text, rather than scrolling to just-below the heading?

configuration/device-tree.md
6by9: Nice and short! :-/

configuration/device-tree.md
```diff
@@ -159,26 +135,28 @@ It is often necessary for one part of the tree to refer to another, and there ar
...
 
-<a name="part2"></a>
 ## Part 2: Device Tree overlays
+<a name="part2"></a>
 
 A modern SoC (System on a Chip) is a very complicated device; a complete Device Tree could be hundreds of lines long. Taking that one step further and placing the SoC on a board with other components only makes matters worse. To keep that manageable, particularly if there are related devices that share components, it makes sense to put the common elements in `.dtsi` files, to be included from possibly multiple `.dts` files.
 
 But when a system like Raspberry Pi supports optional plug-in accessories, such as HATs, the problem grows. Ultimately, each possible configuration requires a Device Tree to describe it, but once you factor in different base hardware (models A, B, A+, and B+) and gadgets only requiring the use of a few GPIO pins that can coexist, the number of combinations starts to multiply rapidly.
```
lurch: Perhaps the "different base hardware" here should now be updated to mention Pi Zero / 2 / 3 / 4? :man_shrugging: 
(or at least a different subset than "models A, B, A+, and B+")

configuration/device-tree.md
pelwell: I find it depends which direction you are jumping - I'll try putting them on the same line.

configuration/device-tree.md
lurch: Are "All Pis" and "Pi 4 only" the only (sesnsible) options, are are there other variants?
Is there any way to indicate "compatible with a Pi1 but not a Pi4" (or "Pi2 but not Pi3"), or is that a nonsensical question?
(I'm only playing devil's advocate here :wink: )

pelwell: For now it's just those two options.

6by9: It was more the wording you'd used that made me think it was working already.

I don't know how it would work within the firmware, but the hardware config between Pi0-3 and Pi4 means the overlays are slightly different, eg enabling more pixelvalves, ddc0/dcc1 instead of i2c2, hdmi0 & hdmi1 instead of hdmi, etc.

It would be nice if we could do a platform agnostic overlay as ```vc4-kms-v3d``` that worked on all platforms, instead of having to use config.txt conditionals.
```
[pi4]
dtoverlay=vc4-kms-v3d-pi4
[!pi4]
dtoverlay=vc4-kms-v3d
[all]
```
if indeed that can work.

We have loaded the base DT, so does /compatible contain a string we can compare against before applying the overlay? Looking in /proc/device-tree/compatible on a running system it appears to.

configuration/device-tree.md
```diff
@@ -332,10 +310,85 @@ Note that a property is assigned the value `false` by not defining it. Boolean p
 name = <&label>,"property?";
 ```
 
-where `label` and `property` are replaced by suitable values. Boolean parameters can cause properties to be created or deleted.
+where `label` and `property` are replaced by suitable values.
+
+Inverted booleans invert the input value before applying it in the same was as a regular boolean; they are declared similarly, but use `!` to indicate the inversion:
```
lurch: "invert the input value before applying it" - is this just a complicated way of saying they set a false value, or is it more subtle / complicated than that? :confused:

configuration/device-tree.md
```diff
@@ -332,10 +310,85 @@ Note that a property is assigned the value `false` by not defining it. Boolean p
...
+
+```
+name = <&label>,"property!";
+```
 
+Boolean parameters can cause properties to be created or deleted, but they can't delete a property that already exists in the base DTB.
+
+#### 2.2.4: Byte string parameters
 <a name="part2.2.4"></a>
-#### 2.2.4 Overlay/fragment parameters
```
lurch: I guess this shows the down-side of using numbered sections (instead of textual names) as anchor-links - any external pages that were previously linking directly to "Overlay/fragment parameters" will now be linking to "Byte string parameters" instead :wink:

configuration/device-tree.md
```diff
@@ -332,10 +310,85 @@ Note that a property is assigned the value `false` by not defining it. Boolean p
...
+
+Bear in mind that parameters do nothing unless they are applied - a default value in a lookup table is ignored unless the parameter name is used without assigning a value.
+
+#### 2.2.7: Lookup tables
+<a name="part2.2.7"></a>
+
+Lookup tables allow parameter input values to be transformed before they are used. They act as associative arrays, rather like switch/case statements:
+```
+phonetic = <&node>,"letter{a=alpha,b=bravo,c=charlie,d,e,='tango uniform'}";
+bus      = <&fragment>,"target:0{0=",<&i2c0>,"1=",<&i2c1>,"}";
```
lurch: Perhaps I'm misunderstanding what this is doing, but should the `,"1=",` above be `,",1=",` ?

configuration/device-tree.md
```diff
@@ -332,10 +310,85 @@ Note that a property is assigned the value `false` by not defining it. Boolean p
...
+```
+A key with no `=value` means to use the key as the value, an `=` with no key before it is the default value in the case of no match, and starting or ending the list with a comma (or an empty key=value pair anywhere) indicates that the unmatched input value should be used unaltered; otherwise, not finding a match is an error.
+
+Note that the comma separator within the table string after a cell integer value is implicit - adding one explicitly creates an empty pair (see above).
```
lurch: Doh, I guess this answers my question! :laughing:

configuration/device-tree.md
```diff
@@ -332,10 +310,85 @@ Note that a property is assigned the value `false` by not defining it. Boolean p
...
+
+Assignments appear at the end of a declaration, and are indicated by a `=`:
+
+```
+str_val  = <&target>,"strprop=value";              // 1
+int_val  = <&target>,"intprop:0=42                 // 2
+int_val2 = <&target>,"intprop:0=",<42>;            // 3
+bytes    = <&target>,"bytestr[=b8:27:eb:01:23:45"; // 4
+```
+Lines 1, 2 and 4 are fairly obvious, but line 3 is more interesting because the value appears as an integer (cell) value. The DT compiler evaluates integer expressions at compile time, which might be convenient (particularly if macro values are used), but the cell can also contain a reference to a label:
```
lurch: Could you clarify if/how this makes the effects of lines 2 and 3 different? Or is this just a pre-cursor to the following paragraph?

configuration/device-tree.md
```diff
@@ -405,56 +479,58 @@ Here are some examples of different types of properties, with parameters to modi
...
+		spibus =      <&test>,"spi:0[0=",<&spi0>,"1=",<&spi1>,"2=",<&spi2>;
+
 		toggle2 =     <0>,"=2";
 		not1 =        <0>,"!1";
-		not2 =        <0>,"!2";
+
+		only1 =       <0>,"+1-2";
+		only2 =       <0>,"-1+2";
+		enable1 =     <0>,"=1";
+		disable2 =    <0>,"!2";
```
lurch: Is the `toggle2` example doing something different to the `enable1` example?
(and I guess similarly `not1` and `disable2`)

configuration/device-tree.md
pelwell: Apart from the compatible strings for device nodes, there is only the top-level node.

There are a number of problems to solve:
1. Fragments that are in legal on all platforms but that we only want to be applied on some platforms. This could be perhaps solved without adding much infrastructure using reserved parameter names that enable and disable fragments.
2. References to labels that are only exist on some platforms. Since symbols are all lumped together without regard to the referencing fragment, solving this one is more tricky - we'd have to delay symbol resolution to the point at which each fragment is applied, and filter out the unwanted fragments in the dtoverlay utility.

configuration/device-tree.md
```diff
@@ -405,56 +479,58 @@ Here are some examples of different types of properties, with parameters to modi
...
-There are some situations where it is convenient to be able to set the same value in multiple locations within the Device Tree. Rather than the ungainly approach of creating multiple parameters, it is possible to add multiple targets to a single parameter by concatenating them, like this:
-
-```
-    __overrides__ {
-        gpiopin = <&w1>,"gpios:4",
-                  <&w1_pins>,"brcm,pins:0";
-        ...
-    };
-```
+For further examples, there is a large collection of overlay source files hosted in the Raspberry Pi/Linux GitHub repository [here](https://github.com/raspberrypi/linux/tree/master/arch/arm/boot/dts/overlays).
```
lurch: Probably remove the slash to make it just "Raspberry Pi Linux GitHub repository" ?

configuration/device-tree.md
pelwell: It means that if you set the parameter to true the property is set to false (i.e. deleted), while setting it to false makes it true. It's for the case where the sense of the parameter name is opposite to the sense of the property, and also makes radio-button property groups possible.

configuration/device-tree.md
pelwell: I'm switching them back...

configuration/device-tree.md
6by9: OK, problem for another day.
For now it might be easier to have a magic string match in the firmware and replace vc4-kms-v3d with vc4-kms-v3d-pi4 if we're on 2711.

configuration/device-tree.md
lurch: Hmmm, that link 404s :slightly_frowning_face: 
Should it have been https://github.com/raspberrypi/linux/tree/rpi-4.19.y/arch/arm/boot/dts/overlays ?

configuration/device-tree.md
```diff
@@ -405,56 +479,58 @@ Here are some examples of different types of properties, with parameters to modi
...
-(example taken from the `w1-gpio` overlay)
+### 2.3: Exporting labels
+<a name="part2.3"></a>
 
-Note that it is even possible to target properties of different types with a single parameter. You could reasonably connect an "enable" parameter to a `status` string, cells containing zero or one, and a proper boolean property.
+The overlay handling in the firmware and the run-time overlay application using the `dtoverlay` utility treat labels defined in an overlay as being private to that overlay. This avoids the need to invent globally unique names for labels (which keeps them short), and it allows the same overlay to be used multiple times without clashing (provided some tricks are used - see <a href="part2.2.9">Special properties</a>).
 
-<a name="part2.2.7"></a>
-#### 2.2.7: Further overlay examples
+Sometime, however, it is very useful to be able to create a label with one overlay and use it from another. Firmware released since Feb 14 2020 has the ability to declare some labels as being global - the `__export__` node:
```
lurch: Sometime -> Sometimes ?

configuration/device-tree.md
pelwell: The end results of lines 2 and 3 are the same, but line 3 could have been written with `,<(6*9)>`, or `<MEANING_OF_LIFE>` with the appropriate macro constant definition. But yes, it is largely a set-up for the next paragraph on labels,

configuration/device-tree.md
pelwell: They should have been deleted.

configuration/device-tree.md
pelwell: Will do.

configuration/device-tree.md
pelwell: Agreed.

configuration/device-tree.md
lurch: "forcing" -> "which forces" ?

configuration/device-tree.md
pelwell: Yes, unfortunately master != default. I was hoping to remove the fixed version number, but for now I'll just make it use 5.4 which should be considered State of the Art.

configuration/device-tree.md
```diff
@@ -405,56 +479,58 @@ Here are some examples of different types of properties, with parameters to modi
...
+Note that DT and ATAGs are mutually exclusive, and passing a DT blob to a kernel that doesn't understand it causes a boot failure. The firmware will always try to load the DT and pass it to the kernel, since all kernels since rpi-4.4.y will not function without a DTB. You can override this by adding `device_tree=` in config.txt, forcing the use of ATAGs, which can be useful for simple "bare-metal" kernels.
 
-A kernel built from the rpi-4.4.y tree (and later) will not function without a DTB, so from the 4.4 releases onwards, any kernel without a trailer is assumed to be DT-capable. You can override this by adding a trailer without the DTOK flag or by putting `device_tree=` in config.txt, but don't be surprised if it doesn't work.
-N.B. A corollary to this is that if the kernel has a trailer indicating DT capability then `device_tree=` will be ignored.
+[ The firmware used to look for a trailer appended to kernels by the `mkknlimg` utility, but support for this has been withdrawn. ]
```
lurch: Does this mean that the "...if the kernel has been tagged by a recent `mkknlimg`..." line below needs tweaking as well?

configuration/device-tree.md
```diff
@@ -463,20 +539,19 @@ dtparam=foo=bar,level=42
 
 This will cause the loader to look for `overlays/acme-board.dtbo` in the firmware partition, which Raspbian mounts on `/boot`. It will then search for parameters `foo` and `level`, and assign the indicated values to them.
 
-The loader will also search for an attached HAT with a programmed EEPROM, and load the supporting overlay from there; this happens without any user intervention.
+The loader will also search for an attached HAT with a programmed EEPROM, and load the supporting overlay from there - either directly or by name from the "overlays" directory; this happens without any user intervention.
```
lurch: Does the "or by name" imply that there's some kind of precedence if both are present?

configuration/device-tree.md
pelwell: Yes, tweaking with extreme prejudice.

configuration/device-tree.md
```diff
@@ -463,20 +539,19 @@ dtparam=foo=bar,level=42
...
 
 With a Device Tree, the kernel will automatically search for and load modules that support the indicated enabled devices. As a result, by creating an appropriate DT overlay for a device you save users of the device from having to edit `/etc/modules`; all of the configuration goes in `config.txt`, and in the case of a HAT, even that step is unnecessary. Note, however, that layered modules such as `i2c-dev` still need to be loaded explicitly.
 
-The flipside is that because platform devices don't get created unless requested by the DTB, it should no longer be necessary to blacklist modules that used to be loaded as a result of platform devices defined in the board support code. In fact, current Raspbian images ship without a blacklist file.
+The flipside is that because platform devices don't get created unless requested by the DTB, it should no longer be necessary to blacklist modules that used to be loaded as a result of platform devices defined in the board support code. In fact, current Raspbian images ship with no blacklist files except for a few WLAN devices where multiple drivers are available.
```
lurch: Would it be appropriate to put the "except ..." part at the end in brackets? :man_shrugging:

configuration/device-tree.md
pelwell: No, it's either/or. If the BLOB start 0xd00dfeed then it's an embedded overlay, otherwise we attempt to interpret it as an overlay name.

configuration/device-tree.md
pelwell: Spose so.

configuration/device-tree.md
```diff
@@ -595,17 +660,19 @@ Two points to note:
...
 
-`dtparam` creates an overlay that has the same effect as using a dtparam directive in `config.txt`. In usage it is largely equivalent to `dtoverlay` with an overlay name of `-`, but there are a few small differences:
+`dtparam` creates an overlay that has largely the same effect as using a dtparam directive in `config.txt`. In usage it is largely equivalent to `dtoverlay` with an overlay name of `-`, but there are a few differences:
 
 1. `dtparam` will list the help information for all known parameters of the base DTB. Help on the dtparam command is still available using `dtparam -h`.
 
 2. When indicating a parameter for removal, only index numbers can be used (not names).
 
-<a name="part3.5.3"></a>
+3. Not all Linux subsystems respond to the addition of devices at runtime - I2C, SPI and sound devices work, but some won't.
```
lurch: Does this imply that the "`dtparam` creates an overlay..." a few lines earlier should be tweaked to "`dtparam` creates and loads an overlay..." or similar?

lurch:
I'm afraid I didn't understand most of that, but hopefully some of my comments were useful anyway!

configuration/device-tree.md
lurch: None of that seems to be mentioned at https://github.com/raspberrypi/hats/blob/master/eeprom-format.md or https://github.com/raspberrypi/hats/blob/master/devicetree-guide.md ?

configuration/device-tree.md
pelwell: Yes - "creates and loads".

configuration/device-tree.md
pelwell: I haven't finished the documentation spree yet - there's more joy to come.

configuration/device-tree.md
pelwell: Replaced with:
> `config.txt` is scanned for user-provided parameters, along with any overlays and their parameters, which are then applied. The loader examines the result...

configuration/device-tree.md
pelwell: Changed to:
> but once you factor in all the different base models and the large number of available accessories, the number...

configuration/device-tree.md
pelwell: Done.

pelwell:
Thanks, everyone - that's been helpful.

The fix-ups have been applied, and I'm happy for the result to be merged.

JamesH65:
LGTM, everyone else OK with the changes?

JamesH65:
No news is good news, so merging.

configuration/device-tree.md
```diff
@@ -635,7 +709,7 @@ The loading of overlays at runtime is a recent addition to the kernel, and so fa
 <a name="part3.6"></a>
 ### 3.6: Supported overlays and parameters
 
-As it is too time-consuming to document the individual overlays here, please refer to the [README](https://github.com/raspberrypi/firmware/blob/master/boot/overlays/README) file found alongside the overlay `.dtbo` files in `/boot/overlays`. It is kept up-to-date with additions and changes.
+As it is too time-consuming to document the individual overlays here, please refer to the [README](https://github.com/raspberrypi/firmware/blob/rpi-5.4.y/boot/overlays/README) file found alongside the overlay `.dtbo` files in `/boot/overlays`. It is kept up-to-date with additions and changes.
```
lurch: Oops, just spotted that this URL 404s @pelwell

pelwell: The perils of search and replace.

PR #1441, opened 2020-03-19
Add (and enhance) cross-links in boot_folder.md

lurch:


lurch:
@pelwell I noticed that https://github.com/raspberrypi/documentation/blob/master/configuration/config-txt/boot.md#arm_64bit says 64-bit isn't for 2836, whereas https://github.com/raspberrypi/documentation/blob/master/configuration/boot_folder.md says it is?
I could fix that in this PR too if you like?

JamesH65:
@lurch This seems to have got lost - OK to merge?

lurch:
Yup, I just checked and the links this PR adds are still correct.

PR #1443, opened 2020-03-23
device-tree.md: Update and extend example overlay

pelwell:
Changed because:
1. The compatible string has changed.
2. Newer versions of dtc generate slightly different output and can
   accept the newer, more friendly syntax variant.
3. We can get more value from the example.

Thanks to @HiassofT for the feedback.

### DIFF
--- a/configuration/device-tree.md
+++ b/configuration/device-tree.md
@@ -172,6 +172,10 @@ A DT overlay comprises a number of fragments, each of which targets one node and
         target = <&i2s>;
         __overlay__ {
             status = "okay";
+			test_ref = <&test_label>;
+			test_label: test_subnode {
+				dummy;
+			};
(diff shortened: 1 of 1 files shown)


JamesH65:
LGTM.

configuration/device-tree.md
```diff
@@ -172,6 +172,10 @@ A DT overlay comprises a number of fragments, each of which targets one node and
         target = <&i2s>;
         __overlay__ {
             status = "okay";
+			test_ref = <&test_label>;
```
lurch: Excessive indentation here?

pelwell: It's TABs vs spaces - now fixed.

pelwell:
Ta.

PR #1453, opened 2020-03-30
Clarify country code in WiFi doc

bman46:


lurch:
Not sure if this is necessary, as it already says `See [Wikipedia](https://en.wikipedia.org/wiki/ISO_3166-1) for a list of country codes.` just underneath?

bman46:
I found it kind of confusing on which one to use (alpha 2, alpha 3, or numeric). Plus, it will help out with the people who just want to copy and paste and not read anything.

JamesH65:
I think this would be OK, if the link below also stated the ISO number. The text used to match in both places, it should continue to match, so if one changes the other does.

bman46:
> I think this would be OK, if the link below also stated the ISO number. The text used to match in both places, it should continue to match, so if one changes the other does.

Good point, added that in.

JamesH65:
Thanks

bman46:
No problem.

PR #1455, opened 2020-04-02
Update README.md

fssantana2:
Each USB port is compliant to the usb.org specification. Each USB compliant port must handle 500mA. As Raspberry have two ports, so the total USB power budget is 1A. The previous text can lead the reader to think that the mentioned power requirement is for the whole Raspberry power consumption, not only USB: "The maximum power the Raspberry Pi peripherals can use is 1 Amp. If you need to connect a device that will take the power requirements above 1 Amp, then you must connect it to an externally-powered USB hub."
My text suggestion makes it clear that it refers only to the USB peripherals: "The maximum USB power the Raspberry Pi peripherals can use is 1 Amp. If you need to connect a USB device that will take the USB power requirements above 1 Amp, then you must connect it to an externally-powered USB hub."


JamesH65:
@lurch Does make it clearer, but isn't the max 1.2A spread over all the ports?

hardware/raspberrypi/power/README.md
```diff
@@ -2,7 +2,7 @@
...
 
-Typically, the model B uses between 700-1000mA depending on what peripherals are connected; the model A can use as little as 500mA with no peripherals attached. The maximum power the Raspberry Pi can use is 1 Amp. If you need to connect a USB device that will take the power requirements above 1 Amp, then you must connect it to an externally-powered USB hub.
+Typically, the model B uses between 700-1000mA depending on what peripherals are connected; the model A can use as little as 500mA with no peripherals attached. The maximum USB power the Raspberry Pi peripherals can use is 1 Amp. If you need to connect a USB device that will take the USB power requirements above 1 Amp, then you must connect it to an externally-powered USB hub.
```
lurch, line 5: I think "The maximum USB power the Raspberry Pi peripherals can use is 1 Amp." would be clearer as "The maximum power the Raspberry Pi USB peripherals can use is 1 Amp."?

lurch:
@JamesH65 The way the original text is written, it's not clear if it's talking about the "maximum power available to downstream USB devices" or the "maximum power that can be sucked in from the microUSB input"? I guess @jimbojr would be the person to ask?
Bear in mind that this paragraph is talking about the original model B released all the way back in 2012! (so perhaps this text isn't worth updating at all?)

JamesH65:
Hmm, I think the entire page needs a revamp, adding the table from the FAQ will mean a lot of this page can be removed, just leaving the hints and tips and how the power is distributed in there. Amazing how much stuff in the docs is rather out of date. We should be able to make this stuff a lot more generic so the descriptions don't change between models, but a simple table does. Should make it easier to update in the future.

fssantana2:
@lurch I just found other page that provides more information about the USB power capacity, but it conflicts in some aspects with the page edited in this pull request: https://www.raspberrypi.org/documentation/hardware/raspberrypi/usb/README.md#powerlimits

The page of this pull request says 1A current capacity (https://www.raspberrypi.org/documentation/hardware/raspberrypi/power/README.md), but the page in the paragraph above says 1.2A supplied to all USB ports in aggregate.

Both pages mention Raspberry 3 so it seems that it relates to the current hardware in fact, not the model from 2012 (my original assumption it refers to the 2 port device is wrong). According to the link on the top of this text, Raspberry 3 current handling is 1.2A in fact.

If this understanding is correct, the text proposed in this pull request should be updated to 1.2A. Please check if you agree.

JamesH65:
Would this complete revamp be better? 

https://github.com/raspberrypi/documentation/pull/1456

fssantana2:
@JamesH65 I agree.

JamesH65:
Could add something about the voltage warning icon as well I suppose.

JamesH65:
OK, I'll close this one then. Thanks to @fssantana2 for pointing out the problem!

PR #1458, opened 2020-04-05
Remove info for gnome-schedule, since outdated

Django987:
- https://askubuntu.com/questions/861689/how-to-install-gnome-schedule-for-16-10
- https://www.raspberrypi.org/forums/viewtopic.php?t=250197


JamesH65:
Thanks - didn't even know that was in there!

PR #1459, opened 2020-04-06
James h65 patch datasheets page

JamesH65:
@lurch Thinking something like this. Not happy yet with it, although cannot place exactly why.

### DIFF
--- a/hardware/raspberrypi/README.md
+++ b/hardware/raspberrypi/README.md
@@ -5,7 +5,7 @@ The hardware in the Raspberry Pi
 - [Schematics](schematics/README.md)
     - Schematics for the Raspberry Pi
 - [Datasheets](datasheets.md)
-    - Currently available datasheets for the Raspberry Pi range
+    - Datasheets for the Raspberry Pi
 - [BCM2835](bcm2835/README.md)
     - The Broadcom processor used in Raspberry Pi 1 and Zero
 - [BCM2836](bcm2836/README.md)
--- a/hardware/raspberrypi/README.md
+++ b/hardware/raspberrypi/README.md
@@ -4,6 +4,8 @@ The hardware in the Raspberry Pi
 
 - [Schematics](schematics/README.md)
     - Schematics for the Raspberry Pi
+- [Datasheets](datasheets.md)
+    - Currently available datasheets for the Raspberry Pi range
 - [BCM2835](bcm2835/README.md)
     - The Broadcom processor used in Raspberry Pi 1 and Zero
 - [BCM2836](bcm2836/README.md)
--- a/hardware/raspberrypi/bcm2837/README.md
+++ b/hardware/raspberrypi/bcm2837/README.md
@@ -4,7 +4,7 @@ This is the Broadcom chip used in the Raspberry Pi 3, and in later models of the
 
 The ARM cores run at 1.2GHz, making the device about 50% faster than the Raspberry Pi 2. The VideoCore IV runs at 400MHz.
 
-Please refer to the following BCM2836 document for details on the ARM peripherals specification which is unchanged.
+Please refer to the following BCM2836 document for details on the ARM peripherals specification, which also applies to the BCM2837.
 
 - [BCM2836 ARM-local peripherals](../bcm2836/QA7_rev3.4.pdf)
 
--- a/hardware/raspberrypi/bcm2837/README.md
+++ b/hardware/raspberrypi/bcm2837/README.md
@@ -4,4 +4,8 @@ This is the Broadcom chip used in the Raspberry Pi 3, and in later models of the
 
 The ARM cores run at 1.2GHz, making the device about 50% faster than the Raspberry Pi 2. The VideoCore IV runs at 400MHz.
 
+Please refer to the following BCM2836 document for details on the ARM peripherals specification which is unchanged.
+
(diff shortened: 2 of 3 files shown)


hardware/raspberrypi/datasheets.md
```diff
@@ -0,0 +1,14 @@
+## Datasheets
+
+### SoC's
+
+The following datasheets are available for the SoC (System On Chip - the main processing chip) devices used on the Raspberry Pi platform.
+
+[BCM2711](bcm2711/rpi_DATA_2711_1p0.pdf) - The SoC used on the Raspberry Pi 4.
+
+[BCM2835](bcm2735/BCM2835-ARM-Peripherals.pdf) - The SoC used on the Raspberry Pi 0 and 1. This datasheet is also applicable to the Raspberry Pi 2 and 3, with the BCM2836 and BC2837 SoC's respectively, as the Videocore GPU is the same.
```
lurch, line 9: BCM2835 was released before BCM2711, so I'd list it first. :man_shrugging: 
Also, I'm assuming the `bcm2735` in the URL is wrong? :laughing:

hardware/raspberrypi/README.md
```diff
@@ -4,6 +4,8 @@ The hardware in the Raspberry Pi
 
 - [Schematics](schematics/README.md)
     - Schematics for the Raspberry Pi
+- [Datasheets](datasheets.md)
+    - Currently available datasheets for the Raspberry Pi range
```
lurch, line 8: The schematics link doesn't say "Currently available schematics for the Raspberry Pi range", so maybe the text here should just say "Datasheets for the Raspberry Pi" ?

lurch:
There's also the "ARM-local peripherals" document on the BCM2836 page - is that worth linking here?

hardware/raspberrypi/datasheets.md
JamesH65: Weird, I thought I checked all the links...

hardware/raspberrypi/README.md
JamesH65: Just trying to add a bit of variety!

JamesH65:
I wasn't sure what to do about that arm stuff. But having taken a longer look, I think we should have it in there. I wonder how much of it is relevant to the generic Quad core arm devices - ie A7->A53->A72.

hardware/raspberrypi/datasheets.md
```diff
@@ -0,0 +1,18 @@
+## Datasheets
+
+### SoC's
+
+The following datasheets are available for the SoC (System On Chip - the main processing chip) devices used on the Raspberry Pi platform.
+
+[BCM2835](bcm2835/BCM2835-ARM-Peripherals.pdf) - The SoC used on the Raspberry Pi 0 and 1. This datasheet is also applicable to the Raspberry Pi 2 and 3, with the BCM2836 and BC2837 SoC's respectively, as the Videocore GPU is the same.
```
lurch, line 7: This feels a little bit clumsy? Perhaps:
The SoC used on the Raspberry Pi 0 and 1. This datasheet is also applicable to the BCM2836 and BC2837 SoCs (used on the Pi2 and Pi3 respectively), as they use the same VideoCore IV GPU.

:question:

lurch:
> I wonder how much of it is relevant to the generic Quad core arm devices - ie A7->A53->A72.

AFAIK it applies to both the BCM2836 and BCM2837, but **doesn't** apply to BCM2711.

hardware/raspberrypi/datasheets.md
lurch: ...or perhaps "Large parts of this datasheet are also applicable to the BCM2836 and BC2837 SoCs ..." ?

JamesH65: Yes, much better. Ta.

JamesH65:
> > I wonder how much of it is relevant to the generic Quad core arm devices - ie A7->A53->A72.
> 
> AFAIK it applies to both the BCM2836 and BCM2837, but **doesn't** apply to BCM2711.

That would be my guess, but would need to confirm.

hardware/raspberrypi/datasheets.md
```diff
@@ -0,0 +1,17 @@
+## Datasheets
+
+### SoC's
```
lurch, line 3: Greengrocer's apostrophe? :wink:

pelwell:
Yes, it's applicable to 2837. If you look at bcm2837.dtsi in the kernel
you'll see that it makes use of the bcm2836-l1-intc and bcm2836-smp drivers,
which make use of the hardware described in that PDF.

JamesH65:
OK, will update the links and text appropriately. Thanks all.

hardware/raspberrypi/bcm2837/README.md
```diff
@@ -4,4 +4,8 @@ This is the Broadcom chip used in the Raspberry Pi 3, and in later models of the
 
 The ARM cores run at 1.2GHz, making the device about 50% faster than the Raspberry Pi 2. The VideoCore IV runs at 400MHz.
 
+Please refer to the following BCM2836 document for details on the ARM peripherals specification which is unchanged.
```
pelwell, line 7: "...which also applies to the BCM2837." is less ambiguous.

JamesH65: Done

JamesH65:
Thanks all.

PR #1462, opened 2020-04-10
Update updating.md

fprades:
apt command lacks clean subcommand. It's apt-get which include clean subcommand.

### DIFF
--- a/raspbian/updating.md
+++ b/raspbian/updating.md
@@ -38,7 +38,7 @@ The kernel and firmware are installed as a Debian package, and so will also get 
 
 ### Running out of space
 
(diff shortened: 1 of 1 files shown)


JamesH65:
Is there a apt version of clean? That would be a better change, but if there is nothing similar, this PR looks fine.

fprades:
I run into this problem while updating. Jessie to Stretch. At least apt 1.0.9.8.5(jessie) doesn't seem to have it.

JamesH65:
Just tried this on a Pi4 with latest Raspbian, no errors apparent, but no idea if it actually did anything.

fprades:
I suspect `apt clean` was added in later releases of apt. Modified the doc to use apt, or apt-get if former not available

JamesH65:
Thanks!

PR #1465, opened 2020-04-12
fixed dd path to device

tensionhead:
There was a typo disk -> rdisk in the
```bash
bash
```
boxes.

### DIFF
--- a/installation/installing-images/mac.md
+++ b/installation/installing-images/mac.md
@@ -61,7 +61,7 @@ If you specify the wrong device in the instructions, you could overwrite your pr
 - Copy the image
 
   ```bash
-  sudo dd bs=1m if=path_of_your_image.img of=/dev/rdiskN conv=sync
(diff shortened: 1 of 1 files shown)


JamesH65:
Has MacOS or whatever its called changed in this respect? Would be odd to have so many typos - so presumably it was rdisk at some point but isn't now?

tensionhead:
Hi James,

in the rest of the mac.md the device path is referred to as /dev/diskN. I
thought maybe it was like an extra security that people don't wipe the
wrong partition..? But now I found this thread:

https://superuser.com/questions/631592/why-is-dev-rdisk-about-20-times-faster-than-dev-disk-in-mac-os-x

So apparently /rdisk is correct for the dd command! It's just confusing for
the reader, so maybe it's good to insert a small note in the .md. I will
retract my pull request.

JamesH65:
LGTM. Merging.

lurch:
Yeah, see also #366 :slightly_smiling_face:

PR #1468, opened 2020-04-13
device-tree.md: Document the overlay map file

pelwell:


pelwell:
I'm not planning on adding any more to this, unless something comes up in review.

JamesH65:
LGTM. Clear and concise!

JamesH65:
@lurch for comment

configuration/device-tree.md
```diff
@@ -448,9 +448,60 @@ When assigning to the `reg` property, the address portion of the parent node nam
...
+
+There is therefore a need for a method of tailoring an overlay to multiple platforms with differing hardware. Supporting them all in a single .dtbo file would require heavy use of hidden ("dormant") fragments and a switch to an on-demand symbol resolution mechanism so that a missing symbol that isn't needed doesn't cause a failure. A simpler solution is to add a facility to map an overlay name to one of several implementation files depending on the current platform.
+
+The overlay map, which is rolling out with the switch to Linux 5.4, is a file that gets loaded by the firmware at start of day. It is written in DTS source format - `overlay_map.dts`, compiled to `overlay_map.dtb` and stored in the overlays directory.
```
lurch, line 458: Maybe non-technical people would get confused by the "start of day" phrase? Perhaps "by the firmware at bootup." instead?
(but then again, I suppose non-technical people aren't going to be reading this part of the documentation anyway? :rofl: )

configuration/device-tree.md
```diff
@@ -448,9 +448,60 @@ When assigning to the `reg` property, the address portion of the parent node nam
...
+        renamed = "disable-bt";
+    };
+
+    lirc-rpi {
+        deprecated = "use gpio-ir";
+    };
+};
+```
+
+Each node has the name of an overlay that requires special handling. The properties of each node are either platform names or one of a small number of special directives. A platform name with no value (an empty property) indicates that the current overlay is compatible with the platform; for example, `vc4-kms-v3d` is compatible with the `bcm2835` platform. A non-empty value for a platform is the name of an alternative overlay to use in place of the requested one; asking for `vc4-kms-v3d` on BCM2711 results in `vc4-kms-v3d-pi4` being loaded instead. Any platform not included in an overlay's node is not compatible with that overlay.
```
lurch, line 487: "compatible with the `bcm2835` platform."

Is it worth clarifying that this also includes the BCM2836 and BCM2837 SoCs? (or is that something that the reader of this page will already know?)

configuration/device-tree.md
```diff
@@ -448,9 +448,60 @@ When assigning to the `reg` property, the address portion of the parent node nam
...
+
+The second example node - `vc4-kms-v3d-pi4` - could be inferred from the content of `vc4-kms-v3d`, but that intelligence goes into the construction of the file, not its interpretation.
+
+In the event that a platform is not listed for an overlay, one of the special directives may apply:
+
+* The `renamed` directive indicates the new name of the overlay (which should be largely compatible with the original), but also logs a warning about the rename.
+
+* The `deprecated` directive contains a brief explanatory error message which will be logged after the common prefix `overlay '...' is deprecated:`.
```
lurch, line 495: Will readers of this page know where these warning / error logs are visible?

configuration/device-tree.md
```diff
@@ -448,9 +448,60 @@ When assigning to the `reg` property, the address portion of the parent node nam
...
+
+Remember: only exceptions need to be listed - the absence of a node for an overlay means that the default file should be used for all platforms.
```
lurch, line 497: This seems like fairly important info - perhaps it should be moved to earlier?

pelwell: Ah ha! This time you really did miss it:
> Each node has the name of an overlay that requires special handling.

configuration/device-tree.md
pelwell: It's in section 4.1 (Debugging), and searching for "log", "error" or "warning" finds it - which I think is enough.

configuration/device-tree.md
pelwell: It's mentioned in section 2.1 (Fragments):

> The `compatible` string identifies this as being for BCM2835, which is the base architecture for the Raspberry Pi SoCs; if the overlay makes use of features of a Pi 4 then `brcm,bcm2711` is the correct value to use, otherwise `brcm,bcm2835` can be used for all Pi overlays.

configuration/device-tree.md
pelwell: And there was me trying to humanise the language a little.

JamesH65: Device tree is not for humans.

pelwell: OK - I'll tweak a few things, stripping out any remaining soul, and add a patch to squash in at merge time.

pelwell: Done.

pelwell:
What's stopping this being merged?

JamesH65:
Nothing!

PR #1472, opened 2020-04-15
update arm versions in kernel building instructions

nagimov:
I'm running Raspbian Buster Lite, 2020-02-12.

On pi 2 I get:
```
pi@raspberrypi:~ $ cat /sys/firmware/devicetree/base/model;echo
Raspberry Pi 2 Model B Rev 1.1
pi@raspberrypi:~ $ lscpu | grep arm
Architecture:        armv7l
```

On pi 3+ I get:
```
pi@raspberrypi:~/mcp3008hwspi $ cat /sys/firmware/devicetree/base/model;echo
Raspberry Pi 3 Model B Plus Rev 1.3
pi@raspberrypi:~/mcp3008hwspi $ lscpu | grep arm
Architecture:        armv7l
```

However [build instructions](https://github.com/raspberrypi/documentation/blob/master/linux/kernel/building.md#raspberry-pi-2-pi-3-pi-3-and-compute-module-3-default-build-configuration) for both Pi 2 and Pi 3+ call for `KERNEL=kernel7` instead of `KERNEL=kernel7l`.

### DIFF
--- a/linux/kernel/building.md
+++ b/linux/kernel/building.md
@@ -52,7 +52,7 @@ make bcmrpi_defconfig
 
 ```bash
 cd linux
-KERNEL=kernel7
+KERNEL=kernel7l
 make bcm2709_defconfig
 ```
 


JamesH65:
@pelwell @popcornmix Have we changed the older stuff to use the LPA kernel as well?

pelwell:
I'm not sure what the "l" in "armv7l" stands for, but (I think) all the "v7" BCM283x processors are also "v7l" so it's not significant to us. The "l" in kernel7l.img _is_ significant because (as @JamesH65 has hinted) it indicates that LPAE (Large Physical Address Extension) is enabled, which is what we need to support BCM2711 on a 32-bit kernel.

There's no benefit (and a small overhead) to enabling LPAE on devices prior to 2711, so we don't - 64-bit kernels don't need it because they support large physical addresses by default.

pelwell:
Which means this PR is erroneous, although a brief note about kernel versions (or a pointer to it if one already exists) might be called for...

lurch:
> I'm not sure what the "l" in "armv7l" stands for

https://stackoverflow.com/questions/29166619/differences-between-arm-versions-armv7-only suggests that the `l` in `armv7l` stands for [Little-Endian](https://en.wikipedia.org/wiki/Endianness). Which is indeed something very different to the [LPAE](https://en.wikipedia.org/wiki/ARM_architecture#LPAE) that the `l` in `kernel7l.img` indicates :wink:  (i.e. it's merely an unfortunate overlapping of initials)

JamesH65:
> > I'm not sure what the "l" in "armv7l" stands for
> 
> https://stackoverflow.com/questions/29166619/differences-between-arm-versions-armv7-only suggests that the `l` in `armv7l` stands for [Little-Endian](https://en.wikipedia.org/wiki/Endianness). Which is indeed something very different to the [LPAE](https://en.wikipedia.org/wiki/ARM_architecture#LPAE) that the `l` in `kernel7l.img` indicates 😉 (i.e. it's merely an unfortunate overlapping of initials)

That should probably be documented somewhere. I have no idea where though.

lurch:
> That should probably be documented somewhere. I have no idea where though.

I guess https://www.raspberrypi.org/documentation/configuration/boot_folder.md is a good candidate? Which perhaps https://www.raspberrypi.org/documentation/linux/kernel/README.md could also cross-link to?

There's also the question of whether it's worth going down the "difference between Raspbian and Debian" rabbit-hole? (as briefly explained at https://wiki.debian.org/ArmPorts )

JamesH65:
https://github.com/raspberrypi/documentation/pull/1473

lurch:
Closing this PR, as it's "resolved" by @JamesH65 's PR.

PR #1473, opened 2020-04-15
Explain lscpu armv7l.

JamesH65:


configuration/boot_folder.md
```diff
@@ -58,6 +58,8 @@ The boot folder will contain various kernel image files, used for the different
 | kernel7l.img | BCM2711 | Pi 4 | Large Physical Address Extension |
 | kernel8.img  | BCM2836, BCM2837, BCM2711 | Pi2/3/4 | Beta 64 bit kernel|
 
+Note: The architecture reported by `lscpu` is `armv7l` for 32 bit systems and `aarch64` for 64 bit systems. The `l` in the `armv7l` case refers to the architecture being little-endian, not `LPAE` as is indicated by the `l` in the `kernel7l.img` filename.
```
lurch, line 61: The table above doesn't explicitly mention 32-bit kernels, so _perhaps_ it's worth clarifying "... for 32 bit systems (i.e. everything except `kernel8.img`), and ..." ?
Should 32-bit and 64-bit be hyphenated?

pelwell: Yes to both suggestions.

PR #1476, opened 2020-04-15
VNC installation clarification

JamesH65:


spl237:
Looks good to me.

lurch:
LGTM. While looking at the full page, there seems to be a stray parenthesis in "or in the office)." which might be worth removing in this PR.

PR #1477, opened 2020-04-16
Fix broken link in tv-hat README.md for VideoLan.org/vlc

Antelope-IT:
Broken link found on live site 16/04/2020 browsing on RPi 4B: Original link appears to be relative to raspberrypi.org and fails. Add protocol to url as per other links on page and now seems to work as expected.


lurch:
Looks like the same error is affecting the "coveragechecker" link too?

andrum99:
Line 45 also contains a broken link to the Digital UK coverage checker - same issue.

JamesH65:
I've fixed the other link, so merging. Thanks.

PR #1479, opened 2020-04-19
Describe enable_jtag_gpio for config.txt

czak:
There seems to be an `enable_jtag_gpio=1` setting for `config.txt` which I've seen mentioned in a few blog posts and Github issues but not documented here.

I guess this is just a shortcut for `gpio=22-27=a4`, but maybe there's more to this and worth documenting?
Or maybe there's a reason why it's undocumented? Not supported on all devices perhaps?

References:
* https://github.com/raspberrypi/firmware/issues/639
* https://www.raspberrypi.org/documentation/hardware/raspberrypi/bcm2835/BCM2835-ARM-Peripherals.pdf (section 6.2 - Alternative Function Assignments)

### DIFF
--- a/configuration/config-txt/gpio.md
+++ b/configuration/config-txt/gpio.md
@@ -44,7 +44,7 @@ if booting over the network or from a USB mass storage device.
 
 ## `enable_jtag_gpio`
 
-Setting `enable_jtag_gpio=1` selects Alt4 mode for GPIO pins 22-27, thus enabling JTAG interface for the ARM CPU.
+Setting `enable_jtag_gpio=1` selects Alt4 mode for GPIO pins 22-27, and flips some internal SoC connections, thus enabling the JTAG interface for the ARM CPU. It works on all models of Raspberry Pi.
 
 | Pin #  | Function |
 | ------ | -------- |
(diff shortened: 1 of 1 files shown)


lurch:
ping @pelwell

pelwell:
`enable_jtag_gpio` does all the necessary setup to enable JTAG access to the ARM complex. This includes switching to Alt4 on 22-27, but it also flips some SoC-specific switches to hook it up. It should work on all devices.

lurch:
I know nothing about JTAG, but does "it also flips some SoC-specific switches" imply that `enable_jtag_gpio` is _required_ if you want to JTAG the ARM? (although I appreciate this is obviously  a fairly niche use-case)

pelwell:
Where I wrote "does all the necessary setup" you should infer "(and no unnecessary setup)".

lurch:
I'm not sure I understand your reply Phil? As already mentioned I know sod all about JTAG, but I think what I was asking is "Is it impossible to use JTAG without setting `enable_jtag_gpio`?"

pelwell:
Yes.

JamesH65:
@pelwell Made some minor changes, if they are accurate and you are happy with the change I will merge.

pelwell:
That looks OK to me.

leo60228:
@pelwell Is it possible to use the Alt5 JTAG pins? I'm using the Alt4 pins for something else already.

PR #1481, opened 2020-04-22
config.txt: Document `armstub` and `arm_peri_high`.

hvenev:


configuration/config-txt/boot.md
```diff
@@ -49,6 +49,16 @@ Note that the 64-bit kernel will only work on the Pi4, Pi3, and Pi2B rev1.2 boar
...
 
+## armstub
+
+`armstub` is the filename on the boot partition from which to load the ARM stub. The default ARM stub is stored in firmware and is selected automatically based on the Pi model and various settings.
+
+The stub is a small piece of ARM code that is run before the kernel. Its job is to set up low-level hardware like the interrupt controller before passing control to the kernel.
+
+## arm_peri_high
+
+Enables "High Peripheral" mode on the Pi 4. A suitable ARM stub and device tree must be installed, or your system will fail to boot.
```
lurch, line 60: In line with the other config.txt option-descriptions, should this say something like:
Set `arm_peri_high` to `1` to enable "High Peripheral" mode on the Pi 4.

?

lurch:
ping @pelwell for comment

pelwell:
You can change the wording to match the style of other entries, but the information is correct. It should also say that arm_peri_high is automatically set to 1 if a suitable DTB is loaded - the override from config.txt is more for bare metal kids.

Issue #1423, opened 2020-03-01
Bridged Access Point instructions do not work for me


Taikuh:
I understand this repo is for documentation, so I will try to keep this issue within that scope. I will further research and ask around the forums to find out more about the technical side of things.

The current master's steps ([here](https://github.com/raspberrypi/documentation/blob/master/configuration/wireless/access-point.md#using-the-raspberry-pi-as-an-access-point-to-share-an-internet-connection-bridge)) nor the steps in the next major revision ([here](https://github.com/raspberrypi/documentation/blob/aaf97e8c5c85e1aeb33bdd5e4119e6330a2e2f51/configuration/wireless/access-point-bridged.md)) work for me. Using #1333 as reference, the following modifications have worked for me:

## Added

```
sudo apt install bridge-utils

sudo nano /etc/systemd/network/bridge-br0.network
```

File contents:

```
[Match]
Name=br0

[Network]
Address=192.168.1.3/24
Gateway=192.168.1.1
DNS=8.8.8.8
```

where:

* `Address` serves as the bridge `br0`'s static IP within my LAN. It is outside my router's DHCP range.

* `Gateway` is my router's IP address within my LAN.

The following addition is sourced from [here](https://github.com/SurferTim/documentation/blob/6bc583965254fa292a470990c40b145f553f6b34/configuration/wireless/access-point.md#using-the-raspberry-pi-as-an-access-point-to-share-an-internet-connection):

```
sudo nano /etc/network/interfaces
```

File contents:

```
auto br0
iface br0 inet manual
bridge_ports eth0 wlan0
```

## Removed

The following two files ([here](https://github.com/raspberrypi/documentation/blob/aaf97e8c5c85e1aeb33bdd5e4119e6330a2e2f51/configuration/wireless/access-point-bridged.md#create-a-bridge-device-and-populate-the-bridge)) are not necessary for me. They don't seem to do anything and removing them doesn't seem to affect my end result.

```
/etc/systemd/network/bridge-br0.netdev
/etc/systemd/network/br0-member-eth0.network
```

Again, something is probably technically wrong from my end. It could be something to do with my router's ipv6 settings that I can't seem to disable. Hopefully this issue will serve as a record and hint for others facing similar problems. If there's any way for me to format or present this issue more clearly, please let me know.

lurch:
ping @epoch1970

epoch1970:
[...]

PR #1485, opened 2020-04-24
Update the just updated access-point-bridged.md

epoch1970:
As per a Mar. 7 comment I made on #1333 that didn't get included in the merge.
Declaring "interface br0" in dhcpcd.conf is mandatory I think. The dhcpcd.conf manpage says so, since br is a "virtual" interface. The original write-up worked without it, but no longer. Uncovered in #1423
Thanks.


JamesH65:
Thanks!

Issue #1439, opened 2020-03-18
Adding  telebit.io 


ravish0007:
I would like to add one more way of accessing pi over the Internet by telebit.io which is created by @coolaj86

JamesH65:
On what page?

ravish0007:
This very page raspberrypi/documentation/blob/master/remote-access/access-over-Internet/README.md

JamesH65:
Hmm, got a lot on that page already, not sure we need any of them to be honest. Will check with comms teams.

lurch:
@JamesH65 See also #1133 and #1421 !

JamesH65:
Currently talking to the comms team to see what to do. My inclination is to get rid of that section completely, or just tell people to Google.

phiney0711:
Hi.  I work for "remote.it" that is also listed on this page.  We would hate to lose our link on this page because we've been there for 5 years and tens of thousands of pi users have downloaded and used our solution.  But I understand if that's best for the community.  May I suggest stop adding new solutions to the page instead of deleting it completely?  I agree there are too many at this point.  Thank you for your consideration.

ravish0007:
telebit.io is so incredible and also open source, It's totally free and costs no charges at all.
I got to know on that by randomly watching this video "Look Ma No Cloud" ( https://www.youtube.com/watch?v=E1Q2MWGCADo ), perhaps you might enjoy this :)

JamesH65:
It's been decided that since we cannot review or even list all possible online options we will remove all references to third party systems.

ravish0007:
Sounds acceptable.

phiney0711:
We understand.  It had gotten a little overwhelming in the number of choices.  

Would you consider leaving it up a few more weeks?  The reason is with Corona Virus and the new "work-from-home" trend we are seeing overwhelming interest in setting up remote access via a Raspberry Pi to office LANs.  It seems to be helping a lot of small businesses as a low-cost solution to access their computers on the office LAN via a Raspberry Pi acting as a bridge (without the need to port forward).  Seems many small business do not have IT skills to port forward their gateway and port forwarding a terrible security issue too.  So a PI + remote.it gives full remote access to all devices/computers on office LAN without port forwarding.  Of course please go ahead and remove all the "solutions" on that page as planned whenever things settle down if that's acceptable.  Thanks you.

joshwget:
[...]

PR #1487, opened 2020-04-24
Update updating.md

eitanchud:
Added additional third-party solution that supports OTA updates for Raspbian out-of-the-box.


lurch:
Does https://github.com/raspberrypi/documentation/issues/1439#issuecomment-601694131 also apply here @JamesH65 ?  :shrug:

JamesH65:
Hmm, quite possibly, didn;t know about that section.

eitanchud:
Would be nice if you decide regarding the third party providers topic and apply the decision on all the sections across the documentation. Also if you could share the final decision would be very interesting :)

JamesH65:
Indeed, I will ask the powers above what our current policy is.

JamesH65:
OK, had spoken to the bosses, and we are going to remove third party recommendations/instructions from the site, except in rare cases where that would be detrimental or we have agreements in place.

PR #1488, opened 2020-04-27
Fix example of default location for ssh keys

FelixGoetze:
The ssh keys are generated on the computer that is used to connect to the Raspberry Pi. The default location for the ssh keys most likely isn't '/home/pi/.ssh/id_rsa', but generally is '~/.ssh/id_rsa'. 
This was confusing, since it could be interpreted as having to generate the ssh keys on the Raspberry Pi itself.

### DIFF
--- a/remote-access/ssh/passwordless.md
+++ b/remote-access/ssh/passwordless.md
@@ -20,7 +20,7 @@ To generate new SSH keys enter the following command:
 ssh-keygen
 ```
 
-Upon entering this command, you'll be asked where to save the key. We suggest you save it in the default location (`/home/pi/.ssh/id_rsa`) by just hitting `Enter`.
+Upon entering this command, you'll be asked where to save the key. We suggest you save it in the default location (`~/.ssh/id_rsa`) by just hitting `Enter`.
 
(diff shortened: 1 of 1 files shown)


remote-access/ssh/passwordless.md
```diff
@@ -20,7 +20,7 @@ To generate new SSH keys enter the following command:
 ssh-keygen
 ```
 
-Upon entering this command, you'll be asked where to save the key. We suggest you save it in the default location (`/home/pi/.ssh/id_rsa`) by just hitting `Enter`.
+Upon entering this command, you'll be asked where to save the key. We suggest you save it in the default location (`~/.ssh/id_rsa`) by just hitting `Enter`.
```
andrum99, line 23: I think that sentence could do with rewording to make it sound a bit more 'professional', for example:

'Upon entering this command, you will be asked where to save the key. We suggest saving it in the default location ('~/.ssh/id_rsa`) by pressing `Enter`.

andrum99: Sorry - please ignore the above comment - the page has multiple similar issues. I will file a separate PR once this integrates.

lurch: That's the joy of documentation written by multiple different people, each with their own styles :wink:

PR #1489, opened 2020-04-27
Update updating.md

JamesH65:
Removed a third party reference and replaced with generic advice.

### DIFF
--- a/raspbian/updating.md
+++ b/raspbian/updating.md
@@ -64,7 +64,7 @@ If moving to a new Pi model (for example the Pi 3B+), you may also need to updat
 
 ## Third-party solutions
 
-This section addresses why third-party solutions are of interest and why [apt](../linux/software/apt.md) is not optimal for all situations. It also covers existing third-party solutions that support Raspbian.
+This section addresses why third-party solutions are of interest and why [apt](../linux/software/apt.md) is not optimal for all situations. Raspberry Pi do not recommend any specific third party tools, prospective users should determine the most suitable tool for their particular requirements.
 
 [Apt](../linux/software/apt.md) is a convenient way of updating the software of your device running Raspbian, but the limitation of this method becomes apparent when you have a larger pool of devices to update, and especially when you do not have physical access to your devices and when they are distributed geographically.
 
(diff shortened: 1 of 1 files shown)


raspbian/updating.md
```diff
@@ -64,7 +64,7 @@ If moving to a new Pi model (for example the Pi 3B+), you may also need to updat
 
 ## Third-party solutions
 
-This section addresses why third-party solutions are of interest and why [apt](../linux/software/apt.md) is not optimal for all situations. It also covers existing third-party solutions that support Raspbian.
+This section addresses why third-party solutions are of interest and why [apt](../linux/software/apt.md) is not optimal for all situations. Raspberry Pi do not recommend any specific third party tools, prospective users should determine the most suitable tool for their particular requirements.
```
andrum99, line 67: Comma splice - replace comma after 'tools' with a full stop.

lurch, line 67: I think I'd be tempted to change "why third-party solutions are of interest" to "why third-party solutions may be of interest" ?  And "third party" in the second sentence should probably be hyphenated?

andrum99: Good catch - 'may be' would be a useful addition.

> And "third party" in the second sentence should probably be hyphenated?

I thought so too, but according to the Oxford English Dictionary (see https://www.lexico.com/definition/third_party) the correct spelling is as-is, i.e. 'third party' without the hyphen.

lurch: That page says that it **is** hyphenated when used as an adjective rather than a noun? :stuck_out_tongue_winking_eye:

andrum99: > That page says that it **is** hyphenated when used as an adjective rather than a noun? 😜

Ooops - sorry - I didn't read the page fully. My bad.

JamesH65: Added hyphen, or as I like to say, subtraction sign.

mirzak:
Hi @JamesH65. I was involved in the addition of the "third-party solution" section and notices this change-set.

I was curious to why you removed the third party reference. It has been shown to work well for the community and has been well maintained over the period that it was present.

I understand the following,

> Raspberry Pi do not recommend any specific third party tools, prospective users should determine the most suitable tool for their particular requirements.

But above text together with a third party reference would make sense to me.

JamesH65:
Part of an ongoing effort to remove third party references throughout our documentation. These sections could be construed as recommendations, but since we do not test any third party products, they cannot be seen as such, and therefor, we feel that removal is the best option.

mirzak:
Thanks for getting back to me. It is understandable that you are not able to test third party products, but this should be in the interest of the community that is linked to ensure it works. 

Eg. I was the one who added this section, and I together with our community have ensured that this worked trough out the majority of the Raspbian releases and would have gladly shared test results with you and Raspberry Pi community to ensure a high quality reference. This eliminates the burden of you testing it.

I do not know if this is open for discussions, but I am at least open for suggestions and interested to keep this.

I am curious, how are the "Third Party Operating System Images" handled in the download section, https://www.raspberrypi.org/downloads/? Is this tested by you, or are you relying on the community?

lurch:
@mirzak I think https://github.com/raspberrypi/documentation/commit/9e64f964bd3f8fbb893f0213fecc74cea6599c45 is part of the  change that instigated this policy - we just had a random list of URLs on there, with no way of assessing the quality of the services they were offering, if they worked with the Pi, or even if the URLs still worked.

If you want somewhere to list your service, you could always try the community [eLinux wiki](https://elinux.org/RPi_Hub) ?

PR #1490, opened 2020-04-27
passwordless.md - style and grammar improvements

andrum99:
The page as written used an informal style. Make it a bit more formal, as befits a piece of technical documentation from a professional organisation.

### DIFF
--- a/remote-access/ssh/passwordless.md
+++ b/remote-access/ssh/passwordless.md
@@ -84,9 +84,9 @@ If this does not work, you can get assistance on the [Raspberry Pi forums](/foru
 
 **Note:** you can also send files over SSH using the `scp` command (secure copy). See the [SCP guide](scp.md) for more information.
 
-## Store the passphrase in the macOS Keychain
+## Store the passphrase in the macOS keychain
 
-If you are using macOS, and after verifying that your new key allows you to connect, you have the option of storing the passphrase for your key in the macOS Keychain. This allows you to connect to your Raspberry Pi without entering the passphrase.
+If you are using macOS, and after verifying that your new key allows you to connect, you have the option of storing the passphrase for your key in the macOS keychain. This allows you to connect to your Raspberry Pi without entering the passphrase.
 
 Run the following command to store it in your keychain:
 
--- a/remote-access/ssh/passwordless.md
+++ b/remote-access/ssh/passwordless.md
@@ -10,7 +10,7 @@ First, check whether there are already keys on the computer you are using to con
 ls ~/.ssh
 ```
 
-If you see files named `id_rsa.pub` or `id_dsa.pub` then you have keys set up already, so you can skip the 'Generate new SSH keys' step (or delete these files with `rm id*` and make new keys).
+If you see files named `id_rsa.pub` or `id_dsa.pub` then you have keys set up already, so you can skip the 'Generate new SSH keys' step below.
 
 ## Generate new SSH keys
 
--- a/remote-access/ssh/passwordless.md
+++ b/remote-access/ssh/passwordless.md
@@ -80,7 +80,7 @@ If you see a message "Agent admitted failure to sign using the key" then add you
 ssh-add
 ```
 
-If this did not work, delete your keys with `rm ~/.ssh/id*` and follow the instructions again.
+If this does not work, you can get assistance on the [Raspberry Pi forums](/forums/viewforum.php?f=63).
 
 You can also send files over SSH using the `scp` command (secure copy). See the [SCP guide](scp.md) for more information.
 
--- a/remote-access/ssh/passwordless.md
+++ b/remote-access/ssh/passwordless.md
@@ -56,7 +56,7 @@ ssh-rsa <REALLY LONG STRING OF RANDOM CHARACTERS> user@host
 ## Copy your public key to your Raspberry Pi
 
 
-On the computer which you will be connecting from, append the public key to your `authorized_keys` file on the Raspberry Pi by sending it over SSH:
+Using the computer which you will be connecting from, append the public key to your `authorized_keys` file on the Raspberry Pi by sending it over SSH:
 
 ```bash
 ssh-copy-id <USERNAME>@<IP-ADDRESS>
--- a/remote-access/ssh/passwordless.md
+++ b/remote-access/ssh/passwordless.md
@@ -56,7 +56,7 @@ ssh-rsa <REALLY LONG STRING OF RANDOM CHARACTERS> user@host
 ## Copy your public key to your Raspberry Pi
 
 
-On the computer which you will be connecting from, append the public key to your `authorized_keys` file on the Raspberry Pi, sending it over SSH:
+On the computer which you will be connecting from, append the public key to your `authorized_keys` file on the Raspberry Pi by sending it over SSH:
 
 ```bash
 ssh-copy-id <USERNAME>@<IP-ADDRESS>
--- a/remote-access/ssh/passwordless.md
+++ b/remote-access/ssh/passwordless.md
@@ -10,7 +10,7 @@ First, check whether there are already keys on the computer you are using to con
 ls ~/.ssh
 ```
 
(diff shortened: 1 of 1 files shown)


JamesH65:
LGTM @lurch?

remote-access/ssh/passwordless.md
```diff
@@ -10,7 +10,7 @@ First, check whether there are already keys on the computer you are using to con
 ls ~/.ssh
 ```
 
-If you see files named `id_rsa.pub` or `id_dsa.pub` you have keys set up already, so you can skip the generating keys step (or delete these files with `rm id*` and make new keys).
+If you see files named `id_rsa.pub` or `id_dsa.pub` then you have keys set up already, so you can skip the 'Generate new SSH keys' step (or delete these files with `rm id*` and make new keys).
```
lurch, line 13: Why is this suggesting deleting the old keys? Is it worth adding a warning here that deleting the old keys _might_ lock the user out of other systems if those keys are being used elsewhere?
(Or is this one of those situations where if the keys _were_ already being used, the user would know they shouldn't delete them?)

remote-access/ssh/passwordless.md
```diff
@@ -38,7 +38,7 @@ authorized_keys  id_rsa  id_rsa.pub  known_hosts
 
 The `id_rsa` file is your private key. Keep this on your computer.
 
-The `id_rsa.pub` file is your public key. This is what you share with machines you want to connect to. When the machine you try to connect to matches up your public and private key, it will allow you to connect.
+The `id_rsa.pub` file is your public key. This is what you share with machines that you connect to: in this case your Raspberry Pi. When the machine you try to connect to matches up your public and private key, it will allow you to connect.
```
lurch, line 41: "you try" -> "you are trying" ?
"matches up" -> "successfully matches up" ?

remote-access/ssh/passwordless.md
```diff
@@ -55,21 +55,22 @@ ssh-rsa <REALLY LONG STRING OF RANDOM CHARACTERS> user@host
 <a name="copy-your-public-key-to-your-raspberry-pi"></a>
 ## Copy your public key to your Raspberry Pi
 
-To copy your public key to your Raspberry Pi, use the following command, on the computer you will be connecting from, to append the public key to your `authorized_keys` file on the Pi, sending it over SSH:
+
+On the computer which you will be connecting from, append the public key to your `authorized_keys` file on the Raspberry Pi by sending it over SSH:
```
lurch, line 59: "On the computer which you will be connecting from, append the public key ..." - might sound like you're appending the public key _on_ the computer your connecting from? Does this alternative work better?
"On the computer which you will be connecting from, run the following command to append the public key ..."

And perhaps it's worth changing "your `authorized_keys` file" to "the `authorized_keys` file" ? :shrug:  (or "your user's `authorized_keys` file" ?)

remote-access/ssh/passwordless.md
```diff
@@ -83,9 +84,9 @@ If this did not work, delete your keys with `rm ~/.ssh/id*` and follow the instr
 
 You can also send files over SSH using the `scp` command (secure copy). See the [SCP guide](scp.md) for more information.
 
-## Let macOS store your passphrase so you don't have to enter it each time
+## Store the passphrase in the macOS Keychain
 
-If you're using macOS and after verifying that your new key allows you to connect, you can optionally choose to store the passphrase for your key in the macOS Keychain. This will make it so that you don't have to enter the passphrase each time you connect to your Pi.
+If you are using macOS, and after verifying that your new key allows you to connect, you can optionally choose to store the passphrase for your key in the macOS Keychain. This allows you to connect to your Raspberry Pi without entering the passphrase.
```
lurch, line 89: "the passphrase for your key" -> "the passphrase for your key (if you've set one)" ?

remote-access/ssh/passwordless.md
andrum99: I've no idea. Seems a bit daft to me. I will remove that bit as it is both unnecessary, and liable to trip up noobs.

remote-access/ssh/passwordless.md
andrum99: I don't think that first one is necessary - just adds more words.
Similarly, 'successfully' would be kind of redundant, and makes it more wordy.

It may be sensible to have a copy editor to look at this page anyway.

remote-access/ssh/passwordless.md
andrum99: I think that needs the first word of the sentence change from 'on' to 'using'.

remote-access/ssh/passwordless.md
andrum99: It should be obvious from the section title that this only applies in the case where the user has set a passphrase. You can't ask macOS to store a passphrase if there isn't one.

remote-access/ssh/passwordless.md
JamesH65: We don't have the CE resource to do any serious (i.e. not just you two guys and me!) copy editing at the moment. I guess we need to start some sort of list of jobs that will need doing when CE resource becomes available, if ever. @aallan is going to be in charge of all documentation in the future, so I guess he is the best place for that.

andrum99: I see. Realistically, such a list would probably be something like "the whole documentation repository", since most of it was written without CE input.

A publicly accessible style guide would be useful - there was one back in the day, but I can't find it now.

I wonder if it would be possible to get hold of a copy editor to work on the documentation on a voluntary basis? I mean a professional copy editor, who is looking to volunteer their services.

JamesH65: Actually, quite a few of the docs have been CE'd - our process before our CE person left to work full time for the Foundation meant all changes for probably 18 months went through CE. - so not only were changes checked, but the entire page in which they appeared as well. We do have an internal style guide, wonder if that could be repurposed specifically for the docs repo. 

I suspect getting a CE person to work for free won't be easy!

remote-access/ssh/passwordless.md
lurch: Meh. It may be obvious to experienced users like yourself, but IMVHO it's always better to be explicit to avoid doubt / prevent confusion :slightly_smiling_face:

remote-access/ssh/passwordless.md
```diff
@@ -79,13 +80,13 @@ If you see a message "Agent admitted failure to sign using the key" then add you
 ssh-add
 ```
 
-If this did not work, delete your keys with `rm ~/.ssh/id*` and follow the instructions again.
+If this does not work, you can get assistance on the [Raspberry Pi forums](/forums/viewforum.php?f=63).
```
lurch, line 83: IMHO this URL should be `https://www.raspberrypi.org/forums/` so that the links work both when the documentation is being viewed via https://www.raspberrypi.org/documentation and also when being viewed via https://github.com/raspberrypi/documentation

Also, I dunno if we want to be linking to a specific sub-forum rather than just "the forums in general"? @JamesH65 ?

remote-access/ssh/passwordless.md
andrum99: I see. That's interesting. One possible avenue is goodmoves.org.uk - it's mostly for paid 3rd sector roles, but it also has volunteering opportunities for professionals on it. Or maybe if there is a national "3rd sector interface" that might be better. Although perhaps there's not enough for a volunteer copy editor to do, given someone, presumably a paid employee, would need to manage them.

andrum99: Definitely a style guide if you can provide a public version. I'm not sure anyone else would use it, but I'm sure there's stuff in there that would assist me, and perhaps save some of the back-and-forth we've had on some things, which takes up the time of paid employees.

remote-access/ssh/passwordless.md
andrum99: I linked to the general forum, as I thought a noob might be put off by the fact that there are so many different forums. I suspect it is actually best to do as you suggested though. I've a feeling the docs usually just link to raspberrypi.org/forums. Let me know if you want this changed and I will submit an additional PR. (Since this one has already integrated).

lurch: No worries, I'll edit it directly.

PR #1493, opened 2020-04-27
Fix instructions for enabling camera

olliebennett:
This reflects the latest `raspi-config` menu navigation required to enable the camera interface.

Not sure if there still needs to be some mention of these "old" instructions, or whether they're simply incorrect.

### DIFF
--- a/usage/camera/README.md
+++ b/usage/camera/README.md
@@ -20,7 +20,7 @@ Open the `raspi-config` tool from the Terminal:
 sudo raspi-config
 ```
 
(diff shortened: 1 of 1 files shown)


andrum99:
A quick search of https://github.com/RPi-Distro/raspi-config reveals that the "interfacing options" menu was added on 30th November 2016, which is almost 4 years ago. So the old instructions are no longer needed.

JamesH65:
Thanks!

PR #1494, opened 2020-04-28
Remote wireless access update

JamesH65:
Changes to text to be more generic and adds link to the remote access page.

### DIFF
--- a/configuration/wireless/headless.md
+++ b/configuration/wireless/headless.md
@@ -24,9 +24,7 @@ Note that some older WiFi dongles don't support 5GHz networks.
 
 More information on the `wpa_supplicant.conf` file can be found [here](wireless-cli.md). See [Wikipedia](https://en.wikipedia.org/wiki/ISO_3166-1) for a list of 2 letter ISO 3166-1 country codes.
 
-## Enabling SSH
+## Remote Access
 
-SSH can be enabled by placing a file called `ssh` in to the boot folder. This flags the Pi to enable the SSH system on the next boot.
-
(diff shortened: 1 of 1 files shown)


configuration/wireless/headless.md
```diff
@@ -24,9 +24,7 @@ Note that some older WiFi dongles don't support 5GHz networks.
 
 More information on the `wpa_supplicant.conf` file can be found [here](wireless-cli.md). See [Wikipedia](https://en.wikipedia.org/wiki/ISO_3166-1) for a list of 2 letter ISO 3166-1 country codes.
 
-## Enabling SSH
+## Remote Access
 
-SSH can be enabled by placing a file called `ssh` in to the boot folder. This flags the Pi to enable the SSH system on the next boot.
-
-See [here](../../remote-access/ssh/README.md#3-enable-ssh-on-a-headless-raspberry-pi-add-file-to-sd-card-on-another-machine) for more details.
+With no keyboard or monitor, you will need some way of accessing the headless Raspberry Pi. There are a number of ways of doing this, and details can be found [here](../../remote-access/README.md).
```
lurch, line 29: I like the change of the second sentence to being more generic, but I wonder if it's worth keeping the first sentence? The stuff about adding a file named `ssh` to the boot folder may be a bit tricky to find otherwise, and I think that most of the remote-access methods require a command-line to setup anyway? And in a headless setup SSH seems to be the only way of getting a command line? :shrug:

JamesH65: The remote-access page has all the ssh stuff in it, so having it here was replication. This paragraph is all about remote access, not specifically SSH although that is a common method.

lurch: Fair enough.

PR #1503, opened 2020-05-01
Add link to scripting

jcjveraa:
There's a reference to scripting without a convenient link


JamesH65:
That scripting page is about as much use as a chocolate teapot, but at least this links to something. So thanks!

PR #1513, opened 2020-05-04
installing-images/mac: Don't use conv=sync with dd

Frogging101:
The conv=sync option is for padding input blocks to the input block size.
This is not what we want. OS X's dd does not have the oflag=fsync option
that we would use on we would use on GNU/Linux, but calling "sync" after dd
is an option.

Fixes: 4dbfa3508837 ("installing-images: dd - use conv=fsync and conv=sync for Linux and Mac (#695)")
Signed-off-by: John Brooks <john@fastquake.com>

### DIFF
--- a/installation/installing-images/mac.md
+++ b/installation/installing-images/mac.md
@@ -61,7 +61,7 @@ If you specify the wrong device in the instructions, you could overwrite your pr
 - Copy the image
 
   ```bash
-  sudo dd bs=1m if=path_of_your_image.img of=/dev/rdiskN conv=sync
+  sudo dd bs=1m if=path_of_your_image.img of=/dev/rdiskN; sync
   ```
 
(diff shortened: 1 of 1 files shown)


XECDesign:
If the image size is not a multiple of the block size, would removing this option cause the final block to be cut off or not?

Frogging101:
> If the image size is not a multiple of the block size, would removing this option cause the final block to be cut off or not?

It won't be cut off, it just won't pad the output to the block size with NULs. Which is fine; that's unnecessary anyway, and I believe that the intent of the original change was to ensure that the write was fully synced before finishing.

XECDesign:
That was the intent, but I have a vague recollection of various tools failing when we released an image of an odd size, so I just wanted to check.

LGTM then

JamesH65:
Thanks!

PR #1516, opened 2020-05-04
Fix documented sensor size for new camera

pstray:
The sensor size was listed as the size of the whole chip.  New numbers
are calulated from usable resolution and cell size.


JamesH65:
Than looks better - put that stuff together in a hurry unfortunately! Thanks for the update.

PR #1523, opened 2020-05-05

timg236:
Reference config.txt and GPIO filters

Bump the version number for network boot to reference the latest production critical update.

### DIFF
--- a/hardware/raspberrypi/bcm2711_bootloader_config.md
+++ b/hardware/raspberrypi/bcm2711_bootloader_config.md
@@ -195,7 +195,7 @@ vcgencmd otp_dump | grep 28: | sed s/.*://g
 ```
 
 ### Installation - firmware update
-Network boot is availble in the latest production bootloader (pieeprom-2020-04-16.bin). Previous beta/stable releases which supported network boot are still available but are deprecated.
+Network boot is available in the latest production bootloader (pieeprom-2020-04-16.bin). Previous beta/stable releases which supported network boot are still available but are deprecated.
 
 ```
 # Update the rpi-eeprom package                                                                                                 
--- a/hardware/raspberrypi/bcm2711_bootloader_config.md
+++ b/hardware/raspberrypi/bcm2711_bootloader_config.md
@@ -32,8 +32,8 @@ sudo rpi-eeprom-update -d -f ./pieeprom-new.bin
 sudo reboot
 ```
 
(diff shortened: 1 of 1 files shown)


JamesH65:
I wonder if there is a better way of providing version numbers, rather than having to edit all those each times it changes. Or have I misunderstood the purpose/interpretation? Perhaps a 'macro' LATEST at top of page, and use that? Dunno. @lurch any thoughts?

timg236:
Those numbers won’t need to change. It’s the first critical release where that feature works or for beta the it’s the first beta release. But we could just have a beta section for experimental stuff where the expectation is that you always use latest beta

JamesH65:
Cool, in that case. LGTM.

hardware/raspberrypi/bcm2711_bootloader_config.md
```diff
@@ -194,24 +195,23 @@ vcgencmd otp_dump | grep 28: | sed s/.*://g
...
 sudo apt upgrade
-                                                                          
-# Network boot is only available in the stable and beta releases.
-# Switch from critical updates to the stable release series.
-# As root:-
-echo FIRMWARE_RELEASE_STATUS="stable" > /etc/default/rpi-eeprom-update
+# Check the current version
+sudo rpi-eeprom-update     
+# Update to latest
+sudo rpi-eeprom-update
```
lurch, line 207: Is this correct? It seems to be showing the same command for both "Check the current version" and "Update to latest" ??

hardware/raspberrypi/bcm2711_bootloader_config.md
```diff
@@ -32,8 +32,8 @@ sudo rpi-eeprom-update -d -f ./pieeprom-new.bin
 sudo reboot
 ```
 
-## Configuration Items
-This section describes all the configuration items available in the bootloader.
+## Configuration Properties
+This section describes all the configuration items available in the bootloader. The syntax is the same as [config.txt](../../configuration/config-txt/) but the properties are specific to the bootloader. [Conditional filters](../../configuration/config-txt/conditional.md) are also supported except for EDID.
```
lurch, line 36: I guess the `[pi2]` filter isn't of much use though :wink:   Are the `gpio` filters available, allowing you to dynamically select which parts of the EEPROM config get used? That's pretty funky if so.

PR #1525, opened 2020-05-05
Update misc.md

timg236:


configuration/config-txt/misc.md
```diff
@@ -17,6 +17,8 @@ Causes the content of the specified file to be inserted into the current file.
 
 For example, adding the line `include extraconfig.txt` to `config.txt` will include the content of `extraconfig.txt` file in the `config.txt` file.
 
+**Include directives are not supported by bootcode.bin or the EEPROM bootloader**
```
lurch, line 20: Cool. Would mentioning which (limited set?) of directives are processed by `bootcode.bin` (and so can't be used inside an include-file) be useful here? Or would that list be liable to change at any time?

timg236: not really, the boot loader config.txt options are already mentioned elsewhere. This is just a generic syntax thing, which could be moved to the syntax page but not really worth shuffling stuff around

PR #1528, opened 2020-05-05
Update building.md

bg-master:
Add gcc-arm-linux-gnueabihf dependency becasue with Ubuntu 20.04 the build failed:
`make: arm-linux-gnueabihf-gcc: Command not found`


lurch:
Did you miss the bit just above where it tells you how to install the toolchain? :confused:

JamesH65:
@bg-master Are you saying that the cloning of the toolchain and setting the path, as mentioned in  a previous paragraph, doesn't work and you need this extra dependency?

bg-master:
D'oh! I did a mistake, after a reboot didn't update the $PATH environment with `source ~/.bashrc` and so the modules installation failed. Sorry, closing the PR.

PR #1534, opened 2020-05-07
Implied installation of git

JJTech0130:
git clone was used earlier in the tutorial, so it’s installation is implied (should the install line be moved earlier in the tutorial?)


JamesH65:
Made the changes as suggested, @lurch for a quick check.

JamesH65:
Thanks for pointing this out!

Issue #1535, opened 2020-05-07
BCM2836 listed as compatible with 64-bit kernel on configuration/boot_folder


trejan:
BCM2836 from the original Pi 2 has Cortex-A7 cores which don't support ARMv8 instructions. However, it is listed as supporting the 64-bit kernel8.img on https://github.com/raspberrypi/documentation/blob/master/configuration/boot_folder.md

Only the Pi 2 v1.2 with BCM2837 should be listed as compatible with kernel8.img

lurch:
Fixed by #1538

PR #1538, opened 2020-05-12
boot_folder.md: BCM2836 doesn't do 64-bit

andrum99:
Correction per #1535.

Also, some minor tweaks to the same table.


JamesH65:
Thanks!

PR #1542, opened 2020-05-13
Corrections to digital gain and shutter speed ranges

6by9:
Digital gain is checked in the code for being <=64.0, not 255.0.
Shutter speed on v2 camera module can go up to 10s, not 6.


JamesH65:
I'm wondering whether we should have a table in there, with max and min for all the cameras. I'll merge this as is, then do that I think.

PR #1543, opened 2020-05-13
Use a table for shutter speeds

JamesH65:


raspbian/applications/camera.md
```diff
@@ -246,10 +246,18 @@ Allows the specification of the area of the sensor to be used as the source for
...
 
 ```
-	--shutter,	-ss		Set shutter speed
+	--shutter,	-ss		Set shutter speed/time
 ```
 
-Sets the shutter speed to the specified value (in microseconds). On Camera Module v1 there is an upper limit of approximately 6000000us (6000ms, 6s), and 10000000us (10000ms, 10s) on Camera Module v2, past which operation is undefined. On the HQ Camera exposure times can be up to 200000000us (200000ms, 200s)
+Sets the shutter open time to the specified value (in microseconds). Shutter speed limits are as follows:
+
+| Camera Version | Max (s) |  Max (microseconds) |
```
lurch, line 254: Might be worth clarifying "Max (s)" as "Max (seconds)" ? :shrug:  Maybe it makes sense to list the microseconds column first, given that that's the value that `-ss` expects?

Is there also a "minimum" shutter speed, or is that a nonsensical question? (I know very little about cameras :wink: )

JamesH65: I just moved the microsecond one to the last column!!! I thought it was more obvious that way round! Clearly it depends on the reader. It also originally had a minimum value, but not yet confirmed the values.

lurch: I guess an alternative might be something like:

| Camera Version |  Max microseconds            |
|----------------|------------------------------| 
| V1 (OV5647)    | 6000000 (i.e. 6 seconds)     |
| V2 (IMX219)    | 10000000 (i.e. 10 seconds)   |
| HQ (IMX477)    | 200000000 (i.e. 200 seconds) |

lurch: And if we're _not_ supplying a minimum value, the sentence under the table should probably be changed to "Using values above these maximums will result in undefined behaviour." ?

JamesH65: Yes, like that version, will change.

lurch:
LGTM, but should we get a final okay from @6by9 ?

PR #1545, opened 2020-05-14
update tvservice.md

Yemtex:
updated -m
there is no CVT group
changed it to CEA


JamesH65:
Wow, good spot! Must have been wrong for quite a while! Thanks!

PR #1550, opened 2020-05-18
WIP: Use visudo to edit any sudoers file

andrum99:
See #1107. 

There are currently 3 pages in this repo which have problems to do with the unsafe editing of sudo-related files:

- /linux/usage/users.md
- /linux/usage/root.md
- /configuration/security.md

There is also /usage/terminal/README.md which contains slightly wrong information (doesn't mention Raspbian has no sudo password on the pi user by default), and a grammatical error (comma splice).

This PR fixes them.

### DIFF
--- a/configuration/security.md
+++ b/configuration/security.md
@@ -90,7 +90,7 @@ and change the `pi` entry (or whichever usernames have superuser rights) to:
 pi ALL=(ALL) PASSWD: ALL
 ```
 
-Then save the file: it will be checked for any syntax errors. If no errors were detected, the file will be saved and you will be returned to the shell prompt (e.g. `pi@raspberry:~ $`). If errors were detected, you will be asked 'what now?' Press the 'enter' key on your keyboard: this will bring up a list of options. You will probably want to use press 'e' for '(e)dit sudoers file again,' so you can edit the file and fix the problem. **Note that choosing option 'Q' will save the file with any syntax errors still in place, which makes it impossible for _any_ user to use the sudo command.**
+Then save the file: it will be checked for any syntax errors. If no errors were detected, the file will be saved and you will be returned to the shell prompt. If errors were detected, you will be asked 'what now?' Press the 'enter' key on your keyboard: this will bring up a list of options. You will probably want to use 'e' for '(e)dit sudoers file again,' so you can edit the file and fix the problem. **Note that choosing option 'Q' will save the file with any syntax errors still in place, which makes it impossible for _any_ user to use the sudo command.**
 
 ## Ensure you have the latest security fixes
 
--- a/configuration/security.md
+++ b/configuration/security.md
@@ -90,7 +90,7 @@ and change the `pi` entry (or whichever usernames have superuser rights) to:
 pi ALL=(ALL) PASSWD: ALL
 ```
 
-Then save the file: it will be checked for any syntax errors. If not errors were detected, the file will be saved and you will be returned to the shell prompt (`username@host:directory $`). If errors were detected, you will be asked 'what now?' - press enter/return and it will show a list of options. You will probably want to use '(e)dit sudoers file again,' so you can go in and see what the problem is.
+Then save the file: it will be checked for any syntax errors. If no errors were detected, the file will be saved and you will be returned to the shell prompt (`username@host:directory $`). If errors were detected, you will be asked 'what now?' - press enter/return and it will show a list of options. You will probably want to use press 'e' for '(e)dit sudoers file again,' so you can edit the file and fix the problem. **Note that choosing option 'Q' will save the file with any syntax errors still in place, which makes it impossible for _any_ user to use the sudo command.**
 
 ## Ensure you have the latest security fixes
 
--- a/configuration/security.md
+++ b/configuration/security.md
@@ -90,7 +90,7 @@ and change the `pi` entry (or whichever usernames have superuser rights) to:
 pi ALL=(ALL) PASSWD: ALL
 ```
 
-Then save the file: it will be checked for any syntax errors. If no errors were detected, the file will be saved and you will be returned to the shell prompt (e.g. `pi@raspberry:~ $`). If errors were detected, you will be asked 'what now?' - press enter/return and it will show a list of options. You will probably want to use press 'e' for '(e)dit sudoers file again,' so you can edit the file and fix the problem. **Note that choosing option 'Q' will save the file with any syntax errors still in place, which makes it impossible for _any_ user to use the sudo command.**
+Then save the file: it will be checked for any syntax errors. If no errors were detected, the file will be saved and you will be returned to the shell prompt (e.g. `pi@raspberry:~ $`). If errors were detected, you will be asked 'what now?' Press the 'enter' key on your keyboard: this will bring up a list of options. You will probably want to use press 'e' for '(e)dit sudoers file again,' so you can edit the file and fix the problem. **Note that choosing option 'Q' will save the file with any syntax errors still in place, which makes it impossible for _any_ user to use the sudo command.**
 
 ## Ensure you have the latest security fixes
 
--- a/configuration/security.md
+++ b/configuration/security.md
@@ -81,13 +81,13 @@ Placing `sudo` in front of a command runs it as a superuser, and by default, tha
 To force `sudo` to require a password, enter:
 
 ```bash
-sudo nano /etc/sudoers.d/010_pi-nopasswd
+sudo visudo /etc/sudoers.d/010_pi-nopasswd
 ```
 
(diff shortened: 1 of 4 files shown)


andrum99:
Ready for review.

JamesH65:
LGTM, @lurch ?

configuration/security.md
```diff
@@ -81,13 +81,13 @@ Placing `sudo` in front of a command runs it as a superuser, and by default, tha
 To force `sudo` to require a password, enter:
 
 ```bash
-sudo nano /etc/sudoers.d/010_pi-nopasswd
+sudo visudo /etc/sudoers.d/010_pi-nopasswd
```
lurch, line 84: Not related to this change, but I just noticed that the previous paragraph on this page mentions "require superuser credential" which should probably be "require superuser credentials" ?

andrum99: Good spot - fixed.

lurch:
I've not tested this (I never use any user other than `pi`!), but the changes here LGTM.

andrum99:
> I've not tested this (I never use any user other than `pi`!), but the changes here LGTM.

I've had a dig into the history of the various files, to see what existing review there has been on them, as I've not really tested them much myself:

- configuration/security.md: `sudo nano` usage to edit a sudoers file comes from original version of page from 2017.
- linux/usage/root.md - error introduced by 'copy edit', presumably on the advice of someone technical, in 2016.
- linux/usage/users/md - error introduced by one of my PRs in 2019

So lurch makes a good point about testing these things! I will test the instructions on these 3 pages this week at some point, probably tomorrow. In the meantime, I will mark it as WIP.

lurch:
Perhaps @StoneTrue32 would also like to review  / comment on this PR? :slightly_smiling_face:

StoneTrue32:
My pleasure!

configuration/security.md
```diff
@@ -76,21 +76,21 @@ sudo deluser -remove-home pi
...
 
 and change the `pi` entry (or whichever usernames have superuser rights) to:
 
 ```bash
-alice ALL=(ALL) PASSWD: ALL
+pi ALL=(ALL) PASSWD: ALL
 ```
 
-Now save the file.
+Then save the file: it will be checked for any syntax errors. If not errors were detected, the file will be saved and you will be returned to the shell prompt (`username@host:directory $`). If errors were detected, you will be asked 'what now?' - press enter/return and it will show a list of options. You will probably want to use '(e)dit sudoers file again,' so you can go in and see what the problem is.
```
lurch, line 93: "If not errors" ? :laughing: 

Probably no need to show an example shell prompt?
Perhaps "press enter/return" should be changed to "press Enter on your keyboard"?
Does the '(e)dit sudoers file again,' mean that the user need to press the E key on their keyboard? (is that worth clarifying? :shrug: )

"you can go in and see what the problem is" sounds a bit vague - perhaps something like "to edit the file and fix the problem" ?

lurch, line 93: Now that you've got this more detailed explanation of `visudo` (nice :+1: ) perhaps it's worth linking to here from the other pages where visudo is mentioned?

andrum99:
Tested on a clean build of Raspbian w/desktop, brought fully up-to-date using apt.

- /configuration/security.md: sudo visudo invokes nano by default. I've added a brief description of the syntax checks visudo performs before it saves the file. I've also tested the sudo breakage that occurs when the file contains syntax errors - it prevents any user from sudo-ing.

- /linux/usage/root.md: tested behaviour of user not in sudoers list - prompted for their password then told their not in the list. I've removed the passing reference to visudo, since it is safer to just add the user to the sudo group. That's also how it is done for the Pi user - the file /etc/sudoers.d/010_pi-nopasswd is only there so that the pi user does not need to enter their password when sudo-ing.

- /linux/usage/users.md - tested useradd and sudo visudo for sudo setup. Looks OK.

lurch:
Thanks for your investigations!

"the file /etc/sudoers.d/010_pi-nopasswd is only there so that the pi user does not need to enter their password when sudo-ing" - does that imply that simply doing `sudo rm /etc/sudoers.d/010_pi-nopasswd` is potentially a cleaner / easier way of ensuring that the Pi user *does* need to enter a password when using `sudo` ? :man_shrugging:  Or is the interaction more subtle than that?

linux/usage/root.md
```diff
@@ -18,6 +18,6 @@ You can also run a superuser shell by using `sudo su`. When running commands as
 
 ## Who can use sudo?
 
-It would defeat the point of the security if anyone could just put `sudo` in front of their commands, so only approved users can use `sudo` to gain administrator privileges. The `pi` user is included in the `sudoers` file of approved users. To allow other users to act as a superuser you can add the user to the `sudo` group with `usermod`, edit the `/etc/sudoers` file, or add them using `visudo`.
+It would defeat the point of the security if anyone could just put `sudo` in front of their commands, so only approved users can use `sudo` to gain administrator privileges. The `pi` user is included in the `sudoers` file of approved users. To allow other users to act as a superuser, add the user to the `sudo` group with `usermod`.
```
lurch, line 21: > add the user to the `sudo` group with `usermod`.

Maybe it's worth providing an example command of how you'd do that?

andrum99:
> "the file /etc/sudoers.d/010_pi-nopasswd is only there so that the pi user does not need to enter their password when sudo-ing" - does that imply that simply doing `sudo rm /etc/sudoers.d/010_pi-nopasswd` is potentially a cleaner / easier way of ensuring that the Pi user _does_ need to enter a password when using `sudo` ? 🤷‍♂️ Or is the interaction more subtle than that?

It is _slightly_ more subtle than that, in that what is in the 010_pi-nopasswd file is enough to give the pi user sudo rights without them being a member of the sudo group. I would rather not recommend rm-ing that file, since the whole point of this PR is to make the instructions safer and less likely to cause the user to end up with a system with a broken sudoers setup. It does mean that the instructions recommend leaving the file with a misleading name (010_pi-nopasswd would tend to imply that the user won't need a passwd, even if the user edits it to change NOPASSWD to PASSWD, which means they **do** need to enter a password).

If someone wants to go beyond what is in the docs and investigate things further, they can do that.

linux/usage/root.md
andrum99: I would rather not get bogged down with adding more documentation about sudo to that page. That paragraph at least gives the reader a starting point - it's not intended to give detailed instructions. They can ask on the forums, or read the man page (linux/usage/commands.md tells them about 'man').

andrum99:
- /usage/terminal/README.md - looks OK, refers reader to /linux/usage/root.md for more information about sudo.

configuration/security.md
andrum99: I've copied that paragraph into linux/usage/users.md, which is the only other explanation of how to use visudo.

configuration/security.md
```diff
@@ -76,21 +76,21 @@ sudo deluser -remove-home pi
...
 
 and change the `pi` entry (or whichever usernames have superuser rights) to:
 
 ```bash
-alice ALL=(ALL) PASSWD: ALL
+pi ALL=(ALL) PASSWD: ALL
 ```
 
-Now save the file.
+Then save the file: it will be checked for any syntax errors. If no errors were detected, the file will be saved and you will be returned to the shell prompt (e.g. `pi@raspberry:~ $`). If errors were detected, you will be asked 'what now?' Press the 'enter' key on your keyboard: this will bring up a list of options. You will probably want to use press 'e' for '(e)dit sudoers file again,' so you can edit the file and fix the problem. **Note that choosing option 'Q' will save the file with any syntax errors still in place, which makes it impossible for _any_ user to use the sudo command.**
```
lurch, line 93: IMHO I still don't think the " (e.g. `pi@raspberry:~ $`)" is necessary.

"want to use press" ??

Good shout on pointing out what goes wrong if you ignore the `visudo` warning! :+1:

configuration/security.md
lurch: I hate seeing duplication in the documentation, as when one location gets improved / fixed later the other location(s) inevitably get missed :confused: 
Although I guess the `visudo` command isn't going to be changing anytime soon, so perhaps that can be ignored in this case?

lurch:
LGTM! :+1:

JamesH65:
LGTM2 so merging.

andrum99:
Thanks. I was going to leave it for 7 to 14 days to see if StoneTrue32 had any comments, but I think it is probably good enough as it is.

StoneTrue32:
Looks great to me!  You guys are awesome - I really appreciate the combing through for other mentions of potentially the same issue.  I even see where it's on the web page.  Thanks for taking care of this!

PR #1555, opened 2020-05-21
Update camera.md

JamesH65:
Update to -t and --stats

### DIFF
--- a/raspbian/applications/camera.md
+++ b/raspbian/applications/camera.md
@@ -273,7 +273,7 @@ DRC changes the images by increasing the range of dark areas, and decreasing the
 By default, DRC is off.
 
 ```
-	--stats,	-st		Display image statistics
+	--stats,	-st		Image statistics use stills capture
 ```
 
 Force recomputation of statistics on stills capture pass. Digital gain and AWB are recomputed based on the actual capture frame statistics, rather than the preceding preview frame.
@@ -448,7 +448,7 @@ Outputs debugging/information messages during the program run.
 	--timeout,	-t		Time before the camera takes picture and shuts down
 ```
 
-The program will run for this length of time, in milliseconds, then take the capture (if output is specified). If not specified, this is set to 5 seconds (-t 5000). Note that low values (less than 500ms, although it does depend on other settings) may not give enough time for the camera to start up and provide enough frames for the automatic algorithms like AWB and AGC to provide accurate results.
+The program will run for this length of time, in milliseconds, then take the capture, which will be saved if an output is specified. If a timeout is not specified, then it is set to 5 seconds (-t 5000). Note that low values (less than 500ms, although it can depend on other settings) may not give enough time for the camera to start up and provide enough frames for the automatic algorithms like AWB and AGC to provide accurate results.
 
 If set to 0, the preview will run indefinitely, until stopped with CTRL-C. In this case no capture is made. 
 
--- a/raspbian/applications/camera.md
+++ b/raspbian/applications/camera.md
@@ -273,7 +273,7 @@ DRC changes the images by increasing the range of dark areas, and decreasing the
 By default, DRC is off.
 
 ```
-	--stats,	-st		Image statistics use stills capture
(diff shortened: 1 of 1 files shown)


raspbian/applications/camera.md
```diff
@@ -448,7 +448,9 @@ Outputs debugging/information messages during the program run.
 	--timeout,	-t		Time before the camera takes picture and shuts down
 ```
 
-The program will run for this length of time, then take the capture (if output is specified). If not specified, this is set to 5 seconds.
+The program will run for this length of time, in milliseconds, then take the capture (if output is specified). If not specified, this is set to 5 seconds (-t 5000). Note that low values (less than 500ms, although it does depend on other settings) may not give enough time for the camera to start up and provide enough frames for the automatic algorithms like AWB and AGC to provide accurate results.
```
lurch, line 451: I wonder if the first sentence ending with "... (if output is specified)" while the second sentence starts with "If not specified, ..." might potentially cause confusion? :shrug:  Would changing the second sentence to start with "If the timeout is not specified, ..." be better?

Maybe "it does depend on other settings" sounds better as "it also depends on other settings" ?

raspbian/applications/camera.md
```diff
@@ -276,7 +276,7 @@ By default, DRC is off.
 	--stats,	-st		Display image statistics
 ```
 
-Displays the exposure, analogue and digital gains, and AWB settings used.
+Force recomputation of statistics on stills capture pass. Digital gain and AWB are recomputed based on the actual capture frame statistics, rather than the preceding preview frame.
```
lurch, line 279: Oh, change in behaviour or just more accurate description of what the option does? Are the stats also still displayed?

lurch:
Minor comments, but LGTM otherwise.

raspbian/applications/camera.md
JamesH65: I think when I wrote up the first version I guessed what it did and never came back to correct it.

raspbian/applications/camera.md
JamesH65: Have changed it around a bit.

raspbian/applications/camera.md
```diff
@@ -273,10 +273,10 @@ DRC changes the images by increasing the range of dark areas, and decreasing the
 By default, DRC is off.
 
 ```
-	--stats,	-st		Display image statistics
+	--stats,	-st		Image statistics use stills capture
```
lurch, line 276: "Image statistics use stills capture" sounds like a nonsense-phrase?

Would something like "Use current-frame statistics for stills capture" be technically correct? :shrug:

raspbian/applications/camera.md
```diff
@@ -448,7 +448,9 @@ Outputs debugging/information messages during the program run.
 	--timeout,	-t		Time before the camera takes picture and shuts down
 ```
 
-The program will run for this length of time, then take the capture (if output is specified). If not specified, this is set to 5 seconds.
+The program will run for this length of time, in milliseconds, then take the capture, which will be saved if an output is specified. If a timeout is not specified, then it is set to 5 seconds (-t 5000). Note that low values (less than 500ms, although it can depend on other settings) may not give enough time for the camera to start up and provide enough frames for the automatic algorithms like AWB and AGC to provide accurate results.
```
lurch, line 451: Feels like maybe there's too many commas in the first sentence? Only a minor niggle though.

"may not give enough time for the camera to start up and provide enough frames for the automatic algorithms like AWB and AGC to provide accurate results" - is that exactly what the `--stats` option is for? (if so, maybe it's worth mentioning that? Or are they slightly different things?)

raspbian/applications/camera.md
JamesH65: Made change which probably makes more sense.

raspbian/applications/camera.md
JamesH65: Different things.

JamesH65: Have rewritten first sentence.

lurch:
:+1:

Issue #1461, opened 2020-04-08
UART documentation does not mention RPi4


rgov:
https://www.raspberrypi.org/documentation/configuration/uart.md says:

> By default, on Raspberry Pis equipped with the wireless/Bluetooth module (Raspberry Pi 3 and Raspberry Pi Zero W)

I believe the RPi4 does have a wireless/Bluetooth module, so this should be expanded to clarify.

rgov:
Generally I find this document very confusing. I've had to read it many times. I'll probably contribute a revision that simplifies it somewhat.

One point I'm confused about is if `enable_uart=1` then the core frequency is fixed to 250MHz so the Mini UART can have a constant baud rate.

But then it says if Mini UART is not the primary UART, you must add `core_freq=250`... to fix the core frequency to 250 MHz so that the Mini UART can have a constant baud rate?

----

The following table summarizes the default configuration of the UARTs.

| Model               | mini UART            | PL011                 |
| ------------------- | -------------------- | --------------------- |
| Raspberry Pi 1      | secondary            | primary               |
| Raspberry Pi 2      | secondary            | primary               |
| Raspberry Pi 3      | primary (disabled)   | secondary (Bluetooth) |
| Raspberry Pi 4      | primary (disabled)   | secondary (Bluetooth) | 
| Raspberry Pi Zero   | secondary            | primary               |
| Raspberry Pi Zero W | primary (disabled)   | secondary (Bluetooth) |
 
In Linux device terms:

| Linux device   | Description                  |
| -------------- | ---------------------------- |
| `/dev/ttyS0`   | mini UART                    |
| `/dev/ttyAMA0` | PL011                        |
| `/dev/serial0` | primary UART (Linux console) |
| `/dev/serial1` | secondary UART               |

JamesH65:
Looks like a number of issues on that page, since its was written before the Pi4 came out. That has different core frequencies etc, and I think more UARTS anyway. I'll need to check what we need to add on this. Thanks for the report.

lurch:
IIRC the 'primary' UART (the one symlinked to `/dev/serial0`) is also the one that appears on GPIOs 14 and 15 (pins 8 and 10) on the Pi's 40-pin header, if you've enabled serial console in `raspi-config`.
And yes, the Pi4 has 6 UARTs in total (one mini-UART and 5 PL011s).

rgov:
Glad to see the page get some attention.

[...]

Issue #1427, opened 2020-03-04
A SSH / UART dongle client for Android suggestion.


green-green-avk:
Um... Not so humble from my side. It's rather a suggestion.
I tried different SSH and UART dongle clients for Android... and found a lack of good input / rendering features.
But what do you think about this one:
https://play.google.com/store/apps/details?id=green_green_avk.anotherterm

Thank you for your patience in advance.

lurch:
Are you suggesting that this should be added to the list of clients on [this page](https://github.com/raspberrypi/documentation/blob/master/remote-access/ssh/android.md) ?

lurch:
Although @JamesH65 do you think it might be worth adding a "serial console" section to https://github.com/raspberrypi/documentation/blob/master/remote-access/README.md ? :man_shrugging:

green-green-avk:
It sounds funny but my initial intent to create the proposed client was the access to Raspberry-Pi through serial to use `raspi-config` and possibly other ncurses software. A long search for a suitable client brings nothing. Another concern was the requirement of a specific screen keyboard, so I just made it built-in.

duckida:
To me that looks like a terminal for the Droid and not an SSH Client

green-green-avk:
Yes, it also has a local pty support.

JamesH65:
> Although @JamesH65 do you think it might be worth adding a "serial console" section to https://github.com/raspberrypi/documentation/blob/master/remote-access/README.md ? 🤷‍♂️

Hmm, does appear to be a gap in the docs there. Might be worth spending some time on that. Not sure we want a list of third party serial terminals though, been trying to get rid of that sort of thing.

aallan:
Policy is no third party products. Closing.

PR #1560, opened 2020-05-23
configuration/uart.md - Make page easier to understand

andrum99:
Ryan Govostes (@rgov) points out, in #1461, that this page is hard to understand, which it is. This PR attempts to make it easier to understand by adding the table he proposed, and some other cleanup.

### DIFF
--- a/configuration/uart.md
+++ b/configuration/uart.md
@@ -1,8 +1,40 @@
-# The Raspberry Pi UARTs
+# UARTs on the Raspberry Pi
 
-The SoCs used on the Raspberry Pi have two built-in UARTs, a [PL011](http://infocenter.arm.com/help/index.jsp?topic=/com.arm.doc.ddi0183g/index.html) and a mini UART. They are implemented using different hardware blocks, so they have slightly different characteristics. However, both are 3.3V devices, which means extra care must be taken when connecting up to an RS232 or other system that utilises different voltage levels. An adapter must be used to convert the voltage levels between the two protocols. Alternatively, 3.3V USB UART adapters can be purchased for very low prices. 
- 
-By default, on Raspberry Pis equipped with the combined wireless LAN/Bluetooth module, the PL011 is connected to the Bluetooth module, while the mini UART is used as the primary UART and will have a Linux console on it. On all other models, the PL011 is used as the primary UART. 
+There are two types of UART available on the Raspberry Pi -  [PL011](http://infocenter.arm.com/help/index.jsp?topic=/com.arm.doc.ddi0183g/index.html) and mini UART. The PL011 is a capable, broadly 16550-compatible UART, while the mini UART has a reduced feature set.
+
+All UARTs on the Raspberry Pi are 3.3V only - damage will occur if they are connected to 5V systems. An adaptor can be used to connect to 5V systems. alternatively, low-cost USB to 3.3V serial adaptors are available from various third parties.
+
+## Pi Zero, 1, 2 and 3 - two UARTs
+
+The Raspberry Pi Zero, 1, 2 and 3 each contain two UARTs as follows:
+
+| Name | Type |
+|------|------|
+|UART0 |PL011 |
+|UART1 |mini UART |
+
+## Pi 4 - six UARTS
+
+The Raspberry Pi 4 has four additional PL011s:
+
+| Name | Type |
+|------|------|
+|UART0 |PL011 |
+|UART1 |mini UART |
+|UART2 |PL011 |
+|UART3 |PL011 |
+|UART4 |PL011 |
+|UART5 |PL011 |
+
+## Primary UART
+
+On the Raspberry Pi, one UART is selected to be present on GPIO 14 (transmit) and 15 (receive) - this is the primary UART. By default, this will also be the UART on which a Linux console may be present. Note that GPIO 14 is pin 8 on the GPIO header, while GPIO 15 is pin 10.
+
+## Secondary UART
+
+The secondary UART is not normally present on the GPIO connecto. By default, the secondary UART is used to connect to the Bluetooth side of the combined wireless LAN/Bluetooth controller, on models to which this is fitted.
+
+## Configuration
 
 The following table summarises the default configuration of the UARTs:
 
@@ -21,14 +53,14 @@ Linux devices on Raspbian:
 |--------------|-------------|
 |`/dev/ttyS0`  |mini UART    |
 |`/dev/ttyAMA0`|PL011        |
-|`/dev/serial0` |primary UART (Linux console) |
+|`/dev/serial0` |primary UART |
 |`/dev/serial1` |secondary UART |
 
 ## Mini UART and CPU core frequency
 
 The baud rate of the mini UART is linked to the core frequency of the VPU on the GPU. This means that, as the VPU frequency governor varies the core frequency, the baud rate of the mini UART also changes. This makes the mini UART of limited use in the default state. By default, if the mini UART is selected for use as the primary UART, it will be disabled. To enable it, add `enable_uart=1` to config.txt. This will also fix the core frequency to 250MHz (unless `force_turbo` is set, when it will be fixed to the VPU turbo frequency). When the mini UART is not the primary UART, for example you are using it to connect to the Bluetooth controller, you must add `core_freq=250` to config.txt, otherwise the mini UART will not work.
 
-The default value of the `enable_uart` flag depends on the actual roles of the UARTs, so that if the PL011 is assigned to the Bluetooth module, `enable_uart` defaults to 0. If the mini UART is assigned to the Bluetooth module, then `enable_uart` defaults to 1. Note that if the UARTs are reassigned using a Device Tree Overlay (see below), `enable_uart` defaults will still obey this rule.
+The default value of the `enable_uart` flag depends on the actual roles of the UARTs, so that if UART0 (the first PL011) is configured as the secondary UART, `enable_uart` defaults to 0. If the mini UART is configured as the secondary UART, then `enable_uart` defaults to 1. Note that if the UARTs are reassigned using a Device Tree overlay (see 'UARTS and Device Tree' below), `enable_uart` defaults will still obey this rule.
 
 ## Disable Linux serial console
 
@@ -40,10 +72,6 @@ By default, the primary UART is assigned to the Linux console. If you wish to us
 1. At the prompt `Would you like a login shell to be accessible over serial?` answer 'No'
 1. At the prompt `Would you like the serial port hardware to be enabled?` answer 'Yes'
 1. Exit raspi-config and reboot the Pi for changes to take effect.
-
-## UART output on GPIO pins
-
-By default, the UART transmit and receive pins are on GPIO 14 and GPIO 15 respectively, which are pins 8 and 10 on the GPIO header.
 
 ## UARTs and Device Tree
 
@@ -61,6 +89,7 @@ For full instructions on how to use Device Tree overlays see [this page](device-
 dtoverlay=disable-bt
 ...
 ```
+
 ## Relevant differences between PL011 and mini UART
 
 The mini UART has smaller FIFOs. Combined with the lack of flow control, this makes it more prone to losing characters at higher baudrates. It is also generally less capable than the PL011, mainly due to its baud rate link to the VPU clock speed.
--- a/configuration/uart.md
+++ b/configuration/uart.md
@@ -1,4 +1,4 @@
-# UARTs on the Raspberry Pi
+# UART configuration
 
 There are two types of UART available on the Raspberry Pi -  [PL011](http://infocenter.arm.com/help/index.jsp?topic=/com.arm.doc.ddi0183g/index.html) and mini UART. The PL011 is a capable, broadly 16550-compatible UART, while the mini UART has a reduced feature set.
 
--- a/configuration/uart.md
+++ b/configuration/uart.md
@@ -38,14 +38,14 @@ The secondary UART is not normally present on the GPIO connector. By default, th
 
 The following table summarises the default configuration of the first two UARTs:
 
-| Model | mini UART | first PL011 (UART0)|
+| Model | first PL011 (UART0)| mini UART |
 |-------|-----------|-------|
-| Raspberry Pi Zero | secondary | primary |
-| Raspberry Pi Zero W | primary (disabled) | secondary (Bluetooth) |
-| Raspberry Pi 1 | secondary | primary |
-| Raspberry Pi 2 | secondary | primary |
-| Raspberry Pi 3 | primary (disabled) | secondary (Bluetooth) |
-| Raspberry Pi 4 | primary (disabled) | secondary (Bluetooth) |
+| Raspberry Pi Zero |  primary | secondary |
+| Raspberry Pi Zero W | secondary (Bluetooth) | primary (disabled) |
+| Raspberry Pi 1 | primary | secondary |
+| Raspberry Pi 2 | primary | secondary |
+| Raspberry Pi 3 | secondary (Bluetooth) | primary (disabled) |
+| Raspberry Pi 4 | secondary (Bluetooth) | primary (disabled) |
 
 Note: by default, the additional UARTs on the Pi 4 are disabled.
 
--- a/configuration/uart.md
+++ b/configuration/uart.md
@@ -2,7 +2,7 @@
 
 The SoCs used on the Raspberry Pis have two built-in UARTs, a [PL011](http://infocenter.arm.com/help/index.jsp?topic=/com.arm.doc.ddi0183g/index.html) and a mini UART. They are implemented using different hardware blocks, so they have slightly different characteristics. However, both are 3.3V devices, which means extra care must be taken when connecting up to an RS232 or other system that utilises different voltage levels. An adapter must be used to convert the voltage levels between the two protocols. Alternatively, 3.3V USB UART adapters can be purchased for very low prices. 
  
-By default, on Raspberry Pis equipped with the combined wireless LAN/Bluetooth module, the PL011 UART is connected to the Bluetooth module, while the mini UART is used as the primary UART and will have a Linux console on it. On all other models, the PL011 is used as the primary UART. 
+By default, on Raspberry Pis equipped with the combined wireless LAN/Bluetooth module, the PL011 is connected to the Bluetooth module, while the mini UART is used as the primary UART and will have a Linux console on it. On all other models, the PL011 is used as the primary UART. 
 
 The primary UART is assigned to the Linux console.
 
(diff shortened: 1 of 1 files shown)


andrum99:
Per the 2711 datasheet (https://www.raspberrypi.org/documentation/hardware/raspberrypi/bcm2711/rpi_DATA_2711_1p0.pdf) Section 11.5, the PL011s are numbered as:

- UART0 (original PL011 present in previous SoCs)
- UART2
- UART3
- UART4
- UART5

which implies the BCM mini uart is UART1 - is that correct?

configuration/uart.md
```diff
@@ -1,46 +1,61 @@
...
+Linux devices on Raspbian:
 
-The default value of the `enable_uart` flag depends on the actual roles of the UARTs, so that if ttyAMA0 is assigned to the Bluetooth module, `enable_uart` defaults to 0. If the mini UART is assigned to the Bluetooth module, then `enable_uart` defaults to 1. Note that if the UARTs are reassigned using a Device Tree Overlay (see below), `enable_uart` defaults will still obey this rule.
+| Linux device | Description |
+|--------------|-------------|
+|`/dev/ttyS0`  |mini UART    |
+|`/dev/ttyAMA0`|PL011        |
+|`/dev/serial0` |primary UART (Linux console) |
+|`/dev/serial1` |secondary UART |
```
lurch, line 25: It's probably worth explaining that the `/dev/serial` devices are actually symlinks to the other devices? (IIRC the symlinks are auto-adjusted based on the model of Pi and devicetree-settings)

configuration/uart.md
```diff
@@ -1,46 +1,61 @@
...
+The SoCs used on the Raspberry Pi have two built-in UARTs, a [PL011](http://infocenter.arm.com/help/index.jsp?topic=/com.arm.doc.ddi0183g/index.html) and a mini UART. They are implemented using different hardware blocks, so they have slightly different characteristics. However, both are 3.3V devices, which means extra care must be taken when connecting up to an RS232 or other system that utilises different voltage levels. An adapter must be used to convert the voltage levels between the two protocols. Alternatively, 3.3V USB UART adapters can be purchased for very low prices. 
  
-By default, on Raspberry Pis equipped with the wireless/Bluetooth module (Raspberry Pi 3 and Raspberry Pi Zero W), the PL011 UART is connected to the Bluetooth module, while the mini UART is used as the primary UART and will have a Linux console on it. On all other models, the PL011 is used as the primary UART. 
+By default, on Raspberry Pis equipped with the combined wireless LAN/Bluetooth module, the PL011 is connected to the Bluetooth module, while the mini UART is used as the primary UART and will have a Linux console on it. On all other models, the PL011 is used as the primary UART.
```
lurch, line 5: The table below shows that on Pi's with bluetooth, the primary UART is disabled by default. That probably needs mentioning in this paragraph?

configuration/uart.md
```diff
@@ -1,46 +1,61 @@
...
+## Mini UART and CPU core frequency
 
-In a default install of Raspbian, the primary UART (serial0) is assigned to the Linux console. Using the serial port for other purposes requires this default behaviour to be changed. On startup, `systemd` checks the Linux kernel command line for any console entries, and will use the console defined therein. To stop this behaviour, the serial console setting needs to be removed from command line.
+The baud rate of the mini UART is linked to the core frequency of the VPU on the GPU. This means that, as the VPU frequency governor varies the core frequency, the baud rate of the mini UART also changes. This makes the mini UART of limited use in the default state. By default, if the mini UART is selected for use as the primary UART, it will be disabled. To enable it, add `enable_uart=1` to config.txt. This will also fix the core frequency to 250MHz (unless `force_turbo` is set, when it will be fixed to the VPU turbo frequency). When the mini UART is not the primary UART, for example you are using it to connect to the Bluetooth controller, you must add `core_freq=250` to config.txt, otherwise the mini UART will not work.
```
lurch, line 29: "selected for use" implies that this is an action performed by the user, whereas it's actually a function of which model of Pi is being used? Perhaps "if the mini UART is being used for the primary UART" would be better? :man_shrugging:

lurch:
Page 178 of that datasheet says "The BCM2711 device has six UARTs. One mini UART (UART1) and five PL011 UARTs (UART0, UART2, UART3, UART4 & UART5)." :smiley:

lurch:
I think part of what makes this page confusing is the way the UARTs are referred to by different names in different places - I wonder it it'd be worth trying to use more consistent naming throughout the page?

configuration/uart.md
andrum99: I've added mention of them being symlinks. The page already mentioned the primary UART being on the GPIO header.

configuration/uart.md
andrum99: I've removed that paragraph altogether - the table is a simpler way of explaining it.

andrum99:
> I think part of what makes this page confusing is the way the UARTs are referred to by different names in different places - I wonder it it'd be worth trying to use more consistent naming throughout the page?

I've gone through the page and tried to address as much of the inconsistency as I can. This is probably the most complicated thing to try and describe! There are so many different combinations it's never going to be perfect. I think the addition of several tables should help.

@rgov - can you possibly have a look at the proposed new version (at https://github.com/raspberrypi/documentation/blob/37513e2d3c5c082ffc291fb04f3286d09bc23007/configuration/uart.md) and give me your thoughts?

andrum99:
I would also like to get a review from someone at RPT who knows how this stuff works. If @JamesH65 could @ in someone appropriate that would be great 🙂

JamesH65:
@pelwell Do you have time to give this change a quick review?

configuration/uart.md
```diff
@@ -1,51 +1,111 @@
...
+Various UART Device Tree overlay definitions can be found in the [kernel GitHub tree](https://github.com/raspberrypi/linux). The two most useful overlays are [`disable-bt`](https://github.com/raspberrypi/linux/blob/rpi-4.19.y/arch/arm/boot/dts/overlays/disable-bt-overlay.dts) and [`miniuart-bt`](https://github.com/raspberrypi/linux/blob/rpi-4.19.y/arch/arm/boot/dts/overlays/miniuart-bt-overlay.dts).
 
-`disable-bt` disables the Bluetooth device and restores UART0/ttyAMA0 to GPIOs 14 and 15. It is also necessary to disable the system service that initialises the modem so it doesn't use the UART: `sudo systemctl disable hciuart`.
+`disable-bt` disables the Bluetooth device and makes the first PL011 (UART0) the primary UART. It is also necessary to disable the system service that initialises the modem, so it doesn't connect to the UART using `sudo systemctl disable hciuart`.
```
lurch, line 96: Comma between "UART" and "using" ?

configuration/uart.md
```diff
@@ -1,51 +1,111 @@
...
+`miniuart-bt` switches the Bluetooth function to use the mini UART, and makes the first PL011 (UART0) the primary UART. Note that this may reduce the maximum usable baud rate (see mini UART limitations below).
 
-There are other UART-specific overlays in the folder. Refer to `/boot/overlays/README` for details on Device Tree Overlays, or run `dtoverlay -h overlay-name` for descriptions and usage information.
+The overlays `uart2`, `uart3`, `uart4`, and `uart5` are used to enable the four additional UARTs on the Pi 4. For details on which GPIO pins these connect to, refer to section 5.3 of the [BCM2711 ARM Peripherals](https://www.raspberrypi.org/documentation/hardware/raspberrypi/bcm2711/rpi_DATA_2711_1p0.pdf) document.
```
lurch, line 100: Might be better to link to the "parent page", rather than the PDF directly, in case later revisions of the PDF get created in future? :shrug:

configuration/uart.md
```diff
@@ -1,51 +1,111 @@
...
+`disable-bt` disables the Bluetooth device and makes the first PL011 (UART0) the primary UART. It is also necessary to disable the system service that initialises the modem, so it doesn't connect to the UART using `sudo systemctl disable hciuart`.
 
-`miniuart-bt` switches the Raspberry Pi 3 and Raspberry Pi Zero W Bluetooth function to use the mini UART (ttyS0), and restores UART0/ttyAMA0 to GPIOs 14 and 15. Note that this may reduce the maximum usable baudrate (see mini UART limitations below). It is also necessary to edit /lib/systemd/system/hciuart.service and replace ttyAMA0 with ttyS0, unless you have a system with udev rules that create /dev/serial0 and /dev/serial1. In this case, use /dev/serial1 instead because it will always be correct. If cmdline.txt uses the alias serial0 to refer to the user-accessible port, the firmware will replace it with the appropriate port whether or not this overlay is used.
+`miniuart-bt` switches the Bluetooth function to use the mini UART, and makes the first PL011 (UART0) the primary UART. Note that this may reduce the maximum usable baud rate (see mini UART limitations below).
```
lurch, line 98: Does this overlay also lock the core clock frequency @pelwell ?

configuration/uart.md
```diff
@@ -1,51 +1,111 @@
...
 
-By default, the UART transmit and receive pins are on GPIO 14 and GPIO 15 respectively, which are pins 8 and 10 on the GPIO header.
+| Model | mini UART | first PL011 (UART0)|
+|-------|-----------|-------|
+| Raspberry Pi Zero | secondary | primary |
+| Raspberry Pi Zero W | primary (disabled) | secondary (Bluetooth) |
+| Raspberry Pi 1 | secondary | primary |
+| Raspberry Pi 2 | secondary | primary |
+| Raspberry Pi 3 | primary (disabled) | secondary (Bluetooth) |
+| Raspberry Pi 4 | primary (disabled) | secondary (Bluetooth) |
```
lurch, line 48: @pelwell On Pis without the bluetooth module, is the miniUART disabled by default, or is it enabled but just not connected to anything?

configuration/uart.md
```diff
@@ -1,51 +1,111 @@
...
+The secondary UART is not normally present on the GPIO connector. By default, the secondary UART is connected to the Bluetooth side of the combined wireless LAN/Bluetooth controller, on models which contain this controller.
 
-Reboot the Raspberry Pi for the change to take effect.
+## Configuration
 
-## UART output on GPIO pins
+The following table summarises the default configuration of the first two UARTs:
 
-By default, the UART transmit and receive pins are on GPIO 14 and GPIO 15 respectively, which are pins 8 and 10 on the GPIO header.
+| Model | mini UART | first PL011 (UART0)|
```
lurch, line 41: @andrum99 Does this table work if the header-row is switched to "Model, Primary UART, Secondary UART" (with the table-cells then showing either mini-UART or PL011-UART) ? Would that be less confusing or more confusing? :wink:

lurch:
I agree that this is pretty complicated to try and explain, but IMHO this page is now looking much better than the previous version. Thanks for your hard work on this @andrum99 :heart:

configuration/uart.md
andrum99: I think the way it is is probably OK, although I have swapped the columns as it makes more sense to list the UARTs in order (mini UART is UART 1).

configuration/uart.md
```diff
@@ -1,51 +1,111 @@
...
+# UART configuration
 
-The SoCs used on the Raspberry Pis have two built-in UARTs, a [PL011](http://infocenter.arm.com/help/index.jsp?topic=/com.arm.doc.ddi0183g/index.html) and a mini UART. They are implemented using different hardware blocks, so they have slightly different characteristics. However, both are 3.3V devices, which means extra care must be taken when connecting up to an RS232 or other system that utilises different voltage levels. An adapter must be used to convert the voltage levels between the two protocols. Alternatively, 3.3V USB UART adapters can be purchased for very low prices. 
- 
-By default, on Raspberry Pis equipped with the wireless/Bluetooth module (Raspberry Pi 3 and Raspberry Pi Zero W), the PL011 UART is connected to the Bluetooth module, while the mini UART is used as the primary UART and will have a Linux console on it. On all other models, the PL011 is used as the primary UART. 
+There are two types of UART available on the Raspberry Pi -  [PL011](http://infocenter.arm.com/help/index.jsp?topic=/com.arm.doc.ddi0183g/index.html) and mini UART. The PL011 is a capable, broadly 16550-compatible UART, while the mini UART has a reduced feature set.
```
pelwell, line 3: I wasn't aware that the PL011 design is based on 16550, and I don't think this factoid adds anything - why mention this and not that the mini-UART is an 8250 clone?

configuration/uart.md
pelwell: No it doesn't - a fact mentioned in the README IIRC, but worth reiterating.

configuration/uart.md
```diff
@@ -1,51 +1,111 @@
...
+`miniuart-bt` switches the Bluetooth function to use the mini UART, and makes the first PL011 (UART0) the primary UART. Note that this may reduce the maximum usable baud rate (see mini UART limitations below).
 
-There are other UART-specific overlays in the folder. Refer to `/boot/overlays/README` for details on Device Tree Overlays, or run `dtoverlay -h overlay-name` for descriptions and usage information.
+The overlays `uart2`, `uart3`, `uart4`, and `uart5` are used to enable the four additional UARTs on the Pi 4. For details on which GPIO pins these connect to, refer to section 5.3 of the [BCM2711 ARM Peripherals](https://www.raspberrypi.org/documentation/hardware/raspberrypi/datasheets.md) datasheet.
```
pelwell, line 100: I would remove the second sentence and run it into the next paragraph, since pin usage for the Pi4 UART overlays is in the README.

configuration/uart.md
```diff
@@ -1,51 +1,111 @@
...
 
-For full instructions on how to use Device Tree Overlays see [this page](device-tree.md). In brief, add a line to the `config.txt` file to enable Device Tree Overlays. Note that the `-overlay.dts` part of the filename is removed.
+There are other UART-specific overlays in the folder. Refer to `/boot/overlays/README` for details on Device Tree overlays, or run `dtoverlay -h overlay-name` for descriptions and usage information.
+
+For full instructions on how to use Device Tree overlays see [this page](device-tree.md). In brief, add a line to the `config.txt` file to enable Device Tree overlays. Note that the `-overlay.dts` part of the filename is removed. For example:
```
pelwell, line 104: Change "to enable Device Tree overlays" to "to apply a Device Tree overlay".

rgov:
There is good information here, thanks for taking this on.

There is this concept of "primary UART" and then there's the "first" UART (UART0) which is a little confusing. Are they or aren't they the same? Instead of saying "first UART" at all can we just always refer to it as UART0?

One thing to please keep in mind is that many Raspberry Pi users are hobbyists. I think some explanatory text on what UART even _is_ would be valuable. E.g., 

> A UART is a hardware device that transmits and receives serial communications as electrical signals. On the Raspberry Pi, UARTs are often used for access to the Linux console from a computer via the `TXD` and `RXD` GPIO pins. They are also used to communicate with other components, such as the Bluetooth controller.

andrum99:
> There is this concept of "primary UART" and then there's the "first" UART (UART0) which is a little confusing. Are they or aren't they the same? Instead of saying "first UART" at all can we just always refer to it as UART0?

They're not the same - The numbers on the UARTs refer to the actual, physical UARTs in the SoC. "Primary" and "Secondary" are roles which can be swapped around.
> 
> One thing to please keep in mind is that many Raspberry Pi users are hobbyists and beginners. I think some explanatory text on what UART even _is_ would be valuable. E.g.,

I don't think that's necessary.

configuration/uart.md
andrum99: That sentence does not claim the design is based on the 16550. The Broadcom datasheet for the 2835 also compares it to the 16550 in a similar way. Comparison to the 16550 is used as a way of saying that it is fairly standard in terms of its capability. If you have a suggestion for a better description I'm happy to use that instead.

pelwell:
I just think that comparing one unknown thing to anotherv potentially unknown thing is unproductive.

pelwell:
That was a reply to a deleted comment.

Primary/secondary and UART0/UART1 are useful concepts - logical vs physical - first/second is not.

configuration/uart.md
andrum99: > No it doesn't - a fact mentioned in the README IIRC, but worth reiterating.

Which part are you disputing?

rgov:
Not to just repeat myself, but please view this from the perspective of someone who has technical familiarity with a regular PC. The Raspberry Pi is meant to be approachable as a "regular" computer. Raspbian is a full desktop environment, and the device supports the familiar HDMI and USB out of the box.

However one thing that people coming from a traditional desktop environment *won't* know about is UART. It's not a term that comes up unless you really get into computer hardware. And since the Raspberry Pi is, to many, a way to dip a toe into that world, it's a good place to be introductory.

Take for instance the documentation for [NFS](https://www.raspberrypi.org/documentation/configuration/nfs.md) which start by at least *defining* what NFS is and why you might configure it. If you wind up at that page, perhaps you already know what NFS is, and the text can be skipped easily. If not, it gives some context and some terminology the reader can use to search for more introductory texts.

Hobbyists often start out wanting a Linux PC that's wired up to LEDs or a button, and they need some guidance connecting the software concepts ("this is a serial device I can write to from Python") to the hardware ("this is a UART wired to some controller module"). Look no further than the popularity of basic explanatory guides from Make magazine or Adafruit.

pelwell:
Sorry - what's a serial device? And what's Python?

Any concept that can be explained unambiguously quickly using your favourite search engine doesn't need explaining, otherwise this becomes Wikipedia.

pelwell:
@andrum99 I don't know why your comments keep disappearing, but I wasn't disputing anything - just answering @lurch's question about whether `miniuart-bt` automatically locks the core frequency (it doesn't).

configuration/uart.md
andrum99: I've added a note that the core clock needs set to a fixed frequency to this paragraph.

andrum99:
I've pressing the "resolve conversation" button, which hides the bits you have commented on, because I've resolved the issue. You can un-hide them using the small "show resolved" on the righthand side of the main part of the page.

You seem to have left the same comment in reply to one of lurch's comments as well, which I've now addressed. Does the page now look correct from a technical perspective?

pelwell:
See Approving review

andrum99:
Thanks you all for your reviews 😁 This one is now ready to roll.

pelwell:
And resolving a discussion after asking a question 
that hasn't yet been answered seems counterproductive.

andrum99:
> And resolving a discussion after asking a question
> that hasn't yet been answered seems counterproductive.

If I've understood you correctly, the one I marked as resolved I marked after I had noticed the same comment against a question lurch had asked you, and figured it was a duplicate. That was premature, as after reading further, it seems the "orphan" comment I was asking about might have been in reply to this question from lurch:

>@pelwell On Pis without the bluetooth module, is the miniUART disabled by default, or is it enabled but just not connected to anything?

I'm not sure what the answer to that is - can you help clarify?

pelwell:
It seems that when one replies to a resolved comment via the link in an email one ends up in the main discussion, hence the confusion 

The Device Tree node of the mini-UART in a non-Bluetooth Pi is disabled by default.

andrum99:
> It seems that when one replies to a resolved comment via the link in an email one ends up in the main discussion, hence the confusion
> 
> The Device Tree node of the mini-UART in a non-Bluetooth Pi is disabled by default.

Apologies for the confusion. I've added that extra information to the page.

andrum99:
OK, that's this one ready to go @JamesH65

JamesH65:
Thanks to all the contributors on this one - looks a lot better now.

lurch:
> hence the confusion

Things also get confusing when you add a comment on a file which changes in a later commit, and GitHub auto-hides the "outdated" comment even though it hasn't been "resolved" yet. :confounded: 
Not saying that's happened here, just something to be aware of!

lurch:
And to respond to @rgov 's comments - UART **is** mentioned in the https://github.com/raspberrypi/documentation/tree/master/glossary :wink: 
And I suggested in #1427 that it might be worth adding more documentation about using a serial console.

PR #1562, opened 2020-05-24
MAC Address `TFTP_PREFIX` Request is Lowercase

sdwilsh:
I'm working on getting my Pi 4 to network boot, and had setup the TFTP server with the MAC address, only to find that it comes across as lower case, not upper case.


timg236:
@lurch LGTM

PR #1574, opened 2020-05-28
apply lossless optimizations to image files

vszakats:
using <https://imageoptim.com/mac>

Screen Shot 2020-05-28 at 15 28 46


JamesH65:
@lurch This seems like a sane thing to do, but I guess each one need to be checked! Thoughts?

aallan:
We'd have to go through each image and check it against the original. But it theory, yes? In practice, I'm not entirely sure saving 6MB is worth the effort that will take?

JamesH65:
> We'd have to go through each image and check it against the original. In theory, yes? In practice, I'm not entirely sure saving 6MB is worth the effort that will take?

My concern is that although I have no reason to believe any malfeasance, someone could slip images in for spamming purposes!

It's 6MB to us for storage, but a lot of bandwidth going out of the website and on users net links given how many people use the docs repo.

vszakats:
I know of only a handful of cases where this process may result in invalid files, and even those were PNG **test** files with rare internal properties. If that helps, I'll go through them and check if there happens to be any such case here. For the rest I can say with a 99.9999% certenity that there the uncompressed bitmaps are all identical. This is based on years working with these tools and having no issues. If that's not convincing, I can run a batch uncompress for before/after state and binary diff the results.

Regarding image metadata: I had the option enabled to clean these off, so if this may be a concern with you and retaining metadata is desired, I can do another run with that reconfigured.

vszakats:
With the help of a short script (attached) that converted all images into RGBA format via ImageMagick and then compared the results before and after the lossless compression, out of the 117 files there were 9 with binary differences. Close inspection of these 9 images using a picture comparer tool (<https://www.kaleidoscopeapp.com>), revealed no visual difference between them.

So, to me this seems safe to merge.

```bash
#!/usr/bin/env bash
c=0
find . -name '*.png' -o -name '*.jpg' -o -name '*.jpeg' -o -name '*.PNG' -o -name '*.JPG' -o -name '*.gif' | sort | while read -r f; do
  cs="$(printf '%03d' "$c")"
  echo "${f}" "${cs}-$(basename "${f}")"
  cp -p "${f}" "${cs}-$(basename "${f}")"
  magick convert "$f" "$cs.rgba"
  ((c++))
done
```

vszakats:
BTW, here on the GitHub page, clicking on the `Files Changed` tab and then clicking on `Rich Diff` next to each individual file, will reveal visual comparison options very much alike Kaleidoscope.

lurch:
I'd be inclined to agree with @aallan that this might not actually be worth it? :shrug: 
Most users will probably only view one or two pages of the documentation, and therefore won't be downloading every image? :wink:

vszakats:
What is the cost of accepting this?

vszakats:
Can't see why a positive-only change is resisted. Closing.

JamesH65:
Why are you closing? It's not resistance to the change, seems sensible to me, its  just making sure everything is right.

vszakats:
I wasn't ready to convince everyone from zero :( I can do it again if there is an agreement for it, but won't be forcing doing optimizations out of the wild. Also the HTTPS fixup would be needed to avoid the mixed-content issues. E.g. on my system with HTTP disabled, some page components appear with broken placeholders.

PR #1577, opened 2020-05-29
Replace Remaining Raspbian References!

lurch:


JamesH65:
Ah, I did wonder about the merge I had to do when pushing the original changes - Github kept giving errors, then decided it was all OK. Looks like it screwed up this merge.

PR #1582, opened 2020-06-02
Update README.md

JamesH65:
Added 8GB data to the FAQ


JamesH65:
Also removed (mostly) references to the 1GB version, except to say why its not longer there.

PR #1583, opened 2020-06-03
Remove no-longer-used image

lurch:


JamesH65:
Just ouf of interest, how do we know this is no longer used?

lurch:
Because it's explicitly referencing dataplicity (in the filename), and you removed all references to that?
Also: https://github.com/raspberrypi/documentation/search?q=dataplicity&type=Code :wink: 

(I've not been specifically looking for unused images, I just noticed while looking at https://github.com/raspberrypi/documentation/tree/master/remote-access/access-over-Internet that there was an `images` folder, which seemed weird given that the `README.md` there doesn't actually contain any images)

JamesH65:
Cool.

PR #1586, opened 2020-06-06
headless add to glossary

JamesH65:


lurch:
Strictly speaking I think "headless" might just mean "without display device"; but without a display device there's not really much you can do with a mouse or keyboard :rofl:

PR #1587, opened 2020-06-06
fix: Corrects capitalisation typo in paragraph of using-github.md

darigovresearch:


lurch:
Thanks!

"It will be read, assessed for technical correctness, passed to copy editors for final checking, and finally merged to the main documentation tree."

I think it's safe to skip the copy-edit check for this one :rofl:

PR #1589, opened 2020-06-07

timg236:
Update to filename for latest beta releases (maybe create symlinks in the repo in future if it works) and point people at the releases page for Raspberry Pi Imager


timg236:
Thanks @JamesH65 @lurch

PR #1590, opened 2020-06-09
UCB-C change to USB-C

SimUKdev:
Correcting what I believe to be a typo in paragraph 2, "UCB-C" should be "USB-C" in "HDMI port adjacent to the UCB-C power input (labelled HDMI0)".


aallan:
Thanks!

PR #1593, opened 2020-06-12
Remove USB_MSD_BOOT_MAX_RETRIES

timg236:
This option has been removed from the bootloader because it's not useful for USB MSD boot.


timg236:
Thanks @jamesh65

PR #1594, opened 2020-06-19
Mention the PSU and Hifiberry AMP2

lbt:
This combination appeared to work fine - but touch was laggy as although it uses a  60W PSU the regulator on the AMP2 used to power the main board does not provide quite enough power via the GPIO pins.


aallan:
We try not to recommend third party products in the documentation.

lbt:
Fair enough. I deliberately put the brand in to make it a good search target :)
I will remove and resubmit

PR #1595, opened 2020-06-19
Added fix for showing hostnames when running nmap

vbhaip:
When running nmap, it won't show the hostnames like the output the documentation shows unless it is run as root. This provides a quick note to let users know that.

### DIFF
--- a/remote-access/ip-address.md
+++ b/remote-access/ip-address.md
@@ -77,7 +77,7 @@ Host is up (0.0030s latency).
 Nmap done: 256 IP addresses (4 hosts up) scanned in 2.41 seconds
 ```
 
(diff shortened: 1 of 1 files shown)


JamesH65:
Tested, looks good to me. 

@lurch

lurch:
I know nothing about nmap, nor why it would need to be run as root in order to display hostnames. @XECDesign ?

aallan:
By default an unprivileged scan with `nmap` uses `-sT` while privileged (root) uses `-sS` mode. Essentially what's happening is that unprivileged `nmap` uses TCP Connect by default, while root uses a Syn scan (as you'd expect). There is also sometimes some weirdness around access to the ARP cache for privilege scans, but that's not what's going on here.

The TL;DR is that `nmap` is working as expected here. I'll merge.

PR #1598, opened 2020-06-22
Touch function is not affected by display_power

lbt:
It's useful and not obvious that touch is available when the screen is "off"


JamesH65:
This is technically correct and a worthwhile addition, but I'm not sure I like the phrasing. @lurch ?

lbt:
Agreed. Is this acceptable?

lurch:
I'm not sure if this ought to say "Raspberry Pi Touch Display" to match the nomenclature on https://www.raspberrypi.org/products/raspberry-pi-touch-display/ ?

LGTM otherwise :+1:

JamesH65:
I've always called it the 7" display! But I guess it should match the sales document!

JamesH65:
Thanks.

PR #1602, opened 2020-06-25
Update README.md

JamesH65:


lurch:
@XECDesign It seems the version in the Raspberry Pi OS repos is still rpi-imager v1.2, not the latest v1.3 ?

PR #1554, opened 2020-05-19
Elaborated on HQ Camera gasket assembly

Adam-Langley:
Described the gasket assembly, added some images, detailed how to protect from dust ingress.

PR #1605, opened 2020-07-01
High quality camera IR filter removal pictures

cygann:
I had to remove the IR filter on a high quality camera and couldn't find pictures or videos for process online. Hopefully someone else might find the pictures detailing the removal process useful.


JamesH65:
We already have a similar PR waiting for copy editing at #1554.

Issue #1596, opened 2020-06-20
is display_lcd_rotate still a valid property?


zjgoodman:
Hello, I have an LCD touchscreen connected to the board of my pi. The display is rendering upside-down for me so I edited `/boot/config.txt` and per [the documentation](https://www.raspberrypi.org/documentation/configuration/config-txt/video.md) set `lcd_rotate=2` and was able to fix it. However I also noticed that there's a section in the documentation that describes the property `display_lcd_rotate`. Based on the description provided in the documentation, this property seems to do the same thing as the `lcd_rotate` property. However, when I set `display_lcd_rotate=2` in my config.txt nothing happened with the display. Is this an old/deleted property that should be removed from the documentation? It's unclear if it's only supported in a particular version.

Here's my firmware version
```
pi@Tohru:~ $ uname -a
Linux Tohru 4.19.118-v7l+ #1311 SMP Mon Apr 27 14:26:42 BST 2020 armv7l GNU/Linux
```

lurch:
Are you using the Pi4? See https://www.raspberrypi.org/documentation/configuration/display_rotation.md

zjgoodman:
Ok so per [the documentation that you linked](https://www.raspberrypi.org/documentation/configuration/display_rotation.md), looks like `display_lcd_rotate` is a legacy driver property:
> Legacy graphics driver (default on models prior to the Pi4).

And [the config.txt video options documentation that I originally linked](https://www.raspberrypi.org/documentation/configuration/config-txt/video.md) doesn't mention anything about that:
> Use `display_lcd_rotate` to rotate or flip the LCD orientation. Parameters are the same as `display_hdmi_rotate`.

And the documentation for `display_hdmi_rotate` does have this paragraph:
> If using the VC4 FKMS V3D driver (this is the default on the Raspberry Pi 4), then 90 and 270 degree rotations are not supported. The Screen Configuration utility provides display rotations for this driver. See [this page](https://www.raspberrypi.org/documentation/configuration/display_rotation.md) for more information.

In my opinion the information about `display_lcd_rotate` being a legacy driver property is _discoverable_ via the documentation but not obvious. If I could propose a change, I'd suggest adding _Legacy graphics driver (default on models prior to the Pi4)_ to the `display_lcd_rotate` property's [video config.txt documentation](https://www.raspberrypi.org/documentation/configuration/config-txt/video.md) 👌

zjgoodman:
[...]

PR #1606, opened 2020-07-01
add note about legacy graphics driver for `display_lcd_rotate` property

zjgoodman:
As noted in #1596 the `display_lcd_rotate` property is only valid for use with the legacy graphics driver, but it's not immediately obvious in the documentation, so I added a note to specify such

### DIFF
--- a/configuration/config-txt/video.md
+++ b/configuration/config-txt/video.md
@@ -663,7 +663,7 @@ If using the VC4 FKMS V3D driver (this is the default on the Raspberry Pi 4), th
 
 ### display_lcd_rotate
 
-For the legacy graphics driver (default on models prior to the Pi4), use `display_lcd_rotate` to rotate or flip the LCD orientation. Parameters are the same as `display_hdmi_rotate`. For the modern graphics driver use `lcd_rotate`.
+For the legacy graphics driver (default on models prior to the Pi4), use `display_lcd_rotate` to rotate or flip the LCD orientation. Parameters are the same as `display_hdmi_rotate`. See also `lcd_rotate`.
 
 ### display_rotate
 
--- a/configuration/config-txt/video.md
+++ b/configuration/config-txt/video.md
@@ -663,7 +663,7 @@ If using the VC4 FKMS V3D driver (this is the default on the Raspberry Pi 4), th
 
 ### display_lcd_rotate
 
-Use `display_lcd_rotate` to rotate or flip the LCD orientation. Parameters are the same as `display_hdmi_rotate`.
+For the legacy graphics driver (default on models prior to the Pi4), use `display_lcd_rotate` to rotate or flip the LCD orientation. Parameters are the same as `display_hdmi_rotate`. For the modern graphics driver use `lcd_rotate`.
 
 ### display_rotate
 


configuration/config-txt/video.md
```diff
@@ -663,7 +663,7 @@ If using the VC4 FKMS V3D driver (this is the default on the Raspberry Pi 4), th
 
 ### display_lcd_rotate
 
-Use `display_lcd_rotate` to rotate or flip the LCD orientation. Parameters are the same as `display_hdmi_rotate`.
+For the legacy graphics driver (default on models prior to the Pi4), use `display_lcd_rotate` to rotate or flip the LCD orientation. Parameters are the same as `display_hdmi_rotate`. For the modern graphics driver use `lcd_rotate`.
```
andrum99, line 666: That's not quite correct - `lcd_rotate` can be used with any of the 3 possible graphics drivers. Also, the documentation does not currently specify what options are available for `lcd_rotate` - it only gives an example of one possible option.

zjgoodman: Per [this page of the documentation](https://www.raspberrypi.org/documentation/configuration/display_rotation.md), `display_lcd_rotate` is a legacy driver property. Note that there is a difference between `lcd_rotate`, `display_lcd_rotate` and `display_rotate` 😆 🔥 
 
legacy

lurch: @JamesH65 will no doubt be able to confirm, but from reading the current docs it sounds like `display_rotate` (which is labelled as deprecated), `display_lcd_rotate` and `display_hdmi_rotate` options apply at the GPU-driver level, and are thus only applicable to the legacy graphics driver; whereas `lcd_rotate` works at the LCD-screen hardware level, and is thus applicable to all graphics drivers. So `lcd_rotate` is actually an orthogonal option to the others, and not a "replacement" for `display_lcd_rotate` ? :shrug: 

@JamesH65 As noted by @andrum99 , it's not clear what the valid values for `lcd_rotate` are - does it accept _all_ of the values that `display_hdmi_rotate` does, or only a subset? Perhaps `lcd_rotate` should also be documented on the `display_rotation.md` page, and then all the `*_rotate` options on `video.md` could just refer to `display_rotation.md` ?

JamesH65: IIRC, lcd_rotate does indeed talk direct to the LCD which does the rotate itself (although it's more like horizontal and vertical flips rather than rotate). The other options use the transposer to do rotation - which is only available when using firmware -  so legacy and I think FKMS, but not sure what nasty interactions will occur in FKMS mode.  There's also no guarantee that any future LCD's will be able to use that lcd_rotate option as its part of the LCD hardware.

andrum99: My personal preference for this PR would be to reword the final sentence like this: 'See also `lcd_rotate`.' The rest looks good to me 😁 Improving the documentation for `lcd_rotate` could be tackled in a separate PR.

zjgoodman: @andrum99 I have updated the PR to use the `See also lcd_rotate` phrasing like you suggested 👌

andrum99: LGTM 😁

zjgoodman:
@andrum99 @JamesH65 @lurch what do we need to do to get this PR merged?

PR #1610, opened 2020-07-05
configuration/config-txt/memory.md - minor fixup

andrum99:
There's an 8GB Pi 4 now. Reword things slightly, and don't enumerate each model of Pi 4. The intention is to make the opening paragraph read simpler, and perhaps make future maintenance of the page slightly easier.

### DIFF
--- a/configuration/config-txt/memory.md
+++ b/configuration/config-txt/memory.md
@@ -2,7 +2,9 @@
 
 ## gpu_mem
 
-Specifies how much memory, in megabytes, to reserve for the exclusive use of the GPU: the remaining memory is allocated to the ARM CPU. The default value is `64`. The recommended maximum values are as follows:
+Specifies how much memory, in megabytes, to reserve for the exclusive use of the GPU: the remaining memory is allocated to the ARM CPU. The default value is `64`. You should set `gpu_mem` to the lowest possible value to ensure best performance. If a particular graphics feature is not working correctly, try increasing the value of `gpu_mem`, being mindful of the recommendations in the above table.
+
+The recommended maximum values are as follows:
 
 | total RAM | `gpu_mem` recommended maximum |
 |-----------|-------------------------------|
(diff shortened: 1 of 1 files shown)


JamesH65:
I'd be inclined to really emphasise that the use of large values for gpu_mem can actually stop the device booting, also emphasis that increasing it does NOT increase performance.

andrum99:
I've address the review comments. The page says the default is 64 but on my systems I'm seeing a default on 76MB now, but I'm not sure if this is Pi 4-specific.

andrum99:
OK, this one should be ready to roll now.

andrum99:
It seems the default is now 76MB on all devices. I've filed https://github.com/RPi-Distro/raspi-config/issues/122 to have `raspi-config` fixed accordingly.

andrum99:
> It seems the default is now 76MB on all devices. I've filed [RPi-Distro/raspi-config#122](https://github.com/RPi-Distro/raspi-config/issues/122) to have `raspi-config` fixed accordingly.

Actually it's just Pi 2 and later. I don't have a Zero to hand to test, but from memory it is 64MB as well.

JamesH65:
LGTM @lurch?

configuration/config-txt/memory.md
```diff
@@ -2,31 +2,36 @@
...
 
-Using `gpu_mem_256`, `gpu_mem_512`, and `gpu_mem_1024` allows you to swap the same SD card between 256MB, 512MB, and 1024MB Pis without having to edit `config.txt` each time:
+On the Raspberry Pi 4 the 3D component of the GPU has its own memory management unit (MMU), and does not use memory from the `gpu_mem` allocation. Instead memory is allocated dynamically within Linux. This may allow a smaller value to be specified for `gpu_mem` on the Pi 4, compared to previous models.
+
+You can also use `gpu_mem_256`, `gpu_mem_512`, and `gpu_mem_1024` to allow swapping the same SD card between Pis with different amounts of RAM without having to edit `config.txt` each time:
 
 ## gpu_mem_256
 
-The `gpu_mem_256` command sets the GPU memory in megabytes for the 256MB Raspberry Pi. (It is ignored if memory size is not 256MB). This overrides `gpu_mem`. The maximum value is `192`, and the default is not set.
+The `gpu_mem_256` command sets the GPU memory in megabytes for the 256MB Raspberry Pi. (It is ignored if memory size is not 256MB). This overrides `gpu_mem`.
```
lurch, line 23: Minor nitpick, but should "the 256MB Raspberry Pi" be replaced with something like "Raspberry Pi devices with 256MB of memory" ? (and again below with 512MB)

lurch:
Might be useful to get feedback from @spl237 or @popcornmix about the default gpu_mem value?

andrum99:
> Might be useful to get feedback from @spl237 or @popcornmix about the default gpu_mem value?

Good idea. Is there a 1GB Pi 1? It might be related to memory size rather than SoC.

popcornmix:
There are no 1GB Pi0/Pi1 devices.

popcornmix:
default gpu_mem is 76 for 1GB (or greater) and 64 for less.

configuration/config-txt/memory.md
andrum99: Good catch. I've used "Raspberry Pis" rather than "Raspberry Pi devices" as the extra word doesn't add anything, and I would rather keep it simple. It also fits in better with the intended chatty, informal style of official Pi material. (With the usual caveat that I'm not entirely convinced that is appropriate for technical docs).

andrum99:
> default gpu_mem is 76 for 1GB (or greater) and 64 for less.

Thanks for the clarification. The page should now be ready to go.

PR #1611, opened 2020-07-06
Update msd.md - power not a Pi-specific problem, so move out of known issues section

andrum99:


lurch:
Not sure why you moved it specifically into the "Raspberry Pi 3B+, Compute Module 3+" section? :confused:

andrum99:
> Not sure why you moved it specifically into the "Raspberry Pi 3B+, Compute Module 3+" section? 😕

That's a good point. Try now.

PR #1615, opened 2020-07-07
Update memory.md - Slight reword, document total_mem

andrum99:
This one's for the pedants 😋

### DIFF
--- a/configuration/config-txt/memory.md
+++ b/configuration/config-txt/memory.md
@@ -2,7 +2,7 @@
 
 ## gpu_mem
 
-Specifies how much memory, in megabytes, to reserve for the exclusive use of the GPU: the remaining memory is allocated to the ARM CPU. For Pis with less than 1GB of memory, the default is `64`; for Pis with 1GB or more of memory the default is `76`. You should set `gpu_mem` to the lowest possible value, to ensure best performance. If a particular graphics feature is not working correctly, try increasing the value of `gpu_mem`, being mindful of the recommended maximums shown below.
+Specifies how much memory, in megabytes, to reserve for the exclusive use of the GPU: the remaining memory is allocated to the ARM CPU. For Pis with less than 1GB of memory, the default is `64`; for Pis with 1GB or more of memory the default is `76`. To ensure best performance, you should set `gpu_mem` to the lowest possible value. If a particular graphics feature is not working correctly, try increasing the value of `gpu_mem`, being mindful of the recommended maximums shown below.
 
 The recommended maximum values are as follows:
 
--- a/configuration/config-txt/memory.md
+++ b/configuration/config-txt/memory.md
@@ -32,7 +32,7 @@ The `gpu_mem_1024` command sets the GPU memory in megabytes for Raspberry Pis wi
 
 ## total_mem
 
-This parameter can be used to force a Raspberry Pi to limit its memory capacity: specify the total amount of RAM, im megabytes, you wish the Pi to use. For example, to make a 4GB Raspberry Pi 4 behave as though it were a 1GB model, use the following:
+This parameter can be used to force a Raspberry Pi to limit its memory capacity: specify the total amount of RAM, im megabytes, you wish the Pi to use. For example, to make a 4GB Raspberry Pi 4B behave as though it were a 1GB model, use the following:
 
 ```
 total_mem=1024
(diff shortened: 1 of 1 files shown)


andrum99:
OK, so it seems total_mem is not actually documented. I've therefore added it to this PR.

configuration/config-txt/memory.md
```diff
@@ -30,6 +30,14 @@ The `gpu_mem_512` command sets the GPU memory in megabytes for Raspberry Pis wit
 
 The `gpu_mem_1024` command sets the GPU memory in megabytes for Raspberry Pis with 1GB or more of memory. (It is ignored if memory size is smaller than 1GB). This overrides `gpu_mem`.
 
+## total_mem
+
+This parameter can be used to force a Raspberry Pi to limit its memory capacity: specify the total amount of RAM you wish the Pi to use. For example, to make a 4GB Raspberry Pi 4 behave as though it were a 1GB model, use the following:
```
lurch, line 35: It's obvious from the example, but maybe the text should clarify that this value is specified in megabytes?

configuration/config-txt/memory.md
```diff
@@ -2,7 +2,7 @@
 
 ## gpu_mem
 
-Specifies how much memory, in megabytes, to reserve for the exclusive use of the GPU: the remaining memory is allocated to the ARM CPU. The default value is `76`, except on the Pi Zero and Pi 1 where it is `64`. You should set `gpu_mem` to the lowest possible value to ensure best performance. If a particular graphics feature is not working correctly, try increasing the value of `gpu_mem`, being mindful of the recommended maximums shown below.
+Specifies how much memory, in megabytes, to reserve for the exclusive use of the GPU: the remaining memory is allocated to the ARM CPU. For Pis with less than 1GB of memory, the default is `64`; for Pis with 1GB or more of memory the default is `76`. To ensure best performance, you should set `gpu_mem` to the lowest possible value. If a particular graphics feature is not working correctly, try increasing the value of `gpu_mem`, being mindful of the recommended maximums shown below.
```
lurch, line 5: I wonder if it's worth rewording "To ensure best performance, you should set `gpu_mem` to the lowest possible value." to something like "To ensure best performance, you should allocate as much RAM as possible to the ARM CPU, by setting `gpu_mem` to the lowest possible value." ? :shrug:

lurch:
> This one's for the pedants

Hello :wave:

configuration/config-txt/memory.md
andrum99: More words for the same meaning.

configuration/config-txt/memory.md
andrum99: Good catch - fixed.

andrum99:
> This one's for the pedants 😋

Of which I am.

configuration/config-txt/memory.md
```diff
@@ -30,6 +30,14 @@ The `gpu_mem_512` command sets the GPU memory in megabytes for Raspberry Pis wit
 
 The `gpu_mem_1024` command sets the GPU memory in megabytes for Raspberry Pis with 1GB or more of memory. (It is ignored if memory size is smaller than 1GB). This overrides `gpu_mem`.
 
+## total_mem
```
lurch, line 33: @JamesH65 Are there any restrictions on the values `total_mem` can take? Does it work on all Pis?

configuration/config-txt/memory.md
lurch: Meh, I just thought it might be nice to explain _why_ selecting a low `gpu_mem` value "ensures best performance". But I'm not bothered if you think it's too superfluous.

andrum99: I've added your suggestion, with some rejigging and adjustment to the wording. It is somewhat wordy, but we've had folk on the forums query things, and get the wrong idea in the past, so I think the level of detail we've got here is probably justified.

andrum99:
Thank's to @lurch for querying my wording on the previous PR, despite it having already integrated. We're iterating quite nicely today 😁

lurch:
> We're iterating quite nicely today :grin:

Uh-huh https://github.com/raspberrypi/documentation/commit/d972afd7def0fbc043b7bc2e910c3770612166f9 :wink:

configuration/config-txt/memory.md
JamesH65: min 128 max memory on board. Will adjust this PR to make that clear.

PR #1618, opened 2020-07-10
Update security.md to clarify that a warning by deluser is not a problem

nroming:
I found the warning that the group `pi` has no more members a bit puzzling. It might be useful to clarify that this is not a problem.

### DIFF
--- a/configuration/security.md
+++ b/configuration/security.md
@@ -73,7 +73,7 @@ This command will delete the `pi` user but will leave the `/home/pi` folder. If 
 ```bash
 sudo deluser -remove-home pi
 ```
-This command will result in a warning that the group `pi` has now more members. This group is, however, removed along with the `pi` user itself by the deluser command, so that the warning can be ignored.
+This command will result in a warning that the group `pi` has no more members. This group is, however, removed along with the `pi` user itself by the deluser command, so that the warning can be ignored.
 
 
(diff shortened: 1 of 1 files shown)


configuration/security.md
```diff
@@ -73,6 +73,8 @@ This command will delete the `pi` user but will leave the `/home/pi` folder. If
 ```bash
 sudo deluser -remove-home pi
 ```
+This command will result in a warning that the group `pi` has now more members. This group is, however, removed along with the `pi` user itself by the deluser command, so that the warning can be ignored.
```
lurch, line 76: "now more members" -> "no more members" ?
I think the second sentence might be more readable as something like "The `deluser` command removes both the `pi` user and the `pi` group though, so the warning can be safely ignored."

(Note that I'm only editing the text for grammar here - I've not tried this on my Pi so I don't know if it's technically correct!)

lurch:
ping @XECDesign

configuration/security.md
nroming: Correct. Sorry for that.

XECDesign:
Not sure I have anything to contribute here, but yes, the information in the PR is correct.

lurch:
> yes, the information in the PR is correct.

Thanks! :+1:

PR #1621, opened 2020-07-14
WIP: bcm2711_bootloader_config.md: document USB_MSD_PWR_OFF_TIME

andrum99:


andrum99:
Paging @timg236 for review.

andrum99:
> Looks great. That was lurking on my todo list so thanks for this!
> The bootloader restricts the min time to 250ms and the max to 5000ms.
> Zero means no power off, most flash devices are fine with zero but the defaults are designed to maximise compatibility.

Thanks for the additional info - I've added this to the PR.

timg236:
LGTM. Thanks!

hardware/raspberrypi/bcm2711_bootloader_config.md
```diff
@@ -373,7 +373,15 @@ Version: pieeprom-2020-06-15.bin - STABLE
 How long to wait in milliseconds before advancing to the next LUN e.g. a multi-slot SD-CARD reader. This is still being tweaked but may help speed up boot if old/slow devices are connected as well as a fast USB-MSD device containing the OS.
 
 Default:  2000 (2 seconds)  
-Version: pieeprom-2020-06-15.bin - STABLE    
+Version: pieeprom-2020-06-15.bin - STABLE
+
+### USB_MSD_PWR_OFF_TIME
+During USB mass storage boot, power to the USB ports is switched off for a short time to ensure the correct operation of USB mass storage devices. Most devices work correctly using the default setting: change this only if you have problems booting from a particular device. Setting `USB_MSD_PWR_OFF_TIME=0` will prevent power to the USB ports being switched off during USB mass storage boot.
```
lurch, line 379: Should "During USB mass storage boot" be changed to "Prior to USB mass storage boot"? I think things would go a bit wonky if USB power was switched off **during** USB boot :laughing:

timg236:
Power is turned off at at the start of the usb boot mode initialisation but only on the first iteration. This would not happen if only network or SD boot were selected.
Although they might happen in future if it resolves issues certain awkward usb devices :)

timg236:
Power is turned off at at the start of the usb boot mode initialisation but only on the first iteration. This would not happen if only network or SD boot were selected.
Although they might happen in future if it resolves issues certain awkward usb devices :)

PR #1622, opened 2020-07-15
Update sdxc_formatting.md to include MBR scheme for Mac

binarygeek:
added note on which scheme to use. No where did it state that MBR was needed. I tried "GUID Partition Map (Table)" and NOOBS was unable to boot with errors

### DIFF
--- a/installation/sdxc_formatting.md
+++ b/installation/sdxc_formatting.md
@@ -16,6 +16,8 @@ Run the Raspberry Pi Imager application, then from the `Choose OS option`, selec
 
 The standard formatting tools built into these operating systems are able to create FAT32 partitions; they might also be labelled as FAT or MS-DOS. Simply delete the existing exFAT partition and create and format a new FAT32 primary partition, before proceeding with the rest of the [NOOBS instructions](noobs.md). On a Mac, this means using the command-line `diskutil` program.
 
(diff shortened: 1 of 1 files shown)


JamesH65:
I have limited knowledge of the Mac, so any comments from others? Seems OK to me.

lurch:
pin @aallan

installation/sdxc_formatting.md
```diff
@@ -16,6 +16,8 @@ Run the Raspberry Pi Imager application, then from the `Choose OS option`, selec
 
 The standard formatting tools built into these operating systems are able to create FAT32 partitions; they might also be labelled as FAT or MS-DOS. Simply delete the existing exFAT partition and create and format a new FAT32 primary partition, before proceeding with the rest of the [NOOBS instructions](noobs.md). On a Mac, this means using the command-line `diskutil` program.
 
+NOTE: For Mac, use `Master Boot Record` for the scheme.
```
andrum99, line 19: Probably best to make this part of the preceding paragraph. Something like "On a Mac, this means using the command-line `diskutil` program: choose the `Master Boot Record` scheme."

JamesH65: I agree, think it would scan better if it were in the preceding paragraph.

PR #1624, opened 2020-07-19
add terminal equivalents to the various commands

tedder:
Adding terminal/shell equivalents to commands. It makes setting up raspios a lot easier since the various options are buried in the script, which doesn't have a manpage or a `--help` option. I'm somewhat torn on the level of detail to give, I thought about prefixing each with "for expert users" or similar.

I've banged my head against the locale and keyboard options, especially, so hopefully Future Me will benefit, if nobody else.

Also adding some examples of why pixel doubling might be wanted.

### DIFF
--- a/configuration/raspi-config.md
+++ b/configuration/raspi-config.md
@@ -2,7 +2,7 @@
 
 This page describes the console based raspi-config application. If you are using the Raspberry Pi desktop then you can use the graphical Raspberry Pi Configuration application from the Preferences menu to configure your Raspberry Pi.
 
-`raspi-config` is the Raspberry Pi configuration tool originally written by [Alex Bradbury](https://github.com/asb). It targets Raspberry Pi OS.
+`raspi-config` is the Raspberry Pi configuration tool originally written by [Alex Bradbury](https://github.com/asb) and lives at [github.com/RPi-Distro/raspi-config](https://github.com/RPi-Distro/raspi-config/tree/master/raspi-config). It targets Raspberry Pi OS.
 
 <a name="usage"></a>
 ## Usage
@@ -60,7 +60,9 @@ Generally speaking, `raspi-config` aims to provide the functionality to make the
 ### Change User Password
 
 The default user on Raspberry Pi OS is ```pi``` with the password ```raspberry```. You can change that here. Read about other [users](../linux/usage/users.md).
- 
+
+From the console, this can be changed with `passwd`.
+
 <a name="network-options"></a>
 ### Network Options
 
@@ -70,6 +72,13 @@ From this submenu you can set the host name, your wireless LAN SSID, and pre-sha
 #### Hostname
 
 Set the visible name for this Pi on a network.
+
+<a name="wifi"></a>
+#### Wireless LAN
+
+Enter SSID and passphrase here.
+
+From the console, after setting the [wireless country](#wireless-country), this can be set [using the wireless CLI documentation](wireless/wireless-cli.md).
 
 <a name="boot-options"></a>
 ### Boot Options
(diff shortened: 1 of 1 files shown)


configuration/raspi-config.md
```diff
@@ -2,7 +2,7 @@
 
 This page describes the console based raspi-config application. If you are using the Raspberry Pi desktop then you can use the graphical Raspberry Pi Configuration application from the Preferences menu to configure your Raspberry Pi.
 
-`raspi-config` is the Raspberry Pi configuration tool originally written by [Alex Bradbury](https://github.com/asb). It targets Raspberry Pi OS.
+`raspi-config` is the Raspberry Pi configuration tool originally written by [Alex Bradbury](https://github.com/asb) and lives at [github.com/RPi-Distro/raspi-config](https://github.com/RPi-Distro/raspi-config/tree/master/raspi-config). It targets Raspberry Pi OS.
```
lurch, line 5: Perhaps "which now lives at" ?
Also, not sure we need the full tree and branch info, rather than just a link to the repo?

configuration/raspi-config.md
```diff
@@ -60,7 +60,9 @@ Generally speaking, `raspi-config` aims to provide the functionality to make the
 ### Change User Password
 
 The default user on Raspberry Pi OS is ```pi``` with the password ```raspberry```. You can change that here. Read about other [users](../linux/usage/users.md).
- 
+
+From the console, this can be changed with `passwd`.
```
lurch, line 64: Isn't that already mentioned on the "users" page that's being linked to? I think this repo generally tries to avoid having duplicate information in multiple places.

lurch:
Having now read the full PR, I'm wondering what the value of these "from the console" additions are? `raspi-config` exists precisely to avoid needing to run console commands...?

> I've banged my head against the locale and keyboard options, especially

Is there a deficiency in raspi-config's capabilities? :man_shrugging: Maybe file a bug at https://github.com/RPi-Distro/raspi-config/issues ?

> Also adding some examples of why pixel doubling might be wanted.

Would the same thing be useful on https://www.raspberrypi.org/documentation/configuration/config-txt/video.md ?

configuration/raspi-config.md
JamesH65: I don;t like the phrase "lives at" either. "Is hosted at" or similar?

JamesH65:
Whilst I understand the thinking behind the change (teach teach teach!), this is a page specifically describing raspi-config, so adding the console commands doesn't seem right, on this particular page. If you are using raspi-config, you don't need the terminal commands.

What was the reasoning behind the change - did raspi-config just not work correctly? I would expect it to just work for the vast majority of users?

tedder:
> What was the reasoning behind the change - did raspi-config just not work correctly? I would expect it to just work for the vast majority of users?

If it's not wanted, that's fine- my objection is that it is presented as the _only_ documented way to do things, especially those that are fairly Pi-specific, like enabling the camera or setting the wifi or keyboard locales. There's certainly a subset of users who want a consistent or repeatable way to configure one or more Pis without needing to write down "human" steps.

JamesH65:
That's fair comment. The issue is therefor, should it be on that page, or elsewhere. Hmm. Let me have a think.

lurch:
> consistent or repeatable way to configure one or more Pis

@tedder Have you seen that `raspi-config` _is_ usable non-interactively? That's exactly how the [GUI version](https://github.com/raspberrypi-ui/rc_gui) of "raspi config" works under the hood.
See e.g. https://github.com/RPi-Distro/raspi-config/blob/master/raspi-config#L2365 and https://github.com/RPi-Distro/raspi-config/issues/121 and https://github.com/raspberrypi-ui/rc_gui/blob/master/src/rc_gui.c#L51

Although of course if you really want to work out what raspi-config _itself_ is doing under the hood, then that's the beauty of everything being OpenSource :heart_eyes:

tedder:
> Have you seen that raspi-config is usable non-interactively?

Yeah, that's what got me started down this path, since I didn't see a manpage or `--help` option. So I read the source and saw the four `--options`, with only two actually being supported. That's why I read through the source to create the additional documentation.

I missed, however, how the `nonint` flag works, [the gui strings give a lot](https://github.com/raspberrypi-ui/rc_gui/blob/master/src/rc_gui.c#L51), which I didn't know 'til you linked to it. One alternative would be to rewrite these individual descriptions (in the PR) to give the `nonint` versions, or to give both versions. I think it's especially good for the dangerous options like ssh and enabling wifi+country to give the nonint version.

> then that's the beauty of everything being OpenSource 

Including the documentation! :)

lurch:
> > then that's the beauty of everything being OpenSource

> Including the documentation! :)

Haha, indeed :grinning:  However much of the Raspberry Pi documentation is beginner-oriented (which IIRC was also one of the initial motivations for `raspi-config`), and I think littering "from the console" references throughout the existing text here might be too confusing for said beginners? @JamesH65 was hinting about having the instructions from this PR on a separate page, and that sounds like a good idea to me.

> dangerous options like ssh

I'm not sure what's "dangerous" about ssh, but [this](https://github.com/raspberrypi/documentation/blob/master/remote-access/ssh/README.md#3-enable-ssh-on-a-headless-raspberry-pi-add-file-to-sd-card-on-another-machine) describes the easiest way of non-interactively enabling SSH on a Pi.

aallan:

tedder:
I'm fine with it being closed (if needed) and happy to see a revamp on the docs.

PR #1626, opened 2020-07-20
Fixed typo (256MB instead of 512MB)

EwanRoycroft:


JamesH65:
Nice catch, thanks. I must have read that so many times and not seen it!

lurch:
Don't panic James, it was only a [recent change](https://github.com/raspberrypi/documentation/pull/1610/files) :wink:  (which I'm kinda surprised that I missed too!)

PR #1628, opened 2020-07-22
Fix missing details with VNC setup

rugk:
* The `/root/.vnc/config.d/vncserver-x11` did not have any `Authentication`entries in my case, so we cannot replace, but append a string
* Add a note on how to restart the vnc server (I hope I am right that I need to restart both `vncserver-virtuald.service` and `vncserver-x11-serviced.service`)
* Add a note on how to open/edit the file.
* Clarify where to find the "VNC Server dialog". 
* Even if you do not connect it to a screen, you can start (and autologin)

### DIFF
--- a/remote-access/vnc/README.md
+++ b/remote-access/vnc/README.md
@@ -94,10 +94,10 @@ If you're connecting from the [compatible VNC Viewer app](https://www.realvnc.co
 If you're connecting from a non-RealVNC Viewer app, you'll first need to downgrade VNC Server's authentication scheme, specify a password unique to VNC Server, and then enter that instead.
 * If you are in front of your Raspberry Pi and can see its screen, open the VNC Server dialog on your Raspberry Pi, select **Menu > Options > Security**, and choose **VNC password** from the **Authentication** dropdown.
 * **Or** if you're configuring your Raspberry Pi remotely from the command line, then to make the changes for Service Mode (the default configuration for the Raspberry Pi):
-  * Open the `/root/.vnc/config.d/vncserver-x11` config file.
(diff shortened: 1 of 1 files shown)


rugk:
Note I did not include the workaround for https://github.com/raspberrypi/noobs/issues/599 in here.

andrum99:
> * Even if you do not connect it to a screen, you can start (and autologin)

Enabling boot to desktop and autologin does **not** force display output. (You've also added this to the virtual desktop section of the VNC instructions, which are for the case where the user has the local display disabled).

andrum99:
> > * Even if you do not connect it to a screen, you can start (and autologin)
> 
> Enabling boot to desktop and autologin does **not** force local display output, nor is it meant to. (You've also added this to the virtual desktop section of the VNC instructions, which is for the case where the user wishes to have the VNC desktop separate from any local display which may or may not be running).

Sorry, I'm talking rubbish. On Pi Zero, 1, 2 and 3 if there is no screen detected on HDMI, the display will come up on composite: on Pi 4 you get no display.

The fact quite a few folk seem to have trouble with VNC means the documentation could probably do with improving.

andrum99:
See #1631

rugk:
> Enabling boot to desktop and autologin does not force local display output, nor is it meant to

Really? In my case it always worked. Do you have a quote/source for this?

JamesH65:
Has this PR been replaced by #1631? @andrum99

andrum99:
> Has this PR been replaced by #1631? @andrum99

That was the intention originally, but I think I need to verify the proposed additions/corrections in this PR as well, since it seems the docs as currently written may not match the current situation.

aallan:

aallan:

PR #1629, opened 2020-07-23
Update mac.md

GokselKUCUKSAHIN:
if and of parameters locations were twisted. Must be if=/'FROM' of=/'TO'.

### DIFF
--- a/installation/installing-images/mac.md
+++ b/installation/installing-images/mac.md
@@ -61,7 +61,7 @@ If you specify the wrong device in the instructions, you could overwrite your pr
 - Copy the image
 
   ```bash
-  sudo dd bs=1m if=path_of_your_image.img of=/dev/rdiskN; sync
+  sudo dd bs=1m if=/dev/rdiskN of=path_of_your_image.img; sync
   ```
 
    Replace `N` with the number that you noted before. Note the ```rdisk``` ('raw disk')


lurch:
Are you sure? `if` is "input file" and `of` is "output file", so I believe `sudo dd bs=1m if=path_of_your_image.img of=/dev/rdiskN; sync` copies the disk-image to your actual SD-card ?

GokselKUCUKSAHIN:
> Are you sure? `if` is "input file" and `of` is "output file", so I believe `sudo dd bs=1m if=path_of_your_image.img of=/dev/rdiskN; sync` copies the disk-image to your actual SD-card ?

In macOS, if=/name_of_diskutil_partition_name of=/img_save_path; 
Input file is SD Card Actual image and Output File() is Created Clone Image from SD Card.

Ekran Resmi 2020-07-23 15 23 04

disk2 is my 16Gig RaspberryPi SD Card. and I unmounted properly. 
On Desktop i have PiImage folder inside ddIMG directory. 
Ekran Resmi 2020-07-23 15 25 10

After Switching if and of parameters dd works just fine. (But very slow :()

lurch:
> In macOS, if=/name_of_diskutil_partition_name of=/img_save_path;
Input file is SD Card Actual image and Output File() is Created Clone Image from SD Card.

Yes, that's correct if you're wanting to create a backup image of your SD card, but the page you're modifying here is talking about copying an image TO your SD card?? ping @aallan in case I'm missing anything obvious.

> After Switching if and of parameters dd works just fine. (But very slow :()

Yup, it'll take a long time because it needs to read the whole 16GB from your SD card (`dd` is a low-level tool and so has no idea which parts of the SD card are "used" or not).

GokselKUCUKSAHIN:
Ohhh. I just realized that. Found this article searching for cloning SD card on MacOS. I guess Cloning and Copying images are too similar approaches. 
We need SD Cloning tool like inside like Raspberry PI OS for other operating systems.
Thanks for your time.

PR #1630, opened 2020-07-23
Add a caveat when using CCache in a CI environment

native-api:
Since the toolchain is `git clone`'d rather than extracted from a tarball, mtimes of its executables are different each time, invalidating CCache's cache if used in a CI environment.
See e.g. https://travis-ci.community/t/bad-cache-hit-rate-ccache/9400 for an example of this problem.

### DIFF
--- a/linux/kernel/building.md
+++ b/linux/kernel/building.md
@@ -125,8 +125,8 @@ If you are using a 32-bit operating system (for example, our Raspberry Pi Deskto
 
 `sudo apt install zlib1g-dev:amd64`
 
-If you are using CCache and a CI environment, instruct CCache to not use the compiler's mtime for cache ID calculations.
-This is because Git intentionally doesn't save file timestamps, so each time you clone the toolchain its file mtimes are different, invalidating CCache's cache when default settings are used.
+If you are using Ccache and a CI environment, instruct Ccache to not use the compiler's mtime for cache ID calculations.
+This is because Git intentionally doesn't save file timestamps, so each time you clone the toolchain its file mtimes are different, invalidating Ccache's cache when default settings are used.
 
 `ccache --set-config=compiler_check=content`
 
--- a/linux/kernel/building.md
+++ b/linux/kernel/building.md
@@ -125,8 +125,8 @@ If you are using a 32-bit operating system (for example, our Raspberry Pi Deskto
 
 `sudo apt install zlib1g-dev:amd64`
 
-If you are using CCache and a CI environment, instruct it to not use compiler's mtime for cache ID calculations.
-This is because Git intentionally doesn't save file timestamps, so each time you clone the toolchain, its file mtimes are different, invalidating CCache's cache with default settings.
+If you are using CCache and a CI environment, instruct CCache to not use the compiler's mtime for cache ID calculations.
+This is because Git intentionally doesn't save file timestamps, so each time you clone the toolchain its file mtimes are different, invalidating CCache's cache when default settings are used.
 
 `ccache --set-config=compiler_check=content`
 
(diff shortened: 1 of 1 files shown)


lurch:
I'm not sure if this is relevant to our kernel building instructions? Pinging @pelwell for comment
a) I don't think this is specific to compiling the Raspberry Pi kernel
b) surely this is a generic CCache problem, and ccache isn't mentioned anywhere else in our documentation?
c) my comments should be taken with a pinch of salt as I've not used ccache myself :wink:

native-api:
This is relevant to the kernel building instructions because
* they instruct to get the toolchain in an unusual way -- `git clone` instead of installing package or unpacking a tarball -- which is the root cause for this problem. It's more relevant to them than CCache documentation because it's them rather than CCache that use an uncommon practice.
* CCache is the standard product to use to save on long compilation times so it's very likely that someone uses it to build the kernel

pelwell:
It seems like good advice to me.

JamesH65:
Probably not a huge number of people actually doing this, but seems a sane change nevertheless.

linux/kernel/building.md
```diff
@@ -125,6 +125,11 @@ If you are using a 32-bit operating system (for example, our Raspberry Pi Deskto
 
 `sudo apt install zlib1g-dev:amd64`
 
+If you are using CCache and a CI environment, instruct it to not use compiler's mtime for cache ID calculations.
```
lurch, line 128: "and a CI environment" -> "(e.g. in a CI environment)" ?
"use compiler's" -> "use the compiler's" ?

linux/kernel/building.md
```diff
@@ -125,6 +125,11 @@ If you are using a 32-bit operating system (for example, our Raspberry Pi Deskto
...
+This is because Git intentionally doesn't save file timestamps, so each time you clone the toolchain, its file mtimes are different, invalidating CCache's cache with default settings.
```
lurch, line 129: Perhaps it's just me, but "invalidating CCache's cache with default settings" makes it sounds like it's the "default settings" that are going into the cache, and making it invalid? Perhaps a small tweak to:
"invalidating CCache's cache when default settings are used" ? (and perhaps the comma after "toolchain" can be removed?)

linux/kernel/building.md
```diff
@@ -125,6 +125,11 @@ If you are using a 32-bit operating system (for example, our Raspberry Pi Deskto
...
+
+`ccache --set-config=compiler_check=content`
```
lurch, line 131: Maybe this should go _between_ the previous two lines? :man_shrugging:

native-api: I was afraid that this would make it unclear which text relates to which code -- since the previous sections are always "chunk of text - chunk of code". If there's no such problem, fine by me.

linux/kernel/building.md
native-api: The CI environment is required, too, so no "e.g.". The problem only appears if you clone the tools anew each build.
Fine by me otherwise.

linux/kernel/building.md
native-api: Fine by me. Overkill or not, it's only better if the formulations are clearer.

linux/kernel/building.md
```diff
@@ -125,6 +125,11 @@ If you are using a 32-bit operating system (for example, our Raspberry Pi Deskto
 
 `sudo apt install zlib1g-dev:amd64`
 
+If you are using CCache and a CI environment, instruct CCache to not use the compiler's mtime for cache ID calculations.
+This is because Git intentionally doesn't save file timestamps, so each time you clone the toolchain its file mtimes are different, invalidating CCache's cache when default settings are used.
```
jrosdahl, line 129: I suggest using the preferred capitalization of Ccache:

```suggestion
If you are using Ccache and a CI environment, instruct Ccache to not use the compiler's mtime for cache ID calculations.
This is because Git intentionally doesn't save file timestamps, so each time you clone the toolchain its file mtimes are different, invalidating Ccache's cache when default settings are used.
```

PR #1635, opened 2020-07-30
Update linux.md

SKelem:
Added sync as part of the dd write commands. Logically, sync should be executed at the end of the dd writes to the SD card. Organizationally, this page has the sync separated from the dd command by at least half a page, separated by a section on verification, and formatted as text, rather than as the essential command that it is.

### DIFF
--- a/installation/installing-images/linux.md
+++ b/installation/installing-images/linux.md
@@ -20,7 +20,7 @@
 - In a terminal window, write the image to the card with the command below, making sure you replace the input file `if=` argument with the path to your `.img` file, and the `/dev/sdX` in the output file `of=` argument with the correct device name. **This is very important, as you will lose all the data on the hard drive if you provide the wrong device name.** Make sure the device name is the name of the whole SD card as described above, not just a partition. For example: `sdd`, not `sdds1` or `sddp1`; `mmcblk0`, not `mmcblk0p1`.
 
     ```bash
-    dd bs=4M if=2020-02-13-raspios-buster.img of=/dev/sdX conv=fsync
+    dd bs=4M if=2020-02-13-raspios-buster.img of=/dev/sdX conv=fsync; sync
     ```
 
 - Please note that block size set to `4M` will work most of the time. If not,  try `1M`, although this will take considerably longer.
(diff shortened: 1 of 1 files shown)


lurch:
I believe the `conv=fsync` option means that the separate `sync` isn't necessary? Given that, the bit about `sync` at the bottom of the page can probably be removed?

SKelem:
Damn! You're right!

I thought I had figured out why my RPi 3 becomes unbootable during the automatic updates.

Steve

Andrew Scheller wrote on 7/30/20 11:49 AM:

lurch:
> why my RPi 3 becomes unbootable during the automatic updates

You could try asking for help on the https://www.raspberrypi.org/forums/ ?

JamesH65:
I believe this PR can be closed as not needed?

SKelem:
Not sure what to say. I really would like to get the RPi to work. It's been sitting in the "hopeless" box for the last two months. I'm just working up the courage to tackle it again. The trouble is that it takes about a day to go from "transfer image to microSD" to "all the updates have been installed, and now it won't boot". I'm considering videoing the process to document how many times it reboots during the installation and update process.

JamesH65:
? I run Raspberry Pi imager, it downloads the files and programs the SD car in about 10 minutes, put it in Pi, run though the setup screen, about 2 minutes, then download any new updates, up to 10 minutes. Then  it just works. If its taking longer than 30 minutes to go from boxed to working, something is going wrong.

JamesH65:
To add, at the most it should reboot once.

SKelem:
Imaging works fine. I do it on Linux, so I'm using "dd". The long part is after the RPi is resizing the partition, installing software, ...

Is there a troubleshooting document?

Steve

James Hughes wrote on 10/5/20 2:39 PM:

JamesH65:
There is a Troubleshooting thread on the forums, and that is really where these sort of issues should be discussed, not in the docs repo. Read the thread, then start  a thread with your specific issues.

lurch:
It might be that you have an old or especially slow SD card? If so, getting a newer (class A1 e.g. https://www.amazon.co.uk/dp/B073JWXGNT/ ) card might make things quicker / smoother?

SKelem:
I made some progress.

No, the SD cards are new, verified really from SanDisk.

The RPi is part of OpenSprinkler. I separated the two boards, and RPi is working properly by itself. I suspect there's a problem with the high-order address bits on the OpenSprinkler board, which would allow software to run properly until the high-order (or middle-order?) address bits are used. The fact that the software does run "for a while" implies that the low-order address bits are okay.

I filed an issue with OpenSprinkler.

Steve

Andrew Scheller wrote on 10/6/20 3:18 AM:

lurch:
Closing PR, as the problem has been identified as being elsewhere.

PR #1637, opened 2020-07-31
Explicitly call-out Vendor Class identifier for PXE

virtualjmills:
Vendor Class identifier was buried in some tcpdump parsed output further down in the page. Explicitly calling it out up-top, in some environments this is used as an initial pass filter by the DHCP server to determine which set of responses / behavior to use when interacting with the PXE Client system.

### DIFF
--- a/hardware/raspberrypi/bootmodes/net.md
+++ b/hardware/raspberrypi/bootmodes/net.md
@@ -6,7 +6,7 @@ We also have a [tutorial about setting up a network boot system](net_tutorial.md
 To network boot, the boot ROM does the following:
 
 * Initialise on-board Ethernet device (Microchip LAN9500 or LAN7500)
-* Send DHCP request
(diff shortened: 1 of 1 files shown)


lurch:
Is this the same for network-booting on both a Pi 3B+ and also a Pi 4B ?
ping @timg236

JamesH65:
@timg236 Are you OK with this change? Seems sane to me.

timg236:
The ROM doesn't do TFTP boot on Pi4 although the behaviour is the same in bootloader. I think this page is really about Pi3 ROM

lurch:
I wonder if that page needs to be "refactored" slightly to show which parts of the net-boot process are different between Pi3 ROM and Pi4 EEPROM (e.g. Pi4 EEPROM probably doesn't request `bootcode.bin`?), and which parts are common to both? :shrug:

aallan:

See #1911 for full details.

PR #1638, opened 2020-08-03
add a tip

ludertal:
maybe it would be useful to add that you need (I had to) enable your usb camera with raspi-config before the system can use it, otherwise the /dev/video0 will not be available !


JamesH65:
USB camera have nothing to do with the Pi camera, so it should not be necessary to enable the Pi camera to have USB cameras working.

PR #1644, opened 2020-08-10
docs: vcgencmd get_throttled: add hex values

tedder:
since the 0th bit is actually value 1, I find it's easier to decompose get_throttled by pulling out the hex values rather than converting to binary. Hopefully this is instructive to others.


lurch:
The example immediately below this table shows how to decode `0x50000` into binary. People might get confused by the hex column you've added not having an entry for `50000`? :shrug: 
I dunno, I'll leave it up to @JamesH65

tedder:
Good point. It's all about finding the precise wording. I added an explanation. I can't remember what the process of "recursively subtracting the largest value" is actually called, I have "greatest common divisor" in my head, which is wrong.

JamesH65:
I'm OK with this.

PR #1648, opened 2020-08-17
Indicate that RESTART (0xf) was added later

iosabi:
Bootloader pieeprom-2020-04-16 added the `BOOT_ORDER` setting but it didn't include the 0xf option. Using 0xf in that version will cause the following error:

Unknown boot mode 0xf 0x0

This option is present in the pieeprom-2020-06-15.bin, which is the same version that added the related setting `MAX_RESTARTS`.


lurch:
ping @timg236

timg236:
Submit a PR :)

iosabi:
@timg236 , I'm confused. Are you asking me to submit a PR? this is a PR already, should I do something different? I was about to file a but about 0xf not working in 2020-04-16 when I realized and then confirmed that it was added later so I sent this PR instead.

iosabi:
ping @timg236

lurch:
This was resolved in #1661

PR #1652, opened 2020-08-25
Declare build instructions as 32-bit

RyanHir:
As the Raspberry Pi OS is beginning support for a 64-bit edition, we should start updating the documentation for building between the two. This will be the first commit by me adding documentation for compiling a 64-bit kernel in the future.

I am wanting to update [this](https://www.raspberrypi.org/documentation/linux/kernel/building.md) page which is about compiling the kernel yourself for the Pi.

As a start I just changed the cross compile title from `Cross-compiling` to `Cross-compiling 32-Bit`

As I work on a guide for 64 bit compiling this should help clarify that this is for the 32-bit OS.

### DIFF
--- a/linux/kernel/building.md
+++ b/linux/kernel/building.md
@@ -92,7 +92,7 @@ sudo cp arch/arm/boot/zImage /boot/$KERNEL.img
 
 **Note**: On a Raspberry Pi 2/3/4, the `-j4` flag splits the work between all four cores, speeding up compilation significantly.
 
-## Cross-compiling
+## Cross-compiling 32-bit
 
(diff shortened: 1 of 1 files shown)


JamesH65:
I think we may need to be a little more fine grained. For example, most of the instruction in that section will be the same for 32 and 64 bit builds. I am not sure of the exact differences in the builds, will find out.

RyanHir:
Thanks for the comment. I am aware of the similarities between the build guides and I am just marking it as 32 bit untill I finish my guide but if it is not what yall are looking for I can close this and make another request when my guide is done which is nested in the 32 bit guide.

RyanHir:
It looks like it is adding commits I don't want to merge yet yet so I will close this pull request

PR #1653, opened 2020-08-25
WIP: Add Build instructions for compiling a 64-Bit kernel

RyanHir:
In this merge request I will be adding instructions for compiling a 64-bit kernel for any device that supports the 64-bit instructions.

As the Raspberry Pi OS moves to be cross arch we should update the build instructions for the kernel to better clarify how to compile for both instruction sets. 

As the build instructions are mostly the same between arches, I will have notes in the markdown that states if X is specific to 32 or 64 bit.

This request is the predecessor of another pull request #1652 that I have done which I should have waited to do.

Tested build on 64-Bit Pi 3 and derivatives
Tested build on 64-Bit Pi 4


RyanHir:
This request is the predecessor of another pull request #1652 that I have done which I should have waited to do.

lurch:
Maybe this is a bit premature, and we ought to avoid adding this to the official documentation until the 64-bit Raspberry Pi OS is out of beta? :shrug:

RyanHir:
That is reasonable, I will leave this pull request up until the 64-Bit Raspberry Pi OS comes out of beta for possible merging into the docs.

lurch:
I've labelled it as WIP, to tag that it's not ready to be merged yet.

JamesH65:
@popcornmix @pelwell Can either of you guys give this a quick once over to check its sane? Seems like a good addition to me.

popcornmix:
Looks reasonable. I notice that "64-Bit Configs" section says "For Pi 3, Pi 3+ or Compute Module 3" and "For Raspberry Pi 4" with the less straightforward Raspberry Pi 2 ignored.

linux/kernel/building.md
```diff
@@ -208,6 +251,17 @@ sudo umount mnt/fat32
...
 ```
 
+#### For 64-Bit
+
+```bash
+sudo cp mnt/fat32/$KERNEL.img mnt/fat32/$KERNEL-backup.img
+sudo cp arch/arm64/boot/Image mnt/fat32/$KERNEL.img
+sudo cp arch/arm64/boot/dts/broadcom/*.dtb mnt/fat32/
+sudo cp arch/arm64/boot/dts/overlays/*.dtb* mnt/fat32/overlays/
+sudo umount mnt/fat32
```
pelwell, line 261: This fails to copy the README.

linux/kernel/building.md
```diff
@@ -103,10 +103,21 @@ You can either do this using VirtualBox (or VMWare) on Windows, or install it di
 
 To build the sources for cross-compilation, make sure you have the dependencies needed on your machine by executing:
 ```bash
-sudo apt install git bc bison flex libssl-dev make libc6-dev libncurses5-dev crossbuild-essential-armhf
+sudo apt install git bc bison flex libssl-dev make libc6-dev libncurses5-dev
 ```
+
 If you find you need other things, please submit a pull request to change the documentation.
 
+#### Install the 32-Bit toolchain for a 32-Bit Kernel
```
pelwell, line 111: On a stylistic note, I don't like the random capitalisation of certain words such as "Bit", "Kernel" and "Config".

linux/kernel/building.md
```diff
@@ -92,7 +92,7 @@ sudo cp arch/arm/boot/zImage /boot/$KERNEL.img
 
 **Note**: On a Raspberry Pi 2/3/4, the `-j4` flag splits the work between all four cores, speeding up compilation significantly.
 
-## Cross-compiling
+## Cross-compiling 32/64-bit
```
pelwell, line 95: Does this add anything?

linux/kernel/building.md
```diff
@@ -190,14 +224,23 @@ sudo mount /dev/sdb6 mnt/fat32
...
-Next, install the modules:
 
+Next, install the kernel modules onto the SD card:
+
+#### For 32-Bit
 ```bash
 sudo env PATH=$PATH make ARCH=arm CROSS_COMPILE=arm-linux-gnueabihf- INSTALL_MOD_PATH=mnt/ext4 modules_install
 ```
 
+#### For 64-Bit:
```
pelwell, line 235: Inconsistent use of a colon.

linux/kernel/building.md
RyanHir: The readme is not generated on a his build so there is no file to copy

linux/kernel/building.md
RyanHir: Resolved in new commit

linux/kernel/building.md
RyanHir: Resolved in new commit

RyanHir:
@popcornmix that change to add pi 2 instructions would not be possible for a 64-bit kernel as the device only has a 32-bit processor

linux/kernel/building.md
RyanHir: Resolved in b66bbe7

linux/kernel/building.md
pelwell: What do you mean not generated in this build? If you are only updating the 64-bit kernel then the README should be copied.

RyanHir: When I build the kernel with the 64bit instructions, there is no README file to copy like the 32bit kernel build. I cant add to the instruction to copy a file that does not exist..

pelwell: The overlays are built _from arch/arm_, and the README must be copied _from arch/arm_. Look at the 32-bit instructions and you might understand.

RyanHir: interesting, i will do another build and check if it exists. If so, i will add it to the build instructions

pelwell: It's a source file - it's just copied.

popcornmix:
Newer revisions of Pi 2 are actually 64-bit (they use the same processor as Pi 3).
You can identify if this is the case from /proc/cpuinfo. But this does complicate the instructions so could be glossed over.

RyanHir:
Being how small I assume the pool of pi 2 users interested in the 64-bit kernel, I don't think it is worth adding to the instructions.

On top of that, the Pi 2 is not supported for the 64-bit Raspberry Pi OS.

But thank you for this info, I was not aware some Pi 2's shipped with a Pi 3 processor.

linux/kernel/building.md
RyanHir: You are right, i was normally building into a separate folder so i got confused.

Fixed in f9e1697

linux/kernel/building.md
```diff
@@ -103,10 +103,21 @@ You can either do this using VirtualBox (or VMWare) on Windows, or install it di
...
 ```
+
 If you find you need other things, please submit a pull request to change the documentation.
 
+#### Install the 32-bit toolchain for a 32-bit kernel
+```bash
+sudo apt install crossbuild-essential-armhf
+```
+
+#### Or, Install the 64-bit toolchain for a 64-bit kernel
```
lurch, line 116: Lowercase i on 'install' here please

lurch:
This only seems to add the cross-compiling instructions for building a 64-bit kernel. Should it also add the instructions for native-compiling a 64-bit kernel? (which I guess will be slightly different depending on whether you're running the 32-bit or 64-bit version of Raspberry Pi OS?)

linux/kernel/building.md
RyanHir: Fixed in 9ba510a

RyanHir:
Adding instructions for compiling 32 -> 32 or 64 -> 64 is quite simple, but in the case of compiler from 32 -> 64 or 64 -> 32, you might as well be using the cross compile instructions as they are not arch specific.

I can add a clarifier to choose the correct configs for compiling a 64 -> 64 build.

linux/kernel/building.md
```diff
@@ -103,10 +103,21 @@ You can either do this using VirtualBox (or VMWare) on Windows, or install it di
 
 To build the sources for cross-compilation, make sure you have the dependencies needed on your machine by executing:
 ```bash
-sudo apt install git bc bison flex libssl-dev make libc6-dev libncurses5-dev crossbuild-essential-armhf
+sudo apt install git bc bison flex libssl-dev make libc6-dev libncurses5-dev
 ```
+
 If you find you need other things, please submit a pull request to change the documentation.
 
+#### install the 32-bit toolchain for a 32-bit kernel
```
lurch, line 111: We **do** want a capital `I` here, because it's the first word in the header :smile:

RyanHir: My bad, miss understood your request. Fixed in f32ba83

linux/kernel/building.md
```diff
@@ -208,6 +251,18 @@ sudo umount mnt/fat32
 sudo umount mnt/ext4
```
lurch, line 251: I wonder if it might make sense to move the `sudo umount mnt/ext4` lines immediately after the `modules_install` lines? :shrug:

RyanHir: While i like that idea, i think that may be outside the scope of this PR, if you fell otherwise I would be glad to change it.

pelwell: I agree - let's not make the perfect the enemy of the good.

lurch:
>  but in the case of compiler from 32 -> 64 or 64 -> 32, you might as well be using the cross compile instructions as they are not arch specific.

Probably worth adding a note to that effect? Something like "If you want to compile a 64-bit kernel on a 32-bit Raspberry Pi OS, or compile a 32-bit kernel on a 64-bit Raspberry Pi OS, then follow the cross-compilation instructions elsewhere on this page." ?
Or do you think that would be too confusing @pelwell @JamesH65 ?

> I can add a clarifier to choose the correct configs for compiling a 64 -> 64 build.

I think that would be useful :+1: 
(as well as perhaps clarifying that the existing instructions are for compiling a 32-bit kernel on a 32-bit Raspberry Pi OS?)

pelwell:
It's not confusing, it's correct. What's needed is a small note near the top of the file saying briefly what cross-compilation is (building for an architecture that isn't the one doing the building), including 64-bit on 32-bit and vice versa as example, and referring to the cross-compilation section.

RyanHir (commented):
EDIT: ~~This probably needs some updating but it is a start~~ This was meant to be a comment on a individual patch, look ahead.

linux/kernel/building.md
```diff
@@ -1,8 +1,8 @@
 # Kernel building
 
-There are two main methods for building the kernel. You can build locally on a Raspberry Pi, which will take a long time; or you can cross-compile, which is much quicker, but requires more setup.
+There are two ways of compiling the linux kernel. One is to compile nativly or cross architecture via cross compiling. Cross compiling is used to compile the kernel for an architecture (such as arm64) on when on another architecture (such as armhf on the Raspberry Pi OS 32-bit).
```
RyanHir, line 3: I meant this paragraph probably needs updating but I added some starting information

lurch: Something like this might be better?

There are two main methods for building the kernel - natively and cross-compiling. Native compilation is the easiest approach, and is used when the architecture you're building the kernel on (host platform) is the same as the architecture you'll be running it on (target platform); such as building a 32-bit Raspberry Pi kernel using 32-bit Raspberry Pi OS, or building a 64-bit Raspberry Pi kernel using 64-bit Raspberry Pi OS. Cross-compiling is used when the "host platform" is different to the "target platform", such as building a 64-bit Raspberry Pi kernel using 32-bit Raspberry Pi OS (or vice versa), or when building a Raspberry Pi kernel (either 32 or 64-bit) on an separate Linux PC. Cross-compilation is slightly more involved, but building a Raspberry Pi kernel is typically much faster when cross-compiled on a separate Linux PC than when built natively on a Raspberry Pi.

I suspect @pelwell and @JamesH65 may have other suggestions / tweaks to the above too. Not sure if what I've suggested here is too wordy?

linux/kernel/configuring.md
```diff
@@ -22,12 +22,18 @@ Once you've got everything set up and ready to go, you can compile and run the `
 $ make menuconfig
 ```
 
-If you're cross-compiling,:
+If you're cross-compiling a 32-Bit kernel:
```
lurch, line 25: Lower-case 'b' on 'bit' here (and also below) please :slightly_smiling_face:

linux/kernel/building.md
pelwell: My attempt:
> The default compilers and linkers that come with an OS are configured to build executables to run on that OS - they are _native_ tools - but that doesn't have to be the case. A cross-compiler is configured to build code for a target other than the one running the build process, and using it is called cross-compilation.
>
> Cross-compilation of the Raspberry Pi kernel is useful for two reasons:
> 1. it allows a 64-bit kernel to be built using a 32-bit OS, and vice versa, and
> 2. even a modest laptop can cross-compile a Pi kernel significantly faster than the Pi itself.
>
> The instructions below are divided into native builds and cross-compilation; choose the section appropriate for your situation - although there are many common steps between the two there are also some important differences.

linux/kernel/configuring.md
RyanHir: Fixed in 4ee1b6a

linux/kernel/building.md
RyanHir: I personally like @pelwell 's attempt as it is not as wordy. Perhaps we should add hyper links to the section of the document in the header?

RyanHir: I will add pelwell's header for now

pelwell:
Is there anything outstanding on this PR before it can be merged? We've recently pointed people at this PR because it's better than the current published version.

lurch:
I think it still needs "compiling a 64-bit kernel on 64-bit Raspberry Pi OS" instructions?

pelwell:
We've not had that before - can't it be added in a follow-up commit?

lurch:
I guess so! Although in that case I guess this PR ought to be renamed "Add Build instructions for cross-compiling a 64-bit kernel" :wink: 

This PR LGTM otherwise - thanks for your work on this @RyanHir

PR #1657, opened 2020-08-28
Update image names

XECDesign:


lurch:
The raspios filenames seem to be missing the `-armhf` part of the filename?
`2020-08-20-raspios-buster-armhf.zip`

XECDesign:
Yeah sorry, my script just sed's the date. I'll update it to get the full filename to avoid changes like this from getting dropped.

PR #1660, opened 2020-08-30
Use distro cross-compilers in kernel build

RyanHir:
The compiler given in the [raspberrypi/tools](https://github.com/raspberrypi/tools/tree/master/arm-bcm2708/arm-rpi-4.9.3-linux-gnueabihf) are fairly old and all the supported versions of Ubuntu and Debian have either the supported version or newer of the compiler offered before.

This can be seen at [here for Ubuntu 16.04](https://packages.ubuntu.com/xenial/gcc-arm-linux-gnueabihf) and [here for Debian Stretch](https://packages.debian.org/stretch/gcc-arm-linux-gnueabihf). Both of which is the oldest supported versions of their respective distro.

While this is not essential, I think that this will help get users setup quicker as it already has the proper paths set.

### DIFF
--- a/linux/kernel/building.md
+++ b/linux/kernel/building.md
@@ -114,11 +114,6 @@ Use the following command to install the cross-arch compiler onto your machine:
 sudo apt install crossbuild-essential-armhf
 ```
 
-If you are using Ccache and a CI environment, instruct Ccache to not use the compiler's mtime for cache ID calculations.
-This is because Git intentionally doesn't save file timestamps, so each time you clone the toolchain its file mtimes are different, invalidating Ccache's cache when default settings are used.
-
-`ccache --set-config=compiler_check=content`
-
 ### Get sources
 
 To download the minimal source tree for the current branch, run:
--- a/linux/kernel/building.md
+++ b/linux/kernel/building.md
@@ -99,20 +99,13 @@ also a Debian distribution, it means many aspects are similar, such as the comma
 
 You can either do this using VirtualBox (or VMWare) on Windows, or install it directly onto your computer. For reference, you can follow instructions online [at Wikihow](http://www.wikihow.com/Install-Ubuntu-on-VirtualBox).
 
-### Install required dependencies
+### Install required dependencies and toolchain
 
 To build the sources for cross-compilation, make sure you have the dependencies needed on your machine by executing:
 ```bash
-sudo apt install git bc bison flex libssl-dev make libc6-dev libncurses5-dev
+sudo apt install git bc bison flex libssl-dev make libc6-dev libncurses5-dev crossbuild-essential-armhf
 ```
 If you find you need other things, please submit a pull request to change the documentation.
-
-### Install toolchain
-
-Use the following command to install the cross-arch compiler onto your machine:
-```bash
-sudo apt install crossbuild-essential-armhf
-```
 
 ### Get sources
 
--- a/linux/kernel/building.md
+++ b/linux/kernel/building.md
@@ -109,18 +109,11 @@ If you find you need other things, please submit a pull request to change the do
 
 ### Install toolchain
 
-Use the following command to download the toolchain to the home folder:
-
+Use the following command to install the cross-arch compiler onto your machine:
 ```bash
-git clone https://github.com/raspberrypi/tools ~/tools
+sudo apt install binutils-arm-linux-gnueabihf crossbuild-essential-armhf
 ```
 
-Updating the $PATH environment variable makes the system aware of file locations needed for cross-compilation. 
-
-```bash
-echo PATH=\$PATH:~/tools/arm-bcm2708/arm-linux-gnueabihf/bin >> ~/.bashrc
-source ~/.bashrc
-```
 If you are using a 32-bit operating system (for example, our Raspberry Pi Desktop for PC), then you may need to install an additional set of libraries:
 
 `sudo apt install zlib1g-dev:amd64`
--- a/linux/kernel/building.md
+++ b/linux/kernel/building.md
@@ -103,7 +103,7 @@ You can either do this using VirtualBox (or VMWare) on Windows, or install it di
 
 To build the sources for cross-compilation, make sure you have the dependencies needed on your machine by executing:
 ```bash
-sudo apt install git bc bison flex libssl-dev make libc6-dev libncurses5-dev
+sudo apt install git bc libssl-dev libncurses5-dev
 ```
 If you find you need other things, please submit a pull request to change the documentation.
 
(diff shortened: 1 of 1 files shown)


lurch:
ping @pelwell

pelwell:
It seems to work, but we should probably retain the old method for platforms that aren't Debian-based.

Any concerns, @popcornmix?

popcornmix:
I think this makes sense. The standard repo toolchains will work just fine (and in fact are required for aarch64 builds).

pelwell:
The rest of the instructions assume `apt`, so I withdraw the request to not delete the previous version.

RyanHir:
The other thing is the documentaton specifically says that it uses Ubuntu and already uses APT so perhaps if we dont want to use the repo tools then we should also change the other instructions or move to a container like build to have a common environment.

While I personally like the idea of using Docker or Podman, i dont think most people wont so for the time being I think we should continue to target Ubuntu (and Debian derivatives)

RyanHir:
And in the case we do want to add support for other distros such as Fedora or Arch. I would not expect it to be that much of a problem to have distro specific instructions as most of the build is distro agnostic.

linux/kernel/building.md
```diff
@@ -109,18 +109,11 @@ If you find you need other things, please submit a pull request to change the do
...
+sudo apt install binutils-arm-linux-gnueabihf crossbuild-essential-armhf
 ```
 
-Updating the $PATH environment variable makes the system aware of file locations needed for cross-compilation. 
-
-```bash
-echo PATH=\$PATH:~/tools/arm-bcm2708/arm-linux-gnueabihf/bin >> ~/.bashrc
-source ~/.bashrc
-```
 If you are using a 32-bit operating system (for example, our Raspberry Pi Desktop for PC), then you may need to install an additional set of libraries:
```
lurch, line 117: Are the above-mentioned cross-compilers available in "Raspberry Pi Desktop for PC" ? If they are, does that mean the additional `zlib1g-dev:amd64` mentioned below isn't needed?

lurch:
ping @native-api I think this PR might mean that the CCache instructions (added in #1630 ) could be removed too?

linux/kernel/building.md
RyanHir: It appears that the compilers are in the 32bit repo's as seen here https://packages.debian.org/stretch/i386/gcc-arm-linux-gnueabihf/download and after spinning up a vm the tools are there when needed. I am not familiar with zlib1g so I will leave that up to @native-api as he appears to be knowledgeable.

RyanHir: So i took another look at the dependencies and it appears that zlib1g is required when pulling the cross compiler as seen [here](https://packages.debian.org/stretch/gcc-6-arm-linux-gnueabihf).

My only concern is that when installing on a 32bit platform it will install the 32bit version of zib1g so for some unknown reason that the amd64 version is required i think that could be a bug with the build system? i am grasping at straws at this point.

native-api:
@lurch Yes, Apt packages use tarballs under the hood to hold files which preserve timestamps upon restore. So the CCache workaround would no longer be needed if Apt packages are used instead of a Git checkout.

native-api:
@lurch I would still double-check though. It's possible that the packages create something on the fly (like symlinks) in the postinstall step which may or may not have the same timestamps and/or confuse Ccache.

linux/kernel/building.md
native-api: Thanks for putting so much faith in me but I've never compiled the Pi kernel with 32-bit tools :-)
All my knowledge comes from diagnosing the problem that I've linked to in my PR.

RyanHir:
I would like to confirm @native-api findings as when i install the compiler on my Debian Sid installation i get a binary that was modified days ago not minutes. So i have reason to believe that ccache will not complain.

```bash
ryan@think:~$ ls -l $(whereis arm-linux-gnueabihf-gcc | awk '{print $2}')
lrwxrwxrwx 1 root root 26 Jul 21 09:16 /usr/bin/arm-linux-gnueabihf-gcc -> arm-linux-gnueabihf-gcc-10
```

linux/kernel/building.md
lurch: IIRC the `zlib1g-dev:amd64` was only needed on the "32-bit Raspberry Pi Desktop for PC" because the compiler-tools being installed from the tools-repo were themselves actually 64-bit (and so needed the 64-bit zlib). See #1413

But as the compiler-tools (installed from the repo) will now be 32-bit, then the compilers will only need the 32-bit zlib.

At least, that's the way I understand it, but @pelwell and @XECDesign probably understand it better than I do!

RyanHir: Thank you for the clarification! That makes the most sense!

RyanHir (commented):
So this change removes some packages from the first pull and allow the meta-package for cross compiling to pull the packages via the --install-recommends.

lurch:
Is there any guarantee that `--install-recommends` will _always_ install all the packages needed? Might be better to be explicit? :shrug:

lurch:
Also, do you want to remove the ccache parts?

RyanHir:
--install-recommends will work unless the maintainer removes the package from the meta-package so maybe that should be revisited.

i have left the ccache part as I was not sure what was wanted, but i presume it is not needed anymore as we have confirmed the binaries hold their timestamp. I will work on that now.

RyanHir:
Would there be benifit by using the build-essential meta-package as it is quite static and contains libc6-dev, make, and the host platform compilers which will be needed for menuconfig which is talked about later. 

A new debian or ubuntu install will not contain the host gcc and the meta package will only install the target gcc. With this configuration then the instructions at https://github.com/raspberrypi/documentation/blob/master/linux/kernel/configuring.md are missing information.

lurch:
Hmmm, now that the "Install toolchain" section is just `sudo apt install crossbuild-essential-armhf` it probably makes sense to just include that in the "Install required dependencies" section? (i.e. append crossbuild-essential-armhf to the end of the other "apt install" line)

RyanHir:
Looking at it, i will agree it would just be better to be all one command

RyanHir:
Would there be benefit for adding a note for non Debian/Ubuntu users about what minimum version of the cross compiler that they should install?

lurch:
The page already talks about running Ubuntu inside a VM, so probably not any need for non Debian/Ubuntu instructions? :shrug:

RyanHir:
perhaps but the instructions are almost completly distro agnostic, if we just gave a list of generic packages like...
```md
 * git
 * bc
 * bison
 * flex
 * gcc-armhf >= 4.9.5
```
for those who have like a alpine build server of sorts. This is kinda far fetched so if it was wanted then i guess another pull request will do. might as well keep this PR clean.

lurch:
The more detailed the documentation is (e.g. listing specific versions of specific packages), the more of a maintenance burden it becomes if / when it has to change :wink: 

I suspect somebody setting up an Alpine Linux build server will already have the technical know-how to convert the Ubuntu instructions to their distro of choice?

RyanHir:
I presume so :D

JamesH65:
Going to merge this, think it makes sense as is.

PR #1666, opened 2020-09-03
WIP: Update FAQ to talk about the Rasberry Pi OS 64bit Edition.

RyanHir:
This is most definitely a WIP, I am just adding a start for what the FAQ will state when the 64-bit OS gets out of beta.

I am talking in present tense as if the OS is released, the URL may need to be changed later.

I am doing this as the FAQ talks as if there is no such supported OS for the Pi and I am just correcting that for the future..

### DIFF
--- a/faqs/README.md
+++ b/faqs/README.md
@@ -298,7 +298,6 @@ Do not use `rpi-update` unless you have been recommended to do so by a Raspberry
 
 ### The processors on the latest Raspberry Pi models are 64-bit, but I cannot find an official 64-bit OS.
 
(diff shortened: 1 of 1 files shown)


lurch:
Once the 64-bit OS comes out of beta, I guess it'll probably be added to the downloads page. At that point, doesn't the "but I cannot find an official 64-bit OS" part become rather redundant?

RyanHir:
Oh wow, i did not think about that. I presume that it is quite redundant if the link is in the same location.

PR #1669, opened 2020-09-10
GPIO permissions information

mfloryan:
When adding a new user they will not be able to program gpio unless they are part of the `gpio` group.


lurch:
Maybe it's worth re-using the text (with appropriate modifications) from https://github.com/raspberrypi/documentation/blob/master/usage/webcams/README.md ?

JamesH65:
Thanks!

PR #1670, opened 2020-09-11
Update headless.md

rohit-karthik:


lurch:
IMHO this change isn't needed

PR #1672, opened 2020-09-12
Adding tips when editing wpa_supplicant

ihatetoregister:
Clarify boot directory and add a warning about newline and file extension

### DIFF
--- a/configuration/wireless/headless.md
+++ b/configuration/wireless/headless.md
@@ -6,7 +6,7 @@ Once an image is created on an SD card, by inserting it into a card reader on a 
 
 ## Setting up wireless networking
 
-You will need to define a `wpa_supplicant.conf` file for your particular wireless network. Put this file in the `/boot` folder, and when the Pi first boots, it will copy that file into the correct location in the Linux root file system and use those settings to start up wireless networking. Depending on the OS and editor you are creating this on, the file could have incorrect newlines or the wrong file extension so make sure you use an editor that accounts for this. Linux expect the line feed (LF) newline character. For more information, see this [Wikipedia article](https://en.wikipedia.org/wiki/Newline). 
+You will need to define a `wpa_supplicant.conf` file for your particular wireless network. Put this file in the `boot` folder, and when the Pi first boots, it will copy that file into the correct location in the Linux root file system and use those settings to start up wireless networking. Depending on the OS and editor you are creating this on, the file could have incorrect newlines or the wrong file extension so make sure you use an editor that accounts for this. Linux expect the line feed (LF) newline character. For more information, see this [Wikipedia article](https://en.wikipedia.org/wiki/Newline). 
 
 `wpa_supplicant.conf` file example:
 ```
--- a/configuration/wireless/headless.md
+++ b/configuration/wireless/headless.md
@@ -6,7 +6,7 @@ Once an image is created on an SD card, by inserting it into a card reader on a 
 
 ## Setting up wireless networking
 
-You will need to define a `wpa_supplicant.conf` file for your particular wireless network. Put this file in the `/boot` folder, and when the Pi first boots, it will copy that file into the correct location in the Linux root file system and use those settings to start up wireless networking. Depending on the OS and editor you are creating this on, the file could have incorrect newlines or the wrong file extension so make sure you use an editor that accounts for this. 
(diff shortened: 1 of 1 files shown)


configuration/wireless/headless.md
```diff
@@ -6,7 +6,7 @@ Once an image is created on an SD card, by inserting it into a card reader on a
 
 ## Setting up wireless networking
 
-You will need to define a `wpa_supplicant.conf` file for your particular wireless network. Put this file in the boot folder, and when the Pi first boots, it will copy that file into the correct location in the Linux root file system and use those settings to start up wireless networking.
+You will need to define a `wpa_supplicant.conf` file for your particular wireless network. Put this file in the `/boot` folder, and when the Pi first boots, it will copy that file into the correct location in the Linux root file system and use those settings to start up wireless networking. Depending on the OS and editor you are creating this on, the file could have incorrect newlines or the wrong file extension so make sure you use an editor that accounts for this.
```
lurch, line 9: Probably useful to clarify what the "correct newlines setting" should be? (i.e. Unix line-ending)

ihatetoregister: Fixed

andrum99: The bit about file extensions got me wondering about whether Raspberry Pi OS would handle the case where the user had unwittingly created the file as `wpa_supplicant.conf.txt`, due to Windows hiding the file extension. It seems it does not.

For the `ssh` file, which may be added to the boot filesystem to enable SSH, it also checks if `ssh.txt` exists and enables SSH if that file exists too. There's no reason why the wpa_supplicant.conf copy service cannot do the same thing for the wpa_supplicant copy. I shall create a PR to do just that. In the meantime, you can remove the reference to the file extension from this PR. (If my PR gets rejected I wil submit a new PR to make the required addition to this part of the documentation).

andrum99: See https://github.com/RPi-Distro/raspberrypi-net-mods/pull/4

ihatetoregister: Yeah, that would be much better. I actually had this issue but on Mac, but I still think it is worth a heads up even if the Pi would be kind to correct it. Hope your PR gets through.

andrum99: > Yeah, that would be much better. I actually had this issue but on Mac, but I still think it is worth a heads up even if the Pi would be kind to correct it. Hope your PR gets through.

How does macOS handle the file extension?

ihatetoregister: Just like Windows it seems, I modified the file with the built in Texteditor and it append .txt, and having 'show all extensions' off in the OS hides this.

andrum99: > Just like Windows it seems, I modified the file with the built in Texteditor and it append .txt, and having 'show all extensions' off in the OS hides this.

Good, so https://github.com/RPi-Distro/raspberrypi-net-mods/pull/4 would fix that case as well.

burtyb:
I'm not sure what calling it "/boot" over "boot" achieves other than adding to the confusion as it won't be "/boot" on the system you're doing the copy on.

andrum99:
> I'm not sure what calling it "/boot" over "boot" achieves other than adding to the confusion as it won't be "/boot" on the system you're doing the copy on.

Good point. It should be called just `boot` on this page.

andrum99:
Strange that this page doesn't mention the ssh{,.txt} file. I suspect we have anothe page somewhere that does. I shall have a nosey.

lurch:
> Strange that this page doesn't mention the ssh{,.txt} file. I suspect we have another page somewhere that does.

https://github.com/raspberrypi/documentation/blob/master/configuration/boot_folder.md and https://github.com/raspberrypi/documentation/blob/master/remote-access/ssh/README.md

configuration/wireless/headless.md
andrum99: While researching dos2unix, I discovered that OSX uses Unix line endings, so that's not a problem for Mac users. (I'm assuming nobody is still using an older version of macOS). Which is handy since, while dos2unix can convert mac to unix, it can't auto-detect whether the file is DOS or Mac. I'm going to add to https://github.com/RPi-Distro/raspberrypi-net-mods/pull/4 so it adds line break conversion usign dos2unix.

ihatetoregister: Yes, that is more of a Windows issue.

PR #1679, opened 2020-09-18
Update external-storage.md

Adephx:
Confusing for newbies, who are unfamilirar with Linux differentiating between upper and lower caps.

### DIFF
--- a/configuration/external-storage.md
+++ b/configuration/external-storage.md
@@ -70,7 +70,7 @@ You can modify the `fstab` file to define the location where the storage device 
     ```
     UUID=5C24-1453 /mnt/mydisk fstype defaults,auto,users,rw,nofail 0 0
     ```
-   Replace "fstype" with the type of your file system, which you found in step 2 of 'Mounting a storage device' above, for example: "ntfs".
+   Replace `fstype` with the type of your file system, which you found in step 2 of 'Mounting a storage device' above, for example: `ntfs`.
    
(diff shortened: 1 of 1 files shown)


configuration/external-storage.md
```diff
@@ -68,9 +68,9 @@ You can modify the `fstab` file to define the location where the storage device
 4. Add the following line in the `fstab` file:
 
     ```
-    UUID=5C24-1453 /mnt/mydisk FSTYPE defaults,auto,users,rw,nofail 0 0
+    UUID=5C24-1453 /mnt/mydisk fstype defaults,auto,users,rw,nofail 0 0
     ```
-   Replace FSTYPE with the type of your file system, which you found in step 2 of 'Mounting a storage device' above.
+   Replace "fstype" with the type of your file system, which you found in step 2 of 'Mounting a storage device' above, for example: "ntfs".
```
lurch, line 73: `ntfs` here should probably be in backticks, rather than double-quotes? Otherwise said newbies might try to use
```
UUID=5C24-1453 /mnt/mydisk "ntfs" defaults,auto,users,rw,nofail 0 0
```
?

Adephx: I agree. Updated.

andrum99:
I think that's 2 good changes there - you've removed the bit in caps, and given an example that makes clear that NTFS needs to be in lower case 👍

JamesH65:
Thanks!

PR #1248, opened 2019-08-16
Added details for Windows 10's «OpenSSH Client»

DandelionSprout:
Now that Windows 10 has finally got a first-party SSH client, I figured out that it'd be helpful to add details about it to the SSH guides.

It's for the most part the same as the SSH guide for Unix, but with some extra info on how to turn on W10's SSH client and without the X window forwarding section.

PR #1691, opened 2020-10-02
Update windows10.md

nikiljos:
About using SSH with WSL


lurch:
I don't have Windows10 myself, but is this different to #1248 ?

nikiljos:
> I don't have Windows10 myself, but is this different to #1248 ?

In #1248 it is mentioned about windows command line, But my edit can be useful to those who already use ubuntu terminal in Windows 10. They don't have to follow the whole steps.

nikiljos:
Shall I make a commit changing it as @JamesH65 said?

ret394:
What exactly is @JamesH65  suggestion here. Revert or append to file?

Current one is a bit better and less confusing than the file in #1248, someone new or young should look at that.

thanks.

andrum99:
I don't think this is a good addition to the documentation. Is it not obvious that if you are running Linux on Windows using WSL, that you can use the Linux command `ssh` within WSL?

PR #1693, opened 2020-10-03
Update FAQ

kompowiec:
EDLC supercapacitors cooperating with a specialized power controller are ideally suited as an energy store for maintaining the power supply. Or use Powerbank.

I'm not sure if it makes sense to include this type of information in this guide, so I did not include it in the end.

### DIFF
--- a/faqs/README.md
+++ b/faqs/README.md
@@ -564,6 +564,8 @@ Most Raspberry Pi models have circuity to detect drops of the incoming power sup
 
 SD cards have a limited lifespan due to the way they work. Under most circumstances, they offer some years of use, but heavy file accessing, or using it as a swap drive, may reduce the SD card's lifespan considerably. Note that there are also fake capacity SD cards being sold that are likely to be unreliable.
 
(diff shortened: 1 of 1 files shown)


lurch:
Is this already covered by https://github.com/raspberrypi/documentation/blob/master/faqs/README.md#where-is-the-onoff-switch ?

BTW @JamesH65 / @spl237 I guess the "To switch off, if you are in the graphical environment, you can either log out from the main menu, exit to the Bash prompt, or open the terminal." sentence probably needs updating!

JamesH65:
@lurch Have fixed the FAQ "To switch off" section to current scheme. 

As for this PR, I am unsure. It is clearly information that may be useful, but is it enough information? Seems a bit like its teases  the reader with something but never fully explains.

kompowiec:
tbh both pages don't explain why pi there is no protection against card damage. If you think it's worth it, I can give you some solutions, but most of the pages in the documentation do not go in this direction, so I refrained from doing so.

aallan:

aallan:
We're deleting the FAQ, see #1913, with the contents distributed throughout the documentation. So closing.

PR #1695, opened 2020-10-09
Update uart.md

bsimmo:
updating DT overlay links to now default branch rpi-5.4.y  from rpi-4.9.
(unless anyone knows how to link to the default so it doesn't need to be changed each time?)

### DIFF
--- a/configuration/uart.md
+++ b/configuration/uart.md
@@ -113,7 +113,7 @@ N.B. Selecting the wrong early console can prevent the Pi from booting.
 
 ## UARTs and Device Tree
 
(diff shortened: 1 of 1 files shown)


lurch:
ping @pelwell 
From a quick `git grep`, looks like `configuration/cmdline-txt.md` also has an `rpi-4.19.y` link - all the other files (with links to https://github.com/raspberrypi/linux ) in this repo are already linking to a `rpi-5.4.y` URL.

pelwell:
Did you really need to involve me in this? Merge this and fix the cmdline link.

JamesH65:
Thanks!

lurch:
Sorry Phil, I was pinging you in case you "knows how to link to the default so it doesn't need to be changed each time".

pelwell:
The only options I can think of are to specifically create a branch to track the current default (e.g. `current`), or to use `HEAD` as the branch in URLs - what do you think, @popcornmix?

PR #1703, opened 2020-10-22
updated the command sequence and location of the service file

ronsmits:
The systemctl structure has been changed since 2018. I updated it to the current way of working.

### DIFF
--- a/linux/usage/systemd.md
+++ b/linux/usage/systemd.md
@@ -26,9 +26,9 @@ WantedBy=multi-user.target
 ```
 So in this instance, the service would run Python 3 from our working directory `/home/pi/myscript` which contains our python program to run `main.py`. But you are not limited to Python programs: simply change the ExecStart line to be the command to start any program/script that you want running from booting.
 
-Copy this file into `/var/run/systemd/system` as root, for example:
+Copy this file into `/etc/systemd/system` as root, for example:
 ```
-sudo cp myscript.service /var/run/systemd/system/myscript.service
+sudo cp myscript.service /etc/systemd/system/myscript.service
 ```
 
 Once this has been copied, you have to inform `systemd` that a new service has been added. This is done with the following command:
--- a/linux/usage/systemd.md
+++ b/linux/usage/systemd.md
@@ -26,12 +26,18 @@ WantedBy=multi-user.target
 ```
 So in this instance, the service would run Python 3 from our working directory `/home/pi/myscript` which contains our python program to run `main.py`. But you are not limited to Python programs: simply change the ExecStart line to be the command to start any program/script that you want running from booting.
 
-Copy this file into `/etc/systemd/system` as root, for example:
+Copy this file into `/var/run/systemd/system` as root, for example:
 ```
-sudo cp myscript.service /etc/systemd/system/myscript.service
+sudo cp myscript.service /var/run/systemd/system/myscript.service
 ```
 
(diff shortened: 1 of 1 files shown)


lurch:
ping @XECDesign

XECDesign:
Is there some upstream documentation explaining why this path would be preferred? And if so, since which version of systemd?

As far as I can tell /var/run/ is a symlink to /run, which is a tmpfs mount. So anything you put there will disappear after reboot.

That seems to match Redhat's documentation (emphasis mine):

> /run/systemd/system/ | Systemd unit files created **_at run time_**. This directory takes  precedence over the directory with installed service unit files.

The same information is mirrored in upstream documentation:/man/systemd.unit.html

> /etc/systemd/system | System units created by the administrator
> /run/systemd/system | Runtime units

https://www.freedesktop.org/software/systemd/man/systemd.unit.html

linux/usage/systemd.md
```diff
@@ -26,12 +26,18 @@ WantedBy=multi-user.target
 ```
 So in this instance, the service would run Python 3 from our working directory `/home/pi/myscript` which contains our python program to run `main.py`. But you are not limited to Python programs: simply change the ExecStart line to be the command to start any program/script that you want running from booting.
 
-Copy this file into `/etc/systemd/system` as root, for example:
+Copy this file into `/var/run/systemd/system` as root, for example:
```
andrum99, line 29: The path was fine as it was. You should not be telling users to put configuration files under `/var/run`, since they will disappear on reboot.

(To be 100% correct the path should probably be `/etc/systemd/user` rather than `system`).

linux/usage/systemd.md
```diff
@@ -26,12 +26,18 @@ WantedBy=multi-user.target
...
 ```
-sudo cp myscript.service /etc/systemd/system/myscript.service
+sudo cp myscript.service /var/run/systemd/system/myscript.service
```
andrum99, line 31: Again, this was fine as it was - the proposed change would be wrong.

linux/usage/systemd.md
```diff
@@ -26,12 +26,18 @@ WantedBy=multi-user.target
...
 ```
 
-Once this has been copied, you can attempt to start the service using the following command:
+Once this has been copied, you have to inform `systemd` that a new service has been added. This is done with the following command:
+
+```
+sudo systemctl daemon-reload
+```
+
+Now you can attempt to start the service using the following command:
```
andrum99, line 40: This is a useful addition, and a step that is easy to forget 👍

linux/usage/systemd.md
ronsmits: http://manpages.ubuntu.com/manpages/xenial/en/man5/systemd.unit.5.html /etc/systemd or /run/systemd which is a symlink to /run.

andrum99: > http://manpages.ubuntu.com/manpages/xenial/en/man5/systemd.unit.5.html /etc/systemd or /run/systemd which is a symlink to /run.

That documentation is for Ubuntu 16: on Raspberry Pi OS `/etc/systemd` and `/var/run/systemd` are two different things - they're not symlinked.

ronsmits: then I will remove the change

aallan:

See #1911 for full details.

aallan:
Anyone against merging?

PR #1713, opened 2020-10-29
Update README.md

7h0mas-R:
There is no overlay spi0-1cs or spi0-2cs (anymore?). On all systems I checked, dtoverlay -a only lists spi0-cs and spi0-hw-cs.

For spi1 and higher the spi<n>-1cs and similar are listed


7h0mas-R:
My fault- just found out, that the systems I am dealing with have been setup with a non-standard raspbian. sorry...

PR #1718, opened 2020-11-02
FAQ: minor corrections, rewrite overclocking answer

andrum99:
Some minor corrections. I've removed the Pi 400 kit from the price list to avoid confusion with the standalone Pi 400 - none of the other kits are listed in this price list.

Also, rewrite overclocking section so it contains the information I believe it should.


JamesH65:
We've been discussing the addition of the Pi400 to the list. It's a new venture in some respects, since we have not prompted kits before, so right now we are undecided as to whether to have it priced in there or not. For the moment I'd like it to stay until a decision is made.

JamesH65:
I think we probably need to add a section anyway to explain the Pi400 kit. Will do that later today.

JamesH65:
#1722

PR #1719, opened 2020-11-03
Update cm-emmc-flashing.md

timg236:
Minor updates for CM4


lurch:
This page still recommends `Win32DiskImager` and `dd` - probably makes sense to recommend Raspberry Pi Imager instead?

PR #1723, opened 2020-11-06
Add note about changing the hostapd hw_mode

lurch:
If I left `channel` set to `7` when changing the `hw_mode` to `a`, hostapd wouldn't start and reported an error about an invalid channel selection. (it later worked when I changed the channel to 36)

However I dunno much about wireless networks, and even less about access points, so pinging @epoch1970 for confirmation.

### DIFF
--- a/configuration/wireless/access-point-bridged.md
+++ b/configuration/wireless/access-point-bridged.md
@@ -7,24 +7,24 @@ If you wish to create a standalone wireless network, consider instead setting up
 
 ```
                                          +- RPi -------+
-                                     +---+ Bridge      |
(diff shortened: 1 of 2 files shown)


epoch1970:
Makes sense to me. I have no 5GHz-capable Pi, so I never came across this bump on the road.
The same note would apply to the bridged AP documentation.

lurch:
Thanks @epoch1970 - I've just made small tweaks to both pages to make them more consistent with each other.

JamesH65:
Is this OK to merge?

lurch:
LGTM :wink:

PR #1734, opened 2020-11-15
Add documentation to update to Buster

MPParsley:


lurch:
ping @XECDesign

XECDesign:
LGTM.

FWIW, it's going to drag in unnecessary dependencies, leave behind obsolete packages and it's possible some things may not work. That can be dealt with if the user knows a fair bit about Debian, but if somebody is following those instructions, they might not and it's setting them up for some frustration and wasted time. So, I'm a bit unsure about the value of having those partial instructions there, but that's just my 2p.

lurch:
So maybe change "Upgrading an existing Stretch image is possible, but is not guaranteed to work in every circumstance." to "Upgrading an existing Stretch image is possible, but is not recommended as it may not work or may leave your system in an inconsistent state." ?

Or would it be better to just get rid of this "Upgrading from XXX to YYY" section entirely? Probably a decision for @JamesH65 to make :wink:

JamesH65:
Hmm, we really don't recommend upgrades, but there again, we don't try and stop people doing stuff if they want. I'd keep it in, but with improved warnings it may not work.

raspbian/updating.md
```diff
@@ -40,11 +40,11 @@ The kernel and firmware are installed as a Debian package, and so will also get
 
 When running `sudo apt full-upgrade`, it will show how much data will be downloaded and how much space it will take up on the SD card. It's worth checking with `df -h` that you have enough free disk space, as unfortunately `apt` will not do this for you. Also be aware that downloaded package files (`.deb` files) are kept in `/var/cache/apt/archives`. You can remove these in order to free up space with `sudo apt clean` (`sudo apt-get clean` in older releases of apt).
 
-### Upgrading from Jessie to Stretch
+### Upgrading from Stretch to Buster
 
-Upgrading an existing Jessie image is possible, but is not guaranteed to work in every circumstance. If you wish to try upgrading a Jessie image to Stretch, we strongly recommend making a backup first — we can accept no responsibility for loss of data from a failed update.
+Upgrading an existing Stretch image is possible, but is not guaranteed to work in every circumstance. If you wish to try upgrading a Stretch image to Buster, we strongly recommend making a backup first — we can accept no responsibility for loss of data from a failed update.
```
MPParsley, line 45:
```suggestion
> :warning: **Warning**: Upgrading an existing Stretch image is possible, but is not guaranteed to work in every circumstance. If you wish to try upgrading a Stretch image to Buster, we strongly recommend making a backup first — we can accept no responsibility for loss of data from a failed update.
```

MPParsley:
Added a clearer indication that the upgrade is at one's own risk.
image

lurch:
I don't think we've used "warning" icons like that on any other pages? It _might_ not work when the documentation gets auto-copied to https://www.raspberrypi.org/documentation/

Based on what @XECDesign and @JamesH65 have said, I think it's worth adding a "is not recommended" caveat in there.

Issue #1098, opened 2019-03-11
mount.cifs gives erroneous "host is down" 


roykimbrell:
"mount.cifs //IP/share /mnt/point -o user=uname" results in an erroneous "host is down" error notice. The solution, as noted at several locations (https://askubuntu.com/questions/915549/16-04-cifs-host-is-down-but-they-are-not, https://www.raspberrypi.org/forums/viewtopic.php?t=210311, and https://access.redhat.com/discussions/3002961), is to add the option vers=1.0. As in

mount.cifs //IP/share /mnt/point -o user=uname,vers=1.0

lurch:
One of the comments on the askubuntu page you linked to https://askubuntu.com/a/1070656 suggests that the SMB version used depends upon which server-host you're connecting to, so presumably adding `vers=1.0` isn't a universal solution?
But it definitely sounds like the sort of thing that would be worth adding to https://github.com/raspberrypi/documentation/blob/master/remote-access/samba.md

roykimbrell:
Yeah. That's the place.

andrum99:
Reading https://askubuntu.com/questions/915549/16-04-cifs-host-is-down-but-they-are-not reveals that this error results from 2 things: SMB protocol version mismatch, and the cifs client on Linux returning a misleading error message. The solution is not as simple as adding vers=1.0, since it depends which version of the SMB protocol the server supports. The need to add vers=x.0 to the mount should certainly be documented, with an explanation that the user may need to try different values for x.

JamesH65:
Fixed by #1739

PR #1739, opened 2020-11-19
Update samba.md

JamesH65:
Fix for #1098

### DIFF
--- a/remote-access/samba.md
+++ b/remote-access/samba.md
@@ -73,6 +73,26 @@ cd windowshare
 ls
 ```
 
+### "Host is down" error
+
+This error is caused by a combination of two things: A SMB protocol version mismatch, and the CIFS client on Linux returning a misleading error message. In order to fix this a version entry needs to be added to the mount command. By default Raspberry Pi OS will only use versions 2.1 and above, which are compatible with Windows 7 and later. Older devices, including some NAS, may require version 1.0:
+
+```
+sudo mount.cifs //IP/share /mnt/point -o user=<uname>,vers=1.0
+```
+
+You may need to try different versions to match up with the server version. Possible values are: 
+
+| Version | Description |
+|---------|-------------|
+|  1.0    | Classic CIFS/SMBv1 protocol |
+|  2.0    | The SMBv2.002 protocol. Windows Vista Service Pack 1, and Windows Server 2008 |
+|  2.1    | The SMBv2.1 protocol. Microsoft Windows 7 and Windows Server 2008R2 |
+|  3.0    | The SMBv3.0 protocol.  Microsoft Windows 8 and Windows Server 2012 |
+| 3.02    | The SMBv3.0.2 protocol. Microsoft Windows 8.1 and Windows Server 2012R2 |
+| 3.11    | The SMBv3.1.1 protocol. Microsoft Windows 10 and Windows Server 2016 |
+| 3       | The SMBv3.0 protocol version and above |
+
 ### Sharing a folder for use by Windows
 
 Firstly, create a folder to share. This example creates a folder called `shared` in the `home` folder of the current user, and  assumes the current user is `pi`.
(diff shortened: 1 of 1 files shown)


remote-access/samba.md
```diff
@@ -105,3 +105,25 @@ workgroup = <your workgroup name here>
...
+You may need to try different versions to match up with the server version. Possible values are: 
+
+| Version | Description |
+|---------|-------------|
+|  1.0    | Classic CIFS/SMBv1 protocol |
+|  2.0    | The SMBv2.002 protocol. Windows Vista Service Pack 1, and Windows Server 2008 |
+|  2.1    | The SMBv2.1 protocol. Microsoft Windows 7 and Windows Server 2008R2 |
+|  3.0    | The SMBv3.0 protocol.  Microsoft Windows 8 and Windows Server 2012 |
+| 3.02/3.0.2 | The SMBv3.0.2 protocol. Microsoft Windows 8.1 and Windows Server 2012R2 |
+| 3.1.1/3.11 | The SMBv3.1.1 protocol. Microsoft Windows 10 and Windows Server 2016 |
```
lurch, line 128: Might be worth editing one of these two lines so that the `x.yz` and `x.y.z` versions are listed in the same order?

remote-access/samba.md
```diff
@@ -105,3 +105,25 @@ workgroup = <your workgroup name here>
 ```
 
 That should be enough to share the folder. On your Windows device, when you browse the network, the folder should appear and you should be able to connect to it.
+
+### "Host is down" error
```
lurch, line 109: It probably makes sense to move this section to below the "Mount the folder on the Raspberry Pi" section rather than below the "Sharing a folder for use by Windows" section?

remote-access/samba.md
```diff
@@ -105,3 +105,25 @@ workgroup = <your workgroup name here>
...
+
+This error is caused by a combination of two things: A SMB protocol version mismatch, and the CIFS client on Linux returning a misleading error message. 
+
+In order to fix this a version entry needs to be added to the mount command. Note that by default the latest 'mount.cifs' on Raspberry Pi OS will attempt to negotiate the highest possible version greater or equal to 2.1 which should work with Windows 7 and above shares, but if connecting to older NAS devices for example, you may need to use an earlier version.
+
+```
+mount.cifs //IP/share /mnt/point -o user=<uname>,vers=1.0
```
lurch, line 116: Looking at the command used in the "Mount the folder on the Raspberry Pi" section, I guess this needs a `sudo`?

remote-access/samba.md
```diff
@@ -105,3 +105,25 @@ workgroup = <your workgroup name here>
 ```
 
 That should be enough to share the folder. On your Windows device, when you browse the network, the folder should appear and you should be able to connect to it.
+
+### "Host is down" error
+
+This error is caused by a combination of two things: A SMB protocol version mismatch, and the CIFS client on Linux returning a misleading error message. 
+
+In order to fix this a version entry needs to be added to the mount command. Note that by default the latest 'mount.cifs' on Raspberry Pi OS will attempt to negotiate the highest possible version greater or equal to 2.1 which should work with Windows 7 and above shares, but if connecting to older NAS devices for example, you may need to use an earlier version.
```
andrum99, line 113: These two paragraphs should be combined. Also, the last sentence is quite long and complex - what about this:

`By default Raspberry Pi OS will only use versions 2.1 or above, which is compatible with Windows 7 and later. Older devices, including some NAS, may require version 1.0:

remote-access/samba.md
andrum99: > Might be worth editing one of these two lines so that the `x.yz` and `x.y.z` versions are listed in the same order?

I would remove the variants with the two dots in them to keep things simple. While you can specify them either way, knowing this doesn't really add anything useful to the page.

aallan:
@JamesH65 Is this good to merge? If so we need to do it within the next couple of weeks, see #1911.

PR #1740, opened 2020-11-20
Update headless.md

Willmish:
The reason I believe this 2 sentences should be added is because I have seen to many open threads of people trying to set their Pis headless and eventually giving up. I myself did not manage to set up my first Pi headless, only when fiddling with it for a long while did I figure out this (trivial?) part. If you don't like the phrasing, please feel free to rephrase it, but I highly recommend you add it to the guide.


JamesH65:
@lurch This LGTM, it doesn't really have any downsides. Might rephrase a bit perhaps but seem OK on the whole to you?

PR #1743, opened 2020-11-23
Update passwordless.md

lprc:
Remark on setting correct directory permissions if connecting via ssh fails

### DIFF
--- a/remote-access/ssh/passwordless.md
+++ b/remote-access/ssh/passwordless.md
@@ -94,12 +94,12 @@ tail -f /var/log/secure
 Nov 23 12:31:26 raspberrypi sshd[9146]: Authentication refused: bad ownership or modes for directory /home/pi
 ```
 
-If the log says `Authentication refused: bad ownership or modes for directory /home/pi` there is a permission problem regarding your home directory. SSH needs your home and `~/.ssh` directory to not have group write access. You can adjust the permissions using `chmod` (make sure to replace `user` with your username, e.g. `pi`):
+If the log says `Authentication refused: bad ownership or modes for directory /home/pi` there is a permission problem regarding your home directory. SSH needs your home and `~/.ssh` directory to not have group write access. You can adjust the permissions using `chmod`:
 
 ```bash
-chmod g-w /home/user
-chmod 700 /home/user/.ssh
-chmod 600 /home/user/.ssh/authorized_keys
+chmod g-w $HOME
+chmod 700 $HOME/.ssh
+chmod 600 $HOME/.ssh/authorized_keys
 ```
 Now only the user itself has access to `.ssh` and `.ssh/authorized_keys` in which the public keys of your remote machines are stored.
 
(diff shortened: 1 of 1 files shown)


remote-access/ssh/passwordless.md
```diff
@@ -84,6 +84,25 @@ If this does not work, you can get assistance on the [Raspberry Pi forums](https
...
+If you can't establish a connection after following the steps above there might be a problem with your directory permissions. First, you want to check the logs for any errors:
+
+```bash
+tail -f /var/log/secure
+# might return:
+Nov 23 12:31:26 raspberrypi sshd[9146]: Authentication refused: bad ownership or modes for directory /home/pi
+```
+
+If the logs says `Authentication refused: bad ownership or modes for directory /home/pi` there is a permission problem regarding your home directory. SSH needs your home and ~/.ssh directory forbid group write access. You can adjust the permissions using `chmod` (make sure to replace `user` with your username, e.g. `pi`):
+
```
JamesH65, line 98: This sentence doesn't read right - where does forbid come in to it?

lprc: I meant that those directories are not allowed to have write permission for groups. Sorry, I'm not a native english speaker and didn't know how to express it any better. Please feel free to correct the sentence and I'll add another commit (if that's the way PRs work).

ajs256: Here's what I would say:
```md
If the log says `Authentication refused: bad ownership or modes for directory /home/pi` there is a permission problem regarding your home directory. SSH needs your home and `~/.ssh` directory to not have group write access. You can adjust the permissions using `chmod` (make sure to replace `user` with your username, e.g. `pi`):
```

remote-access/ssh/passwordless.md
```diff
@@ -84,6 +84,25 @@ If this does not work, you can get assistance on the [Raspberry Pi forums](https
...
+```bash
+tail -f /var/log/secure
+# might return:
+Nov 23 12:31:26 raspberrypi sshd[9146]: Authentication refused: bad ownership or modes for directory /home/pi
+```
+
+If the log says `Authentication refused: bad ownership or modes for directory /home/pi` there is a permission problem regarding your home directory. SSH needs your home and `~/.ssh` directory to not have group write access. You can adjust the permissions using `chmod` (make sure to replace `user` with your username, e.g. `pi`):
+
+```bash
+chmod g-w /home/user
```
lurch, line 100: I don't think this first `chmod` is necessary?

remote-access/ssh/passwordless.md
```diff
@@ -84,6 +84,25 @@ If this does not work, you can get assistance on the [Raspberry Pi forums](https
...
+chmod 700 /home/user/.ssh
```
lurch, line 101: Given that the documentation here specifically targets Raspberry Pi OS, perhaps it makes more sense to use `/home/pi/.ssh` etc. in the instructions rather than `/home/user/.ssh` ?
Alternatively, I think you could just use `$HOME/.ssh` ? :shrug:

lprc: Sure, using `$HOME` sounds good. I like that there's no loss in generality as opposed to using `/home/pi/` directly.

lurch:
LGTM :+1:

PR #1749, opened 2020-12-02
Make syntax match other blocks in installation instructions

ajs256:
I added the `bash` highlight keyword to the code block under Eject to make it match the other shell blocks.


lurch:
Thanks!

PR #1755, opened 2020-12-08
Update all dt-blobs for adding cameras / displays to CM 1/3/4

6by9:
Currently the blobs. I'm still writing the text.
If someone else wants to do CM4IO photos then that would be useful as I've got a pre-production CM4IO (some connectors moved).

I'm in the process of testing out the blobs as they've all been rebuilt from scratch. Quickest/easiest to compare against the dt-blob.dts in firmware/extras/dt-blob.dts rather than against the older versions here.


6by9:
Blobs all verified on CM3 and CM4. I'll trust my copy/paste skills for CM1.

@JamesH65 over to you. I'll sort out the docs in the meantime.

6by9:
@JamesH65 Can we merge these blob changes even without the docs update? It at least means that people can get the displays/camera working if they have a little common sense?

JamesH65:
No problem, I'd forgotten these blobs were in the docs repo. Doesn't sound like the best place for them in the long term TBH, but OK for the moment.

PR #1765, opened 2020-12-20
Update libcamera README.md

kamomil:
1. Add note that it is impossible to use dtoverlay command to load the sensor overlay
2. Add the needed pip3 installations

### DIFF
--- a/linux/software/libcamera/README.md
+++ b/linux/software/libcamera/README.md
@@ -24,7 +24,7 @@ dtoverlay=imx219
 core_freq_min=250
 ```
 
-If you are using a sensor other than the `imx219` you will need to supply the alternative name here (for example, `ov5647` for the V1 camera, or `imx477` for the HQ Cam for which support will be available shortly).
+If you are using a sensor other than the `imx219` you will need to supply the alternative name here (for example, `ov5647` for the V1 camera, or `imx477` for the HQ Cam for which support will be available shortly). It is impossible to load the sensor overlay dynamically with `dtoverlay` command as the firmware updates any override nodes that are named ["cam0-pwdn" or "cam0-pwdn-ctrl"](https://github.com/raspberrypi/linux/blob/rpi-5.10.y/arch/arm/boot/dts/overlays/imx219-overlay.dts#L111) (and cam1...) to point to the correct GPIO for that particular platform.
 
 **NOTE**: after rebooting, control of the camera system will be passed to the ARM cores, and firmware-based camera functions (such as raspistill and so forth) will no longer work. Setting `/boot/config.txt` back and rebooting will restore the previous behaviour.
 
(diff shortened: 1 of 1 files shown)


JamesH65:
@6by9 @naushir I have no idea whether this is appropriate or not, any comments?

linux/software/libcamera/README.md
```diff
@@ -24,7 +24,7 @@ dtoverlay=imx219
 core_freq_min=250
 ```
 
-If you are using a sensor other than the `imx219` you will need to supply the alternative name here (for example, `ov5647` for the V1 camera, or `imx477` for the HQ Cam for which support will be available shortly).
+If you are using a sensor other than the `imx219` you will need to supply the alternative name here (for example, `ov5647` for the V1 camera, or `imx477` for the HQ Cam for which support will be available shortly). It is impossible to load the sensor overlay dynamically with `dtoverlay` command as the firmware updates any override nodes that are named ["cam0-pwdn" or "cam0-pwdn-ctrl"](https://github.com/raspberrypi/linux/blob/rpi-5.10.y/arch/arm/boot/dts/overlays/imx219-overlay.dts#L111) (and cam1...) to point to the correct GPIO for that particular platform.
```
6by9, line 27: Details of what is fixed up is not required.

It is not possible to load the sensor overlays dynamically with the `dtoverlay` command as the firmware fixes up the overlays to point to the shutdown GPIO to the correct GPIO for the specific platform.

I'm just wondering if we're better off creating the regulator definitions in the base DT files with the correct GPIOs configured, and then all overlays can just reference those rather than needing to have things fixed up. Doing that would allow loading with `dtoverlay`. The original fixup was written when the GPIO was configured as a shutdown/reset GPIO of the sensor driver, rather than the control GPIO of the regulator.

linux/software/libcamera/README.md
```diff
@@ -37,7 +37,7 @@ sudo apt install libboost-dev
 sudo apt install libgnutls28-dev openssl libtiff5-dev
 sudo apt install meson
 sudo apt install qtbase5-dev libqt5core5a libqt5gui5 libqt5widgets5
-sudo pip3 install pyyaml
+sudo pip3 install pyyaml ninja ply
 ```
```
naushir, line 41: Can confirm this is now required.

linux/software/libcamera/README.md
6by9: Hopefully this should be fixed for imx219 and imx477 by https://github.com/raspberrypi/linux/pull/4051

ov5647 doesn't use the regulator framework, so that still needs the firmware fixups. A driver fix wouldn't go amiss there.

aallan:

See #1911 for full details.

aallan:
@kamomil  Can you resolve the merge conflicts so this can be merged ahead of #1911. Thanks!

aallan:

PR #1766, opened 2020-12-27
Change jail.conf to jail.local

kevlg:
It is done to avoid merging problems when upgrading fail2ban


lurch:
Merged, as the line immediately above says `sudo nano /etc/fail2ban/jail.local` so the mention of `jail.conf` was clearly an error.
Thanks!

PR #1768, opened 2021-01-01
Updated Fat32 Format link as domain has changed

Uncraftablecreeper:


JamesH65:
Good spot and thanks for the fix.

PR #1769, opened 2021-01-01
Installation: deprecate NOOBS, tweak page links

andrum99:
Just noticed this page is still recommending NOOBS, so this change aims to update it to point folk to Raspberry Pi Imager instead. I've also made some attempt to make use of the new Foundation-provided content, which somewhat supercedes what is in the installation section of the docs.

The new Foundation-provided page at https://www.raspberrypi.org/software/ is simpler than anything in this repo, so I've linked straight to that for Raspberry Pi Imager. It also has a handy video, which should explain the (fairly obvious) process in enough detail for most users.

I've tweaked the links to the 3 sub-pages as follows:

1. `NOOBS` - still linked, but added a note that it is now deprecated.
2. `Installing Images` - I've added a bit of blurb pointing folk to that page if they need some extra help, as in most cases raspberrypi.org/software/ should be all folk need. I've also pointed folk to the forums for extra help as well.
3. `SD cards` - I've labelled this "further info about SD cards", since in most cases the info on this page should not really be needed, but it's there if folk require it.

### DIFF
--- a/installation/README.md
+++ b/installation/README.md
@@ -4,7 +4,6 @@ We recommend using [Raspberry Pi Imager](https://www.raspberrypi.org/software/) 
 
 * You can find detailed help with installing OS images [here](installing-images/). For further assistance, check out our [forums](https://www.raspberrypi.org/forums/).
 
-* [Further information about SD cards](sd-cards.md)
+* [Further information about SD cards](sd-cards.md).
 
-
-**Note:** [NOOBS](noobs.md) is still available, but is no longer recommended for new installations.
+* [NOOBS](noobs.md) is still available, but is no longer recommended for new installations.
--- a/installation/README.md
+++ b/installation/README.md
@@ -1,6 +1,6 @@
 # Installation
 
-We recommend using [Raspberry Pi Imager](https://www.raspberrypi.org/software/) to install an operating system on your Raspberry Pi. As well as Raspberry Pi OS, you can use it to install other operating systems such as Ubuntu, and LibreELEC (Kodi media centre).
+We recommend using [Raspberry Pi Imager](https://www.raspberrypi.org/software/) to install an operating system on your Raspberry Pi. As well as Raspberry Pi OS, you can use it to install other operating systems such as Ubuntu or LibreELEC (Kodi media centre).
 
 * You can find detailed help with installing OS images [here](installing-images/). For further assistance, check out our [forums](https://www.raspberrypi.org/forums/).
 
--- a/installation/README.md
+++ b/installation/README.md
@@ -6,4 +6,5 @@ We recommend using [Raspberry Pi Imager](https://www.raspberrypi.org/software/) 
 
 * [Further information about SD cards](sd-cards.md)
 
-[NOOBS](noobs.md) is still available, but is no longer recommended for new installations.
+
+*Note:* [NOOBS](noobs.md) is still available, but is no longer recommended for new installations.
--- a/installation/README.md
+++ b/installation/README.md
@@ -7,4 +7,4 @@ We recommend using [Raspberry Pi Imager](https://www.raspberrypi.org/software/) 
 * [Further information about SD cards](sd-cards.md)
 
 
-*Note:* [NOOBS](noobs.md) is still available, but is no longer recommended for new installations.
+**Note:** [NOOBS](noobs.md) is still available, but is no longer recommended for new installations.
--- a/installation/README.md
+++ b/installation/README.md
@@ -6,4 +6,4 @@ We recommend using [Raspberry Pi Imager](https://www.raspberrypi.org/software/) 
 
 * [Further information about SD cards](sd-cards.md).
 
-* [NOOBS](noobs.md) is still available, but is no longer recommended for new installations.
+* [NOOBS](noobs.md)
--- a/installation/README.md
+++ b/installation/README.md
@@ -1,11 +1,9 @@
 # Installation
 
-Some simple guides to setting up the software on your Raspberry Pi.
+We recommend using [Raspberry Pi Imager](https://www.raspberrypi.org/software/) to install an operating system on your Raspberry Pi. As well as Raspberry Pi OS, you can use it to install other operating systems such as Ubuntu, and LibreELEC (Kodi media centre).
 
-Beginners should start with [NOOBS](noobs.md), which gives the user a choice of operating system from the standard distributions. The recommended distribution for normal use is Raspberry Pi OS. Alternatives are available, such as LibreELEC (Kodi media centre) or Arch Linux.
+* You can find detailed help with installing OS images [here](installing-images/). For further assistance, check out our [forums](https://www.raspberrypi.org/forums/).
 
-## Contents
+* [Further information about SD cards](sdcards.md)
 
-- [NOOBS](noobs.md)
-- [Installing images](installing-images/README.md)
-- [SD Cards](sd-cards.md)
+[NOOBS](noobs.md) is still available, but is no longer recommended for new installations.
--- a/installation/README.md
+++ b/installation/README.md
@@ -4,6 +4,6 @@ We recommend using [Raspberry Pi Imager](https://www.raspberrypi.org/software/) 
 
 * You can find detailed help with installing OS images [here](installing-images/). For further assistance, check out our [forums](https://www.raspberrypi.org/forums/).
 
-* [Further information about SD cards](sdcards.md)
+* [Further information about SD cards](sd-cards.md)
 
 [NOOBS](noobs.md) is still available, but is no longer recommended for new installations.
--- a/installation/noobs.md
+++ b/installation/noobs.md
@@ -1,6 +1,6 @@
 # NOOBS
 
-**Note:** We now recommend that you use [Raspberry Pi Imager](https://www.raspberrypi.org/software/) to install an operating system, instead of using NOOBS. If you are buying an SD card with software pre-installed on it, then we do still recommend buying one with NOOBS on it.
+**Note: We now recommend that you use [Raspberry Pi Imager](https://www.raspberrypi.org/software/) to install an operating system, instead of using NOOBS. If you are buying an SD card with software pre-installed on it, then we do still recommend buying one with NOOBS on it.**
 
 **New Out Of Box Software (NOOBS)** is an operating system installation manager for the Raspberry Pi.
 
--- a/installation/noobs.md
+++ b/installation/noobs.md
@@ -14,7 +14,7 @@ SD cards with NOOBS pre-installed are available from many of our distributors an
 
 ## What's included in NOOBS
 
-NOOBS includes Raspberry Pi OS. It can also be used to install other operating systems: this requires that the Pi has a working internet connection to download the OS.
+NOOBS includes Raspberry Pi OS, and LibreELEC which is a Kodi media player system. It can also be used to install other operating systems: this requires that the Pi has a working internet connection to download the OS.
 
 ## NOOBS development
 
--- a/installation/noobs.md
+++ b/installation/noobs.md
@@ -1,6 +1,8 @@
 # NOOBS
 
-**New Out Of Box Software (NOOBS)** is an easy operating system installation manager for the Raspberry Pi.
+**Note:** We now recommend that you use [Raspberry Pi Imager](https://www.raspberrypi.org/software/) to install an operating system, instead of using NOOBS. If you are buying an SD card with software pre-installed on it, then we do still recommend buying one with NOOBS on it.
+
+**New Out Of Box Software (NOOBS)** is an operating system installation manager for the Raspberry Pi.
 
 ![NOOBS OS selection](images/noobs.png)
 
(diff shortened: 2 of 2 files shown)


installation/README.md
```diff
@@ -1,11 +1,9 @@
 # Installation
 
-Some simple guides to setting up the software on your Raspberry Pi.
+We recommend using [Raspberry Pi Imager](https://www.raspberrypi.org/software/) to install an operating system on your Raspberry Pi. As well as Raspberry Pi OS, you can use it to install other operating systems such as Ubuntu, and LibreELEC (Kodi media centre).
 
-Beginners should start with [NOOBS](noobs.md), which gives the user a choice of operating system from the standard distributions. The recommended distribution for normal use is Raspberry Pi OS. Alternatives are available, such as LibreELEC (Kodi media centre) or Arch Linux.
+* You can find detailed help with installing OS images [here](installing-images/). For further assistance, check out our [forums](https://www.raspberrypi.org/forums/).
 
-## Contents
+* [Further information about SD cards](sdcards.md)
```
lurch, line 7: Typo? This page previously linked to `sd-cards.md`

andrum99: Oops - good catch. Fixed.

lurch:
ping @XECDesign in case he has any thoughts / comments.

XECDesign:
No issues with the PR from my end.

installation/README.md
```diff
@@ -1,11 +1,9 @@
 # Installation
 
-Some simple guides to setting up the software on your Raspberry Pi.
+We recommend using [Raspberry Pi Imager](https://www.raspberrypi.org/software/) to install an operating system on your Raspberry Pi. As well as Raspberry Pi OS, you can use it to install other operating systems such as Ubuntu, and LibreELEC (Kodi media centre).
```
lurch, line 3: "Ubuntu, and LibreELEC" should probably be "Ubuntu or LibreELEC" ?

lurch:
I can't decide whether I like the bullet-points or not (IMHO it looks weird having only two bullet-points, and not having NOOBS bulleted), so I'll leave the final decision to @JamesH65 :wink:

andrum99:
> I can't decide whether I like the bullet-points or not (IMHO it looks weird having only two bullet-points, and not having NOOBS bulleted), so I'll leave the final decision to @JamesH65 😉

Yeah it's not perfect, but I thought I would put something up as a starting point at least, particularly as this page is still recommending NOOBS. I see that some resellers are still selling SD cards with NOOBS on - e.g. pimoroni.com.

I also noticed a commit in the firmware repo (https://github.com/raspberrypi/firmware/commit/919aee0ed75f7db48a38b8b96c13228a7584cfd7) which says this:

> firmware: sdcard: Remove legacy NOOBS support to support booting from primary partition 4

But I don't know enough about how NOOBS works to know if that is going to cause problems if folk try to use NOOBS to run the next version of Raspberry Pi OS. Not sure what the plan is with that one.

Back to the documentation itself, and I noticed if you have a look at the raspberrypi.org/help, which is linked to from the website's new and 'improved' top menu under `Software` -> `Help`, you can see it has Foundation-authored sections entitled `Setting uo your Pi` and `Using your Pi`. Underneath that you then have the Trading-owned documentation repo top links, which start with `Setup / Quickstart` and `Installation`. I've looked at this issue of duplication of content between the Foundation and Trading stuff before and it seems to me it would perhaps be worthwhile trying to either combine them somehow, or perhaps just remove some of the lower-trafficked, more technical stuff from those two sections of the Trading-owned docs site, or somehow refer folk to the trading stuff for the more technical details, but I've never quite come up with a good way of doing it. My edits to this page are my first attempt to try to refer folk to the Foundation content as much as possible, as it is more beginner-oriented and seems to be more up-to-date, but also keep some of the technical stuff in the Trading docs repo in case folk need to refer to it. Ideally it probably needs someone inside the organisation to look at how best to rationalise the docs, as obviously having two sets of installation and setup documentation is potentially a maintenance overhead. Having said that I'm not sure it's that big a problem, and the fact it hasn't really been addressed much would tend to suggest it's not really a problem.

Apologies for rather random ramble - just my thoughts as ever 😁

installation/README.md
andrum99: Fixed.

andrum99:
OK, so after playing about with this a bit, I think lurch is right - it looks better with NOOBS as a third bullet point. Should be ready to roll now.

andrum99:
I've tweaked my edit to not mention that NOOBS is deprecated, since resellers are still selling cards with NOOBS on it. This PR should therefore be ready to roll.

JamesH65:
LGTM, any other comments?

lurch:
> I've tweaked my edit to not mention that NOOBS is deprecated, since resellers are still selling cards with NOOBS on it.

If we've deprecated NOOBS (which we seem to have done, given that there are no links to it from the Download pages?), I'm not sure that some resellers still selling NOOBS SD-cards is a valid reason to **not** say that it's deprecated in our documentation? But I suspect that that's a decision for @aallan or @ghollingworth to make.

aallan:
NOOBS is deprecated, we should therefore say that in the documentation.

andrum99:
I think someone should probably speak to a director about this, since there are clear business implications in terms of official resellers still selling cards with NOOBS on it. I would expect the correct approach is probably to send out a specific notice to all official resellers that NOOBS is being deprecated before announcing it on the website and in the docs. There doesn't need to be a blog post obviously, but it would be good to have a notice at the bottom of the main download page. Ideally the notice to official resellers would have been done already, e.g. when the website switched to recommending Raspberry Pi Imager instead of NOOBS. It may have been done already of course.

Having a notice on the download page would clearly be helpful to users as well. A day or two ago a user popped up on the forums wondering why NOOBS was nowhere to be found: https://www.raspberrypi.org/forums/viewtopic.php?f=63&t=303015.

ghollingworth:
NOOBS has now been deprecated as a suitable tool for users to install themselves onto their SD cards.  We've done this because we believe rpi-imager is a much better solution in this particular use case.

But it is still the preferred image for pre-imaged SD cards.  This is because our approved resellers require something that they can easily communicate with their customers how to get various different images to install onto the card.  In future, this function will be replaced with something else that we've been working on in the background...

So I'm happy with adding a note explaining that full NOOBS will still be pre-imaged onto official Raspberry Pi SD cards, but it is no longer the preferred method of installing images to the SD card.

Gordon

andrum99:
Thank you for that explanation Gordon - that's the first time I've seen it mentioned that NOOBS is actually still the preferred option for resellers who sell pre-imaged SD cards. I will create some suitable wording.

andrum99:
One further question @ghollingworth: given there is now no download page for NOOBS, are you happy for me to change this PR so that it removes the instructions for how users can install NOOBS to an SD card themselves? (Since they can't really download it without a download page, although it is still available to download from the download server itself).

aallan:
End users shouldn't be using NOOBS. They should be using the Imager, the documentation should reflect that.

ghollingworth:
Agree, removing those instructions should be fine and it will stop people getting confused!

:)

andrum99:
Thanks everyone. I've added changes to the NOOBS page per Gordon's recommendations.

installation/noobs.md
```diff
@@ -1,50 +1,20 @@
...
-
-##### Linux
-
-For Linux users we recommend `gparted` (or the command line version `parted`). Norman Dunbar has written up [instructions](http://qdosmsq.dunbar-it.co.uk/blog/2013/06/noobs-for-raspberry-pi/) for Linux users.
+SD cards with NOOBS pre-installed are available from many of our distributors and independent retailers, including [Pimoroni](https://shop.pimoroni.com/products/noobs-8gb-sd-card), [Adafruit](https://www.adafruit.com/products/1583), and [Pi Hut](http://thepihut.com/collections/raspberry-pi-sd-cards-and-adapters/products/noobs-preinstalled-sd-card).
 
 ## What's included in NOOBS
 
-The following operating systems are currently included in NOOBS:
+The following operating systems are currently included in NOOBS. Note that only Raspberry Pi OS is supplied on the SD card - all other choices require that you have an internet connection to download the operating system.
```
lurch, line 17: ping @XECDesign to confirm which OS images are included on the preloaded NOOBS SD-cards.

installation/noobs.md
```diff
@@ -1,50 +1,20 @@
 # NOOBS
 
-**New Out Of Box Software (NOOBS)** is an easy operating system installation manager for the Raspberry Pi.
+**Note: We now recommend that you use [Raspberry Pi Imager](https://www.raspberrypi.org/software/) to install an operating system, instead of using NOOBS. If you are buying an SD card with software pre-installed on it, then we do still recommend buying one with NOOBS on it.**
```
lurch, line 3: The "If you are buying ... then we do still recommend ..." text feels a little bit awkward. Perhaps this paragraph should just say something like:

**Note: We now recommend that you use [Raspberry Pi Imager](https://www.raspberrypi.org/software/) to install an operating system onto a blank SD card, instead of using NOOBS.**

installation/README.md
```diff
@@ -1,11 +1,9 @@
 # Installation
 
-Some simple guides to setting up the software on your Raspberry Pi.
+We recommend using [Raspberry Pi Imager](https://www.raspberrypi.org/software/) to install an operating system on your Raspberry Pi. As well as Raspberry Pi OS, you can use it to install other operating systems such as Ubuntu or LibreELEC (Kodi media centre).
```
lurch, line 3: Should probably make the "Raspberry Pi OS" text a link to `../raspbian/README.md` ?

installation/README.md
```diff
@@ -1,11 +1,9 @@
...
-Beginners should start with [NOOBS](noobs.md), which gives the user a choice of operating system from the standard distributions. The recommended distribution for normal use is Raspberry Pi OS. Alternatives are available, such as LibreELEC (Kodi media centre) or Arch Linux.
+* You can find detailed help with installing OS images [here](installing-images/). For further assistance, check out our [forums](https://www.raspberrypi.org/forums/).
 
-## Contents
+* [Further information about SD cards](sd-cards.md).
 
-- [NOOBS](noobs.md)
-- [Installing images](installing-images/README.md)
-- [SD Cards](sd-cards.md)
+* [NOOBS](noobs.md)
```
lurch, line 9: Do we still need/want the link to the NOOBS page? :shrug:

installation/README.md
```diff
@@ -1,11 +1,9 @@
 # Installation
 
-Some simple guides to setting up the software on your Raspberry Pi.
+We recommend using [Raspberry Pi Imager](https://www.raspberrypi.org/software/) to install an operating system on your Raspberry Pi. As well as Raspberry Pi OS, you can use it to install other operating systems such as Ubuntu or LibreELEC (Kodi media centre).
 
-Beginners should start with [NOOBS](noobs.md), which gives the user a choice of operating system from the standard distributions. The recommended distribution for normal use is Raspberry Pi OS. Alternatives are available, such as LibreELEC (Kodi media centre) or Arch Linux.
+* You can find detailed help with installing OS images [here](installing-images/). For further assistance, check out our [forums](https://www.raspberrypi.org/forums/).
 
-## Contents
+* [Further information about SD cards](sd-cards.md).
```
lurch, line 7: Feels like this should be something like "You can find more detailed information about SD card requirements **here**" and be merged with the previous "You can find..." sentence, rather than having separate bullet points? :shrug: 
(just looks a bit out-of-context otherwise, IMHO)

installation/noobs.md
andrum99: I'm not sure there's much point spending too much time on revising this page. WRT this specific comment, it's probably easier if I just delete the list of alternative OSes, as it will no doubt change again in the future.

installation/noobs.md
andrum99: Gordon asked that the page mention NOOBS is still recommended for pre-installed SD cards.

installation/README.md
andrum99: No - that would complicate it. Better to just link straight to the download page, as I have done with the Imager link.

installation/README.md
andrum99: Yes - NOOBS needs to still be there as it is still officially supported for pre-installed cards. Not mentioning it at all in the docs would suggest that it is not supported.

installation/README.md
andrum99: Again, I don't think it's worth spending too much time working on this page. Your wording is perhaps better, but it's also more wordy so I don't think it's worth changing.

lurch: Fair enough, I just thought that https://github.com/raspberrypi/documentation/blob/4839e89445ee97293a399a3c6abf46dbd806f008/installation/README.md looked a bit odd - but maybe that's just me.

installation/noobs.md
lurch: I agree it makes sense to remove the list of alternative OSes and replace with more generic-sounding text.
The only reason I was querying the OSes on preloaded SD cards is that at one point "full NOOBS" included both Raspberry Pi OS and LibreELEC, and I have no idea if that's still the case or not...

installation/noobs.md
lurch: s/recommended/available/ ? :wink: 

**Note: We now recommend that you use Raspberry Pi Imager to install an operating system onto a blank SD card, instead of using NOOBS. Alternatively, you can buy an SD card with NOOBS pre-installed.**

:grey_question:

andrum99: I thought about that, but I think it is probably better to still say NOOBS is recommended for pre-installed cards. The reason for this is that otherwise it sounds like we are saying that you can use NOOBS on a pre-installed card, but that it is not recommended, which is not the case.

installation/noobs.md
andrum99: I've removed the list of OSes, and also updated the latst version info as it was out of date. I'm downloading NOOBS now to check if it has LibreELEC built in.

installation/noobs.md
lurch: Fair enough, perhaps simply removing the "do still" from your original text resolves the "awkward feeling" I had :slightly_smiling_face:

installation/noobs.md
andrum99: You're correct - LibreELEC is included in NOOBS.

installation/noobs.md
andrum99: Fixed.

andrum99:
I've added mention of LibreELEC being included in NOOBS, since it now is.

aallan:
Is there any reason why this should not be merged in? Apart from the conflicts? If not we should do so in the next couple of weeks. See #1911.

andrum99:
> Is there any reason why this should not be merged in? Apart from the conflicts? If not we should do so in the next couple of weeks. See #1911.

Looking at the current state of the pages, it seems this PR has been superceded by other changes and is no longer necessary - closing.

PR #1774, opened 2021-01-06
change example to put file in /lib/systemd/system

scyto:
Some changes for consideration. It seems better practice to put the .service file in /lib/systemd/system and let system d create the symlinks when enable is used.

But i defer to others if this is silly for some reason.

I also added an example of using alias - super useful when you want a longer descriptive .service filename but you want the stop and start name to be shorter.

I know both of these about preference not right/wrong so accept / reject PR as you see fit :-).

### DIFF
--- a/linux/usage/systemd.md
+++ b/linux/usage/systemd.md
@@ -23,12 +23,13 @@ User=pi
 
 [Install]
 WantedBy=multi-user.target
+Alias=ms.service
 ```
 So in this instance, the service would run Python 3 from our working directory `/home/pi/myscript` which contains our python program to run `main.py`. But you are not limited to Python programs: simply change the ExecStart line to be the command to start any program/script that you want running from booting.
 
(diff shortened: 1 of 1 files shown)


andrum99:
Debian's documentation, at https://wiki.debian.org/systemd#Creating_or_altering_services says this:

>  System administrators should put new or heavily-customized unit files in /etc/systemd/system.

It also says that /lib/systemd/system is for unit files installed by packages.

I'm not an expert on systemd, but it looks like /etc/systemd/system is correct in this instance.

scyto:
@andrum99 thanks, the idea being that /lib/ should be used for ones that ship with the OS - got it.

I kinda like that the disabled ones sat in /lib and got sym linked only when enabled - i think it tweaked my OCD when doing ls -l - that mine were bright green vs the color of all the others ones. :-)

Closing.  Does that reject the PR if not, can someone do that?

PR #1775, opened 2021-01-06
Update net_tutorial.md to allow DNS server to be used for internal systems. 

Treah:
Specifying Port=0 is not required and will actually cause problems if you are using this for other systems on the network needing DNS services. It should be omitted allowing the default port to be used. I assume this is probably originally put in here to prevent users from making a DNS server that is a security risk but there should at least be a comment about it as I have added here.

### DIFF
--- a/hardware/raspberrypi/bootmodes/net_tutorial.md
+++ b/hardware/raspberrypi/bootmodes/net_tutorial.md
@@ -170,6 +170,7 @@ sudo nano /etc/dnsmasq.conf
 Then replace the contents of `dnsmasq.conf` with:
 
 ```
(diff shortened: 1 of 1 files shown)


aallan:

See #1911 for full details.

Treah:
> I'd encourage you to wrap this PR up in the next week or so as we're in the process of transitioning the documentation from the current Markdown-based source format to Asciidoc. At some point soon — probably around the end of June, beginning of July — we will freeze the current documentation repo. After that time contributions and PRs based on the Markdown source will not be accepted, and any PRs that are still open will be closed.
> 
> See #1911 for full details.

How would I go about merging this? I do not have write access to the repo. This is just a simple change in the documentation.

PR #1776, opened 2021-01-08
Quickstart instruction fetches the wrong file

jjp-tlg:
The quickstart instruction downloads the blob source (.dts) file rather than the compiled binary (.bin) file.


lurch:
ping @6by9

6by9:
Yup, absolutely right.

PR #1779, opened 2021-01-09
Clarified and corrected example

FransTheekrans:
I used this guide to add some actions to the su cron, but they didn't run. I made some mistakes, and based on these mistakesm I propose updating the example as follows:

1.  added "sh" to shell script command in cron. Without this the shell script doesn't run
2. Clarified the example to make clear distinction between minutes and hours in the crontab file. Common sense would expect the first input to be hours, the second to be minutes. This is not the case in crontab. Altough the documentation was correct, the example was set at midnight (00:00), which did not highlight the pitfall.

### DIFF
--- a/linux/usage/cron.md
+++ b/linux/usage/cron.md
@@ -39,10 +39,10 @@ The layout for a cron entry is made up of six components: minute, hour, day of m
 For example:
 
 ```
-0 0 * * *  /home/pi/backup.sh
+15 0 * * *  sh /home/pi/backup.sh
 ```
 
-This cron entry would run the `backup.sh` script every day at midnight.
+This cron entry would run the `backup.sh` script, located in /home/pi/backup, every day at 15 minutes past midnight.
 
 ### View scheduled tasks
 


linux/usage/cron.md
```diff
@@ -39,10 +39,10 @@ The layout for a cron entry is made up of six components: minute, hour, day of m
 For example:
 
 ```
-0 0 * * *  /home/pi/backup.sh
+15 0 * * *  sh /home/pi/backup.sh
```
lurch, line 42: I think the `sh` is only needed if your script file doesn't have it's executable bit set? (i.e. you haven't done `chmod +x backup.sh` yet)

linux/usage/cron.md
```diff
@@ -39,10 +39,10 @@ The layout for a cron entry is made up of six components: minute, hour, day of m
...
 ```
 
-This cron entry would run the `backup.sh` script every day at midnight.
+This cron entry would run the `backup.sh` script, located in /home/pi/backup, every day at 15 minutes past midnight.
```
lurch, line 45: A "`backup.sh` script, located in /home/pi/backup" would be `/home/pi/backup/backup.sh`, but the crontab example shows `/home/pi/backup.sh` :confused:

linux/usage/cron.md
FransTheekrans: You are absolutely right, same on the comment below. I will recall the request, update and put in a new one. Thanks!

FransTheekrans:
Sh not needed, and incorrect folder designation

PR #1784, opened 2021-01-14
reorder arguments in all `dd` commands to match

DataWearsAHood:
for easier visual comparison of various `dd` incantations (for sticklers, etc.)


JamesH65:
Thanks!

PR #1796, opened 2021-01-22
Increase NFS rsize/wsize for netboot tutorial

ndanyluk:
rsize and wsize normally default to 1048576 (1MB) for NFS mounts, but for an NFS root device they default to 4096 (4KB).  This can lead to a significant performance degradation if the user expects the normal defaults.

### DIFF
--- a/hardware/raspberrypi/bootmodes/net_tutorial.md
+++ b/hardware/raspberrypi/bootmodes/net_tutorial.md
@@ -239,7 +239,7 @@ sudo systemctl restart nfs-kernel-server
 Edit `/tftpboot/cmdline.txt` and from `root=` onwards, and replace it with:
 
 ```
-root=/dev/nfs nfsroot=10.42.0.211:/nfs/client1,vers=4.1,proto=tcp rw ip=dhcp rootwait elevator=deadline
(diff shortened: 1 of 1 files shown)


JamesH65:
@timg236 Do you know anything about this? I have no idea.

timg236:
I've never tried it

ndanyluk:
@JamesH65 @timg236 rsize and wsize are parameters that govern the negotiated packet size for NFS filesystems.  They cap at 1MB (the param is in bytes).  With a typical NFS mount on a Pi, this defaults to 1MB.  However, the NFSROOT kernel directive has different default values.  In the case of the kernel NFSROOT, these parameters default to 4KB which fragments every read/write to the NFS root since each packet will be capped at 4KB rather than a max at 1MB.

[NFSROOT kernel.org documentation](https://www.kernel.org/doc/Documentation/filesystems/nfs/nfsroot.txt)

timg236:
I think it would be better to put that text in there explaining that to improve performance you can add those options

aallan:

aallan:
After some consideration I'm closing.

PR #1798, opened 2021-01-25
include Secure ShellFish as suggested client

palmin:
Note that I'm the developer of Secure ShellFish.


aallan:
In light of #1911 is this PR is not relevant. This feels very much non-Raspberry Pi specific material and therefore outside the scope of the more lightweight documentation set.

PR #1813, opened 2021-02-05
hardware/cmio: Add dt-blob files and docs for running the DSI display…

6by9:
… on DISP0

Default was only for DISP1, but the firmware does support either.
Adds dt-blob files and descriptive text for the options.


JamesH65:
Do we normally supply the bin file? Would it be worth having examples for the pin defines or do we rely of people looking at the dts file?

Otherwise LGTM.

6by9:
https://github.com/raspberrypi/documentation/tree/master/hardware/computemodule has .bin files for all the other alternatives, therefore I added it for this new one.

I'm expecting people to be prepared to look at the provided samples rather than to provide examples in the docs.

JamesH65:
Ta!

Issue #1810, opened 2021-02-02
Proposal: trim down documentation repo


andrum99:
The documentation repo is home to the source of the documentation section of the official Pi website, raspberrypi.org. Originally, the documentation section was the only official source of long-form information about Raspberry Pi: the other main sections of the website were the blog, and the forums. Over time, much more long-form content was added: projects.raspberrypi.org, MagPi content, and teaching resources to name a few. There was also, quite early on, a split to create a wholely owned subsiduary company - Raspberry Pi (Trading) Ltd, which was given custody of the documentation repo.

Raspberry Pi (Trading) Ltd is responsible for the trading aspects of what Raspberry Pi does: broadly, this means products. Trading does this to support the charitable mission of the Raspberry PI Foundation. So far, so good. However, since the documentation section of the website, and this documentation repo, was created before the creation of Trading, it contains documentation which should be managed by the Foundation. There is also duplication between the other parts of the website and the documentation section.

My proposal is to refocus the documentation repo, and the documentation section of the website, on technical product information, and either delete other content it currently contains, or move it so that it is under the control of the Foundation.

Pros:

- Reduced maintenance requirement (of docs repo)
- Reduce duplication of content and effort
- Make things less complicated for users

Cons:

- Potential loss of useful documentation
- Dead links will be created when cull occurs

My proposal for the various sections of the documentation repo is as follows: (sections being defined by the images present at raspberrypi.org/documentation):

- Setup / quickstart - remove. The main download page at https://www.raspberrypi.org/software/ is really good, and we've got a much better install tool now - Raspberry Pi Imager. The setup / quickstart part of the docs doesn't actually say very much, and I'm not sure how many people actually read it anyway. The "help" box on https://www.raspberrypi.org/software/ should perhaps be enhanced to add a link to the forums.

- Installation - remove. Again, there's not much here, and the installation process is much better than it was. Note that for this section, and the "setup / quickstart" section, there's also dedicated equivalent sections elsewhere on the website: they even have prominent links right on the help page (https://www.raspberrypi.org/help/): "Get started with Raspberry Pi" and "Setting up your Pi". Also, there should be a link to https://www.raspberrypi.org/help/ in the help box on https://www.raspberrypi.org/software/.

[...]

PR #1816, opened 2021-02-06
Setup: remove content, point to getting started guide instead

andrum99:


andrum99:
Changed my mind - I've gone for the nuclear option of just deleting the page content and pointing to the much more up-to-date getting started guide instead.

lurch:
I guess this is up to @aallan given that he's said he already has plans to overhaul/refresh the documentation :slightly_smiling_face:  #1810

andrum99:
> I guess this is up to @aallan given that he's said he already has plans to overhaul/refresh the documentation 🙂 #1810

Indeed. I think when I saw the reference to NOOBS on this page I thought it was worth just pointing to the much better 'getting started guide' rather than trying to edit the page, given the getting started guide is far better. An alternative would simply be to remove reference to NOOBS from this page.

JamesH65:
Going to be a while before @aallan is going to have time to look at this I suspect, so in the meantime we should just carry on as before. 

WRT this change, its true the Getting Started guide is much better, and duplicating stuff seems pointless, so I guess this is OK.

aallan:
> Going to be a while before @aallan is going to have time to look at this I suspect, so in the meantime we should just carry on as before.

Agreed. I don't see the overhaul project starting in earnest for another mont or so at the earliest.

lurch:
I just had a look and https://projects.raspberrypi.org/en/projects/raspberry-pi-setting-up/1 doesn't mention the composite video output available on the AV connector, but perhaps that's too much of a niche case to be worth documenting? :man_shrugging:

andrum99:
> I just had a look and https://projects.raspberrypi.org/en/projects/raspberry-pi-setting-up/1 doesn't mention the composite video output available on the AV connector, but perhaps that's too much of a niche case to be worth documenting? 🤷‍♂️

You would need to ask whoever wrote that page. I suspect the answer is as you suspect - hardly anyone uses composite. Also, Pi Zero and 400 don't do composite. (Although Zero has an unpopulated header for it).

I'm in two minds about mentioning composite - given there's no yellow (or black) composite port on the almost all Pis, unless you happen to read the spec you're not going to know it's there. On the other hand, it's something the vast majority of users would never use anyway.

aallan:
If it's currently undocumented, it's probably a sufficiently niche case that we shouldn't spend time documenting it at this point.

lurch:
The only reason I brought this up was because https://www.raspberrypi.org/documentation/setup/monitor-connection.md (which this PR removes) currently says:
"Others models (Raspberry Pi B+ and later) combine the audio out and composite out on to the same 3.5mm jackplug. This requires a particular type of lead, with audio left on the tip, audio right on ring 1, ground on ring 2, and video on the sleeve. This is the same as leads used on the Zune, and on Apple devices."

But with how ubiquitous HDMI displays are now, I'm happy to accept that this probably isn't worth documenting.

andrum99:
> The only reason I brought this up was because https://www.raspberrypi.org/documentation/setup/monitor-connection.md (which this PR removes) currently says:
> "Others models (Raspberry Pi B+ and later) combine the audio out and composite out on to the same 3.5mm jackplug. This requires a particular type of lead, with audio left on the tip, audio right on ring 1, ground on ring 2, and video on the sleeve. This is the same as leads used on the Zune, and on Apple devices."
> 
> But with how ubiquitous HDMI displays are now, I'm happy to accept that this probably isn't worth documenting.
> 
> (EDIT: see also https://github.com/raspberrypi/documentation/blob/master/configuration/config-txt/video.md and https://github.com/raspberrypi/documentation/blob/master/faqs/README.md#pi-video-displays )

I'm not sure if the fact that https://www.raspberrypi.org/documentation/setup/monitor-connection.md mentions composite makes any real difference - most people don't read the documentation pages. To me, if you're going to mention composite it should be on the relevant product pages, with perhaps a passing reference in the getting started guide.

I'm happy to close this PR and instead file a PR to just remove reference to NOOBS from this particular page if folk think that's a better solution.

JamesH65:
Anecdotally, I think we still have a surprising number of users especially third world, that complain like mad if composite threatens to leave. Also the Retro gaming community do like their composite displays.

andrum99:
> Anecdotally, I think we still have a surprising number of users especially third world, that complain like mad if composite threatens to leave. Also the Retro gaming community do like their composite displays.

It sounds like I should shelve this PR and instead submit one that just removes reference to NOOBS from this page. I will now do that.

If it is decided that this PR is the best approach, then it can of course be resurrected.

andrum99:
See #1819

PR #1817, opened 2021-02-06
Installation/sd-cards.md: trim down

andrum99:
Trim down the page to remove reference to NOOBS, and remove out-of-date content.

### DIFF
--- a/installation/sd-cards.md
+++ b/installation/sd-cards.md
@@ -3,3 +3,18 @@
 Raspberry Pi computers use a micro SD card, except for very early models which use a full-sized SD card.
 
 **Note:** Because of a hardware limitation in the Raspberry Pi Zero, 1 and 2, the boot partition on the SD card must be 256GB or smaller otherwise the device will not boot up. Later models of Raspberry Pi 2 with a BCM2837 SoC, Raspberry Pi 3, 4 and 400 do not have this limitation. This does not affect Raspberry Pi OS, which always uses a small boot partition.
+
+## Capacity
+
+We recommend using an SD card of 8GB or greater capacity with Raspberry Pi OS. If you are using the lite version of Raspberry Pi OS, you can use a 4GB card. Other operating systems have different requirements: for example, LibreELEC can run from a smaller card. Please check with the supplier of the operating system to find out what capacity of card they recommend.
+
+## Troubleshooting
+
+We recommend buying an official Raspberry Pi SD card from one of our official resellers. The official SD cards are micro SD cards and are supplied with an adaptor which allows them to be used in a full-sized SD card slot.
+
+If you are having problems with your SD card:
+
+1. Make sure you are using a genuine SD card. The best way to avoid fake SD cards is to always buy from a reputable supplier.
+2. Make sure you are using a good quality power supply: we recommend using an official Raspberry Pi power supply.
+3. The cable from the power supply unit to the Raspberry Pi can also cause problems. This is usually due to the resistance of the wires in the USB power cable; to save money, USB cables have as little copper in them as possible, which causes a voltage drop across the cable.
+4. Make sure you shut down the operating system correctly before you power down the Raspberry Pi.
--- a/installation/sd-cards.md
+++ b/installation/sd-cards.md
@@ -1,29 +1,7 @@
 # SD cards
 
-The Raspberry Pi should work with any compatible SD card, although there are some guidelines that should be followed:
+Raspberry Pi computers use a micro SD card, except for very early models which use a full-sized SD card.
 
-## SD card size (capacity) 
+## Limitation on Raspberry Pi Zero, 1 and 2 computers
 
-For installation of Raspberry Pi OS with desktop and recommended software (Full) via NOOBS the minimum card size is 16GB. For the image installation of Raspberry Pi OS with desktop and recommended software, the minimum card size is 8GB. For Raspberry Pi OS Lite image installations we recommend a minimum of 4GB. Some distributions, for example LibreELEC and Arch, can run on much smaller cards. If you're planning to use a card of 64GB or more with NOOBS, see [this page](sdxc_formatting.md) first.
-
-**Note:** Because of a constraint in versions of SoC's used in the Raspberry Pi Zero, 1 and 2, the SD card partition size limit is 256GB. From the Raspberry Pi 3 onwards this limitation does not apply. 
-
-## SD card class
-
-The card class determines the sustained write speed for the card; a class 4 card will be able to write at 4MB/s, whereas a class 10 should be able to attain 10 MB/s. However, it should be noted that this does not mean a class 10 card will outperform a class 4 card for general usage, because often this write speed is achieved at the cost of read speed and increased seek times.
-
-## SD card physical size
-
-The original Raspberry Pi Model A and Raspberry Pi Model B require full-size SD cards. From the Model B+ (2014) onwards, a micro SD card is required.
-
-## Troubleshooting
-
-We recommend buying the Raspberry Pi SD card which is available [here](https://shop.pimoroni.com/products/noobs-8gb-sd-card), as well as from other retailers; this is an 8GB class 6 micro SD card (with a full-size SD adapter) that outperforms almost all other SD cards on the market and is a good value solution.
-
-If you are having trouble with corruption of your SD cards, make sure you follow these steps:
-
-1. Make sure you are using a genuine SD card. There are many cheap SD cards available which are actually smaller than advertised or which will not last very long.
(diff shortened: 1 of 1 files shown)


andrum99:
I've corrected the information about the 256GB SD card partition size limit, since it seems the docs as currently written are incorrect. 

See https://www.raspberrypi.org/forums/viewtopic.php?f=63&t=166412&p=1814340#p1814206.

installation/sd-cards.md
```diff
@@ -1,29 +1,9 @@
...
-4. Make sure you are shutting your Raspberry Pi down properly before powering it off. Type `sudo halt` and wait for the Pi to signal it is ready to be powered off by flashing the activity LED.
-5. Finally, corruption has been observed if you are overclocking the Pi. This problem has been fixed previously, although the workaround used may mean that it can still happen. If after checking the steps above you are still having problems with corruption, please let us know.
+There is a hardware limitation on Pi Zero, 1 and 2 computers which means that the boot partition must be 256GB or less, otherwise the Pi will not boot. Most operating systems use a small boot partition so this is not normally a problem. However, when using NOOBS, you will normally start by formatting the SD card as a single FAT32 partition which takes up the whole SD card. On SD cards larger than 256GB this will result in the Pi being unable to boot. This problem can be avoided by simply not using NOOBS; alternatively, create the FAT32 partition so that it has a size of 256GB or less.
```
lurch, line 9: Feels like a lot of verbiage when all you needed to do was change the original text to "**Note:** Because of a constraint in versions of SoC's used in the Raspberry Pi Zero, 1 and 2, the SD card boot partition size limit is 256GB. From the Raspberry Pi 3 onwards this limitation does not apply." ?
And I guess you _could_ add "However this only affects NOOBS, and can be resolved by creating the FAT32 partition smaller than 256GB." ?

installation/sd-cards.md
```diff
@@ -2,8 +2,4 @@
...
 
-## Using very large SD cards on Raspberry Pi Zero, 1 and 2 computers
-
-All Raspberry Pi computers can be used with a very large capacity - greater than 256GB. There is however a limitation which you may encounter if you use NOOBS on these devices.
-
-There is a hardware limitation on Pi Zero, 1 and 2 computers which means that the boot partition must be 256GB or less, otherwise the Pi will not boot. Most operating systems use a small boot partition so this is not normally a problem. However, when using NOOBS, you will normally start by formatting the SD card as a single FAT32 partition which takes up the whole SD card. On SD cards larger than 256GB this will result in the Pi being unable to boot. This problem can be avoided by simply not using NOOBS; alternatively, create the FAT32 partition so that it has a size of 256GB or less.
+**Note:** Because of a hardware limitation in the Raspberry Pi Zero, 1 and 2, the boot partition on the SD card must be 256GB or smaller otherwise the device will not boot up. Later models of Raspberry Pi 2 which use a BCM2737 SoC do not have this limitation. This does not affect Raspberry Pi OS, which always uses a small boot partition.
```
lurch, line 5: BCM2837? :wink:

installation/sd-cards.md
andrum99: Thanks - you're right. I've trimmed it down - the new version mostly uses your suggested wording, but I've removed all reference to NOOBS as well, since it is now not the recommended solution anyway - no point complicating things.

installation/sd-cards.md
andrum99: I thought that looked wrong.

installation/sd-cards.md
```diff
@@ -1,29 +1,5 @@
...
-3. Make sure you are using a good quality USB cable for the power supply. When using a lower quality power supply, the TP1->TP2 voltage can drop below 4.75V. This is generally due to the resistance of the wires in the USB power cable; to save money, USB cables have as little copper in them as possible, and as much as 1V (or 1W) can be lost over the length of the cable.
-4. Make sure you are shutting your Raspberry Pi down properly before powering it off. Type `sudo halt` and wait for the Pi to signal it is ready to be powered off by flashing the activity LED.
-5. Finally, corruption has been observed if you are overclocking the Pi. This problem has been fixed previously, although the workaround used may mean that it can still happen. If after checking the steps above you are still having problems with corruption, please let us know.
+**Note:** Because of a hardware limitation in the Raspberry Pi Zero, 1 and 2, the boot partition on the SD card must be 256GB or smaller otherwise the device will not boot up. Later models of Raspberry Pi 2 which use a BCM2837 SoC do not have this limitation. This does not affect Raspberry Pi OS, which always uses a small boot partition.
```
lurch, line 5: I wonder if this should clarify that the Pi3 and Pi4 also don't have this limitation? (perhaps it's better to be explicit than implicit - otherwise this might just look like outdated information which hasn't yet been updated to cover the Pi3?) :shrug:

andrum99: Good point.

installation/sd-cards.md
```diff
@@ -1,29 +1,5 @@
...
-4. Make sure you are shutting your Raspberry Pi down properly before powering it off. Type `sudo halt` and wait for the Pi to signal it is ready to be powered off by flashing the activity LED.
-5. Finally, corruption has been observed if you are overclocking the Pi. This problem has been fixed previously, although the workaround used may mean that it can still happen. If after checking the steps above you are still having problems with corruption, please let us know.
+**Note:** Because of a hardware limitation in the Raspberry Pi Zero, 1 and 2, the boot partition on the SD card must be 256GB or smaller otherwise the device will not boot up. Later models of Raspberry Pi 2 which use a BCM2837 SoC, Raspberry Pi 3, 4 and 400 do not have this limitation. This does not affect Raspberry Pi OS, which always uses a small boot partition.
```
lurch, line 5: It's only a minor tweak, but I'd be tempted to replace "which use a" with "with a" ?

andrum99: Fixed.

installation/sd-cards.md
```diff
@@ -1,29 +1,5 @@
...
-2. Make sure you are using a good quality power supply. You can check your power supply by measuring the voltage between TP1 and TP2 on the Raspberry Pi; if this drops below 4.75V when doing complex tasks then it is most likely unsuitable.
-3. Make sure you are using a good quality USB cable for the power supply. When using a lower quality power supply, the TP1->TP2 voltage can drop below 4.75V. This is generally due to the resistance of the wires in the USB power cable; to save money, USB cables have as little copper in them as possible, and as much as 1V (or 1W) can be lost over the length of the cable.
-4. Make sure you are shutting your Raspberry Pi down properly before powering it off. Type `sudo halt` and wait for the Pi to signal it is ready to be powered off by flashing the activity LED.
-5. Finally, corruption has been observed if you are overclocking the Pi. This problem has been fixed previously, although the workaround used may mean that it can still happen. If after checking the steps above you are still having problems with corruption, please let us know.
```
lurch, line 29: I agree that a lot of this is superfluous now (especially the bit about measuring TP1 voltage!!) but I wonder if it's worth keeping the tips about using a genuine SD card, using a good quality power supply, and shutting down before removing power?

Probably also doesn't hurt to retain the "miminum 8GB card size" requirement for Raspberry Pi OS and 4GB for Raspberry Pi OS Lite?

andrum99: My thinking was that the main getting started guide does not have any such information about power supplies, and most folk will never read this page of the docs so I though tit was best not to have too much content here as that simply means there's more to maintain. The minimum size requirements should be on the page at https://www.raspberrypi.org/software/operating-systems/ - there should be no need to have them here.

However, I think to be on the safe side it is probably best to change the PR as you suggest.

lurch (approved):
Thanks @andrum99 LGTM now :+1:

andrum99:
Thanks lurch. I would suggest Gordon, or someone else who knows the details of the boot ROM bug in the BCM2835 and BCM2836 which limits the size of the boot partition, have a look at this PR to ensure that we get it right this time.

aallan:

See #1911 for full details.

andrum99:
This one needs confirmation of the exact nature of the ROM bug in the BCM2835 and BCM2836 which puts an upper limit on the size of the FAT partition that the device can boot from, then it's good to go. Not sure who would know that - perhaps someone at Pi Towers could ping a suitable engineer, e.g. Gordon or Tim.

aallan:
@ghollingworth Can you comment?

ghollingworth:
So the original information is correct but could be worded better:

The device will not boot from a FAT32 formatted partition greater than 256GB.

Gordon

andrum99:
> So the original information is correct but could be worded better:
> 
> The device will not boot from a FAT32 formatted partition greater than 256GB.
> 
> Gordon

Thanks for the confirmation.

PR #1820, opened 2021-02-09
Documentation update (re. GL driver on Pi 3s)

davidplowman:


davidplowman:
@JamesH65 Hi James, would appreciate this being merged when you have a few moments, assuming you're happy with it. It fixes the rather glaring omission that I failed to mention, namely that you need to change over to the fake KMS driver on Pi 3s for the libcamera-apps preview window to work!

PR #1825, opened 2021-02-12
Update Integrator Programme info

helenlynn:
We now have a dedicated page for the Integrator Programme, so I've cut down the info here in favour of sending people over to the new page. @JamesH65 please could you sense-check and merge?


JamesH65:
LGTM

PR #1828, opened 2021-02-18
Remove "users" option from fstab

aripollak:
I haven't found a "users" option; it's normally "user", but I don't think that's a good default anyway. "auto" and "rw" are also already included in "defaults," so I removed those.

### DIFF
--- a/configuration/external-storage.md
+++ b/configuration/external-storage.md
@@ -68,10 +68,9 @@ You can modify the `fstab` file to define the location where the storage device 
 4. Add the following line in the `fstab` file:
 
     ```
-    UUID=5C24-1453 /mnt/mydisk fstype defaults,user,nofail 0 0
+    UUID=5C24-1453 /mnt/mydisk fstype defaults,nofail 0 0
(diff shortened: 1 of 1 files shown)


lurch:
ping @XECDesign and @spl237 for clarification / confirmation, and in case this has knock-on effects on other parts of Raspberry Pi OS (e.g. the eject icon on the taskbar).

XECDesign:
> t means that any user can unmount the filesystem

I'd assume that would be the desired behaviour, so I can't really comment on the PR. I wouldn't change it, but it depends on what the intent of the page is.

aripollak:
FWIW, I don't think filesystems in /etc/fstab show up under the eject icon?

lurch:
If XECDesign says not to change it, then perhaps a better fix would be to just add a note to this page explaining the impact of the "users" option?

aallan:

aripollak:
Yeah, seems like this can be closed if the doc in question will be removed soon.

PR #1831, opened 2021-02-23
bootloader: Simplify EEPROM write protect documentation

timg236:
This is a standard feature now that CM4 is released. Remove duplication
of settings and make the description of the confusing internaction
between /WP and the Write Status Register less long winded.


JamesH65:
LGTM

PR #1832, opened 2021-02-28
Update access-point-bridged.md

tomwilde:
Had some difficulty with this - this is what enabled the daemon instead of it failing.

https://unix.stackexchange.com/questions/119209/hostapd-will-not-start-via-service-but-will-start-directly


aallan:

epoch1970:
Never been necessary to me on a clean copy of raspios.

aallan:

PR #1845, opened 2021-03-12
Update bootflow_2711.md

andrum99:
Missed a stray hyphen on my last PR.


lurch:
You _could_ argue that the hyphen here is okay, and is turning "boot-mode" into a compound noun?

andrum99:
> You _could_ argue that the hyphen here is okay, and is turning "boot-mode" into a compound noun?

If you did then you would be wrong: compound nouns don't need a hyphen. (It's also `boot mode` without the hyphen everywhere else).

lurch:
> (It's also boot mode without the hyphen everywhere else).

Fair enough. I always like consistency :wink:

Issue #1754, opened 2020-12-07
UART: Enabling of UART1 for CM3+ does not work


boraozgen:
Hi all,

I spent a couple of hours today trying to enable UART1 (Mini UART) for my CM3+. I tried the `enable_uart=1` and `core_freq=250` settings as depicted in the documentation to no avail. After reading some forum posts and github issues, the actual solution for me was to set `dtoverlay=uart1,txd1_pin=32,rxd1_pin=33` in `config.txt`. I'm not sure if this is CM specific, but I think it should be added to the documentation.

Cheers

lurch:
ping @pelwell

pelwell:
That's the correct behaviour. Rather than enable a lot of devices on the off-chance that customers want them, the CM DTS files opt to leave most optional devices disabled. It was always expected that customers would customise the DTS/DTB to suit their application, or provide a custom overlay incorporating the necessary changes.

boraozgen:
It was not clear to me that UART1 is one of the "changes" from the main board. Anyway, thanks for the clarification and you can close the issue if you see no need for additions.

pelwell:
UART1 is "changed" from the "main board" (I assume you mean Pi 3B) because CM3 has no built-in Bluetooth interface. CM3 isn't a variant of the Pi 3, it's a completely different board using the same SoC.

boraozgen:
You are right, but still the documentations seems to be written for all RPi based boards including CM, however one cannot follow it for CM due to the missing DT modification step. If this is supposed to be an implicit knowledge, you are right.

pelwell:
I think the Compute Module + CMIO combination is being used by more end-users than I would have expected; I have no objection to adding a user-friendly comment to the UART documentation.

michaelstoops:
I also found the UART documentation confusing, particularly the invention of "primary" and "secondary" UARTs, in addition to their hardware numbering, device types, device node paths symbolic links.

I have a draft of a major revision for that document, but I want to step away a time and proofread it before I submit the changes.

michaelstoops:
I posted a pull request: https://github.com/raspberrypi/documentation/pull/1849

PR #1849, opened 2021-03-20
Major rewrite of UART documentation in response to issues/1754.

michaelstoops:
Like it says, this is a major rewrite of the UART documentation. Nominally, this is because of /https://github.com/raspberrypi/documentation/issues/1754. But also, I found this documentation confusing when I needed it a couple weeks ago, so I gave a shot at fixing the whole thing.

I got rid of the terminology of "primary" and "secondary" devices, which seemed to appear only in this doc, not in Broadcom's docs or others. Instead, I attempted to consistently refer to hardware devices by their canonical hardware names: UART0, etc. When discussing the serial port on the GPIO header, I referred to it as `/dev/serial0`, because that's how it will appear to the reader in Linux.

As you review this, please look first at the structure and clarity for the expected audience. Does this meet the reader's need? My intent was to give:
* a brief introduction to what a UART is in general, and relate it to other things they may know
* then to discuss the UARTs present in Raspberry Pi devices
* then to discuss applications and other how-to instructions

Once we've settled on the structure, I think we need to look at the accuracy of the information. This is a big change and it contains many assertions of fact. I've attempted to validate them against reality. You can see my methodology and data at https://github.com/michaelstoops/documentation/blob/1754-testing/configuration/uart-testing.md. It should be easy to reproduce if you have the appropriate tools.

Lastly, let's look at minor issues of style: formatting and the like. I had to decide which things to 'quote' and which to mark as `inline code`, which I don't think is quite clear in the style guide. I gave a reasonable attempt, but I also edited and re-edited a lot, so I'm not sure they're all consistent in the end.

Your feedback is welcome.

PR #1850, opened 2021-03-20
configuration/UART.md - minor simplifications

andrum99:


andrum99:
I've also added a quick mention of what is required on the various models of compute module, which fixes #1754.

aallan:
How does this relate to #1849?

andrum99:
> How does this relate to #1849?

It's a reaction to #1849, which purports to solve #1754 but doesn't. This one does. #1849 also adds lots of extra verbage, and if it were up to me that would not get integrated.

I will leave this PR open should you decide it is worth integrating: it contains minor additions to add information which some forum users have requested, and minor corrections, and I believe it is ready to integrate as is.

aallan:
So for clarity, you're arguing this PR should be merged instead of #1849? So it's an either or? Thanks, that clarifies the situation!

andrum99:
> So for clarity, you're arguing this PR should be merged instead of #1849?

That is indeed my position.

aallan:
> That is indeed my position.

Gotcha.

PR #1855, opened 2021-03-22
Update advice to use https in the repository definition.

bjhulst:


lurch:
This documentation repo is only for documenting Raspberry Pi OS, in which case there should be no need for the user to edit the repository link?? ping @XECDesign just in case I've misunderstood.

XECDesign:
The PR doesn't explain how https is more secure given that package lists are signed and it doesn't explain how to switch to https.

But even if the above points are addressed, I'd argue this isn't necessary. I suspect the number of used who are concerned about somebody spying on them to find out what packages they're installing is fairly small.

lurch:
There's also the advantage that http is easier to cache, which may be beneficial to some people (e.g. schools with multiple Pis) ?

bjhulst:
I updated the text a bit.

lurch:
Thanks for the clarification, but IMHO this feels a bit too vague / waffly to add to the documentation? I'm inclined to agree with XECDesign's "I'd argue this isn't necessary", unless you can give good reasons otherwise?

bjhulst:
No advice it is then.

lurch:
We **do** appreciate your contribution, but we try to keep the documentation "tight" and not include (too much) unnecessary information.

Issue #1787, opened 2021-01-17
Documentation /documentation/blob/master/installation/installing-images/mac.md contains an error


kevinnivek:
In the document, the following snippet : 

`sudo dd bs=1m if=path_of_your_image.img of=/dev/rdiskN; sync`

Should be written as : 

`sudo dd bs=1m of=path_of_your_image.img if=/dev/rdiskN; sync`

The "if" and "of" in the snippet are backwards.

lurch:
What makes you think they're backwards? "if" = "Input File" and "of" = "Output File", so for copying a disk-image onto an SD-card, `if=path_of_your_image.img of=/dev/rdiskN` looks correct to me?

michaelstoops:
To clarify, I think you're talking about:
https://github.com/raspberrypi/documentation/blob/d3e51178b5f27b90efb9f8ac62b269c8748c4b35/installation/installing-images/mac.md#L64

I agree with @lurch and the existing documentation. The context is for writing an image from file to SD card, so the `if` is the image file and `of` is the block device.

I think this issue can be closed.

lurch:
Closing as invalid.

PR #1856, opened 2021-03-22
Update mac.md

esackowitz:
The usage of "if" and "of" command line switches for dd utility command on MAC was listed backwards. I have corrected that.

 if=file  Read input from file instead of the standard input
of=file  Write output to file instead of the standard output.

### DIFF
--- a/installation/installing-images/mac.md
+++ b/installation/installing-images/mac.md
@@ -61,7 +61,7 @@ If you specify the wrong device in the instructions, you could overwrite your pr
 - Copy the image
 
   ```bash
-  sudo dd bs=1m if=path_of_your_image.img of=/dev/rdiskN; sync
+  sudo dd bs=1m of=path_of_your_image.img if=/dev/rdiskN; sync
   ```
 
    Replace `N` with the number that you noted before. Note the ```rdisk``` ('raw disk')


JamesH65:
AFAIK, the original is correct, 'of' is output file, the destination, and 'if' is the input file, the image to write.

andrum99:
The page is indeed correct as it is currently written.

lurch:
This has a bizarre similarity to #1787 !  I wonder if there's a reason or if it's just coincidence?

andrum99:
Well this one comes from an account with this PR as its first and only github contribution, so maybe not. Im any case, it should be closed

lurch:
Closing as invalid.

esackowitz:
The usage of "if" and "of" command line switches for dd utility command on MAC was listed backwards. I have corrected that.

if=file Read input from file instead of the standard input
of=file Write output to file instead of the standard output.

andrum99:
> The usage of "if" and "of" command line switches for dd utility command on MAC was listed backwards. I have corrected that.
> 
> if=file Read input from file instead of the standard input
> of=file Write output to file instead of the standard output.

They are not backwards - the command provides instructions for writing an image *to* a drive. The page as written tells the user to command that the dd utility take the image `path_of_your_image.img` as it's input (`if=`) and output it (`of=`) to the destination drive at `/dev/rdiskN`.

esackowitz:
Please look again.  It was specifically the example given that was reversed.  Not the definition of the terms.

This is what was in the original text which claims the path of the image is in the "if" parameter. "if" is NOT the "path of your image.img". "if" is the read operation and should reference the device/disk you are taking the image of. "of" is the write operation which is the path and name of the image file.

ORIGINAL TEXT:
"sudo dd bs=1m if=path_of_your_image.img of=/dev/rdiskN; sync"

CORRECTED TEXT:
"sudo dd bs=1m **of**=path_of_your_image.img **if**=/dev/rdiskN; sync"

If you use the instructions as written, you will get an error. "No such file or directory" because the inputs were reversed as implied by the path/image text.

PR #1858, opened 2021-03-24
Fix note about uncompressed arm64 images

waveform80:
The Pi's built-in bootloader gained the ability to unpack gzip'd kernel images in Dec 2020 (I think? Sometime around then anyway); in fact, this is what raspios is currently using for kernel8.img.

### DIFF
--- a/configuration/config-txt/boot.md
+++ b/configuration/config-txt/boot.md
@@ -39,7 +39,7 @@ Set the `disable_commandline_tags` command to `1` to stop `start.elf` from filli
 
 ## arm_64bit
 
-If set to non-zero, forces the kernel loading system to assume a 64-bit kernel, starts the processors up in 64-bit mode, and sets `kernel8.img` to be the kernel image loaded, unless there is an explicit `kernel` option defined in which case that is used instead. Defaults to 0 on all platforms. **NOTE**: 64-bit kernels must be uncompressed image files.
+If set to non-zero, forces the kernel loading system to assume a 64-bit kernel, starts the processors up in 64-bit mode, and sets `kernel8.img` to be the kernel image loaded, unless there is an explicit `kernel` option defined in which case that is used instead. Defaults to 0 on all platforms. **NOTE**: 64-bit kernels may be uncompressed image files or a gzip archive of an image (which can still be called kernel8.img; the bootloader will recognize the archive from the signature bytes at the beginning).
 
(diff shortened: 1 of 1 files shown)


JamesH65:
@pelwell I have no idea whether this is correct now.

pelwell:
64-bit kernel decompression has been supported since September 2020, so I'm happy with this update. For completeness (and as a slight correction/clarification) it's also worth mentioning that there is an implicit arm_64bit=1 if:

a. an explicit kernel name is given (`kernel=...`) that ends "8.img", or
b. there is no explicit kernel name and (judging by the filenames in the "OS" directory - usually `/boot`) there is a 64-bit kernel and no 32-bit kernel.

configuration/config-txt/boot.md
```diff
@@ -39,7 +39,7 @@ Set the `disable_commandline_tags` command to `1` to stop `start.elf` from filli
 
 ## arm_64bit
 
-If set to non-zero, forces the kernel loading system to assume a 64-bit kernel, starts the processors up in 64-bit mode, and sets `kernel8.img` to be the kernel image loaded, unless there is an explicit `kernel` option defined in which case that is used instead. Defaults to 0 on all platforms. **NOTE**: 64-bit kernels must be uncompressed image files.
+If set to non-zero, forces the kernel loading system to assume a 64-bit kernel, starts the processors up in 64-bit mode, and sets `kernel8.img` to be the kernel image loaded, unless there is an explicit `kernel` option defined in which case that is used instead. Defaults to 0 on all platforms. **NOTE**: 64-bit kernels may be uncompressed image files or a gzip archive of an image (which can still be called kernel8.img; the bootloader will recognize the archive from the signature bytes at the beginning).
```
lurch, line 42: Is this now the same logic as the 32-bit kernel? (If it is, then perhaps this behaviour isn't worth pointing out specifically for the 64-bit kernel? :shrug: )

pelwell: No it's not the same. ARM32 kernel decompression (or not) is handled by the kernel image itself, whereas on ARM64 the loader is expected to take care of it.

lurch: Ahhh, thanks for clarifying.

aallan:

See #1911 for full details.

aallan:
> 64-bit kernel decompression has been supported since September 2020, so I'm happy with this update. 

Is this ready to merge @pelwell ?

pelwell:
Yes it's fine - is that not clear?

PR #1864, opened 2021-04-01
[Wi-Fi AP bridged] New section added about firewalls

theriverman:
After executing the steps in this document the user may end up with a seemingly working wireless AP, but that AP will not grant access to the internet because the traffic is blocked.
Without giving a notice about this potential pitfall here, the reader may think that they've misconfigured something meanwhile in reality they've just installed and enabled a firewall years ago that they already forgot about.

Having this elaborate hint added in the end, we can potentially save the reader from hours of headache. Just like in my case :-)

### DIFF
--- a/configuration/wireless/access-point-bridged.md
+++ b/configuration/wireless/access-point-bridged.md
@@ -186,4 +186,13 @@ Once your Raspberry Pi has restarted, search for wireless networks with your wir
 
 If your wireless client has access to the local network and the internet, congratulations on setting up your new access point!
 
+<a name="routing-config"></a>
+## Allow traffic on br0 in your firewall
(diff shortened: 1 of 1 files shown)


aallan:

See #1911 for full details.

epoch1970 (commented):
This is irrelevant to raspios AFAIK. Raspios ships with routing disabled and open firewall. There is nothing to configure. UFW isn’t shipped w/ raspios either, you’d need to add instructions on installing that…

aallan:
> This is irrelevant to raspios AFAIK. Raspios ships with routing disabled and open firewall. There is nothing to configure. UFW isn’t shipped w/ raspios either, you’d need to add instructions on installing that…

Closing.

PR #1865, opened 2021-04-06
Update unix.md

Kepsz:
Minor addition, just to make it clear blank password is not an option for ssh.


aallan:
Does anyone have a reason for not merging this minor change? @lurch @JamesH65 ? In light of #1911 we merge it, or don't in the next week or so.

andrum99:
This additional warning is worth having, but I would put it earlier in the page, and put it in its own paragraph as the 2nd paragraph on the page (i.e. after 'You can use SSH...'), like this:

`**Note:** If the pi user has a blank password you will not be able to log in using SSH.`

Although thinking about it, do we really need to document how to connect using SSH in the docs? Is that type of thing not best left to MagPi articles and such like?

aallan:
Seems a minor change. Will merge ahead of #1911 unless anyone objects?

PR #1869, opened 2021-04-13
Link camera page to libcamera page

lurch:
I'm unsure if the wording I've used is correct or if it needs tweaking.
ping @JamesH65 and @6by9

### DIFF
--- a/hardware/camera/README.md
+++ b/hardware/camera/README.md
@@ -60,7 +60,7 @@ The Raspberry Pi Camera Modules are official products from the Raspberry Pi Foun
 
 ## Software features
 
-Full camera software documentation can be found [here](../../raspbian/applications/camera.md).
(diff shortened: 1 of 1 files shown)


JamesH65:
Are we at a point where ware happy referring to the old camera stack as legacy ie We can now do the majority of stuff with libcamera that we can do with the original suite?

lurch:
This issue https://github.com/raspberrypi/userland/issues/688 is what inspired me to create this PR. I was just kinda surprised that there was no link from the camera page to the libcamera page. But if libcamera isn't "fully ready" yet (I've never actually used it myself), perhaps that was deliberate? :shrug: 
Feel free to tweak the wording as you see fit.

6by9:
libcamera is the route that we will push for any 64 bit system, and really for any 32bit Pi2/3/4.
Pi0/1 will probably be stuck on the old stack as they don't have the processing grunt for running the control algos on the ARM.

David P has written replacement versions of raspistill and vid for libcamera, although I haven't tried them. The gap is the Python bindings, but that is being addressed.

aallan:

See #1911 for full details.

PR #1871, opened 2021-04-13
Update datasheets page

lurch:


JamesH65:
What is the point of having a document naming scheme if people don't use it....bah humbug.

lurch:
ping @aallan ?

aallan:
Merged, because after all that's where the datasheet **is**. The whole page is changing post #1911 anyway.

PR #1874, opened 2021-04-16

andrum99:
The description of BCM-USB-MSD boot mode was not quite specific enough when it came to specifying which devices (Pi 4, CM4) have which type of USB connector. Also, the NVME boot description assumed there was a PCIe `socket`, which may not be the case on 3rd-party boards, so change that to `interface`.


lurch:
ping @timg236

timg236:
LGTM

PR #1877, opened 2021-04-18

andrum99:
CM4 IO board has a micro USB socket for booting from a USB MSD connected to the BCM2711's XHCI controller, not USB type A.


andrum99:
Ignore me - I got confused. The page is correct as currently written.

PR #1878, opened 2021-04-19
Update overclocking.md

timg236:
@popcornmix @JamesH65 

Update overclocking documentation to reflect state-of-the-art firmware on Pi 4. IIRC enable_tvout doesn't make much difference to core_freq these days and the core-clock can be over-clocked, but perhaps less so than V3D.

### DIFF
--- a/configuration/config-txt/overclocking.md
+++ b/configuration/config-txt/overclocking.md
@@ -1,6 +1,6 @@
 # Overclocking options in config.txt
 
-**NOTE:** Setting any overclocking parameters to values other than those used by [raspi-config](../raspi-config.md#overclock) may set a permanent bit within the SoC, making it possible to detect that your Pi has been overclocked. The specific circumstances where the overclock bit is set are if `force_turbo` is set to `1` and any of the `over_voltage_*` options are set to a value > `0`. See the [blog post on Turbo Mode](https://www.raspberrypi.org/blog/introducing-turbo-mode-up-to-50-more-performance-for-free/) for more information. **Pi 4** overclocking options may be subject to change in the future.
+**NOTE:** Setting any overclocking parameters to values other than those used by [raspi-config](../raspi-config.md#overclock) may set a permanent bit within the SoC, making it possible to detect that your Pi has been overclocked. The specific circumstances where the overclock bit is set are if `force_turbo` is set to `1` and any of the `over_voltage_*` options are set to a value > `0`. See the [blog post on Turbo Mode](https://www.raspberrypi.org/blog/introducing-turbo-mode-up-to-50-more-performance-for-free/) for more information. 
 
 The kernel has a [CPUFreq](https://www.kernel.org/doc/html/latest/admin-guide/pm/cpufreq.html) driver with the "powersave" governor enabled by default, switched to "ondemand" during boot, when [raspi-config](../raspi-config.md) is installed. With "ondemand" governor, CPU frequency will vary with processor load. You can adjust the minimum values with the `*_min` config options or disable dynamic clocking by applying a static scaling governor ("powersave" or "performance") or with `force_turbo=1`. For more information [see this section of the documentation](../../hardware/raspberrypi/frequency-management.md).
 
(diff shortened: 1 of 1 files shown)


configuration/config-txt/overclocking.md
```diff
@@ -39,22 +39,22 @@ Overclocking and overvoltage will be disabled at runtime when the SoC reaches 85
...
-| gpu_freq_min | 250    | 250 | 250 | 250   | 250   | 500  |
-| h264_freq_min|  250   | 250 | 250 | 250   | 250   | 500  |
-| isp_freq_min |  250   | 250 | 250 | 250   | 250   | 500  |
-| v3d_freq_min |  250   | 250 | 250 | 250   | 250   | 500  |
-| sdram_freq_min |400   | 400 | 400 | 400   | 400   | 400  |
+| Option       | Pi 0/W | Pi1 | Pi2 | Pi3   | Pi3A+/Pi3B+ | Pi4/CM4  | Pi 400 |
+| ---          | :---:  | :---: | :---: | :----:  | :-----: | :----: | :----: |
+| arm_freq     | 1000   | 700 | 900 | 1200  | 1400  | 1500 | 1800 | 
+| core_freq    | 400    | 250 | 250 | 400   | 400   | 500/550 | 500/550 |
+| h264_freq    | 300    | 250 | 250 | 400   | 400   | 500 | 500|
```
popcornmix, line 46: I believe these should be `500/550` for all gpu clocks (after "platform: Increase gpu freqs to 550 with enable_4kp60")

configuration/config-txt/overclocking.md
```diff
@@ -39,22 +39,22 @@ Overclocking and overvoltage will be disabled at runtime when the SoC reaches 85
...
+| isp_freq     | 300    | 250 | 250 | 400   | 400   | 500 | 500 |
+| v3d_freq     | 300    | 250 | 250 | 400   | 400   | 500 | 500 |
+| hevc_freq    | N/A    | N/A | N/A | N/A   | N/A   | 500 | 500 |
+| sdram_freq   | 450    | 400 | 450 | 450   | 500   | 3200 | 3200 |
+| arm_freq_min | 700    | 700 | 600 | 600   | 600   | 600  | 600 |
+| core_freq_min| 250    | 250 | 250 | 250   | 250   | 250 | 250 |
```
popcornmix, line 52: I think this is 200 on 2711.

andrum99:
Do we want to mention CM4, given we don't list CM1 and 3 on this page? Perhaps to keep the table simple we should not mention the CMs in the table and just say 'CM4 is the same as Pi 4', similarly for CM1 and 3, underneath?

configuration/config-txt/overclocking.md
```diff
@@ -73,22 +73,23 @@ This table lists the default `over_voltage` settings for the various Pi models.
...
+| Display option | Max `core_freq` |
 | -------------- | --------: |
 | Default        | 500 |
-| enable_tvout | 360 |
 | hdmi_enable_4kp60 | 550 |
 
-Changing `core_freq` in `config.txt` is not supported on the Pi 4, any change from the default will almost certainly cause a failure to boot.
-
-It is recommended when overclocking to use the individual frequency settings (`isp_freq`, `v3d_freq` etc) rather than `gpu_freq`, as since it attempts to set `core_freq` (which cannot be changed on the Pi 4), it is not likely to have the desired effect.
+* Overclocking requires the latest firmware release. Please make a backup before using [rpi-update](../../raspbian/applications/rpi-update.md)
```
andrum99, line 89: Does it require an rpi-update release? Do we want to be pointing people to rpi-update - if so a warning about it installing pre-release software would be useful.

configuration/config-txt/overclocking.md
```diff
@@ -73,22 +73,23 @@ This table lists the default `over_voltage` settings for the various Pi models.
 | --- | --- | --- |
 | Pi 1 | 0 | 1.2V |
 | Pi 2 | 0 | 1.2-1.3125V |
-| Pi 3 | 0 | 1.2-1.3125V |
+| Pi 3 | 0 | 1.2-1.3125V 
+| Pi 4 | 0 | 0.88V |
 | Pi Zero | 6 | 1.35V |
 
 #### Specific to Pi 4B
```
andrum99, line 80: Heading needs changed. Probably this should say `Pi 4B, Pi 400 and Compute Module 4` given we've specified specific products further up the page.

configuration/config-txt/overclocking.md
lurch: The warning on https://github.com/raspberrypi/documentation/blob/master/raspbian/applications/rpi-update.md is probably already sufficient?

configuration/config-txt/overclocking.md
lurch: Didn't we agree on another PR recently to just use "Pi 4" as shorthand for "Pi 4B, Pi 400 and Compute Module 4" ?

andrum99: Yes, but the default ARM clock is different on Pi 400, so Tim has given the Pi 400 its own column in the table. Perhaps the table can be simplified by just having the Pi 4 column headed simply `Pi 4` and for the ARM clock list it as `1500 (1800 on Pi 400)`? Then this heading would just be `Pi 4`.

configuration/config-txt/overclocking.md
andrum99: > The warning on https://github.com/raspberrypi/documentation/blob/master/raspbian/applications/rpi-update.md is probably already sufficient?

Good point. In which case the suggestion on this page to make a backup is not necessary.

configuration/config-txt/overclocking.md
lurch: > having the Pi 4 column headed simply Pi 4 and for the ARM clock list it as 1500 (1800 on Pi 400)

Yes, that might make things simpler as it looks like all the other values are the same.

configuration/config-txt/overclocking.md
andrum99: > I think this is 200 on 2711.

That's what I'm seeing on my Pi 4 with the current default/critical firmware. (Namely 2020-09-03).

configuration/config-txt/overclocking.md
timg236: Does tvout still make a difference or is that 500 (or close to) after the recent clock improvments?

configuration/config-txt/overclocking.md
timg236: I'll just delete the second sentance

configuration/config-txt/overclocking.md
andrum99: > I believe these should be `500/550` for all gpu clocks (after "platform: Increase gpu freqs to 550 with enable_4kp60")

There's no explanation of why there are two values here (i.e. 500/550). I think I'm right in saying the first number is the standard clock, while the second is with `hdmi_enable_4kp60`. To make the table less complicated, my suggestion would be to just list 500 as the clock speed, but put a superscripted 1 after it, then below the table have `1. If hdmi_enable_4kp60=1 is set in config.txt then this clock defaults to 550MHz.`

Superscripted characters need to be inserted using HTML, since there seems to be no markdown equivalent. The HTML would be `<sup>1</sup>`.

popcornmix: tvout doesn't affect core clock

aallan:

See #1911 for full details.

aallan:
Does anyone have opinions about this one? @lurch @andrum99 @popcornmix ?

JamesH65:
@timg236 Needs rebasing, but apart from that I think its good enough to merge.

aallan:
@timg236 If you can rebase, I'll tag as ready to merge and then go ahead and do that before #1911 happens.

timg236:
@aallan The latest rpi-update changes some of the clock configurations so I'll rebase and update. Just need to finish getting the firmware changes out of the door!

configuration/config-txt/overclocking.md
timg236: Added CM4, Pi400. There's no difference in voltage at 1.5 GHz

configuration/config-txt/overclocking.md
timg236: Changed to 500. There's a separate section for 4kp60 which notes that max core freq is increased to 550 MHz

JamesH65:
LGTM, but needs rebasing.

Issue #1760, opened 2020-12-12
Raspberry Pi Revision c03114 not in the list


tom-2015:
Seems like revision c03114 is missing in the documentation.
[https://www.raspberrypi.org/documentation/hardware/raspberrypi/revision-codes/README.md](url)

Does anyone know if there are other revision numbers missing?

JamesH65:
Probably - that list is never exhaustive, I just update it every now and again. Using the table above to correctly decode the revision number.

JamesH65:
I've updated the table to include the code you have.

Issue #1726, opened 2020-11-07
please the revision code of rpi 400


dogi:
.. to https://github.com/raspberrypi/documentation/blob/master/hardware/raspberrypi/revision-codes/README.md

JamesH65:
That list is not meant to be exhaustive - we update it as and when. The important part is the Pi400 type code which has been updated. If you want to identify a board, you should NOT be using the entire revision code (as it include build location which is irrelevant to board identification and can change as locations change), but extracting the correct bits - e,.g. the type.

JamesH65:
Closing, as revision code will get added over time, but the important stuff (type) has already been updated.

Issue #1725, opened 2020-11-07
please add the revision codes of CM4 module


dogi:
.. in https://github.com/raspberrypi/documentation/blob/master/hardware/raspberrypi/revision-codes/README.md

JamesH65:
Will get added as and when - the important stuff has already been added.

PR #1886, opened 2021-04-22
Add missing revision codes for CM4

cr-chsn1:
Added missing revision codes for CM4.


lurch:
It's entirely @JamesH65 's call, but see also #1725 #1726 and #1760 :shrug:

aallan:
@JamesH65 any thoughts? Due to #1911 if it's not merged in the next week or two this PR will be closed.

PR #1894, opened 2021-05-03
Update datasheets.md

samuelt619:
Changed [Line 7]:

"This datasheet is also applicable to the BCM2836 and BC2837 SoCs [...]"

to

"This datasheet is also applicable to the BCM2836 and BCM2837 SoCs [...]"


lurch:
Thanks :+1:

Issue #1893, opened 2021-05-03
DSI Touchscreen pages still reference ft5406


6by9:
https://github.com/raspberrypi/documentation/blob/master/hardware/display/troubleshooting.md references grepping for ft5406 to confirm the touchscreen is working.

The module was renamed to raspberrypi-ts when it was upstreamed in 2018 as part of the 5.0 release.

andrum99:
On latest Raspberry Pi OS with kernel 5.4.35-v7l+ there's no such module. I do however have one named `/lib/modules/5.4.35-v7l+/kernel/drivers/input/touchscreen/edt-ft5x06.ko`.

I've decompiled the rpi-ts5406.dtbo file and it provides no useful information as to which kernel module the screen uses. Perhaps someone in the known could confirm that the above is the correct module: I will then file a PR to update the docs.

6by9:
5.4.35 is not the latest kernel.

https://github.com/raspberrypi/firmware/tree/master/modules/5.10.33-v7l%2B/kernel/drivers/input/touchscreen lists 
raspberrypi-ts.ko

andrum99:
> 5.4.35 is not the latest kernel.
> 
> https://github.com/raspberrypi/firmware/tree/master/modules/5.10.33-v7l%2B/kernel/drivers/input/touchscreen lists
> raspberrypi-ts.ko

Apologies - I was trying to verify and chose the wrong version from the directory listing. I've corrected the PR per your original statement.

aallan:
Closed associated PR, so closing the issue.

PR #1896, opened 2021-05-04
rpi touch display: fix incorrect kernel module in troubleshooting page

andrum99:
See #1893

### DIFF
--- a/hardware/display/troubleshooting.md
+++ b/hardware/display/troubleshooting.md
@@ -31,7 +31,7 @@ sudo reboot
 If you want to make sure your Pi has detected your touchscreen, try running:
 
 ```bash
-dmesg | grep -i edt-ft5x06
+dmesg | grep -i raspberrypi-ts
 ```
 
 You should see a couple of lines that look like this:
--- a/hardware/display/troubleshooting.md
+++ b/hardware/display/troubleshooting.md
@@ -31,7 +31,7 @@ sudo reboot
 If you want to make sure your Pi has detected your touchscreen, try running:
 
 ```bash
-dmesg | grep -i ft5406
+dmesg | grep -i edt-ft5x06
 ```
 
 You should see a couple of lines that look like this:


lurch:
ping @6by9

6by9:
Nope.
edt-ft5x06 is for use with full KMS where the kernel will be directly polling the panel, not for when doing it via the firmware.

andrum99:
Fixed per https://github.com/raspberrypi/documentation/issues/1893#issuecomment-831839010

6by9:
I don't have a touchscreen connected at present, but the supposed result from that grep being 
```
[ 5.224267] rpi-ft5406 rpi_ft5406: Probing device
[ 5.225960] input: FT5406 memory based driver as /devices/virtual/input/input3
``` 
looks incorrect too.

"FT5406 memory based" would be coming from https://github.com/raspberrypi/linux/blob/rpi-4.19.y/drivers/input/touchscreen/rpi-ft5406.c#L253
whilst raspberrypi-ts in 5.4 and 5.10 set it to "raspberrypi-ts" https://github.com/raspberrypi/linux/blob/rpi-5.4.y/drivers/input/touchscreen/raspberrypi-ts.c#L181

The "Probing device" line doesn't exist either with raspberrypi-ts (https://github.com/raspberrypi/linux/blob/rpi-4.19.y/drivers/input/touchscreen/rpi-ft5406.c#L174 for rpi-ft5406)

Sorry, but this needs to be checked against a system with a panel actually connected.

aallan:
In light of #1911 is this PR still relevant?

andrum99:
> In light of #1911 is this PR still relevant?

Per 6by9:

> Sorry, but this needs to be checked against a system with a panel actually connected.

Which I don't have. If this particular page is going to survive the upcoming reorg of the docs then this PR is worth keeping, otherwise feel free to close.

(If you're keeping this page, I would nominate 6by9 to investigate, since he raised the original issue - #1893 😉).

aallan:

PR #1897, opened 2021-05-05
cmio-display updates

6by9:
Tweaks based on questions asked at https://www.raspberrypi.org/forums/viewtopic.php?f=98&t=311167

### DIFF
--- a/hardware/computemodule/cmio-display.md
+++ b/hardware/computemodule/cmio-display.md
@@ -95,6 +95,10 @@ DISPLAY_DSI_PORT
 The I2C port, SDA and SCL pin numbers are self explanatory. DISPLAY_DSI_PORT
 selects between DSI1 (the default) and DSI0.
 
+Once all the required changes have been made to the ```dts``` file, it needs to be compiled and placed on the boot partition of the device.
+
+Instructions for doing this can be found on the [Pin Configuration](../../configuration/pin-configuration.md) page.
+
 ### Sources
 - [dt-blob-disp1-only.dts](dt-blob-disp1-only.dts)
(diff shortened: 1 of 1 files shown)


hardware/computemodule/cmio-display.md
```diff
@@ -95,8 +95,12 @@ DISPLAY_DSI_PORT
 The I2C port, SDA and SCL pin numbers are self explanatory. DISPLAY_DSI_PORT
 selects between DSI1 (the default) and DSI0.
 
+Once all the required changes have been made to the ```dts``` file, it needs to be compiled and placed on the boot partition of the device.
+
+Instructions for doing this can be found on the [Pin Configuration](https://github.com/raspberrypi/documentation/blob/master/configuration/pin-configuration.md) page.
+
```
JamesH65, line 101: this needs to be a relative link so that it works both in github and on the docs site.

6by9: Doh. Sorry.

6by9:
Updated.

hardware/computemodule/cmio-display.md
```diff
@@ -95,8 +95,12 @@ DISPLAY_DSI_PORT
 The I2C port, SDA and SCL pin numbers are self explanatory. DISPLAY_DSI_PORT
 selects between DSI1 (the default) and DSI0.
 
+Once all the required changes have been made to the ```dts``` file, it needs to be compiled and placed on the boot partition of the device.
```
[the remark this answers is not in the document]
lurch: (I'll ignore this for now, as I see there are also several other places in this file where triple-backticks are used, when only single-backticks would be needed!)

PR #1901, opened 2021-05-13
Add disable_poe_fan to config.txt documentation

matthijskooijman:
This is a little low on details and mostly based on various forum posts
(I do not have a PoE to test with), but I did confirm that setting this
option indeed prevents a probe on I²C.

### DIFF
--- a/configuration/config-txt/boot.md
+++ b/configuration/config-txt/boot.md
@@ -113,7 +113,7 @@ On the Raspberry Pi 4B, if this value is set to `0` then the interrupts will be 
 
 ## force_eeprom_read
 
-Set this option to `0` to prevent the firmware from trying to read an I2C HAT EEPROM (connected to pins ID_SD & ID_SC) at powerup.
+Set this option to `0` to prevent the firmware from trying to read an I2C HAT EEPROM (connected to pins ID_SD & ID_SC) at powerup. See also [`disable_poe_fan`](misc.md#disable_poe_fan).
 
 ## os_prefix
 <a name="os_prefix"></a>
--- a/configuration/config-txt/misc.md
+++ b/configuration/config-txt/misc.md
@@ -22,3 +22,8 @@ For example, adding the line `include extraconfig.txt` to `config.txt` will incl
 ## max_usb_current
 
 **This command is now deprecated and has no effect.** Originally certain models of Raspberry Pi limited the USB ports to a maximum of 600mA. Setting `max_usb_current=1` changed this default to 1200mA. However, all firmware now has this flag set by default, so it is no longer necessary to use this option.
+
+## disable_poe_fan
+<a name="disable_poe_fan"></a>
+
+Set this option to `1` to prevent control of the PoE HAT fan through I2C (on pins ID_SD & ID_SC). Without this, a probe on the I2C bus will happen at startup, even when not using the PoE HAT.


configuration/config-txt/boot.md
```diff
@@ -113,7 +113,11 @@ On the Raspberry Pi 4B, if this value is set to `0` then the interrupts will be
 
 ## force_eeprom_read
 
-Set this option to `0` to prevent the firmware from trying to read an I2C HAT EEPROM (connected to pins ID_SD & ID_SC) at powerup.
+Set this option to `0` to prevent the firmware from trying to read an I2C HAT EEPROM (connected to pins ID_SD & ID_SC) at powerup. See also `disable_poe_fan`.
+
+## disable_poe_fan
+
+Set this option to `1` to prevent control of the PoE HAT fan through I²C on pions ID_SD & ID_SC). Without this, a probe on the I²C bus will happen at startup, even when not using the PoE HAT.
```
lurch, line 120: pions -> pins ? :wink:

lurch: @JamesH65 Should this use `I2C` (instead of `I²C`) to match the previous paragraph?

matthijskooijman: Fixed the typo, and I also changed to `I2C` (typing `I²C` is just a habit, but everywhere else uses I2C, so better be consistent). I forced pushed the fixes.

matthijskooijman:
I just realized that I added this to `boot.md` next to `force_eeprom_read`, since for me it was something I needed to make the i2c-0 bus silent at boot, but in reality this is of course not really a boot-related config directive, since AFAIU it primarily disables poe fan control at runtime and the fact that his also influences boot is secondary.

So, I suspect this should be in another file, but I'm not sure which exactly. None seem appropriate, so then I guess it should be in `misc.md`?

configuration/config-txt/boot.md
```diff
@@ -113,7 +113,11 @@ On the Raspberry Pi 4B, if this value is set to `0` then the interrupts will be
 
 ## force_eeprom_read
 
-Set this option to `0` to prevent the firmware from trying to read an I2C HAT EEPROM (connected to pins ID_SD & ID_SC) at powerup.
+Set this option to `0` to prevent the firmware from trying to read an I2C HAT EEPROM (connected to pins ID_SD & ID_SC) at powerup. See also `disable_poe_fan`.
+
+## disable_poe_fan
+
+Set this option to `1` to prevent control of the PoE HAT fan through I2C on pins ID_SD & ID_SC). Without this, a probe on the I2C bus will happen at startup, even when not using the PoE HAT.
```
lurch, line 120: Seems to be a stray `)` ?

matthijskooijman: Fixed, sorry for the noise... Seems I over-estimated the trivialness of this change and under-estimated the need for doing a proper review of my own commit... :-(

lurch:
> So, I suspect this should be in another file, but I'm not sure which exactly. None seem appropriate, so then I guess it should be in misc.md?

IMHO that's a decision for @JamesH65

aallan:

See #1911 for full details.

matthijskooijman:
> I'd encourage you to wrap this PR up in the next week or so

Thanks for pointing this out, but I'm waiting for an opinion on where best to put this bit of documentation, apparently from @JamesH65. But lacking a response, I'll go ahead and follow my own idea and just move the documentation to misc.md and see if someone wants to merge it like that :-) If another place turns out to be more appropriate, it can always be moved later (after asciidoc conversion).

matthijskooijman:
Ok, I moved the new documentation to `misc.md`, since it did not really belong in `boot.md` IMHO.

With that, I'm happy with the PR and consider it ready to merged. Not sure whom to ping for that, though :-)

configuration/config-txt/boot.md
```diff
@@ -113,7 +113,7 @@ On the Raspberry Pi 4B, if this value is set to `0` then the interrupts will be
 
 ## force_eeprom_read
 
-Set this option to `0` to prevent the firmware from trying to read an I2C HAT EEPROM (connected to pins ID_SD & ID_SC) at powerup.
+Set this option to `0` to prevent the firmware from trying to read an I2C HAT EEPROM (connected to pins ID_SD & ID_SC) at powerup. See also `disable_poe_fan`.
```
andrum99, line 116: You probably want this to be a hyperlink, in which case you would need to add an anchor before the heading for `disable_poe_fan` in misc.md. For example:

`<a name="disable-poe-fan"></a>
## disable_poe_fan`

You can then link to using Markdown syntax, e.g.

See also [`disable_poe_fan`](misc.md)

aallan:
> With that, I'm happy with the PR and consider it ready to merged. Not sure whom to ping for that, though :-)

Me in the final instance. :)

configuration/config-txt/boot.md
matthijskooijman: Thanks for the suggestion. I considered this, but could quickly only see examples of links *within* a page, not across pages. However, looking more closely, it seems links across pages are also used, so I followed those examples and amended my commit with a link added. Seems to work in the github rendering of the markdown, so I think it's good now :-)

matthijskooijman:
@aallan I made one more change to this PR suggested by @andrum99, should be good to merge now.

PR #1904, opened 2021-05-17
Update security.md

thecolorjay:


configuration/security.md
```diff
@@ -75,21 +75,20 @@ sudo deluser -remove-home pi
...
-```bash
-sudo visudo /etc/sudoers.d/010_pi-nopasswd
-```
-
-and change the `pi` entry (or whichever usernames have superuser rights) to:
+To force `sudo` to require a password, use:
 
 ```bash
-pi ALL=(ALL) PASSWD: ALL
+sudo sed -i '/pi/alice/g' /etc/sudoers.d/010_pi-nopasswd
```
lurch, line 91: IIRC it's always recommended to use the `visudo` command for editing the sudoers file, as it guards against accidental mistakes?

configuration/security.md
```diff
@@ -182,28 +187,36 @@ To disable the firewall, and disable start up on boot, use:
 sudo ufw disable
 ```
 
-Allow a particular port to have access (we have used port 22 in our example):
+What you'll want to do is deny all incoming requests by default, and allow all outgoing requests by default. This will allow you to allow only specific IP addresses to connect in on a specific port, and allow any and all IP addresses to connect out on any and all ports.
```
lurch, line 190: Perhaps "will let you allow" reads better than "will allow you to allow" ? :wink:

configuration/security.md
```diff
@@ -182,28 +187,36 @@ To disable the firewall, and disable start up on boot, use:
...
-sudo ufw allow 22
+sudo ufw default allow outgoing
 ```
 
-Denying access on a port is also very simple (again, we have used port 22 as an example):
+To disable logging, use:
 
 ```bash
-sudo ufw deny 22
+sudo ufw logging off
```
lurch, line 207: Perhaps this PR would be more useful if it simply added to the existing content, rather than replacing existing content? :confused:

configuration/security.md
thecolorjay: I think you're able to change it back to what it was, but there's no mistake.

configuration/security.md
thecolorjay: Yeah I'll change it.

configuration/security.md
thecolorjay: Actually I rearranged it, it's still there but it's been changed and it's on a new line (244-246).

PR #1909, opened 2021-05-26
Add schematic for original boards

mhjacobson:
* README.md: add entry for original Model B/A schematic

* Raspberry-Pi-Schematics-R1.0.pdf: add; copied from
<https://www.raspberrypi.org/app/uploads/2012/04/Raspberry-Pi-Schematics-R1.0.pdf>


hardware/raspberrypi/schematics/README.md
```diff
@@ -30,6 +30,10 @@ Schematics for the various Raspberry Pi board versions:
 
 - [Revision 1.1](rpi_SCH_1aplus_1p1_reduced.pdf)
 
+## Raspberry Pi Model B/A
+
+- [Revision 1.0](Raspberry-Pi-Schematics-R1.0.pdf)
```
mhjacobson, line 35: I left the filename as it was in the original source, even though it doesn't match all the others, since I figured that makes its provenance clearer.

lurch:
I didn't realise that this wasn't already in the docs?!

Hmmm, did a bit of digging and it turns out this _did_ used to be in the docs (so I wasn't misremembering), but it got removed (along with a couple of others) in https://github.com/raspberrypi/documentation/pull/642
So this is a decision for @JamesH65 or @aallan

Issue #1911, opened 2021-06-07
IMPORTANT: DOCUMENTATION REPO UPDATE


aallan:
All,

We're in the process of transitioning the documentation from the current Markdown-based source format to Asciidoc. This will help us automate the creation of a new and improved documentation site which will fit with the newer website styling and have much improved navigation.

In the process, we will also be addressing structural issues and reducing the overall amount of documentation (e.g. see #1810) to reduce the overall load in maintaining the documentation site. The original documentation was created when there was a lot less third-party content around to support the Raspberry Pi, and we're going to be removing a fair chunk of the existing text that really isn't relevant anymore, and better dealt with elsewhere on the web. The core documentation will remain, but a great deal of non-Raspberry Pi specific documentation will be removed. 

As much as possible we'll also be removing references to third-party products, and the policy going forward is that we won't be recommending specific tools other than things that ship with Raspberry Pi OS.

At some point soon — probably around the end of June, beginning of July — we will freeze the current documentation repo. After that time contributions and PRs based on the Markdown source will not be accepted, and any PRs that are still open will be closed.

There will then be some rearranging behind the scenes and afterwards the current Markdown-based documentation repo will be replaced with an Asciidoc-based equivalent. We're intending to transfer (relevant) Issues from this repo to the new repo, but all PRs will be lost.

The freeze while we sort things out behind the scenes should hopefully not be more than a couple of weeks. Contributions based on the new Asciidoc-based source will be welcomed after the transition is made public. However one priority going forward will be to keep this documentation site much leaner than its current form, so large additions, or new documentation topics, will face additional scrutiny.

andrum99:
Will the source to the Pico / RP2040 documentation webpages be included in the new documentation source repo? Just wondering as that also lives under raspberrypi.org/documentation.

aallan:
> Will the source to the Pico / RP2040 documentation webpages be included in the new documentation source repo? Just wondering as that also lives under raspberrypi.org/documentation.

The Pico/RP2040 HTML documentation is currently generated from YAML source. For now, it'll stay as it is. However in the longer run, yes, I'm looking to migrate those pages to Asciidoc and integrate with the mainline documentation. Probably expect that around Q3 or Q4 this year.

aallan:
Just a quick progress report for anyone following along. We've now merged or closed about two-thirds of the outstanding PRs against this repo. I'd expect to merge or close all of the remaining PRs before we freeze the repo for contributions.

[...]

PR #1912, opened 2021-06-08
Removed content that will be unsupported going forward.

aallan:


aallan:
The start of the great cull ahead of #1911

JamesH65:
LGTM

aallan:
💥

PR #1913, opened 2021-06-08
Delete the FAQ

aallan:
We'll salvage some of the content from the FAQ for use elsewhere, but the FAQ is going away post-#1911.


JamesH65:
Why is the FAQ going away? Always seemed to have useful information in it.

aallan:
> Why is the FAQ going away? Always seemed to have useful information in it.

Some of the information in it will get reused elsewhere in the documentation.

lurch:
Will the FAQ link in the website-footer still redirect to somewhere appropriate?

aallan:
I might leave an abbreviated FAQ in place for that. Or just get a ticket to get the link removed.

aallan:
Superceeded by internal tooling.

JamesH65:
One section that will definitely need to be moved somewhere new is the power consumption stuff - I don't think that is anywhere else in the docs, and it's a very common question.

aallan:
> One section that will definitely need to be moved somewhere new is the power consumption stuff - I don't think that is anywhere else in the docs, and it's a very common question.

Thanks! I'd spotted that one too. I've got a post-it note of sections I'm going to snag stuck to my monitor. If you think of any others that aren't duplicated elsewhere let me know.

lurch:
image

PR #1916, opened 2021-06-18
MISO / MOSI -->  SDO / SDI

erffrfez:
to match the changed definitions as at https://www.oshwa.org/a-resolution-to-redefine-spi-signal-names/

I'm not actually sure if the combination COPI / CIPO might actually be more appropriate than the SDO / SDI that I used


aallan:
We'll probably be looking at this sort of thing as part of #1911.

erffrfez:
Hi
I think that the change to the newer publishing system has been
implemented, but
https://www.raspberrypi.org/documentation/computers/raspberry-pi.html#spi-hardware
still has reference to MISO and MOSI,and Master and Slave

is that correct please, and what are the plans to update the text now?

Issue #1917, opened 2021-06-19
Raspicam Time-lapse tool no longer available


Charlesworth:
In the "Stitching images together" [section](https://www.raspberrypi.org/documentation/usage/camera/raspicam/timelapse.md), `libav-tools` is no longer install-able on Raspberry Pi OS:

```
Package libav-tools is not available, but is referred to by another package.
This may mean that the package is missing, has been obsoleted, or
is only available from another source
However the following packages replace it:
  ffmpeg

E: Package 'libav-tools' has no installation candidate
```

I have re-written this section instead using `ffmpeg`, which further simplifies the section as the same tool can be used both on the Raspberry Pi and "On another Linux computer". I'll attach the relevant PR with the changes. Thanks!

Charlesworth:
Relevant PR: #1918

PR #1918, opened 2021-06-19
Update camera timelapse video creation

Charlesworth:
Re issue #1917 update the time-lapse documentation to use `ffmpeg` to replace `libav-tools`, which is no longer install-able on Raspberry Pi OS:

```
Package libav-tools is not available, but is referred to by another package.
This may mean that the package is missing, has been obsoleted, or
is only available from another source
However the following packages replace it:
  ffmpeg

E: Package 'libav-tools' has no installation candidate
```

Tested on Raspberry Pi OS Lite Kernel version: 5.10

### DIFF
--- a/usage/camera/raspicam/timelapse.md
+++ b/usage/camera/raspicam/timelapse.md
@@ -5,9 +5,11 @@ To create a time-lapse video, you simply configure the Raspberry Pi to take a pi
 ## Using raspistill's inbuilt time-lapse mode
 
 The `raspistill` application has a built in time-lapse mode, using the `--timelapse` (or `-tl`) command line switch. The value that follows the switch is the time between shots in milliseconds:
+
 ```
 raspistill -t 30000 -tl 2000 -o image%04d.jpg
 ```
+
 Note the `%04d` in the output filename: this indicates the point in the filename where you want a frame count number to appear. So, for example, the command above will produce a capture every two seconds (2000ms), over a total period of 30 seconds (30000ms), named image0001.jpg, image0002.jpg, and so on, through to image0015.jpg.
 
 The `%04d` indicates a four-digit number, with leading zeros added to make up the required number of digits. So, for example, `%08d` would result in an eight-digit number. You can miss out the `0` if you don't want leading zeros.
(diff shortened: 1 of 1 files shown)


aallan:
LGTM. @JamesH65 ?

JamesH65:
My only thought was that the glob stuff may need more explanation - most users will not understand the meaning.

usage/camera/raspicam/timelapse.md
```diff
@@ -38,45 +40,30 @@ Make sure that you use e.g. `%04d` to make `raspistill` output each image to a n
...
+First you will need to install `ffmpeg`, you may also need to update and upgrade your Pi to successfully install.
 
-Although it will be slow (due to encoding in software rather than using the Raspberry Pi hardware acceleration), you can stitch your JPEG images together using various available tools. This documentation will use `avconv`, which needs to be installed.
-```
-sudo apt install libav-tools
 ```
-Now you can use the tools to convert your JPEG files in to an H264 video file:
+sudo apt update
+sudo apt full-upgrade
+sudo apt install ffmpeg
```
lurch, line 50: @XECDesign Should we be recommending users to do a `sudo apt full-upgrade` before doing `sudo apt install ffmpeg` ?

usage/camera/raspicam/timelapse.md
```diff
@@ -38,45 +40,30 @@ Make sure that you use e.g. `%04d` to make `raspistill` output each image to a n
...
+On a Raspberry Pi 3, this can encode a little more than two frames per second. The performance of other Pi models will vary. The parameters used are:
 
-```
-mencoder -nosound -ovc lavc -lavcopts vcodec=mpeg4:aspect=16/9:vbitrate=8000000 -vf scale=1920:1080 -o timelapse.avi -mf type=jpeg:fps=24 mf://@stills.txt
-```
+- `-r 10` Set frame rate (Hz value) to ten frames per second in the output video.
+- `-f image2` Set ffmpeg to read from a list of image files specified by a pattern.
+- `-pattern_type glob` Select the pattern type used to interpret the provided filenames.
+- `-i 'image_*.jpg'` The input file specification (to match the files produced during the capture).
+- `-s 1280x720` Scale to 720p. You can also use 1920:1080, or lower resolutions, depending on your requirements. Note the Pi can only play back up to 1080p video, but if you are intending to play back at, for example, 4K, you could set that here.
```
lurch, line 65: I suspect the `1920:1080` ought to be `1920x1080` to match the syntax of `1280x720` ?
Also, I believe "the Pi can only play back up to 1080p video" is no longer true for the Pi4?

usage/camera/raspicam/timelapse.md
Charlesworth: I removed this bit. I initially tested with Raspberry Pi lite which required this, testing with a fresh copy of Raspberry Pi OS desktop, `ffmpeg` was actually already installed so update and upgrade was not required

Charlesworth:
Thanks for the feedback everyone. I added some more explanation on `-pattern_type glob`, corrected the scaling flag notes and updated the installation instructions to remove the update and upgrade.

aallan:
Ready to merge @JamesH65 @lurch ?

JamesH65:
LGTM.

PR #1926, opened 2021-06-24
Drop the weird binaries

aallan:
Gordon said I could delete these!


aallan:
The windows installer now lives at https://github.com/raspberrypi/usbboot/raw/master/win32/rpiboot_setup.exe.

PR #1929, opened 2021-06-28
Update access-point-bridged.md

PegasusNZ:
br_filter blocks traffic on the bridge for DHCP
these changes remove the block

### DIFF
--- a/configuration/wireless/access-point-bridged.md
+++ b/configuration/wireless/access-point-bridged.md
@@ -174,6 +174,38 @@ To use the 5 GHz band, you can change the operations mode from `hw_mode=g` to `h
 
 Note that when changing the `hw_mode`, you may need to also change the `channel` - see [Wikipedia](https://en.wikipedia.org/wiki/List_of_WLAN_channels) for a list of allowed combinations.
 
+Check iptables are not filtering bridge traffic with br_netfilter
+
+
+```
+cat /proc/sys/net/bridge/bridge-nf-call-arptables 
+1
+cat /proc/sys/net/bridge/bridge-nf-call-iptables
+1
+cat /proc/sys/net/bridge/bridge-nf-call-ip6tables
+1
+```
+if they return 1 or true as above you can temporarily disable for testing by these commands:
+```
+sudo echo 0 | tee /proc/sys/net/bridge/bridge-nf-call-arptables
+0
+sudo echo 0 | tee /proc/sys/net/bridge/bridge-nf-call-iptables
+0
+sudo echo 0 | tee /proc/sys/net/bridge/bridge-nf-call-ip6tables
+0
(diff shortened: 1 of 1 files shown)


aallan:
> Not technically competent enough to know what this means, but it's conditional, so LGTM.

I was pretty much thinking the same.

configuration/wireless/access-point-bridged.md
```diff
@@ -174,6 +174,38 @@ To use the 5 GHz band, you can change the operations mode from `hw_mode=g` to `h
...
+cat /proc/sys/net/bridge/bridge-nf-call-arptables 
+1
+cat /proc/sys/net/bridge/bridge-nf-call-iptables
+1
+cat /proc/sys/net/bridge/bridge-nf-call-ip6tables
+1
+```
+if they return 1 or true as above you can temporarily disable for testing by these commands:
+```
+sudo echo 0 | tee /proc/sys/net/bridge/bridge-nf-call-arptables
```
Malvineous, line 190: This will run `echo` as root and then redirect the output to the file as a normal user, so it will likely fail with an access denied error.  It should be `echo 0 | sudo tee ...` instead.

configuration/wireless/access-point-bridged.md
```diff
@@ -174,6 +174,38 @@ To use the 5 GHz band, you can change the operations mode from `hw_mode=g` to `h
...
+0
+sudo echo 0 | tee /proc/sys/net/bridge/bridge-nf-call-iptables
+0
+sudo echo 0 | tee /proc/sys/net/bridge/bridge-nf-call-ip6tables
+0
+```
+or edit
```
Malvineous, line 197: Editing this file will make the change permanent, but you've said above that these commands are only temporary.  I think you should probably make it clear that changing `sysctl.conf` should only be done once the reader is happy to make the change persist across a reboot.

Malvineous (changes requested):
Can't confirm the `proc/` changes either but I spotted a couple of issues that should probably be fixed.

aallan:
I'm probably going to be freezing the current markdown-based documentation tree, and deleting any pending PRs, at the end of the week (see #1911 for more information as to why). So if we're going to merge this, probably need to finalise it before Wednesday or Thursday.

aallan:
I'm going to be freezing the current Markdown-based repo for PRs on Monday morning, see #1911 for details of what's going on. If this PR isn't merged before then it will unfortunately be deleted. Although we'll be happy to get a new PR against the new Asciidoc-based repo after that goes public — which will probably at the start of August.

aallan:
Closed due to #1911. Happy to have a new PR against the new Asciidoc-based repo after that goes public — which will probably at the start of August.

PR #1950, opened 2021-07-20
Menu label reference correction, update

inspector71:
v1.4 uses the label `Raspberry Pi 4 EEPROM boot recovery` rather than `Bootloader`.

### DIFF
--- a/hardware/raspberrypi/booteeprom.md
+++ b/hardware/raspberrypi/booteeprom.md
@@ -20,7 +20,7 @@ Raspberry Pi Imager provides a GUI for updating the bootloader and selecting the
 2. Select a spare SD card. The contents will get overwritten!
 3. Launch `Raspberry Pi Imager`
 4. Select `Misc utility images` under `Operating System`
-5. Select `Bootloader`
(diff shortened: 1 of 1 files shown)


aallan:
Apologies, but the documentation repo is currently frozen, see #1911 for details. We're not accepting pull requests at this time. If you want to resubmit we'll be happy to review a replacement PR against the new Asciidoc-based repo after that goes public — which will probably at the start of August.

lurch:
> v1.4 uses the label `Raspberry Pi 4 EEPROM boot recovery` rather than `Bootloader`.

Just for the sake of completeness, that text actually comes from either http://downloads.raspberrypi.org/os_list_imagingutility.json or http://downloads.raspberrypi.org/os_list_imagingutility_v2.json

@XECDesign Perhaps it makes sense to unify the text in those two files, where it makes sense?

XECDesign:
@ghollingworth Is it time to replace the first file with a single entry asking users to download the latest version yet? Seems like it has been long enough.

ghollingworth:
I agree, lets do that, people should move onto the latest version of rpi-imager!

Gordon

PR #1952, opened 2021-07-21
Update README.md

apal6981:
Took away comment on the config line code. When the comment is included on the same line after pasting the code verbatim the SPI bus was not loaded correctly.


lurch:
Apologies, but the documentation repo is currently frozen, see #1911 for details. We're not accepting pull requests at this time. If you want to resubmit we'll be happy to review a replacement PR against the new Asciidoc-based repo after that goes public — which will probably at the start of August.

PR #1957, opened 2021-07-22
Update docs to avoid rabbit-holing

jherico:
Some very basic unstated assumptions exist for this document.  While the steps were easy to follow on Ubuntu, none of it worked trying to enable a screen and when I'd find guides on debugging they all seemed to assume the presence of `raspi-gpio` and `vcdbg`, which are apparently unavailable for Aarch64.

### DIFF
--- a/hardware/computemodule/cmio-camera.md
+++ b/hardware/computemodule/cmio-camera.md
@@ -4,6 +4,8 @@
 
 These instructions are intended for advanced users, if anything is unclear please use the [Raspberry Pi Camera forums](https://www.raspberrypi.org/forums/viewforum.php?f=43) for technical help.
 
+These instructions are intended to users running the official 32-bit [Raspberry Pi OS](https://www.raspberrypi.org/software/).  When tested against a 64-bit Ubuntu 20 or 21 release, they did not function, and many of the tools that debugging guides or forum posts assume you have (`raspi-gpio`, `vcdbg`) are either not installed or completely unavailable.
+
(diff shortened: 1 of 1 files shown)


JamesH65:
Unfortunately this repo is currently not accepting PR's as we are moving to a completely new doc system.

With regard to this change, our documentation is specifically only for Raspberry Pi OS so I think its unnecessary. If we wanted to put in this warning in every place where someone else's distro did something different they would be absolutely everywhere. Whilst it is an unstated assumption that our documentation applies to our OS, I don't think its a bad assumption.

jherico:
The documentation has distinct sections for "Raspi OS" and "Linux", suggesting that not all the documentation is specific to the former, and this particular doc is under the "Hardware" section which offers no hint that it's specific to the distribution.

Googling around there are lots of dead-ends that hint that it might be possible to get the missing binaries onto an Ubuntu install via the addition of RasPi PPA, and there's even a repo under this org that has binaries, but in fact they're all 32 bit arm, and no 64 bit versions exist as far as I can tell.  I mean seriously.... I get that they're closed source, but Aarch64 is 10 years old and no one has recompiled this shit?

Also, if you're not going to accept PRs anymore because you're migrating, _archive the repository_, maybe even update the README.

Sorry if this comes across as Snarky, but I burned most of a day yesterday trying to get a screen to work with the CM4 dev board and failing because I had the audacity to run a 64 bit OS.

JamesH65:
The documentation is being completely revamped, but will continue to be hosted on this repo, so it has not been archived. 

If moving to 64bit was as easy as "recompiling this shit", we would have "recompiled this shit" years ago. But it's not, as has been written about many times on the forums.

PR #1944, opened 2021-07-12
Replace Chrome App with Chrome Extension

rickard-von-essen:
As announced publicly by the [Chromium blog](https://blog.chromium.org/2020/08/changes-to-chrome-app-support-timeline.html). Google is dropping support for Chrome Apps in June 2021. Please use the _Chromebook Recovery Utility extension_ instead.

PR #1958, opened 2021-07-24
Change Chromebook Recovery Utility Link

39p:
General support for Chrome Apps are over, and this app will not work anytime soon (see https://blog.chromium.org/2020/08/changes-to-chrome-app-support-timeline.html, the app also links to this blog post with a red text saying that chrome apps will be killed off). Fortunately, Google has made the same utility as an extension, so the updated link should work,


39p:
Nevermind, i didn't see the issue saying that the wiki is frozen

lurch:
Also suggested in #1944 :slightly_smiling_face:

39p:
I didnt see that as well, whoops.

PR #1960, opened 2021-08-09
Track page views with Google Analytics

mudge:
To keep continuity with the previous documentation site, add the Google Analytics Global Site Tag with the Raspberry PI property tracking ID.  Use Google's IP anonymization when recording any analytics (see https://support.google.com/analytics/answer/2763052?hl=en)


aallan (approved):
LGTM

PR #1964, opened 2021-08-09
Remove the FAQ link from the site footer

mudge:
As the new documentation site has no FAQ section, remove the now-redundant link from the site footer.


aallan (approved):
Bye Bye!

PR #1966, opened 2021-08-09
Fix page titles

lurch:


makefiles/shared.ninja
```diff
@@ -8,7 +8,7 @@ rule copy
   command = cp $in $out
 
 rule create_categories_page
-  command = echo "---\nlayout: boxes\n---\n" > $out
+  command = echo "---\nlayout: boxes\n---\n:doctitle: $title" > $out
```
mudge, line 11: Is this generating the [front matter](https://jekyllrb.com/docs/front-matter/) for the index page? If so, should the key here be `title` rather than `:doctitle` so that it can be used as `page.title`?

lurch: https://docs.asciidoctor.org/asciidoc/latest/document/title/
I tested this PR locally, and it correctly sets the title for `documentation/index.html`, `documentation/computers/index.html` & `documentation/computers/linux_kernel.html`

mudge (approved):

PR #1989, opened 2021-08-09
Update boot.adoc

timg236:
The Pi4 bootloader has supported start_db since at least 2018-11 so this note can be removed to avoid confusion.


aallan (approved):
LGTM

PR #1993, opened 2021-08-09
Update conditional.adoc

timg236:
Document the new board-type filter now that it is available in APT stable firmware.


aallan (approved):
LGTM

aallan:
Will go live to the site next time I push to production.

PR #1996, opened 2021-08-09
Update frequency-management.adoc

andrum99:
Grammar fixups

### DIFF
--- a/documentation/asciidoc/computers/raspberry-pi/frequency-management.adoc
+++ b/documentation/asciidoc/computers/raspberry-pi/frequency-management.adoc
@@ -22,7 +22,7 @@ Due to possible system stability problems involved with running an undervoltage,
 | property=value | Description
 
 | dvfs=1
-| allow under voltage
+| allow undervoltage
 
 | dvfs=2
 | fixed voltage for default operating frequencies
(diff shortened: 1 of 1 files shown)


lurch:
I'm not sure if the de-hyphenation is appropriate here? :shrug:

andrum99:
> I'm not sure if the de-hyphenation is appropriate here? 🤷

Hyphens are only meant to be used where:

- the words being hyphenated form a compound adjective, which is not the case here
- the words being hyphenated would otherwise be confused with another word which has a different meaning, e.g. co-op versus coop

There are some exceptions, but those don't apply in this case.  See https://blog.oup.com/2018/08/linguistics-where-to-put-hyphens/

aallan:
undervoltage should be all one word, rather than a compound.

andrum99:
> undervoltage should be all one word, rather than a compound.

I use OUP's free online service lexico.com to check spelling, and it doesn't have undervoltage as a word, although it does have overvoltage. This implies `under voltage` is correct in this case, where it is used as a compound noun.

aallan:
> I use OUP's free online service lexico.com to check spelling, and it doesn't have undervoltage as a word, although it does have overvoltage. This implies `under voltage` is correct in this case, where it is used as a compound noun.

It's in Merriam Webster as a word? https://www.merriam-webster.com/dictionary/undervoltage

andrum99:
> > I use OUP's free online service lexico.com to check spelling, and it doesn't have undervoltage as a word, although it does have overvoltage. This implies `under voltage` is correct in this case, where it is used as a compound noun.
> 
> It's in Merriam Webster as a word? https://www.merriam-webster.com/dictionary/undervoltage

Merriam Webster is an American dictionary.

aallan (approved):
LGTM

PR #2004, opened 2021-08-09
Update boot-eeprom-rpi4.adoc

andrum99:
- various grammar fixes.
- removed reference to the capacity of the bootloader EEPROM, since jamesh is on record (in one of the forums) as saying that there is no guarantee that the capacity of the EEPROM won't change. 
- clarified meaning of sentence which says when red HDMI screen will be shown to make it clear this happens when a bootloader EEPROM update fails.
- removed all occurrences of `SPI` from this page - it's an implementation detail and there is potential for confusion if people are also looking for information about configuring the SPI interfaces on the Pi 4.

I've also had to convert another table from markdown to asciidoc since it has not been done as part of the switch to asciidoc. It might be worth doing an automated scan of the repo for any more such tables, since they are unreadable once rendered into HTML by the current toolchain.

### DIFF
--- a/documentation/asciidoc/computers/raspberry-pi/boot-eeprom-rpi4.adoc
+++ b/documentation/asciidoc/computers/raspberry-pi/boot-eeprom-rpi4.adoc
@@ -1,6 +1,6 @@
 == Raspberry Pi 4 Boot EEPROM
 
-Raspberry Pi 4 and 400 computers use an EEPROM to boot the system. All other models of Raspberry Pi computer use the `bootcode.bin` file located in the boot filesystem.
+Raspberry Pi 4, 400 and Compute Module 4 computers use an EEPROM to boot the system. All other models of Raspberry Pi computer use the `bootcode.bin` file located in the boot filesystem.
 
 NOTE: The scripts and pre-compiled binaries used to create the `rpi-eeprom` package which is used to update the Raspberry Pi 4 bootloader and VLI USB controller EEPROMs is available https://github.com/raspberrypi/rpi-eeprom/[on Github].
 
--- a/documentation/asciidoc/computers/raspberry-pi/boot-eeprom-rpi4.adoc
+++ b/documentation/asciidoc/computers/raspberry-pi/boot-eeprom-rpi4.adoc
@@ -210,6 +210,7 @@ See also xref:raspberry-pi.adoc#raspberry-pi-4-boot-flow[Raspberry Pi 4 boot-flo
 | The VLI805 USB firmware EEPROM image - ignored on 1.4 board revision which does not have a dedicated VLI EEPROM
 
 | vl805.sig| The sha256 checksum of vl805.bin
+|===
 
 * If the bootloader update image is called `pieeprom.upd` then `recovery.bin` renames itself to `recovery.000` and resets the CPU. Since `recovery.bin` is no longer present the ROM loads the newly updated bootloader from SPI EEPROM and the OS is booted as normal.
 * If the bootloader update image is called `pieeprom.bin` then `recovery.bin` will stop after the update has completed. On success the HDMI output will be green and the green activity LED is flashed rapidly. If the update fails, the HDMI output will be red and an xref:configuration.adoc#led-warning-flash-codes[error code] will be displayed via the activity LED.
--- a/documentation/asciidoc/computers/raspberry-pi/boot-eeprom-rpi4.adoc
+++ b/documentation/asciidoc/computers/raspberry-pi/boot-eeprom-rpi4.adoc
@@ -212,7 +212,7 @@ See also xref:raspberry-pi.adoc#raspberry-pi-4-boot-flow[Raspberry Pi 4 boot-flo
 | vl805.sig| The sha256 checksum of vl805.bin
 |===
 
-* If the bootloader update image is called `pieeprom.upd` then `recovery.bin` renames itself to `recovery.000` and resets the CPU. Since `recovery.bin` is no longer present the ROM loads the newly updated bootloader from EEPROM and the OS is booted as normal.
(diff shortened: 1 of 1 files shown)


lurch:
> I've also had to convert another table from markdown to asciidoc since it has not been done as part of the switch to asciidoc. It might be worth doing an automated scan of the repo for any more such tables

Good idea. According to a couple of quick `git grep` commands, it looks like we've only got markdown-tables in:
 * documentation/asciidoc/computers/raspberry-pi/boot-eeprom-rpi4.adoc
 * documentation/asciidoc/computers/raspberry-pi/spi-bus-on-raspberry-pi.adoc

...both of which you've already found, hooray :tada: 

I can only assume that our conversion-tool missed these because those tables weren't marked up with "pipe triple-dash pipe", but a different number of dashes instead? :man_shrugging:

andrum99:
> > I've also had to convert another table from markdown to asciidoc since it has not been done as part of the switch to asciidoc. It might be worth doing an automated scan of the repo for any more such tables
> 
> Good idea. According to a couple of quick `git grep` commands, it looks like we've only got markdown-tables in:
> 
> * documentation/asciidoc/computers/raspberry-pi/boot-eeprom-rpi4.adoc
> * documentation/asciidoc/computers/raspberry-pi/spi-bus-on-raspberry-pi.adoc
> 
> ...both of which you've already found, hooray 🎉
> 
> I can only assume that our conversion-tool missed these because those tables weren't marked up with "pipe triple-dash pipe", but a different number of dashes instead? 🤷‍♂️

Markdown uses 3 or more dashes between pipes, so your conversion tool should probably be modified, assuming it's not one which was written specifically for the one off job conversion job on this repo which has already happened.

documentation/asciidoc/computers/raspberry-pi/boot-eeprom-rpi4.adoc
```diff
@@ -1,6 +1,6 @@
 == Raspberry Pi 4 Boot EEPROM
 
-The Raspberry Pi 4 has an SPI-attached EEPROM (4MBits/512KB), which contains code to boot up the system and replaces `bootcode.bin` previously found in the boot partition of the SD card. Note that if a `bootcode.bin` is present in the boot partition of the SD card in a Pi 4, it is ignored.
+Raspberry Pi 4 and 400 computers use an EEPROM to boot the system. All other models of Raspberry Pi computer use the `bootcode.bin` file located in the boot filesystem.
```
lurch, line 3: Didn't we once decide that "Raspberry Pi 4" was a suitable substitute for "Raspberry Pi 4 Model B, Raspberry Pi 400 and Compute Module 4" ? :shrug:  (but of course ultimately this is an @aallan decision)

lurch, line 3: And perhaps "use an EEPROM" should be "use code stored in an EEPROM" ?

andrum99: > And perhaps "use an EEPROM" should be "use code stored in an EEPROM" ?

More words for essentially the same meaning.

andrum99: > Didn't we once decide that "Raspberry Pi 4" was a suitable substitute for "Raspberry Pi 4 Model B, Raspberry Pi 400 and Compute Module 4" ? 🤷 (but of course ultimately this is an @aallan decision)

Yes, but in this case I'm trying not to get into a situation where someone with a Pi 400 thinks the Pi 4 stuff doesn't apply to their Pi 400: the next sentence says all `other Pi models...`. I could enumerate all the other Pi models, but I think it's better to specifically call out the two models that are the exceptions rather than list all of the other models which are not.

documentation/asciidoc/computers/raspberry-pi/boot-eeprom-rpi4.adoc
```diff
@@ -183,29 +183,43 @@ rpi-eeprom-config --config boot.conf --out new.bin pieeprom.bin
...
+| pieeprom.sig
+| The sha256 checksum of bootloader image (pieeprom.upd/pieeprom.bin)
+
+| vl805.bin
+| The VLI805 USB firmware EEPROM image - ignored on 1.4 and later board revisions which do not have a dedicated VLI EEPROM
+
+| vl805.sig| The sha256 checksum of vl805.bin
+|===
+
+* If the bootloader update image is called `pieeprom.upd` then `recovery.bin` renames itself to `recovery.000` and resets the CPU. Since `recovery.bin` is no longer present the ROM loads the newly updated bootloader from EEPROM and the OS is booted as normal.
```
lurch, line 215: "renames itself to `recovery.000`" -> "renames itself to `recovery.000` after the update has completed" ?

andrum99: I've reworded this. You may wish to have timg236 check this over to ensure the exact meaning of my various changes matches how things actually work.

documentation/asciidoc/computers/raspberry-pi/boot-eeprom-rpi4.adoc
lurch: CM4 also uses an EEPROM.

andrum99: > CM4 also uses an EEPROM.

I was deliberately not mentioning CM4 since traditionally the documentation does not specifically refer to compute modules in pages which are not specific to the compute module family of devices: upon reflection this is probably a Bad Thing &trade;

aallan: Maybe, but the Pi documentation is [over here](https://www.raspberrypi.org/documentation/computers/raspberry-pi.html), while the CM documentation is still [over there](https://www.raspberrypi.org/documentation/computers/compute-module.html), and will probably continue to be separated. Mainly due to the different audiences for the two products.

aallan (approved):
Merging!

PR #2010, opened 2021-08-10
Redirect old categories to new pages

mudge:
Now the structure of the documentation site has changed, set up redirects from the old structure's top-level categories to the most relevant page in the new structure.

Note that the first matching `RedirectMatch` takes precedence so the more specific redirect needs to come first (e.g.  `/documentation/hardware/tv-hat/` before `/documentation/hardware/`).


aallan (approved):
LGTM

PR #2012, opened 2021-08-10
Redirect compliance page to PIP

mudge:
While we now redirect https://www.raspberrypi.org/compliance to the Product Information Portal, we should also redirect the old "Product compliance and safety" page as it is directly linked in various places such as forum posts, etc.

As redirects are processed in order, we need the most specific matches to come first. As the redirects generated from the datasheets CSV are specific URLs, they should therefore appear first in our .htaccess with the "extra" configuration coming second (as this tends to be broader).

Without this, the redirects for the top-level documentation categories break the more specific redirects for PDFs that now live at https://datasheets.raspberrypi.org

### DIFF
--- a/documentation/htaccess_extra.txt
+++ b/documentation/htaccess_extra.txt
@@ -1,4 +1,5 @@
 <IfModule mod_alias.c>
+Redirect 302 "/documentation/hardware/raspberrypi/conformity.md" "https://pip.raspberrypi.org"
 RedirectMatch 302 "^/documentation/faqs" "/documentation/"
 RedirectMatch 302 "^/documentation/hardware/raspberrypi/compliance/" "https://pip.raspberrypi.org"
 RedirectMatch 302 "^/documentation/hardware/camera/" "/documentation/accessories/camera.html"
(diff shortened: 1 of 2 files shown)


aallan (approved):
LGTM

aallan:
I've merged to develop, but will hold off pushing to production till @lurch says okay.

lurch:
Will this appending behaviour _always_ be suitable, or is it worth having e.g. `htaccess_extra_prepend.txt` and `htaccess_extra_append.txt` ? :thinking: :man_shrugging:

lurch:
LGTM (I can't explicitly approve, as the PR has already been merged).

aallan:
I'll push to production now.

lurch:
Alternatively, perhaps it's worth removing the `create_htaccess.py` entirely and just have a static `.htaccess` file? Up to you @mudge :wink:

mudge:
> Will this appending behaviour _always_ be suitable, or is it worth having e.g. `htaccess_extra_prepend.txt` and `htaccess_extra_append.txt` ? 🤔 🤷‍♂️

As long as the redirects generated from `datasheets.csv` are always for full paths, they should always come first so I think it is safe to always _append_ the extra content.

If we end up needing to prepend configuration for some reason, I'd be tempted to simplify and go back to a static `.htaccess` file and separate out the useful datasheet link checking logic that exists in the current toolchain.

PR #2015, opened 2021-08-10
Add redirects for Pi 4 mechanical drawings

mudge:
Now the Raspberry Pi 4 mechanical drawings have moved to https://datasheets.raspberrypi.org, add a redirect from their old URLs to the new location.


aallan (approved):
LGTM

PR #2016, opened 2021-08-10
Fixes #1846 ?

### LINKED ISSUE #1846: computers/raspberry-pi/boot-msd.adoc issue
In: 
documentation/hardware/raspberrypi/bootmodes/msd.md

and on: 
https://www.raspberrypi.org/documentation/hardware/raspberrypi/bootmodes/msd.md

it says: 
Check if your Pi 4B has the required bootloader EEPROM version
To check if your Pi 4B has the required bootloader EEPROM version, power it up with no SD card inserted and a display attached to one of the HDMI ports. The Pi 4B will display a diagnostic screen on the attached display, which includes the bootloader EEPROM version at the top of the screen. The bootloader must be dated Sep 3 2020 or later to support USB mass storage boot. If the diagnostic screen reports a date earlier than Sep 3 2020, or there is no diagnostic screen shown, you will need to update the bootloader EEPROM first to enable USB mass storage boot.

USB mass storage boot on the Pi 4B requires Raspberry Pi OS 2020-08-20 or later.

so it would seem to say that if you have this Sep 3 2020 boot loader and a 2021 version of the Raspberry Pi OS, you should be able to boot from the USB mass storage. 

later in these instructions it says: 
After preparing the storage device, connect the drive to the Raspberry Pi and power up the Pi, being aware of the extra USB power requirements of the external drive. After five to ten seconds, the Raspberry Pi should begin booting and show the rainbow splash screen on an attached display. Make sure that you do not have an SD card inserted in the Pi, since if you do, it will boot from that first.

See the bootmodes documentation for the boot sequence and alternative boot modes (network, USB device, GPIO or SD boot).

... but when I have the correct software on my Raspberry Pi 4 8gb unit and boot from the USB device with no SD card mounted it just says install an SD card. 

I had to go into the raspberry-config stuff and set the bootloader order to boot from the USB mounted device before I could get the Pi to boot from the USB mounted device. 

The instructions seem to say that the boot order is optional... but the USB drive was not used at all until I changed the boot order. 

I am not sure why I had to mess with the boot order... no SD card was installed... a USB Drive was... and it did not try and use the USB drive untill I booted from the SD Card, changed the boot order and then booted with no SD card attached.

lurch:


aallan:
> Oh, #1846 was closed while I was making this, so perhaps this should be closed too?

I'll go ahead and merge. :)

PR #2018, opened 2021-08-10
Add redirects for moved PDFs

mudge:
Now that mechanical drawings, schematics and other documents have moved to https://datasheets.raspberrypi.org, add more redirects from their previous location to the new one.


aallan (approved):
LGTM

PR #2020, opened 2021-08-10
Don't run build twice for pull requests

mudge:
To prevent the GitHub Actions build running twice for pull requests raised by contributors with write access to this repository, update our configuration to only run the build for pull requests or pushes to the master branch specifically. This way, pushes to other branches (e.g.  feature branches) will not trigger a build.


aallan (approved):
I trust you!

PR #2024, opened 2021-08-10
Footer accessibility changes

capjamesg:
Add aria label to newsletter form for accessibility and add type="hidden" class to VerificationCodeX item so it is not narrated as a form to screen reader users.

These changes have been made in response to suggestions made by the WAVE Accessibility testing tool, specifically:

```
Missing form label

A form control does not have a corresponding label.
```

### DIFF
--- a/jekyll-assets/_includes/footer.html
+++ b/jekyll-assets/_includes/footer.html
@@ -63,11 +63,11 @@
                     <span class="c-footer-newsletter__legend">
                         Sign up to our newsletter
                     </span>
-                    <input class="c-footer-newsletter__input" name="email" placeholder="Your email here" required="" type="email">
+                    <input class="c-footer-newsletter__input" name="email" placeholder="Your email here" required="" type="email" aria-label="Enter your email address">
                     <input type="hidden" name="htmlemail" value="1">
                     <input type="hidden" name="list[3]" value="signup">
                     <input type="hidden" name="listname[3]" value="Raspberry Pi Weekly">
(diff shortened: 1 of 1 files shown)


aallan:
@mudge Is this part of the Foundation owned footer stuff?

mudge:
> @mudge Is this part of the Foundation owned footer stuff?

Yes, the footer is taken from https://www.raspberrypi.org

Re the `VerificationCodeX` input: I suspect this is a honeypot for our [phpList](https://www.phplist.com) subscription form and adding `type="hidden"` (note there is also a `type="text"` attribute on the `<input>` too) might affect that. Perhaps [`aria-hidden`](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/ARIA_Techniques/Using_the_aria-hidden_attribute) would be more appropriate?

aallan:
> Yes, the footer is taken from https://www.raspberrypi.org

Ah. In which case it's beyond our control. We can pass your comments on to the Foundation. But we can't fix this, sorry!

capjamesg:
No worries! It would be great if you passed my comments along. There is also an issue of colour contrast in the footer privacy / cookies / brand guidelines to note while I'm here :)

aallan:
> No worries! It would be great if you passed my comments along. There is also an issue of colour contrast in the footer privacy / cookies / brand guidelines to note while I'm here :)

Yup, saw that. That we can look at. 😄

PR #2027, opened 2021-08-10
Add some more redirects (and fix some broken ones)

lurch:
I picked an arbitrary "[old commit](https://github.com/raspberrypi/documentation/tree/c7475c40099af57ae4400e20e0286d34a05e807b)" from before we started moving files about, as I guess these "oldest" links are the ones that are most likely to be broken. And then I added redirects (that we didn't already have) for all the non-markdown and non-image files that existed in the repo at that time.


aallan (approved):
More of them!? 🤦‍♂️

lurch:
Hopefully we've got them all now! :wink: :crossed_fingers: 
I did a
```bash
find -type f -print|grep -vE '\.(md|png|jpg|gif|git)' | grep -v '/compliance/' | sort > files.txt
```
from a local checkout of the commit mentioned above.

PR #2030, opened 2021-08-11
Move under voltage detection away from CM4

dp111:


aallan (approved):
I guess? I was never sure why that was there in the first place?

PR #2040, opened 2021-08-11
Typo

dp111:


aallan (approved):
Would have sworn it was a 'c' not an 's' and the 's' still doesn't look right. ¯\_(ツ)_/¯

PR #2049, opened 2021-08-12
Replace icons with full sized versions

aallan:
Now the CSS is fixed up we can replace the smaller category icons with their full sized versions.


aallan:
Merging YOLO

PR #2062, opened 2021-08-12
Fix syntax typo

lurch:
And reorder things to match reality


aallan (approved):
This was copy and pasted directly from https://datasheets.raspberrypi.org/keyboard-mouse/getting-started-with-keyboard-and-mouse.pdf. So the order of things comes from there. 🤷‍♂️ 

That said the typo is all me. 😆

lurch:
Hmmm, does this mean that a "PadLock" locks the amount of padding displayed? :lock: :laughing:

PR #2064, opened 2021-08-13
Adjusting alignment of copyright text

nelliemckesson:
This adjusts the alignment of the copyright text. It's a bit tricky, since the text itself needs to be left-aligned, and isn't actually long enough to fill it's containing box. I went with a "least offensive" max-width that still looks slightly wider on the box page:

Screen Shot 2021-08-12 at 6 16 33 PM

but slightly shorter on the actual documentation pages:

Screen Shot 2021-08-12 at 6 17 14 PM

(also figured out I can emulate larger screens with Chrome tools 😎 )

@aallan take a look and let me know if you can think of something better!

### DIFF
--- a/jekyll-assets/css/style.css
+++ b/jekyll-assets/css/style.css
@@ -409,10 +409,11 @@ div.videoblock iframe {
 }
 
 div.legal {
-  margin: 80px 10% 0 10%;
-  max-width: 1300px;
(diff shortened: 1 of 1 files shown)


aallan (approved):

aallan:
Yup, that works! Thanks!

PR #2073, opened 2021-08-13
Add instructions for building locally on 64bit

ghollingworth:
I've done this once, but do need to just check that the instructions are correct, unless someone else gets there first...

### DIFF
--- a/documentation/asciidoc/computers/linux_kernel/building.adoc
+++ b/documentation/asciidoc/computers/linux_kernel/building.adoc
@@ -86,6 +86,7 @@ For Raspberry Pi 4 default build configuration (64 bit)
 cd linux
 KERNEL=kernel8
 make bcm2711_defconfig
+----
 
 ===== Customising the Kernel version using `LOCALVERSION`
 
@@ -101,7 +102,7 @@ You can also change that setting graphically as shown in xref:linux_kernel.adoc#
 
 ==== Building the Kernel
 
-Build and install the kernel, modules, and Device Tree blobs; this step can take a *long* time depending on the Pi model in use, for the 32 bit kernel:
+Build and install the kernel, modules, and Device Tree blobs; this step can take a *long* time depending on the Pi model in use.  For the 32-bit kernel:
 
 [,bash]
 ----
@@ -113,7 +114,7 @@ sudo cp arch/arm/boot/dts/overlays/README /boot/overlays/
 sudo cp arch/arm/boot/zImage /boot/$KERNEL.img
 ----
 
-For the 64 bit kernel:
+For the 64-bit kernel:
 
 [,bash]
 ----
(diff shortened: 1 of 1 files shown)


documentation/asciidoc/computers/linux_kernel/building.adoc
```diff
@@ -80,6 +80,13 @@ KERNEL=kernel7l
...
 ----
 
+For Raspberry Pi 4 default build configuration (64 bit)
+[,bash]
+----
+cd linux
+KERNEL=kernel8
+make bcm2711_defconfig
+
 ===== Customising the Kernel version using `LOCALVERSION`
```
andrum99, line 90: `Kernel` -> `kernel`

ghollingworth: I've not changed this because there are lots of headings here with capitalized Kernel.  @aallan should I remove all the mixed capitalization from headings (there's a mix-mash)

aallan: We generally follow the AP Style Guide, so we use [title case](https://en.wikipedia.org/wiki/Title_case) in section headings. Kernel **should** be capitalised in this instance.

aallan: …but so should "version" and "using".

aallan (approved):

aallan:
Merging to `develop`, will go to the live site when I push to `master`.

documentation/asciidoc/computers/linux_kernel/building.adoc
```diff
@@ -106,6 +114,18 @@ sudo cp arch/arm/boot/dts/overlays/README /boot/overlays/
...
+For the 64-bit kernel:
+
+[,bash]
+----
+make -j4 Image modules dtbs
+sudo make modules_install
+sudo cp arch/arm64/boot/dts/*.dtb /boot/
+sudo cp arch/arm64/boot/dts/overlays/*.dtb* /boot/overlays/
+sudo cp arch/arm64/boot/dts/overlays/README /boot/overlays/
+sudo cp arch/arm64/boot/zImage /boot/$KERNEL.img
```
lurch, line 126: Comparing this to the cross-compile instructions, should the `zImage` here be `Image` ?
(which would also seem to match up with the `make -j4 Image modules dtbs` line above?)
ping @ghollingworth

ghollingworth: Yes, I missed that one...  As I said I still could do with someone (me) going through the instructions again just to make sure my changes are valid.  Unfortunately I don't have a 64-bit version running at home at the moment and would need to install a new one...  May be able to do that tomorrow morning.

PR #2072, opened 2021-08-13
Add "Edit on GitHub" links to each page / section. Fixes #2029

lurch:
Quite a bit going on here... :sweat_smile: 
* I've added machinery to `.github/workflows/build.yml` for installing Python modules from `requirements.txt`
* I've added `requirements.txt` to install the [pyyaml](https://pyyaml.org/wiki/PyYAMLDocumentation) module
* ...so that we can make use of values in `_config.yml` from the python build-scripts
* ...which I've done in several places :slightly_smiling_face: 
* The way the "build" works has changed so that every `*.adoc` file now gets copied across to a subdirectory of `build/jekyll` (previously the files in `build/jekyll` just directly-included the adoc files straight from `documentation/asciidoc`)
* ...because the first header in each of the `build/jekyll/*.adoc` files now has an autogenerated "Edit this on GitHub" link added immediately underneath it :tada: 
* ...with the text in that link coming from `jekyll-assets/_includes/github_edit.adoc`

I'm going to mark this as draft because we want some @nelliemckesson styling applied to these new links, before this gets merged :slightly_smiling_face: 
Screenshot from 2021-08-13 16-34-35

PR #2092, opened 2021-08-18
Adding styles for the edit link

nelliemckesson:
See notes in #2072


aallan:
Merged into #2072 for style!

PR #2065, opened 2021-08-13
Add heading anchor links

nelliemckesson:
@aallan @lurch @jackbenwillis this is the first step to add those heading links as requested in #1991 . I'm basically copying how github does it. I still need to add the styles to hide/show on hover and get the alignment right etc, and also test it a bunch, but there are a few things I need from y'all, so I wanted to get the ball rolling on that.

* @lurch : Can you add `_plugins` to the list of dirs to copy during the make process?
* @jackbenwillis Can you give me an SVG icon to use for the link image? Here's what it looks like currently (completely stolen from github):

Screen Shot 2021-08-12 at 7 36 32 PM

PR #2093, opened 2021-08-18
Push to production

aallan:
Some big changes in this push;

- #2072 
- #2065 
- Some other minor stuff


jekyll-assets/css/style.css
```diff
@@ -361,6 +361,7 @@ h6 .anchor::before {
   font-style: italic;
   margin-top: -12px;
   margin-bottom: 20px;
+  margin-left: 5px;
```
lurch, line 364: Errr, that's "pushed 5 pixels to the right" :stuck_out_tongue_winking_eye:

aallan: Left, right, all the same. Yes. 5 pixels to the right. 🤦‍♂️

Issue #2098, opened 2021-08-18
Get GitHub CI to "fail" PRs against master


lurch:
Is there a way to configure the workflows thing so that if anybody PRs directly against `master` (as happened in #2022 ) we get a red cross instead of a green tick? We always want any commits to go into `develop` before going into `master`, so that the branches don't get out of sync (see https://github.com/raspberrypi/documentation/commit/b689e018867e7f27f97858ae3f58ae5b3f8dadc0 which was a cherry-pick and #2097 which then brought the branches even again).

PR #2100, opened 2021-08-18
Ensure PRs based on master can only come from the develop branch

mudge:
GitHub: https://github.com/raspberrypi/documentation/issues/2098

Add a check to the GitHub Actions workflow to check that all pull requests use "develop" as their base branch. If not, fail the workflow and prevent merging.

Note that we still need to allow pull requests to be raised between "develop" and "master" so we skip this check if the current branch is "develop".

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -10,6 +10,12 @@ jobs:
   build:
     runs-on: ubuntu-latest
     steps:
+      - name: Only allow pull requests based on master from the develop branch
+        if: ${{ github.base_ref == 'master' && github.ref != 'refs/heads/develop' }}
+        run: |
+          echo "Pull requests based on master can only come from the develop branch"
+          echo "Please check your base branch as it should be develop by default"
+          exit 1
       - uses: actions/checkout@v2
       - uses: actions/setup-python@v2
         with:


mudge:
Tested this by first raising the PR against `master` (it failed) and then updated the base to `develop` and force-pushed an amended commit to trigger a passing build.

Note this doesn't test that raising PRs between `develop` and `master` will work.

mudge:
Marking this as a draft as we'll need to ensure the build that runs when pushing to `master` succeeds (at the moment, I suspect the check will fail as neither the ref nor the base ref will be `develop`).

.github/workflows/build.yml
```diff
@@ -10,6 +10,11 @@ jobs:
   build:
     runs-on: ubuntu-latest
     steps:
+      - name: Ensure the base branch is correct
+        if: github.ref != 'refs/heads/develop' && github.base_ref != 'develop'
```
mudge, line 14: We need to ensure that builds that run on push to `master` (to perform a live deploy) skip this check.

aallan: Okay, so what's the steps to test this do you think?

lurch:
I think that "ensure that PRs are against the `develop` branch" isn't what we actually want; I think we only want "ensure that PRs are **not** against the `master` branch" ?
Because there's been a couple of multi-part PRs where I've done the initial PR against `develop`, and @nelliemckesson has created a PR against my branch. So I then merge Nellie's PR into my branch, and Alasdair then merges my branch into `develop`. See e.g. #2072 and #2092 
I think if we had "PRs not against `develop` automatically fail", then I wouldn't be able to merge Nellie's PR back up into my branch?

Perhaps this is going to be more complicated than I originally thought? :confused:

mudge:
@lurch I’m out until Friday now but that makes sense and being more specific should hopefully avoid surprising side-effects, e.g. we could update the `if` condition to `github.base_ref == 'master' && github.ref != 'refs/heads/develop'`. That should only affect pull requests where the base reference is “master” and fail if the branch is not “develop”.

I’m not sure what the base_ref would be for the typical “master” push but we could also add a predicate to only run that step on pull requests. The question of what happens if a fork uses the branch name “develop” is a pretty interesting one we could test out.

lurch:
Rather than hard-coding `github.repository == 'raspberrypi/documentation'`, I wonder if we could check that the repository being pushed **to** is the same as the repository that the branch is coming **from**?
The reason I ask is that I can also see these checks being useful for `pico-sdk` which uses a similar "PRs to `develop`, which eventually gets merged into `master`" strategy. (so obviously the less that needs to be edited when moving this logic between different repos, the better!)

lurch:
I've just tried doing a bit of reading up on this, and _perhaps_ what we want is:
```
      - name: Check PRs to master only come from develop
        if: github.base_ref == 'master' && (github.head_ref != 'develop' || github.repository_owner != 'raspberrypi')
        run: |
          echo "Only the develop branch is allowed to pull-request to master"
          exit 1
```
(which in hindsight, is very similar to what @mudge suggested above)

mudge:
Please free to push to this branch but having researched a bit more, I think using `base_ref` and `ref` alone should work:

* `base_ref` will be null for non-pull requests so the step will be skipped for pushes
* `ref` will be of the form `refs/heads/develop` for branches and `refs/pulls/123…` for pull requests. This hopefully means `refs/heads/develop` only ever means _our_ `develop` and not anyone else’s.

This means the condition should be:

```
if: github.base_ref == 'master' && github.ref != 'refs/heads/develop'
```

lurch:
I'm happy for you to do what you think is best @mudge (you understand this much better than me), when you get back from holiday :slightly_smiling_face:

mudge:
Update to only check that pull requests based on `master` come from our `develop` branch. This will allow merges between any other branches and should not affect the push to `master` we use to deploy.

Tests:

* [Failing build when trying to base a non-`develop` PR off `master`](https://github.com/raspberrypi/documentation/runs/3379955932?check_suite_focus=true)
* [Passing build when trying to base a non-`develop` PR off `develop`](https://github.com/raspberrypi/documentation/runs/3379963097?check_suite_focus=true)
* [Failing build when trying to base a PR off `master` when using a fork with the branch name `develop`](https://github.com/raspberrypi/documentation/runs/3380027720?check_suite_focus=true)

lurch (approved):
LGTM!

aallan:
> @aallan if you're happy, I'd suggest we merge this and then do a live deploy as soon as is convenient to check the workflow still works as expected between develop and master. In the worst case, we can drop this step from the workflow if it is causing problems.

Sorry, been out ill. It's now 5pm on a Friday, which doesn't sound like a good time to do it. Therefore suggest Monday, are you around then if it all goes sideways?

PR #2102, opened 2021-08-19
First attempt with new libcamera and libcamera-apps documentation - D…

davidplowman:
First attempt at some new libcamera documentation DO NOT MERGE!!

Obviously there's still loads to do before it's anywhere near merging. I've written a lot of new stuff, and it's all in sections so can be moved about easily enough. For now I've just included everything into the big camera.adoc page so that it can be viewed conveniently. Here's what I think we need to do:

1. There's still more stuff to write. A bit more about TFLite once the code is all done, and there will be a very very long section on Python Bindings once we have those.
2. Then we probably need to figure out exactly how to present it all. The libcamera stuff needs to be deleted from what was there previously (as I've now duplicated it in the new stuff), and where it talks generally about raspicam apps we need to add libcamera-apps too. But as things stand it's making for a very long page, so we probably want to think about that.
3. Finally there will be lots of improvements required to the formatting, the text, more pictures etc. etc.


documentation/asciidoc/accessories/camera/libcamera_3rd_party_tuning.adoc
```diff
@@ -0,0 +1,15 @@
+=== Camera Tuning and supporting 3rd Party Sensors
+
+==== The Camera Tuning File
+
+Most of the image procesing applied to frames from the sensor is done by the hardware ISP (Image Signal Procesor). This processing is governed by a set of _control algorithm_ and these in turn must have a wide range of parameters supplied to them. These parameters are tuned specifically for each sensor and are collected together in a JSON file known as the _camera tuning file_.
```
lurch, line 5: "a set of _control algorithm_" -> "a set of _control algorithms_" ?

documentation/asciidoc/accessories/camera/libcamera_3rd_party_tuning.adoc
```diff
@@ -0,0 +1,15 @@
...
+
+This _tuning file_ can be inspected and edited by users. Using the `--tuning-file` command line option, users can point the system at completely custom camera tuning files.
+
+==== 3rd Party Sensors
+
+`libcamera` makes it possible to support 3rd party sensors (that is, sensors other than Raspberry Pi's officially supported sensors) on the Raspberry Pi platform. To accomplish this, a working open source sensor driver must be provided, which the authors are happy to submit to the Linux kernel. Then are a couple of extra files need to be added to `libcamera` which supply device-specific information that is available from the kernel drivers, including the previously discussed camera tuning file.
```
lurch, line 11: "Then are a ..." ?

documentation/asciidoc/accessories/camera/libcamera_apps_intro.adoc
```diff
@@ -0,0 +1,34 @@
...
+* _libcamera-vid_ A video capture application.
+* _libcamera-raw_ A basic application for capturing raw (unprocessed Bayer) frames directly from the sensor.
+
+Raspberry Pi's `libcamera-apps` are not only command line applications that make it easy to capture images and video from the camera, they are also examples of how users can create their own libcamera-based applications with custom functionality to suit their own requirements. The source code for the `libcamera-apps` is freely available under a BSD 2-Clause licence at https://github.com/raspberrypi/libcamera-apps[].
+
+==== More about `libcamera`
+
+`libcamera` is an open source Linux community project. More information is available at the https://libcamera.org[`libcamera` website].
+
+The `libcamera` source code can be found and checked out https://git.linuxtv.org/libcamera.git/[here].
```
lurch, line 23: I think @aallan doesn't like the use of "here" as link-text.

documentation/asciidoc/accessories/camera/libcamera_apps_getting_started.adoc
```diff
@@ -0,0 +1,37 @@
...
+| `dtoverlay=ov5647`
+
+| V2 camera (IMX219)
+| `dtoverlay=imx219`
+
+| HQ camera (IMX477)
+| `dtoverlay=imx477`
+
+| IMX290
+| `dtoverlay=imx290,clock-frequency=74250000` or `dtoverlay=imx290,clock-frequency=37125000`
```
lurch, line 20: How would a user choose which of these two options to use?

documentation/asciidoc/accessories/camera/libcamera_apps_getting_started.adoc
```diff
@@ -0,0 +1,37 @@
...
+
+| IMX327
+| `dtoverlay=imx290,clock-frequency=74250000` or `dtoverlay=imx290,clock-frequency=37125000`
```
lurch, line 23: Was this a cut'n'paste typo, or does the IMX327 indeed use the exact same dtoverlay as IMX290?

documentation/asciidoc/accessories/camera/libcamera_apps_post_processing.adoc
```diff
@@ -0,0 +1,216 @@
...
+{
+    "sobel_cv":
+    {
+        "ksize": 5
+    },
+    "negate":
+    {
+    }
+}
+----
```
lurch, line 31: This is effectively a "dictionary" structure (to use the Python naming) and in most programming languages a dictionary is unordered, so if a script was to read in this JSON file and then write it out again, I think it woud be perfectly valid for it to get written out as:
```
{
    "negate":
    {
    },
    "sobel_cv":
    {
        "ksize": 5
    }
}
```
which conflicts somewhat with your "the stages will be run in the order given" :confused: 
In strict "JSON syntax" if you want to preserve ordering, I think you'd need to use an array `[ ... ]` rather than a dictionary `{ ... }`.
(And apologies if this isn't the right place to be discussing this!)

documentation/asciidoc/accessories/camera/libcamera_apps_post_processing.adoc
```diff
@@ -0,0 +1,216 @@
...
+
+Furthermore, with the post-processing framework being completely open, Raspberry Pi welcomes the contribution of new and interesting stages from the community and would be happy to host them in our `libcamera-apps` repository. The stages that are currently available are documented below.
+
+NOTE: the `libcamera-apps` supplied with the operating system will be built without any optional 3rd party libraries (such as OpenCV), meaning that certain post-processing stages that rely on them will not be enabled. To use these stages, please follow the instructions for xref:camera.adoc#building-libcamera-and-libcamera-apps[building `libcamera-apps` for yourself].
+
+==== `negate` stage
+
+The `negate` stage requires no 3rd party libraries.
+
+It may be more performant on Pi 3 and 32-bit OS Pi 4 devices if recompiled using `-DENABLE_COMPILE_FLAGS_FOR_TARGET=armv8-neon`. (Please see the xref:camera.adoc#building-libcamera-and-libcamera-apps[build instructions].)
```
lurch, line 49: "32-bit OS Pi 4 devices" sounds a bit weird IMHO.
@aallan can confirm, but I think the recommended phrasing would be something like "On Raspberry Pi 3 and Raspberry Pi 4 computers running a 32-bit OS, it may be more performant if ..." ?

documentation/asciidoc/accessories/camera/libcamera_apps_post_processing.adoc
```diff
@@ -0,0 +1,216 @@
...
+
+image::images/nohdr.jpg[Image without HDR processing]
+
+With HDR: (use `libcamera-still -o test.jpg --ev -2 --denoise cdn_off --post-process-file hdr.json`)
+
+image::images/hdr.jpg[Image with DRC processing]
+
+==== `motion_detect` stage
+
+The `motion_detect` stage works by analysing frames from the low resolution image stream, which must be configured for it to work. It compares a region of interest ("roi") in the frame frame to the corresponding part of a previous one and if enough pixels are sufficiently different, that will be taken to indicate motion. The result is added to the metadata under "motion_detect.result".
```
lurch, line 168: "frame frame"

documentation/asciidoc/accessories/camera/libcamera_apps_post_processing.adoc
```diff
@@ -0,0 +1,216 @@
...
+| roi_y | y-offset of the region of interest for the comparison
+| roi_width | width of the region of interest for the comparison
+| roi_height | height of the region of interest for the comparison
+| difference_m | Linear coefficient used to construct the threshold for pixels being different
+| difference_c | Constant coefficient used to construct the threshold for pixels being different according to threshold = difference_m * pixel_value + difference_c
+| frame_period | The motion detector will run only this many frames
+| hskip | The pixel tests are subsampled by this amount horizontally
+| vksip | The pixel tests are subsampled by this amount vertically
+| region_threshold | The proportion of pixels (or "regions") which must be categorised as different for them to count as motion
+| verbose ! Print messages to the console, including when the "motion"/"no motion" status changes
```
lurch, line 186: Should the `!` here have been a `|` ?

documentation/asciidoc/accessories/camera/libcamera_apps_post_processing_opencv.adoc
```diff
@@ -0,0 +1,107 @@
...
+	"scale" : 1.0,
+	"thickness" : 2,
+	"alpha" : 0.3
+    }
+}
+----
+
+Example:
+
+image::images/annotate.jpg[Image with Sobel filter]
```
lurch, line 106: Wrong alt-text here?

documentation/asciidoc/accessories/camera/libcamera_apps_post_processing_tflite.adoc
```diff
@@ -0,0 +1,48 @@
+=== Post-Processing with TensorFlow Lite
+
+These stages requires TensorFlow Lite (TFLite) libraries to be installed exporting the C++ API. Unfortunately the TFLite libraries are not normally distributed conveniently in this form, however, one place where they can be downloaded is https://lindevs.com/install-precompiled-tensorflow-lite-on-raspberry-pi/[here]. Please follow the installation instruction given on that page.
+
+==== `object_classify_tf` stage
+
+`object_classify_tf` uses a Google MobileNet v1 model to classify objects in the camera image. It can be obtained from https://storage.googleapis.com/download.tensorflow.org/models/mobilenet_v1_2018_08_02/mobilenet_v1_1.0_224_quant.tgz[], which will need to be uncompressed. You will also need the `labels.txt` file which can be found in https://storage.googleapis.com/download.tensorflow.org/models/mobilenet_v1_1.0_224_frozen.tgz[].
```
lurch, line 7: Just as an FYI, appending an empty `[]` to the end of a URL is [optional](https://docs.asciidoctor.org/asciidoc/latest/macros/url-macro/) in AsciiDoctor.

documentation/asciidoc/accessories/camera/libcamera_apps_post_processing_tflite.adoc
```diff
@@ -0,0 +1,48 @@
+=== Post-Processing with TensorFlow Lite
+
+These stages requires TensorFlow Lite (TFLite) libraries to be installed exporting the C++ API. Unfortunately the TFLite libraries are not normally distributed conveniently in this form, however, one place where they can be downloaded is https://lindevs.com/install-precompiled-tensorflow-lite-on-raspberry-pi/[here]. Please follow the installation instruction given on that page.
```
lurch, line 3: Another "here" link-text/

documentation/asciidoc/accessories/camera/libcamera_apps_post_processing_writing.adoc
```diff
@@ -0,0 +1,43 @@
+=== Writing your own Post-Processing Stages
+
+The `libcamera-apps` _post-processing framework_ is not only very flexible but is meant to make it easy fo users to create their own custom post-processing stages. It is easy to include algorithms and routines that are already available both in OpenCV and TensorFlow Lite.
```
lurch, line 3: "fo users" -> "for users"

documentation/asciidoc/accessories/camera/libcamera_apps_post_processing_writing.adoc
```diff
@@ -0,0 +1,43 @@
...
+
+Some helpful hints on writing your own stages:
+
+* Generally, the `Process` method should not take too long as it will block the imaging pipeline and may cause stuttering. When time-consuming algorithms need to be run, it may be helpful to delegate them to another asynchronous thread.
+
+* When delegating work to another thread, the way image buffers are handled currently means that they will need to be copied. For some applciations, such as image analysis, it may be viable to use the "low resolution" image stream rather than full resolution images.
+
+* The post-processing framework adds multi-threading parallelism on a per-frame basis. This is helpful in improving throughput if you want to run on every single frame. Some functions may supply parallelism within each frame (such as OpenCV and TFLite). In these cases it would probably be better to serialise the calls so as to suppress the per-frame parallelism.
+
+* Most steams, and in particular the low resolution stream, have YUV420 format. These formats are sometimes not ideal for OpenCV or TFLite so there may sometimes need to be a conversion step.
```
lurch, line 31: steams -> streams ?

documentation/asciidoc/accessories/camera/libcamera_apps_writing.adoc
```diff
@@ -0,0 +1,74 @@
...
+
+```
+	app.SetPreviewDoneCallback(std::bind(&LibcameraEncoder::QueueRequest, &app, _1));
+```
+
+==== `libcamera-raw`
+
+`libcamera-raw` is not so very different from `libcamera-vid`. It too uses a encoder, though this time it is a "dummy" encoder called the `NullEncoder`. This just treats the input image directly as the output buffer and is careful not to say it is finished with the input until the output callback has dealt with it first.
+
+This time, however, we do not forward anything to the preview window, though we could have display the (processed) video stream if we had wanted.
```
lurch, line 57: "could have display" -> "could have displayed" ?

documentation/asciidoc/accessories/camera/libcamera_differences.adoc
```diff
@@ -0,0 +1,36 @@
...
+
+Whilst the `libcamera-apps` attempt to emulate most features of the legacy `Raspicam` applications, there are some differences. Here we list the principal ones that users are likely to notice.
+
+* The use of Boost program_options doesn't allow multi-character short versions of options, so where these were present they have had to be dropped. The long form options are named the same, and any single character short forms are preserved.
+
+* `libcamera-still` and `libcamera-jpeg` do not show the capture image in the preview window.
+
+* `libcamera` performs its own camera mode selection, so the `--mode` option is not supported. It deduces camera modes from the resolutions requested. There is still work ongoing in this area.
+
+* The following features of the legacy apps are not supported as the code has to run on the ARM now. But note that a number of these effects are now provid provided by the xref:camera.adoc#post-processing[post-processing] mechanism.
```
lurch, line 11: "provid provided" -> "provided"

documentation/asciidoc/accessories/camera/libcamera_hello.adoc
```diff
@@ -0,0 +1,66 @@
...
+
+[,bash]
+----
+libcamera-hello --info-text "red gain %rg, blue gain %bg"
+----
+will display the current red and blue gain values.
+
+For the HQ camera, use `--info-text "%focus"` to display the focus measure, which will be helpful for focusing the lens.
+
+A full description of the `--info-text` parameter is given xref:camera.adoc#common-command-line-options[below].
```
lurch, line 66: Perhaps using "below" as link-text is unwise, in case the order of the sections gets shuffled about? :shrug:

documentation/asciidoc/accessories/camera/libcamera_jpeg.adoc
```diff
@@ -0,0 +1,38 @@
...
+
+==== Exposure Control
+
+All the `libcamera-apps` allow the user to run the camera with fixed shutter speed and gain. For example
+
+[,bash]
+----
+libcamera-jpeg -o test.jpg -t 2000 --shutter 20000 --gain 1.5
+----
+would capture an image with an exposure of 20ms and a gain of 1.5x. Note that the gain will be applied as _analogue gain_ within the sensor up until it reaches the maximum analogue gain permitted by the kernel sensor driver, after which the remainder will be applied a digital gain.
```
lurch, line 29: "applied a digital" -> "applied as digital" ?

documentation/asciidoc/accessories/camera/libcamera_options_common.adoc
```diff
@@ -0,0 +1,388 @@
+=== Common Command Line Options
+
+The following options apply across all the `libcamera-apps` with similar or identical semantics, unless noted otherwise.
+
+----
+	--timeout, -t			Delay before application stops automatically <milliseconds>
+----
+
+The `--timeout` option specifies how long the application runs before it stops, whether it is recording a video or showing a preview. In the case of still image capture, the application will show the preview window for this long before capturing the output image.
```
lurch, line 9: Does this have a default value if not set explicitly?

documentation/asciidoc/accessories/camera/libcamera_options_common.adoc
```diff
@@ -0,0 +1,388 @@
...
+
+----
+	--rawfull,			Force sensor to capture in full resolution mode
+----
+
+This option forces the sensor to be driven in its full resolution readout mode for still and video capture, irrespective of the requested output resolution (given by `--width` and `--height`). It has no effect for `libcamera-hello`.
+
+Using this option often incurs a frame rate penalty, as larger resolution frames are slower to read out.
+
+Example: `libcamera-raw -t 2000 --segment 1 --rawfull -o test%03d.raw` will cause multiplie full resolution raw frames to be captured. On the HQ camera each frame will be about 18MB in size. Without the `--rawfull` option the default video output resolution would have caused the 2x2 binned mode to be selected, resulting in 4.5MB raw frames.
```
lurch, line 119: multiplie -> multiple

documentation/asciidoc/accessories/camera/libcamera_options_common.adoc
```diff
@@ -0,0 +1,388 @@
...
+----
+	--sharpness,			Set image sharpness <number>
+----
+
+The given `<number>` adjusts the image sharpness. The value zero means that no sharpening is applied, the value 1.0 uses the default amount of sharpening, and values greater than 1.0 use extra sharpening.
+
+Example: `libcamera-still -o test.jpg --sharpness 2.0`
+
+----
+	--contrast,			Set image constrat <number>
```
lurch, line 169: constrat -> contrast

documentation/asciidoc/accessories/camera/libcamera_options_common.adoc
```diff
@@ -0,0 +1,388 @@
...
+
+Example: `libcamera-still -o test.jpg --contrast 1.5`
+
+----
+	--brightness,			Set image brightness <number>
+----
+
+The given `<number>` adjusts the image brightness. The value -1.0 produces an (almost) black image, the value 1.0 produces an almost entirely white image and the value 0.0 produces standard image brightness.
+
+Note that the brightness parameter adds (or subtracts) an offset from all pixels in the output image. The `--ev` options is often more appropriate.
```
lurch, line 182: "`--ev` options" -> "`--ev` option" ?

documentation/asciidoc/accessories/camera/libcamera_options_common.adoc
```diff
@@ -0,0 +1,388 @@
...
+
+Example: `libcamera-still -o test.jpg --ev 0.3`
+
+----
+	--shutter,			Set the exposure time in microseconds <number>
+----
+
+The shutter time is fixed to the given value. The gain will still be allowed to vary (unless that is also fixed).
+
+Note that this shutter time may not be achieved if the camera is running at a frame rate that is too fast to allow it. In this case the `--framerate` option may be used to lower the frame rate. The maximum possible shutter times for the official Raspberry Pi supported cameras are:
```
lurch, line 208: Does each camera also have a minimum shutter time? :shrug:  (or is that irrelevant, as it would produce images that are too dark to be useful?)

documentation/asciidoc/accessories/camera/libcamera_options_common.adoc
```diff
@@ -0,0 +1,388 @@
...
+Note that there are circumstances where the digital gain can go above 1 even when the analogue gain limit is not exceeded. This can occur when
+
+* Either of the colour gains goes below 1.0, which will cause the digital gain to settle to 1.0/min(red_gain,blue_gain). This means that the total digital gain being applied to any colour channel does not go below 1.0, as that would cause discolouration artifacts.
+* The digital gain can vary slightly while the AEC/AGC changes, though this effect should be only transient.
+
+----
+	--metering,			Set the metering mode <string>
+----
+
+Sets the metering mode of the AEC/AGC algorithm. This may one of the following values
```
lurch, line 244: Does this have a default value if not explicitly set?

documentation/asciidoc/accessories/camera/libcamera_options_still.adoc
```diff
@@ -0,0 +1,128 @@
...
+
+Use the current date and time to construct the output file name, in the form MMDDhhmmss.jpg, where MM = 2-digit month number, DD = 2-digit day number, hh = 2-digit 24-hour hour number, mm = 2-digit minute number, ss = 2-digit second number. (`libcamera-still` only.)
+
+Example: `libcamera-still --datetime`
+
+----
+	--timestamp,			Use system timestamps for the output file names
+----
+
+Uses the current system timestamp as the the output file name. (`libcamera-still` only.)
```
lurch, line 51: Does this need a bit more explanation? i.e. what exactly is a system timestamp?

documentation/asciidoc/accessories/camera/libcamera_options_still.adoc
```diff
@@ -0,0 +1,128 @@
...
+
+Example: `libcamera-still -e png -o test.png`
+
+----
+	--raw,		-r		Save raw file
+----
+
+Save a raw Bayer file in DNG format alongside the usual output image. The file name is given by replacing the output file name extension by `.dng`. These are standard DNG files, and can be processed with standard tools like _dcraw_ or _RawTherapee_, among others. (`libcamera-still` only.)
+
+The image data in the raw file is exactly what came out of the sensor, without no processing whatsoever either by the ISP or anything else. The EXIF data saved in the file, among other things, includes:
```
lurch, line 115: "without no processing"? :upside_down_face:

documentation/asciidoc/accessories/camera/libcamera_options_vid.adoc
```diff
@@ -0,0 +1,108 @@
...
+
+Set the target bitrate for the H.264 encoder, in _bits per second_.
+
+Example: `libcamera-vid -b 10000000 --width 1920 --height 1080 -o test.h264`
+
+----
+	--intra,	-g		Intra-frame period (H.264 only) <number>
+----
+
+Sets the frequence of I (Intra) frames in the H.264 bitstream, as a number of frames. The default value is 60.
```
lurch, line 23: frequence -> frequency ?

documentation/asciidoc/accessories/camera/libcamera_options_vid.adoc
```diff
@@ -0,0 +1,108 @@
...
+
+Set the H.264 level. The value may be `4`, `4.1` or `4.2`.
+
+Example: `libcamera-vid --width 1920 --height 1080 --level 4.1 -o test.h264`
+
+
+
+
+----
+	--keypress,	-k		Capture image when Enter pressed
```
lurch, line 47: Capture image -> Capture video?
(and the same also for the --signal option below)

documentation/asciidoc/accessories/camera/libcamera_options_vid.adoc
```diff
@@ -0,0 +1,108 @@
...
+
+One convenient little trick is to pass a very small duration parameter (namely, `--segment 1`) which will result in each frame being written to a separate output file. This makes it easy to do "burst" JPEG capture (using the MJPEG codec), or "burst" raw frame capture (using `libcamera-raw`).
+
+Example: `libcamera-vid -t 100000 --segment 10000 -o test%04d.h264`
+
+----
+	--circular,			Write the video recording into a circular buffer.
+----
+
+THe video recording is written to a circular buffer which is written to disk when the application quits.
```
lurch, line 98: THe -> The

How big is this circular buffer?

documentation/asciidoc/accessories/camera/libcamera_python.adoc
```diff
@@ -0,0 +1,5 @@
+=== Python Bindings for `libcamera`
+
+Python bindings for `libcanera` are currently in development.
```
lurch, line 3: libcanera :upside_down_face:

documentation/asciidoc/accessories/camera/libcamera_vid.adoc
```diff
@@ -0,0 +1,78 @@
...
+
+==== Encoders
+
+There is support for motion JPEG, and also for uncompressed and unformatted YUV420, for example
+[,bash]
+----
+libcamera-vid -t 10000 --codec mjpeg -o test.mjpeg
+libcamera-vid -t 10000 --codec yuv420 -o test.data
+----
+In both cases the `--codec` parameter determines the output format, not the extension of the output file.
```
lurch, line 26: Should this --codec be listed in libcamera_options_vid.adoc ?

davidplowman:
Wow, thanks for the detailed checking, I wasn't expect that so quickly! Anyway, I'll make those various corrections, and looking forward to hearing what folks think about the very long page - or whether we want to do some chopping up.

lurch:
> Wow, thanks for the detailed checking, I wasn't expect that so quickly!

I sometimes get carried away with my eye-for-detail :wink:

documentation/asciidoc/accessories/camera/libcamera_apps_post_processing.adoc
davidplowman: Hmm, interesting, I don't really have a good answer for that. I suppose the situation is that *our* implementation reads them in order, and that the order matters to us so I won't change this just for the moment. For reference, what would this look like using `[ ... ]`?

davidplowman:
@lurch I've worked my way through that long list, so thank you for those! One thought I had while doing this, should I list all the options in alphabetical order... is that something we normally do?

documentation/asciidoc/accessories/camera/libcamera_apps_post_processing.adoc
lurch: Well if each of the steps has a unique name (which it should for a dictionary data-structure), then I guess you could perhaps do something like:
```json
{
    "config":
    {
        "sobel_cv":
        {
            "ksize": 5
        },
        "negate":
        {
        }
    },
    "order": ["sobel_cv", "negate"]
}
```
i.e. "config" is an unordered dictionary, and "order" specifies what order each of those operations should be performed in.

But if you'd ever want to do the same step in multiple places (perhaps with different options for each version of the same operation), i.e. something like this in your current syntax:
```json
{
    "sobel_cv":
    {
        "ksize": 5
    },
    "negate":
    {
    },
    "sobel_cv":
    {
        "ksize": 3
    }
}
```
then having a dictionary with multiple keys named the same is invalid JSON syntax (if I throw that example into https://jsonlint.com/ it complains with `Error: Duplicate key 'sobel_cv'`).

So In that case maybe you'd want to do something like:
```json
{
    "operations":
    [
        {
            "name": "sobel_cv",
            "config":
            {
                "ksize": 5
            }
        },
        {
            "name": "negate",
            "config":
            {
            }
        },
        {
            "name": "sobel_cv",
            "config":
            {
                "ksize": 3
            }
        }
    ]
}
````
i.e. an (ordered) array of dictionaries.
:man_shrugging: 

You _could_ get rid of the outer-dictionary i.e.
```json
[
    {
        "name": "sobel_cv",
        "config":
        {
            "ksize": 5
        }
    },
    {
        "name": "negate",
        "config":
        {
        }
    },
    {
        "name": "sobel_cv",
        "config":
        {
            "ksize": 3
        }
    }
]
```
but IMHO leaving it in (as in the prior example) gives you more flexibility for the future (i.e. you can add additional top-level dictionary keys without affecting the "operations" key, which you couldn't do if the top-level data structure was an array).

lurch:
Looking at e.g. https://www.raspberrypi.org/documentation/accessories/camera.html#command-line-options the options don't appear to be listed in any particular order? :shrug:  Perhaps it makes sense to list options so that the most-commonly-used are mentioned first, and the least-commonly-used are mentioned last? Or perhaps it doesn't really matter so just leave them in the order you currently have?

documentation/asciidoc/accessories/camera/libcamera_apps_post_processing.adoc
```diff
@@ -0,0 +1,216 @@
...
+
+image::images/negate.jpg[Image with negate]
+
+==== `hdr` stage
+
+The `hdr` stage implements both HDR (high dynamic range) imaging and DRC (dynamic range compression). The terminology that we use here regards DRC as operating on single images, and HDR works by accumulating multiple under-exposed images and then performing the same algorithm as DRC.
+
+The `hdr` stage has no dependencies on 3rd party libraries.
+
+It may be more performant on Pi 3 and 32-bit OS Pi 4 devices if recompiled using `-DENABLE_COMPILE_FLAGS_FOR_TARGET=armv8-neon` (please see the xref:camera.adoc#building-libcamera-and-libcamera-apps[build instructions]). Specifically, the image accumulation stage will run quicker and result in fewer frame drops, though the tonemapping part of the process is unchanged.
```
lurch, line 73: Could you also use the updated phrasing here please?

documentation/asciidoc/accessories/camera/libcamera_apps_post_processing_tflite.adoc
```diff
@@ -0,0 +1,48 @@
+=== Post-Processing with TensorFlow Lite
+
+These stages requires TensorFlow Lite (TFLite) libraries to be installed exporting the C++ API. Unfortunately the TFLite libraries are not normally distributed conveniently in this form, however, one place where they can be downloaded is https://lindevs.com/install-precompiled-tensorflow-lite-on-raspberry-pi/[lindevs.com]. Please follow the installation instruction given on that page.
```
lurch, line 3: "installation instruction" -> "installation instructions"

documentation/asciidoc/accessories/camera/libcamera_options_common.adoc
```diff
@@ -0,0 +1,388 @@
...
+| 6000000 (i.e. 6s)
+
+| V2 (IMX219)
+| 10000000 (i.e. 10s)
+
+| HQ (IMX477)
+| 200000000 (i.e. 200s)
+|===
+
+Using values above these maximums will result in undefined behaviour. Cameras will also have different minimum exposure times, though in practice this is not important as they are all low enough to exposure bright scenes appropriately.
```
lurch, line 224: Given that this is talking about shutter times, should the "minimum exposure times" here be "minimum shutter times" ? (as you can tell, I don't know much about cameras :wink: )
Also, should "enough to exposure bright scenes" be "enough to expose bright scenes" ?

documentation/asciidoc/accessories/camera/libcamera_options_vid.adoc
```diff
@@ -0,0 +1,129 @@
...
+----
+	--keypress,	-k		Toggle between recording and pausing
+----
+
+Pressing Enter will toggle `libcamera-vid` between recording the video stream and not recording it (i.e. discarding it). The application starts off in the recording state, unless the `--initial` options specifies otherwise. Typing `x` and Enter causes `libcamera-vid` to quit.
+
+Example: `libcamera-vid -t 0 -o test.h264 -k`
+
+----
+	--signal,	-s		Capture image when SIGUSR1 received
```
lurch, line 68: "Capture image when" -> "Toggle between recording and pausing when" ?

documentation/asciidoc/accessories/camera/libcamera_options_vid.adoc
```diff
@@ -0,0 +1,129 @@
...
+
+This option causes the H.264 sequence headers to be written into every I (Intra) frame. This is helpful because it means a client can understand and decode the video sequence from any I frame, not just from the very beginning of the stream. It is recommended to use this option with any output type that breaks the output into pieces (`--segment`, `--split`, `--circular`), or transmits the output over a network.
+
+Example: `libcamera-vid -t 0 --keypress --inline --split -o test%04d.h264`
+
+----
+	--listen,			Wait for an incoming TCP connection
+----
+
+This option is provided for streaming over a network using TCP/IP. Using `--listen`` will cause `libcamera-vid` to wait for an incoming client connection before starting th video encode process, which will then be forwarded to that client.
```
lurch, line 127: Looks like there's accidentally two backticks after --listen

lurch:
As a more general point, perhaps it make sense to exclude the comma after the long-option, when there's no short-option? e.g. change
```
----
	--level,			H.264 level <string>
----
```
to
```
----
	--level			H.264 level <string>
----
```
?

lurch:
I've gone through and clicked "Resolved" on each of the points that you've fixed :+1:

documentation/asciidoc/accessories/camera/libcamera_apps_post_processing.adoc
aallan: Yup. We need to avoid using "Pi" anywhere in the documentation — in **all** cases it should be "Raspberry Pi".

documentation/asciidoc/accessories/camera/libcamera_apps_intro.adoc
aallan: Nope. Not allowed (and I'm going through the existing documentation trying to remove 'em all). Documentation should be written as if it were going to be printed out and the links didn't work. Then add the links in afterwards.

documentation/asciidoc/accessories/camera/libcamera_apps_post_processing.adoc
davidplowman: Hmm, would something like this be valid?
```
{
    "sobel_cv":
    {
        "ksize": 5
    },
    "negate":
    {
    },
    "order": ["sobel_cv", "negate"]
}
```
or am I abusing the syntax too much? It has the advantage of "fixing" the problem by allowing a precise order to be given, but is backwards compatible if the "order" is omitted, so long as we preserve the order when reading (which we do). Though you could never have a stage actually named "order", but TBH I don't think that bothers me!

lurch: Yeah, I guess that could work, as long as your parser was updated to treat:
```json
{
    "sobel_cv":
    {
        "ksize": 5
    },
    "negate":
    {
    },
    "order": ["sobel_cv", "negate"]
}
```
and
```json
{
    "negate": {},
    "order": ["sobel_cv", "negate"],
    "sobel_cv": {"ksize": 5}
}
```
as being identical :slightly_smiling_face: 
_Would_ you ever want to do the same operation multiple times (as in my simplistic `["sobel_cv", "negate", "sobel_cv"]` example) ?
I know even less about image-processing than I do about cameras :laughing:

JamesH65: Shouldn't the order be moved outside the block, therefor being at the same hierarchical level as the list of stages?  That way its name doesn't clash, and there are parser ordering issues.

davidplowman: Possibly, but I don't think the JSON parser (we use the Boost one) will accpet that.

To answer the other question, I don't think we have any stages that we want to run multiple times, though I agree it's not impossible to imagine such a scenario. If that became important we could introduce some syntax that would allow it ("negate.0" and "negate.1" perhaps), but it's not something I feel we need to worry about at the moment.

lurch: >> Shouldn't the order be moved outside the block, therefor being at the same hierarchical level as the list of stages?

> Possibly, but I don't think the JSON parser (we use the Boost one) will accpet that.

Yeah, the outermost `{  }` is already the top-level hierarchy of the JSON file, so it's impossible to have anything else at the "same level". (If I'm understanding what James is asking for)

JamesH65: OK, perhaps you need to add an extra item in each stage which is the position or order of it in the list, i.e. explicitly state it 
 since JSON has no implicit ordering I assume.

`
{
    "sobel_cv":
    {
        "ksize": 5
        "order": 0
    },
    "negate":
    {
        "order": 1
    },
}
`

aallan: Sounds like you shouldn't be using JSON at all really? Like XML JSON is explicitly non-ordered.

davidplowman: OK, thanks for the comments. For now I propose to leave this alone (we've been using JSON files elsewhere for quite a while and treat them in the same way), but adding an "order" field in future seems reasonable as it resolves the issue and is backwards compatible.

aallan:
Accidentally got closed via rogue reference to the PR from #2129. Reopened.

davidplowman:
I've been through and updated stuff that didn't talk about libcamera now to include it. Whilst that's fine I've now got a situation as follows.

1. A very long "libcamera and libcamera-apps" section
2. A long "raspicam commands" section

And now a number of smaller sections (e.g. "Timelapse", "Shooting raw", "Long exposures" etc.) that say "if you're using libcamera do this..." and "otherwise if you're using raspicam do that...".

I'm thinking I need to split those sections up and move them into the libcamera/raspicam sections. What do folks think?

lurch:
They sound like "editorial decisions" for @aallan to make?

aallan:
> I'm thinking I need to split those sections up and move them into the libcamera/raspicam sections. What do folks think?

If the sections are essentially `if (libcamera) do foo, else if (raspicam) do bar` then there should be a top level libcamera section, and a top level raspicam section and each of these top level sections should have a number of applications sections (e.g. "Timelapse", "Shooting raw", "Long exposures" etc.). The thing that's top level here, for now at least until raspicam goes away, is the choice between libcamera or raspicam.

davidplowman:
So I think I've been sat on this long enough and it does correctly reflect both pre- and post- Bullseye images. Are folks happy to merge it all?

documentation/asciidoc/accessories/camera/camera_hardware.adoc
```diff
@@ -2,6 +2,32 @@
 
 The Raspberry Pi Camera Modules are official products from the Raspberry Pi Foundation. The original 5-megapixel model was https://www.raspberrypi.org/blog/camera-board-available-for-sale/[released] in 2013, and an 8-megapixel https://www.raspberrypi.org/products/camera-module-v2/[Camera Module v2] was https://www.raspberrypi.org/blog/new-8-megapixel-camera-board-sale-25/[released] in 2016. For both iterations, there are visible light and infrared versions. A 12-megapixel https://www.raspberrypi.org/products/raspberry-pi-high-quality-camera/[High Quality Camera] was https://www.raspberrypi.org/blog/new-product-raspberry-pi-high-quality-camera-on-sale-now-at-50/[released] in 2020. There is no infrared version of the HQ Camera, however the xref:camera.adoc#raspberry-pi-hq-camera-filter-removal[IR Filter can be removed] if required.
 
+=== Installing a Raspberry Pi camera
```
lurch, line 5: FYI once #2159 has been merged then "nested includes" will work properly, i.e. `camera/camera_hardware.adoc` will be able to include `camera_installing.adoc`, so there won't be any need to copy'n'paste the contents of adoc files into other adoc files. Letting you know in case that makes things any easier :wink:

documentation/asciidoc/accessories/camera/camera_installing.adoc
```diff
@@ -14,41 +14,45 @@ Depending on the model, the camera may come with a small piece of translucent bl
...
 === Enabling the Camera
 
-==== Using the desktop
-
-Select `Preferences` and `Raspberry Pi Configuration` from the desktop menu: a window will appear. Select the `Interfaces` tab, then click on the `enable camera` option. Click `OK`. You will need to reboot for the changes to take effect.
-
-==== Using the command line
-
-Open the `raspi-config` tool from the terminal:
+Before proceeding, we recommend ensuring that your kernel, GPU firmware and applications are all up to date. You'll need an internet connection for this to work correctly.
```
lurch, line 17: Perhaps this should link to https://www.raspberrypi.org/documentation/computers/os.html#using-apt rather than repeating the apt command-lines? :shrug:

documentation/asciidoc/accessories/camera/camera_installing.adoc
```diff
@@ -14,41 +14,45 @@ Depending on the model, the camera may come with a small piece of translucent bl
...
-sudo raspi-config
+sudo apt update
+sudo apt full-upgrade
 ----
 
-Select `Interfacing Options` then `Camera` and press `Enter`. Choose `Yes` then `Ok`. Go to `Finish` and you'll be prompted to reboot.
+==== Using `libcamera`
 
-=== Setting up the Camera Software
+Raspberry Pi's three official cameras (V1, V2 and HQ camera) will be detected and enabled automatically when using a Raspberry Pi OS based on _Bullseye_ or later.
```
lurch, line 27: I wonder if we ought to delay merging this until after we've released a Raspberry Pi OS Bullseye image, to avoid any user-confusion?

davidplowman: I did check with Gordon and he was OK with it... after all it's no secret that Bullseye is coming and trailing some of its new features isn't a bad thing?

documentation/asciidoc/accessories/camera/camera_installing.adoc
```diff
@@ -14,41 +14,45 @@ Depending on the model, the camera may come with a small piece of translucent bl
...
 ----
 
-Now you need to enable camera support using the `raspi-config` program you will have used when you first set up your Raspberry Pi.
+You should see a camera preview window for about 5 seconds.
+
+==== Using the legacy stack
+
+===== On the desktop
+
+Select `Preferences` and `Raspberry Pi Configuration` from the desktop menu: a window will appear. Select the `Interfaces` tab, then click on the `enable camera` option. Click `OK`. You will need to reboot for the changes to take effect.
```
lurch, line 44: Perhaps "click on the `enable camera` option" should be "click on the `Camera: Enabled` option" ?

documentation/asciidoc/accessories/camera/camera_usage.adoc
```diff
@@ -1,16 +1,30 @@
...
+* 3rd parties can add support directly for their own cameras.
+* Nearly all aspects of the camera tuning can be changed by users.
+* It integrates much more conveniently with other standard Linux APIs.
+* Raspberry Pi supply a set of `libcamera-apps` which emulate most of the features of the legacy applications.
+* It is fully supported on 64-bit operating systems.
+
+Reasons to consider staying with an older OS and using the legacy _Raspicam_ stack might include:
+
+* It may perform better on Raspberry Pi 2 and Raspberry Pi Zero devices, as it offloads more to the GPU and is less dependent on the ARM cores.
+* `libcamera` is still missing certain features, most notably Python bindings. Whilst this work is in progress, users who need Python will have to stay with the legacy stack for the time being.
```
lurch, line 28: "users who need Python" - is that referring to the [picamera](https://picamera.readthedocs.io/) project? Perhaps it's worth mentioning that by name? :man_shrugging:

documentation/asciidoc/accessories/camera/camera_usage.adoc
```diff
@@ -1,16 +1,30 @@
...
+
+Apart from some general information on setting up the camera, most of this documentation is divded into sections introducing xref:camera.adoc#libcamera-and-libcamera-apps[libcamera and the libcamera-apps], and the xref:camera.adoc#Raspicam-commands[legacy _Raspicam_ apps]. Once you've xref:camera.adoc#installing-a-raspberrypi-pi-camera[installed your camera module], please click on one of these two links.
```
lurch, line 30: Perhaps change "introducing" to "documenting" ?

documentation/asciidoc/accessories/camera/camera_installing.adoc
davidplowman: Though when it's only 2 lines maybe going to a whole new page is kinda annoying? But I don't mind changing it...

documentation/asciidoc/accessories/camera/camera_usage.adoc
davidplowman: Yep, I'll add that!

documentation/asciidoc/accessories/camera/gstreamer.adoc
```diff
@@ -0,0 +1,44 @@
+=== Using Gstreamer
+
+_Gstreamer_ is a Linux framework for reading, processing and playing multimedia files. There is a lot of information and many tutorials at the https://gstreamer.freedesktop.org/[_gstreamer_ website]. Here we show how `libcamera-vid` (and similarly `raspivid`) can be used to stream video over a network.
```
lurch, line 3: Only a minor niggle but it seems a bit inconsistent to mention "stream video over a network" here, but all the examples only send and receive via `localhost` ? Perhaps it's worth including at least one example where the server and client are different computers?

documentation/asciidoc/accessories/camera/libcamera_apps_getting_started.adoc
```diff
@@ -0,0 +1,60 @@
...
+
+==== If you do need to alter the configuration
+
+You may need to alter the camera configuration in your `/boot/config.txt` file if:
+
+* You are using a 3rd party camera (the manufacturer's instructions should explain the changes you need to make).
+
+* You are using an official Raspberry Pi camera but wish to use a non-standard driver/overlay.
+
+* You are using an older version (_Buster_ or earlier) of the operating system (in which case we would recommend upgrading first if this is possible).
```
lurch, line 28: Maybe we need instructions somewhere that a user can use to determine if they're running a 'Buster' or a 'Bullseye' image?

aallan: Yup and that should go here, https://www.raspberrypi.org/documentation/computers/os.html#updating-and-upgrading-raspberry-pi-os, I guess right at the start of the updating and upgrading bit.

aallan: Does `cat /etc/os-release` still work?

documentation/asciidoc/accessories/camera/libcamera_apps_post_processing_tflite.adoc
```diff
@@ -0,0 +1,146 @@
...
+
+[cols=",^"]
+|===
+| top_n_results | How many results to show
+| refresh_rate | The number of frames that must elapse before the model is re-run
+| threshold_high | Confidence threshold (between 0 and 1) where objects are considered as being present
+| threshold_low | Confidence threshold which objects must drop below before being discarded as matches
+| model_file | Pathname to the tflite model file
+| labels_file | Pathname to the file containing the object labels
+| display_labels | Whether to display the object labels on the image. Note that this causes `annotate.text` metadata to bve inserted so that the text can be rendered subsequently by the `annotate_cv` stage
```
lurch, line 19: bve -> be

documentation/asciidoc/accessories/camera/libcamera_apps_writing.adoc
```diff
@@ -0,0 +1,74 @@
...
+	app.SetPreviewDoneCallback(std::bind(&LibcameraEncoder::QueueRequest, &app, _1));
+```
+
+==== `libcamera-raw`
+
+`libcamera-raw` is not so very different from `libcamera-vid`. It too uses a encoder, though this time it is a "dummy" encoder called the `NullEncoder`. This just treats the input image directly as the output buffer and is careful not to say it is finished with the input until the output callback has dealt with it first.
+
+This time, however, we do not forward anything to the preview window, though we could have displayed the (processed) video stream if we had wanted.
+
+The use of the `NullEncoder` is possibly overkill in this application as we could probably just sent the image straight to the `Output` object. However, it serves to underline the general principle that it is normally a bad idea to do too much work directly in the event loop, and time-consuming processes are often better left to other threads.
```
lurch, line 59: "just sent" -> "just send" ?

documentation/asciidoc/accessories/camera/libcamera_options_common.adoc
```diff
@@ -0,0 +1,398 @@
...
+| Camera Version | Max (microseconds)
+
+| V1 (OV5647)
+| 6000000 (i.e. 6s)
+
+| V2 (IMX219)
+| 10000000 (i.e. 10s)
+
+| HQ (IMX477)
+| 200000000 (i.e. 200s)
```
lurch, line 231: Is this the same info as the "Maximum Exposure Times" table in camera_hardware.adoc ? (where it claims the HQ camera supports up to 230 seconds?)

documentation/asciidoc/accessories/camera/libcamera_options_common.adoc
```diff
@@ -0,0 +1,398 @@
...
+
+When outputting to files with an incrementing counter (e.g. `%d` in the output file name), wrap the counter back to zero when it reaches this value.
+
+Example: `libcamera-vid -t 0 --codec mjpeg --segment 1 --wrap 100 -o image%d.jpg`
+
+----
+	--flush					Flush output files immediately
+----
+
+`--flush` causes output files to be flushed to disk as soon as every frame is written, rather that waiting for the system to do it.
```
lurch, line 386: "rather that" -> "rather than"

documentation/asciidoc/accessories/camera/libcamera_options_still.adoc
```diff
@@ -0,0 +1,128 @@
...
+
+Select the still image encoding to be used. Valid encoders are:
+
+* `jpg` - JPEG (the default)
+* `png` - PNG format
+* `bmp` - BMP format
+* `rgb` - binary dump of uncompressed RGB pixels
+* `yuv420` - binary dump of uncompressed YUV420 pixels.
+
+Note that this option determines the encoding and that the extension of the output file name is ignored for this purpose. (`libcamera-still` only.)
```
lurch, line 105: Does this affect the extension of the filenames used by the `--datetime` and `--timestamp` options?

documentation/asciidoc/accessories/camera/libcamera_options_common.adoc
```diff
@@ -0,0 +1,398 @@
...
+`--output` sets the name of the output file to which the output image or video is written. Besides regular file names, this may take the following special values:
+
+* `-` - write to stdout
+* `udp://` - a string starting with this is taken as a network address for streaming
+* `tcp://` - a string starting with this is taken as a network address for streaming
+* a string containing a `%d` directive is taken as a file name where the format directive is replaced with a count that increments for each file that is opened.
+
+Examples:
+
+`libcamera-vid -t 100000 --segment 10000 -o chunk%04d.h264` records a 100 second file in 10 second segments, where each file is named `chunk.h264` but with the inclusion of an incrementing counter.
```
lurch, line 370: I wonder if it's worth explaining what the `04` in the `%04d` part means?

documentation/asciidoc/accessories/camera/libcamera_options_vid.adoc
```diff
@@ -0,0 +1,129 @@
...
+
+`libcamera-vid -t 10000 --codec mjpeg -o test.mjpeg`
+
+`libcamera-vid -t 10000 --codec yuv420 -o test.data`
+
+----
+	--keypress,	-k		Toggle between recording and pausing
+----
+
+Pressing Enter will toggle `libcamera-vid` between recording the video stream and not recording it (i.e. discarding it). The application starts off in the recording state, unless the `--initial` options specifies otherwise. Typing `x` and Enter causes `libcamera-vid` to quit.
```
lurch, line 63: "options specifies" -> "option specifies" ?

documentation/asciidoc/accessories/camera/libcamera_options_vid.adoc
```diff
@@ -0,0 +1,129 @@
...
+
+Example: `libcamera-vid -t 0 -o test.h264 -k`
+
+----
+	--signal,	-s		Toggle between recording and pausing when SIGUSR1 received
+----
+
+The SIGUSR1 signal will toggle `libcamera-vid` between recording the video stream and not recording it (i.e. discarding it). The application starts off in the recording state, unless the `--initial` options specifies otherwise. SIGUSR2 causes `libcamera-vid` to quit.
```
lurch, line 71: "options specifies" -> "option specifies" ?

documentation/asciidoc/accessories/camera/raspicam.adoc
```diff
@@ -2,6 +2,34 @@
 
 `raspistill`, `raspivid` and `raspiyuv` are command line tools for using the camera module.
 
+=== Enabling the Camera
+
+Before using any of the _Raspicam_ applications, the camera must be enabled.
+
+==== On the desktop
+
+Select `Preferences` and `Raspberry Pi Configuration` from the desktop menu: a window will appear. Select the `Interfaces` tab, then click on the `enable camera` option. Click `OK`. You will need to reboot for the changes to take effect.
```
lurch, line 11: Seems to be duplicated info in camera_installing.adoc ?

davidplowman: Huh, I seem to have confused myself a bit by moving the content of camera_installing.adoc into camera_hardware.adoc but then forgetting to remove that file. I'll fix that up along with everything else!

documentation/asciidoc/accessories/camera/camera_installing.adoc
lurch: Depends on how strongly @aallan feels about the "duplication of content" :wink:

documentation/asciidoc/accessories/camera/camera_installing.adoc
aallan: We've already had a couple of emails and forum posts bemoaning the documentation doesn't tell them how to update to Bullseye. I might put a holding paragraph into the update & upgrade section tell folks that Bullseye isn't out yet. This would certainly need a `[NOTE]` block saying that Bullseye isn't out and to watch the blog (or something like that).

JamesH65: We've not on the whole published official instructions on updating from one major version to another because we cannot be sure it will work in all cases. When we have said something its with LOTS of try this at your own peril stuff etc. recommendation has always been get a new SD card and a new installation.

aallan: Yup. There are instructions somewhere but they have heavy warnings around them.

davidplowman:
Thanks again for the giant mega-review! I think I've fixed up everything that was highlighted so it's definitely looking much better now.

aallan:
> Thanks again for the giant mega-review! I think I've fixed up everything that was highlighted so it's definitely looking much better now.

I think what I'm going to do is create a new branch in the documentation repo from `develop`, merge this PR to that new branch rather than `develop` proper, and then we can bring this up in the local web server more easily and poke at it while we're all still committing smaller fixes to `develop`.

aallan:
Merged onto the new `new_camera_documentation` branch so we can all take a look using `make serve_html` locally more easily. Tweaks, rewrites, and comments on that branch going forward please! 😃

lurch:
I've just rebased that `new_camera_documentation` branch against the latest `develop` and force-pushed.

aallan:
> I've just rebased that `new_camera_documentation` branch against the latest `develop` and force-pushed.

I didn't branch from the head of `develop`? I sort of thought I had, did I forget to pull before I branched (again)?

Oddly my local checkout of the branch is 3 commits ahead of origin now. Git. 🤷

PR #2103, opened 2021-08-20
Test that new workflow check rejects PRs based on master from other repositories' develop branch

mudge:
Confirm that the new step in our GitHub Actions workflow to prevent merges into `master` from branches other than `develop` rejects pull requests from forks when the fork's branch name also happens to be`develop` (see @lurch's [comment](https://github.com/raspberrypi/documentation/pull/2100#issuecomment-901254248)).

This build should fail.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -10,6 +10,12 @@ jobs:
   build:
     runs-on: ubuntu-latest
     steps:
+      - name: Only allow pull requests based on master from the develop branch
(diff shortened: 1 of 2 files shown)


lurch:
Just out of curiosity, why is this saying that one build failed and one build succeeded? 
Screenshot from 2021-08-20 11-22-44

mudge:
The exact same commit (`0c07000`) passed when run from [the `raspberrypi/documentation` PR](https://github.com/raspberrypi/documentation/pull/2100) but failed when run from this PR. I guess status checks are per commit (which makes sense) but it looks a little confusing here.

PR #2108, opened 2021-08-20
prefix hex number with 0x

dp111:


aallan:
Hmm. While I agree in theory, I'm not sure we should start unless we intend go through and do this everywhere.

dp111:
Got to start somewhere. Other hex numbers on the page are prefixed. As we
come across them they can be done.

aallan:
> Got to start somewhere. Other hex numbers on the page are prefixed.

Fair enough.

aallan:
@dp111 I merged this, but then noticed you'd based it against the `master` branch. So I've reverted the PR. Please rebase against `develop`. All PRs should be against the `develop` branch. A push to `master` runs the build and pushes live to production.

PR #2134, opened 2021-08-30
fix dtb and zImage  "No such file or directory"

liuyunli:
While building the Kernel
Build and install the kernel, modules, and Device Tree blobs, find the errors:

cp: cannot stat 'arch/arm64/boot/dts/*.dtb': No such file or directory
cp: cannot stat 'arch/arm64/boot/zImage': No such file or directory

There is wrong path and file name in the bash script.

Signed-off-by: Yunli Liu <994605959@qq.com>


lurch:
Relates to #2073

aallan:
Superceded by #2149.

PR #2135, opened 2021-08-31
Push to production

aallan:


.github/workflows/stale.yml
```diff
@@ -16,7 +16,11 @@ jobs:
...
-        stale-issue-message: 'Stale issue message'
-        stale-pr-message: 'Stale pull request message'
-        stale-issue-label: 'no-issue-activity'
-        stale-pr-label: 'no-pr-activity'
+        stale-issue-message: 'This issue has been automatically marked as stale because it has not had recent activity. It will be closed if no further activity occurs. Thank you for your contributions.'
+        stale-pr-message: 'This pull request has been automatically marked as stale because it has not had recent activity. It will be closed if no further activity occurs. Thank you for your contributions.'
+        stale-issue-label: 'stale issue'
+        stale-pr-label: 'stale pull request'
+        exempt-issue-labels: 'ready to merge,read for copy-edit,paused,in progress,linked pull request'
+        exempt-pr-labels: 'ready to merge,read for copy-edit,paused,in progress'
```
lurch, line 24: I guess these should have said "ready for ..." rather than "read for ..." ?

aallan: Fixed in 33ffe2b

PR #2136, opened 2021-09-01
Redirect Pico landing site to microcontroller docs

mudge:
Now information about RP2040 and Raspberry Pi Pico have been folded into the documentation site, redirect requests for the old "Getting Started" landing site to the Microcontrollers section.

Note this won't have any effect until we remove the configuration from Cloudflare (both a page rule which currently handles the redirect from /documentation/pico to /documentation/rp2040 and a worker route which directs requests to the Getting Started site). Once that is gone, we'll be able to control all redirects under /documentation from this project.


aallan (approved):
LGTM

PR #2138, opened 2021-09-01
Redirect old UF2 URLs to datasheets

mudge:
To catch any direct links to the UF2 files on the old RP2040/Pico Getting Started site, redirect to their latest location on datasheets.raspberrypi.org.


aallan (approved):
LGTM

lurch:
@mudge It doesn't really make any difference, but I wonder if some of these would have been more appropriate in `documentation/redirects/datasheets.csv` ? :man_shrugging:  (although that would have made them 301 redirects rather than 302 redirects, if that makes any difference?)

PR #2143, opened 2021-09-03
Add missing KERNEL env variable

Lblenner:
Similarly to `Build sources` above, `make` needs the env variable `KERNEL` to be set accordingly


pelwell:
I think you are mistaken. The KERNEL environment variable is only their to allow a single set of `cp` commands to be used for multiple architectures - I know, because I put it there in the first place.

PR #2146, opened 2021-09-05
Update samba.adoc

alexknop:
updated smb.conf script to change "writable" to the correct command, "writeable"


andrum99:
Per the samba documentation at https://www.samba.org/samba/docs/current/man-html/smb.conf.5.html both `writable` and `writeable` will work, so this PR is not required.

Also, both lexico.com and Merriam Webster's online dictionaries say `writable` is the correct English spelling, so the page is correct as currently written.

PR #2149, opened 2021-09-09
Fix local 64-bit installation instructions

pelwell:
The new 64-bit local installation instructions were clearly copied from the 32-bit versions, ignoring the differences for the 64-bit kernel.

### DIFF
--- a/documentation/asciidoc/computers/linux_kernel/building.adoc
+++ b/documentation/asciidoc/computers/linux_kernel/building.adoc
@@ -120,10 +120,10 @@ For the 64-bit kernel:
 ----
 make -j4 Image modules dtbs
 sudo make modules_install
-sudo cp arch/arm64/boot/dts/*.dtb /boot/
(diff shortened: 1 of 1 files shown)


lurch:
Thanks @pelwell 
Relates to #2073 and I guess this supersedes #2134 ?

pelwell:
It's exactly the same patch. Did somebody really expect a speedy review from the original author? Just merge it already.

aallan:
> Did somebody really expect a speedy review from the original author? 

No, but I was expecting someone from the software team to tell me that #2134 was correct and good to merge, I didn't know whether they were right or not?

aallan (approved):

pelwell:
> I was expecting someone from the software team to tell me that #2134 was correct and good to merge

How were we supposed to know of its existence?

aallan:
I'd assigned it to @ghollingworth for review. I think I emailed him as well.

PR #2153, opened 2021-09-14
Adding classes for floats and widths

nelliemckesson:
@aallan this should fix the floats issue. You're right that it needed styles, but there's also some funniness in how asciidoctor applies the classes/styles to the converted HTML. The solution that made the most sense to me was to add some custom classes to control the width of elements (the built-in asciidoctor width attributes don't work in this case, which I'm happy to get into but don't want to overwhelm you...). There are 9 new role names/classes (`w10`, aka 10% of the container width, up to `w90`, aka 90% of the container width) that you can apply to asciidoc images along with either `float: left` or `float: right`, to get the desired effect.

See the asciidoc in this PR for an example of how to use them.


aallan:
Thanks @nelliemckesson this works great. A slight problem, one which "no, that's not solvable, live with it" is a perfectly reasonable answer, is that headers are also wrapped. Is there any way to stop wrapping when we hit a header block. So that once we get to a header the text is back to "normal"..? i.e. this doesn't happen? 

Screenshot 2021-09-14 at 09 21 04

PR #2154, opened 2021-09-14
Adding non-JS TOC

### LINKED ISSUE #2013: ToC non-functional with Javascript disabled
In #1935 @Gadgetoid complains that the new site is "a little bit completely broken without Javascript" which I was kinda surprised by.
So I gave it a try myself with JS turned off, and it seems that https://www.raspberrypi.org/documentation/ works okay (with the exception of the search box, because that's entirely implemented by javascript), https://www.raspberrypi.org/documentation/accessories/ works okay, and the _contents_ of e.g. https://www.raspberrypi.org/documentation/accessories/display.html works okay.
AFAICT it seems that the only bit that doesn't work is the exapanding and collapsing of sections in the Table of Contents on the left? _Perhaps_ we should tweak things so that **if** JS is disabled, the ToC is fully-expanded rather than being fully-collapsed?

nelliemckesson:
@aallan this ought to fix #2013 . I've used the proposed solution to apply the `hidden` class on load via JS, and also tweaked some other JS bits so that the TOC will be scrollable, etc. Note that this will probably nullify the weird-scroll-issue fix, but only if JS is turned off (seems like an acceptable sacrifice to me).

### DIFF
--- a/jekyll-assets/_includes/nav.html
+++ b/jekyll-assets/_includes/nav.html
@@ -9,8 +9,26 @@
       {% if item.sections[0] %}
         {% if page.url == item.path %}
         <div class="itemcontents" id="toc"></div>
+        <div class="itemcontents noJS">
+          <ul class="sectlevel1">
+            {% for entry in item.sections %}
+              <li>
+                <a href="{{ site.baseurl }}{{ item.path }}#{{ entry.anchor }}">{{ entry.heading }}</a>
+                {% if entry.subsections[0] %}
+                  <ul>
+                    {% for subentry in entry.subsections %}
+                      <li>
+                        <a href="{{ site.baseurl }}{{ item.path }}#{{ subentry.anchor }}">{{ subentry.heading }}</a>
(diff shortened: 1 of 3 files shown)


jekyll-assets/_includes/nav.html
```diff
@@ -9,8 +9,26 @@ <h3 class="contentshead" onclick="expandAndCollapse(this)">{{ item.title }}</h3>
       {% if item.sections[0] %}
         {% if page.url == item.path %}
         <div class="itemcontents" id="toc"></div>
+        <div class="itemcontents noJS">
```
nelliemckesson, line 12: If JS is enabled, the first div here (the empty one) will be auto-filled by the tocify library. This new div is a backup only if JS is disabled.

aallan (commented):
LGTM

aallan (approved):

lurch:
Just given this a bit of testing - seems to work great! :tada:

PR #2156, opened 2021-09-15
Prevent extraneous HTML files being created

lurch:
This is a "bugfix" relating to #2072


aallan (approved):

aallan:
What was the bug? ...and good to merge?

PR #2177, opened 2021-09-30
Security and other fixes

aallan:
Merged the `dependabot` alerts into one branch and had a look. Seems to look okay, so I'm going to go ahead and merge into `develop` unless someone else sees a problem?


lurch (approved):
I just cloned this branch locally, did a `bundle install` and then `make`, and it produces an identical `documentation/html/` directory as before :+1:

PR #2226, opened 2021-10-19
Remove narrow motor from Build HAT support

jncraton:
The "narrow motor" in the Batmobile is the 6138854 motor already included in this list.


aallan:
Paging @topshed for comment.

Issue #2122, opened 2021-08-23
Inconsistency/inaccuracy in fixed focus distance?


leoshmu:
Hello, 

I wanted to double check something. In the Official Raspberry PI Camera Guide (https://magpi.raspberrypi.org/books/camera-guide/pdf), the v2 module is listed as being in focus from 50cm to infinity, implying a hyperfocal distance of 1m. However in the documentation table on this github repo, the table lists a focus of 1m to infinity for the v1 camera, and presumably that number applies to the v2 camera too. That would imply a hyperfocal distance of 2m.

Should the table in the repo list hyperfocal distance =1 m and then both sources are consistent?

Thanks for considering

aallan:
Ping @davidplowman.

aallan:
@leoshmu Is this answered by #2102 by any chance?

leoshmu:
@aallan thanks for the suggestion, but I don't think so, I didn't see an update to the documentation regarding fixed focus distance in that commit. In that commit the camera specification is here: https://github.com/davidplowman/documentation/blob/new-camera-doc/documentation/asciidoc/accessories/camera/camera_hardware.adoc and it still says fixed focus 1m to infinity for v1, a and a blank for v2 (i think indicating that v1 and v2 are same), and then something else for HQ camera. 

The issue I'm wondering about is that this documentation disagrees with the official camera module book published by raspberry pi foundation, which in chapter 14 states that the focus is 50cm to infinity. 

The stock position of the lens is I think in the hyperfocal position - for a lens focused at hyperfocal distance H, the image will be in focus at a distance H/2 to infinity. That is why I'm wondering if either the official camera module book incorrectly assumed that the hyperfocal distance was 1m and concluded that therefore focus is 50cm to infinity, or if this repository wrote 1 m to infinity fixed focus when it meant to say 1m hyperfocal distance. 

Your response is much appreciated, hoping we can figure this out!

Best, 
Leo

davidplowman:
Hi Leo, yes I agree that it's not quite consistent, so we should sort that out. As you have seen, we're doing a lot of work on documentation, though it's more a case of writing new stuff than fixing old stuff, but we'll put this on the list of things to do!

leoshmu:
Thanks David, 
[...]

PR #2261, opened 2021-11-01
camera: Minor clarifications on depth of field

davidplowman:
These are very small corrections, prompted by https://github.com/raspberrypi/documentation/issues/2122.

The table row marked "Fixed focus" is really giving a depth of field, so I've renamed it.

For the v1 sensor I believe "1m to infinity" is about right, but I've added "approx." because there is a significant element of judgement as regards what you believe "in focus" means.

For the v2 sensor I've mentioned that it's adjustable. Because it was previously blank, I think people possibly inferred it to mean "the same as v1", which isn't true. Though in its default configuration, it shouldn't be very different (again, up to a subjective interpretation of what that means!).

### DIFF
--- a/documentation/asciidoc/accessories/camera/camera_hardware.adoc
+++ b/documentation/asciidoc/accessories/camera/camera_hardware.adoc
@@ -117,9 +117,9 @@ Then, please follow the relevant setup instructions either for xref:camera.adoc#
 |
 |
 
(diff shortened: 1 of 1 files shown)


aallan (approved):

leoshmu:
Thanks so much for updating this.

PR #2263, opened 2021-11-01
Fix Boot Partition Size Typo

SonicMastr:


documentation/asciidoc/computers/getting-started/sd-cards.adoc
```diff
@@ -3,7 +3,7 @@
 
 Raspberry Pi computers use a micro SD card, except for very early models which use a full-sized SD card.
 
-WARNING: Because of a hardware limitation in the Raspberry Pi Zero, 1 and 2, the boot partition on the SD card must be 256GB or less otherwise the device will not boot up. Later models of Raspberry Pi 2 — with a BCM2837 SoC — along with the Raspberry Pi 3, 4, Zero 2, and the Raspberry Pi 400 do not have this limitation. This does not affect Raspberry Pi OS, which always uses a small boot partition.
+WARNING: Because of a hardware limitation in the Raspberry Pi Zero, 1 and 2, the boot partition on the SD card must be 256MB or less otherwise the device will not boot up. Later models of Raspberry Pi 2 — with a BCM2837 SoC — along with the Raspberry Pi 3, 4, Zero 2, and the Raspberry Pi 400 do not have this limitation. This does not affect Raspberry Pi OS, which always uses a small boot partition.
```
lurch, line 6: I don't think this `256MB` limit is correct, otherwise NOOBS (which in the non-Lite version has/had a multi-gigabyte boot partition) would never have worked? :thinking:

lurch:
ping @ghollingworth

documentation/asciidoc/computers/getting-started/sd-cards.adoc
SonicMastr: Looking at the partition configuration files used by NOOBS to install different images, all boot partitions are 256MB or smaller:

http://downloads.raspberrypi.org/raspbmc/partitions.json
http://downloads.raspberrypi.org/openelec/partitions.json
http://downloads.raspberrypi.org/raspbian_full/partitions.json
http://downloads.raspberrypi.org/riscos/partitions.json

Also, NOOBS doesn't work in a conventional manner to other Distros, instead taking advantage of a built in check to force the Pi into recovery mode which then ultimately loads NOOBS

ghollingworth:
Agree with @lurch, the limit is 256G for a single FAT partition otherwise you wouldn't be able to boot NOOBS.

PR #2270, opened 2021-11-05
Various tweaks and tidyups

lurch:
* Use the {cpp} macro consistently https://docs.asciidoctor.org/asciidoc/latest/attributes/character-replacement-ref/
 * Get rid of the redundant :pp: definitions
 * Get rid of leading blank lines
 * Add missing spaces in a few places
 * Fix bullet-points in the Pico "Blink an LED" example
 * Change [,shell] and [,console] to [,bash] for consistency

### DIFF
--- a/documentation/asciidoc/accessories/build-hat.adoc
+++ b/documentation/asciidoc/accessories/build-hat.adoc
@@ -1,4 +1,3 @@
-
 include::build-hat/introduction.adoc[]
 
 include::build-hat/preparing-build-hat.adoc[]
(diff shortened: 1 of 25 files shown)


aallan:
Where do you define the `{cpp}` macro?

lurch:
It's built-in, which is why I included that asciidoctor link :slightly_smiling_face:

aallan:
Huh. Learn something new every day! 🤷‍♂️

aallan (approved):

PR #2276, opened 2021-11-09
Update updating page now that Bullseye is released

lurch:
ping @aallan and @spl237


JamesH65:
LGTM, but there are awful lot of others places in the docs that are going to need updating for Bullseye. Lots of config.txt stuff no longer works, for example.

aallan (approved):
Even assuming there is more, we should at least do this!

PR #2281, opened 2021-11-11
Update overclocking.adoc

timg236:
Document arm_boost


aallan:
Do we really want to add an entire (almost identical) column to that table? There is only one change. A `[NOTE]` block below the table might make more sense?

timg236:
> Do we really want to add an entire (almost identical) column to that table? There is only one change. A `[NOTE]` block below the table might make more sense?

You could say the same thing about Pi400 though, or alternatively, arm_boost turns a Pi4B into a Pi400

aallan (approved):

lurch:
> arm_boost turns a Pi4B into a Pi400

That would be quite an impressive Transformers-style trick! :laughing: :robot:

PR #2229, opened 2021-10-22
Updates following feedback from LEGO

topshed:
Updated compatibility list
Changed refs to LPF2

PR #2286, opened 2021-11-15
 updated compat doc

topshed:
This supersedes [#2229 ](https://github.com/raspberrypi/documentation/pull/2229). Updated and revised compatibility. All links switched from BrickOwl to BrickLink.

### DIFF
--- a/documentation/asciidoc/accessories/build-hat/compat.adoc
+++ b/documentation/asciidoc/accessories/build-hat/compat.adoc
@@ -1,48 +1,46 @@
 == Device Compatibility
 
-The Build HAT library supports all the LEGO® Technic™ devices included in the SPIKE™ Portfolio, along with those from the LEGO® Mindstorms Robot Inventor kit and other devices that use an LPF2 connector.
+The Build HAT library supports all the LEGO® Technic™ devices included in the SPIKE™ Portfolio, along with those from the LEGO® Mindstorms Robot Inventor kit and other devices that use an PoweredUp connector.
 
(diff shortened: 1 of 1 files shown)


aallan (approved):
LGTM

aallan:
Although @topshed should that `Link` be in the column to the left of where it is right now..? 

Screenshot 2021-11-15 at 16 44 34

documentation/asciidoc/accessories/build-hat/compat.adoc
```diff
@@ -1,48 +1,46 @@
 == Device Compatibility
 
-The Build HAT library supports all the LEGO® Technic™ devices included in the SPIKE™ Portfolio, along with those from the LEGO® Mindstorms Robot Inventor kit and other devices that use an LPF2 connector.
+The Build HAT library supports all the LEGO® Technic™ devices included in the SPIKE™ Portfolio, along with those from the LEGO® Mindstorms Robot Inventor kit and other devices that use an PoweredUp connector.
```
lurch, line 3: nitpick: "use an PoweredUp" -> "use a PoweredUp" ?

aallan: We probably need to add a note about LPF2 in somewhere. As far as I can see this is now the first place it's called PowerUp.

topshed: LPF2 is an internal name only - PoweredUp is the correct term we should be using.

PR #2293, opened 2021-11-18
Add : Usage command <rm> with <-i> option

RuffaloLavoisier:
I added an option to prevent accidents that accidentally delete important files when using the rm command.


aallan:
Going forward we're deprecating the more generic documentation around using the Linux operating system, ahead of removing these sections entirely at some point in the future as part of a larger update to the documentation site. 

This move is happening as we feel these sort of more general topics are, ten years on from when the documentation was initially written, now much better covered elsewhere on the web. 

As such, unless we're not accepting PRs against these sections unless they're correcting errors. Apologies.

RuffaloLavoisier:
@aallan Okay, thank you.

PR #2303, opened 2021-11-24
Update camera_auto_detect usage in the camera documentation.

naushir:
To manually override the automatic overlay detection, we must delete the
camera_auto_detect parameter from config.txt.

### DIFF
--- a/documentation/asciidoc/accessories/camera/libcamera_apps_getting_started.adoc
+++ b/documentation/asciidoc/accessories/camera/libcamera_apps_getting_started.adoc
@@ -57,6 +57,6 @@ To override the automatic camera detection, _Bullseye_ users will also need to d
 
 `camera_auto_detect=1`
 
-if present.
+if present in the `config.txt` file. If you need to do edit this file then your Raspberry Pi will need to be rebooted. 
 
-If you need to edit this file then the Pi will need to be rebooted.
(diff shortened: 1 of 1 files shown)


lurch:
Oh, does `camera_auto_detect=0` not work as you might expect it to? Is that something that'll be getting "fixed"?

lurch:
(This also appears in the config.txt documentation https://www.raspberrypi.com/documentation/computers/config_txt.html#camera_auto_detect )

naushir:
> Oh, does `camera_auto_detect=0` not work as you might expect it to? Is that something that'll be getting "fixed"?

 `camera_auto_detect=0` disables the boot time detection completely. This would be used specifically for users that want boot times to be as low as possible, and want to avoid any i2c probing.

lurch (approved):

naushir:
> (This also appears in the config.txt documentation https://www.raspberrypi.com/documentation/computers/config_txt.html#camera_auto_detect )

Yes, I did find that reference in my grep of the tree.  That snippet is still valid and needs no change, but do you think I should explicitly say about removing it if using a manual overlay?

lurch:
I think it should be fine as-is - I guess anybody interested in the nitty-gritty will be reading the camera-specific pages anyway?

aallan (approved):
LGTM

aallan:
I think its fine as is @naushir. If they want more detail they should go read the camera-specific section.

PR #2309, opened 2021-12-01
FKMS and KMS

### LINKED ISSUE #2271: Display can now work as separate monitor
https://github.com/raspberrypi/documentation/blob/develop/documentation/asciidoc/accessories/display/display_intro.adoc says that the touch display cannot work as a separate X monitor, but with Pi 4 and Buster, it now can. I have tested, and it has worked.

aallan:
Fixes #2271.

### DIFF
--- a/documentation/asciidoc/accessories/display/display_intro.adoc
+++ b/documentation/asciidoc/accessories/display/display_intro.adoc
@@ -18,9 +18,9 @@ xref:display.adoc#legacy-support[Legacy support for Raspberry Pi 1 Model A/B]
 
 LCD displays have an optimum viewing angle, and depending on how the screen is mounted it may be necessary to change the orientation of the display to give the best results. By default, the Raspberry Pi Touch Display and Raspberry Pi are set up to work best when viewed from slightly above, for example on a desktop. If viewing from below, you can physically rotate the display, and then tell the system software to compensate by running the screen upside down.
 
-==== FKMS Mode
+==== KMD and FKMS Mode
 
-FKMS mode is used by default on the Raspberry Pi 4B. FKMS uses the DRM/MESA libraries to provide graphics and 3D acceleration.
+KMD and FKMS mode is used by default on the Raspberry Pi 4B. KMD and FKMS uses the DRM/MESA libraries to provide graphics and 3D acceleration.
 
 To set screen orientation when running the graphical desktop, select the `Screen Configuration` option from the `Preferences` menu. Right click on the DSI display rectangle in the layout editor, select Orientation then the required option.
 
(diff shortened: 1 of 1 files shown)


JamesH65:
LGTM

documentation/asciidoc/accessories/display/display_intro.adoc
```diff
@@ -18,9 +18,9 @@ xref:display.adoc#legacy-support[Legacy support for Raspberry Pi 1 Model A/B]
 
 LCD displays have an optimum viewing angle, and depending on how the screen is mounted it may be necessary to change the orientation of the display to give the best results. By default, the Raspberry Pi Touch Display and Raspberry Pi are set up to work best when viewed from slightly above, for example on a desktop. If viewing from below, you can physically rotate the display, and then tell the system software to compensate by running the screen upside down.
 
-==== FKMS Mode
+==== KMS and FKMS Mode
 
-FKMS mode is used by default on the Raspberry Pi 4B. FKMS uses the DRM/MESA libraries to provide graphics and 3D acceleration.
+KMS and FKMS modes are used by default on the Raspberry Pi 4B. KMS and FKMS use the DRM/MESA libraries to provide graphics and 3D acceleration.
```
lurch, line 23: Is it worth a brief explanation of the difference between KMS and FKMS? Or is that beyond the scope of this documentation? :shrug:

aallan: Do we have an explanation for it somewhere else in the documentation? Because if not, we probably should do.

lurch: I had a quick look (before posting the previous comment) and the nearest I could find were:
https://www.raspberrypi.com/documentation/computers/configuration.html#fake-or-full-kms-graphics-driver
https://www.raspberrypi.com/documentation/accessories/display.html#fkms-mode
neither of which are _particularly_ informative. ping @JamesH65

lurch: And even https://www.raspberrypi.com/news/raspberry-pi-os-debian-bullseye/ mentions KMS and Legacy, without explaining where FKMS fits in? :confused:

JamesH65:
FKMS is "Fake" KMS, it provides the KMS API to 2D graphics (HW composition etc, NOT 3D) and display, but instead of controlling the hardware, it makes calls in to the firmware on the Videocore which in turns does the HW access. In KMS, all the HW accesses are made from ARM space, so the firmware is not involved. It means the entire thing is now OSS, and all control of attached HW like DPI and DSI displays is in the kernel space, so people can now write their own drivers etc.

documentation/asciidoc/accessories/display/display_intro.adoc
JamesH65: FKMS is "Fake" KMS, it provides the KMS API to 2D graphics (HW composition etc, NOT 3D) and display, but instead of controlling the hardware, it makes calls in to the firmware on the Videocore which in turns does the HW access. In KMS, all the HW accesses are made from ARM space, so the firmware is not involved. It means the entire thing is now OSS, and all control of attached HW like DPI and DSI displays is in the kernel space, so people can now write their own drivers etc.

JamesH65: Legacy is where everything is controlled by the firmware, no standard KMS API. The firmware takes complete control of all the hardware (HVS, 3D etc) and this is accessed using fairly non-standard mechanisms, via VCHIQ and some ARM side shim libraries.

aallan:
Sort of feels like we should have a section talking about this somewhere between "HDMI Configuration" and "Rotating your Display in configuration.adoc?

lurch:
With maybe a nice picture explaining how things fit together? And could maybe also explain how different Pis and different OS versions default to using different graphics stacks? :shrug:

PR #2319, opened 2021-12-15
Fix JavaScript error on pages with a TOC

mudge:
When loading a documentation page with a table of contents (e.g.  https://www.raspberrypi.com/documentation/computers/getting-started.html), the following JavaScript error is thrown in the console:

    TypeError: e.indexOf is not a function. (In 'e.indexOf(" ")', 'e.indexOf' is undefined)

This is caused by a unsupported mix of dependencies for Tocify: specifically, we're attempting to load jQuery 3.6.0 which removed an API used by Tocify (see https://github.com/gfranko/jquery.tocify.js/issues/103).

Fix this by loading versions of jQuery and jQuery UI that are supported by Tocify.


aallan (approved):

lurch:
Ah! I always wondered why I kept seeing that error in the JS console. Nice one @mudge :+1:

PR #2322, opened 2021-12-16
Push to production

aallan:


lurch:
Now that this has been pushed live, it looks like the text for the "NOTE" might be a different colour than the text for the links? Or perhaps I'm getting old and my eyes are just playing tricks on me? :man_shrugging: 
Screenshot from 2021-12-16 13-28-15

ping @mudge

mudge:
@lurch if you inspect that element (a `<div>` with class `title`), you'll see it has a `color` of `var(--red)` (which is set to `#cd2355`) which is the same as the `color` for `<a>` tags.

aallan:
Could be the [NOTE] being in bold just makes it look more red?

lurch:
Yeah, probably just the font anti-aliasing that's confusing my eyes :eyes:

PR #2323, opened 2021-12-16
Fixed small typo

mcmarto:


JamesH65:
I believe the original is correct given the context?

mcmarto:
Ah, my mistake - I must have misread it originally. Apologies. I'll close this.

PR #2329, opened 2022-01-02
Navigate to parent directory before pip install

kevinfreeman1101:
Guide instructed users to navigate to the `documentation` folder, but forgot to have users `cd` to the parent folder before instructing pip to install `requirements.txt`.

### DIFF
--- a/README.md
+++ b/README.md
@@ -96,6 +96,7 @@ $ bundle install
 ```
 (which may take several minutes), followed by,
 ```
+$ cd ..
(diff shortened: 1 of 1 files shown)


lurch:
This is correct as-is, I think you're simply confusing the top-level `documentation` directory you get when you clone the repo (which is what the instructions are referring to), with the `documentation` directory that's _inside_ the repo :wink: 

However I'll leave this PR open for now to remind me to add some clarification to the README.

lurch:
See #2346

Issue #1987, opened 2021-08-09
Search bar only returns the first 10 results


aallan:
The built-in search bar will return only the first 10 results and there is no visual indicator that only the first 10 results are being shown…

<img width="1399" alt="Screenshot 2021-08-09 at 15 10 06" src="https://user-images.githubusercontent.com/53625/128720207-98974736-215f-4d70-9d56-8b25c37ce0d5.png">

…need to think about how to work around this UI/UX problem.

andrum99:
I hadn't clocked that the drop-down list of pages wasn't just a preview. Based on how Chrome works, I was expecting the list that appears as you type to just be a preview, then the actual proper search would then take place once you pressed enter or clicked on the search button.

aallan:
> I hadn't clocked that the drop-down list of pages wasn't just a preview. Based on how Chrome works, I was expecting the list that appears as you type to just be a preview, then the actual proper search would then take place once you pressed enter or clicked on the search button.

This might be the solution to carrying a full index of the site around on all of the pages. We can carry our current (fairly small) index around, and then if the user pushes "Enter" then we do a full text search (rather than a section heading search)? Maybe? Thoughts from @lurch, @mudge, and @nelliemckesson welcome here about how practical that is?

lurch:
I'm afraid "adding a search function to a site" is something I have no familiarity/experience with! But I'm happy to add any toolchain / index-generation support code that might be needed for that.

I guess if we're adding a "press `Enter` to do a full-search" feature, then we ought to make that more explicit / obvious by adding a `Search` button to the RHS of the search-box? (which would do the same thing as pressing `Enter`)

aallan:
> I'm afraid "adding a search function to a site" is something I have no familiarity/experience with! 

Yeah, I think that bit is more of a @nelliemckesson or perhaps @mudge thing. It's possible we could third-party it and use a Google Custom Search Engine — the look&feel of the return results can be customised a fair bit by CSS.

> But I'm happy to add any toolchain / index-generation support code that might be needed for that.

Thanks!

> I guess if we're adding a "press `Enter` to do a full-search" feature, then we ought to make that more explicit / obvious by adding a `Search` button to the RHS of the search-box? (which would do the same thing as pressing `Enter`)

Yup.

andrum99:
> It's possible we could third-party it and use a Google Custom Search Engine — the look&feel of the return results can be customised a fair bit by CSS.

[...]

PR #2315, opened 2021-12-08
WIP Google Programmatic Search Engine for search

mudge:
In order to provide a better searching experience, experiment with replacing the Simple Jekyll Search-powered search box with one from Google Programmatic Search Engine.

PR #2333, opened 2022-01-04
Replace Simple Jekyll Search with Algolia DocSearch

mudge:
In order to provide a better search experience and address some of the usability issues raised in https://github.com/raspberrypi/documentation/issues/1987, replace our Simple Jekyll Search-powered search box with one provided by [Algolia DocSearch][0].

We use the default, [out of the box theme provided by DocSearch v3][1] as customising it will involve rewriting/overriding the CSS provided directly by Algolia or overriding the default options in JS to render our own markup. As it is, my hope is that it provides a better user experience than the current search box.

  [0]: https://docsearch.algolia.com
  [1]: https://docsearch.algolia.com/docs/styling

Screenshot 2022-01-04 at 16 18 30


aallan:
Alternative to #2315.

lurch:
I'll leave any functionality and Look'n'Feel comments to @aallan and @jackbenwillis (other than to say that this obviously gives **much** better search results than our previous header-only-search did), but from a technical PoV the code-changes here (including removal of the old static search) look fine :+1:  Well done @mudge with the Python and Ninjabuild hacking.

The only slight drawback (which I guess is unavoidable with _any_ cloud-based search system) is that when I'm viewing http://127.0.0.1:4000/documentation/ for local testing, the search results take me directly to a page on https://www.raspberrypi.com/documentation/ (so we'll just have to remember to double-check which version we're looking at when doing local testing).

mudge:
Is there any value in getting this deployed without building out our own custom look and feel (which is [not fully documented by Algolia](https://docsearch.algolia.com/docs/styling))? The Crawler has been running daily with success and it might be that the improved search is worth the non-Raspberry Pi branded UI?

If we were to try to customise the appearance, I'd recommend we stick to customising only the [official variables exposed by DocSearch](https://github.com/algolia/docsearch/blob/next/packages/docsearch-css/src/_variables.css) which is mostly colours and sizing rather than anything more significant like disabling the modal altogether, etc.

aallan:
> Is there any value in getting this deployed without building out our own custom look and feel (which is not fully documented by Algolia)? The Crawler has been running daily with success and it might be that the improved search is worth the non-Raspberry Pi branded UI?

That sounds like a @jackbenwillis + me decision rather than just a me decision. I'll bring it up in the Comms standup this morning. 

> If we were to try to customise the appearance, I'd recommend we stick to customising only the official variables exposed by DocSearch which is mostly colours and sizing rather than anything more significant like disabling the modal altogether, etc.

I'd be happy to deploy it immediately so long as we get the colours sorted out so it looks brand-ish, and I guess we have to try and make it uses Roboto font (if that's possible, if it doesn't already).

mudge:
@aallan @jackbenwillis here's a version with the DocSearch primary colour set to Raspberry Red (#cd2355) and the Algolia logo set to our red tint (#d75a64).

https://user-images.githubusercontent.com/287/151948525-eed1ba4e-6c7c-49fb-8df8-b2ab8200d177.mov

aallan (approved):
I think that's deployable now, and we can talk about doing further customisation later if needed.

PR #2345, opened 2022-01-12
Check that all internal links and anchors are valid. Fixes #2332

### LINKED ISSUE #2332: Investigate incorrect anchors
I'm not sure how #2328 "slipped through the net", so investigate it further and possibly add a script to check that there are no other incorrect anchors.

lurch:
WRT the "slipping through the net" - some of these were genuine typos, and some were because in the old MarkDown version of the documentation, each H1 (top heading on the page) was a valid anchor, whereas now only H2 and below are valid anchors.

The link-checking is done as part of the "build nav-menu" script, which means any future link-typos will cause the build to fail, and be caught by the CI.


aallan (approved):
LGTM

RuffaloLavoisier:
Great !

PR #2348, opened 2022-01-13
Typo in example command

j-calvert:
Took better part of an hour to track down that this was the cause of 

```
[a2702fc8] main stream debug: creating access: stream://dev/stdin
[a2702fc8] main stream debug: looking for access module matching "stream": 29 candidates
[a2702fc8] main stream debug: no access modules matched
```


davidplowman:
Thanks for spotting this, LGTM!

aallan (approved):

PR #2378, opened 2022-01-24
Add Thin clients to the site footer

mudge:
While updating the "For industry" section of the footer with a new page, use sentence case more consistently.


aallan (approved):
LGTM

PR #2390, opened 2022-02-01
Override DocSearch colours with Raspberry Pi brand

mudge:
To make sure the Algolia DocSearch widget matches our brand colours, update it to use Raspberry Red as its primary colour and use our red tint for the Algolia logo.


aallan (approved):
Oops.

PR #2392, opened 2022-02-01
Remove duplicated words

RuffaloLavoisier:
Is this the right way to fix it? 
# Please let me know if I can improve it more!


lurch:
I think in this case the original was correct? @aallan ?

aallan:
So"that that" is not great wording, and personally I try and avoid it in written English with judicious rewording, but it is correct.

lurch:
Thanks again for all your other PRs though @RuffaloLavoisier :heart:

RuffaloLavoisier:
@lurch @aallan 
Thank you for checking 😆

PR #2364, opened 2022-01-18
Adding Build HAT .NET documentation

Ellerbach:
Adding Build HAT .NET documentation.

- There is a native .NET support for Build HAT. See https://github.com/dotnet/iot/tree/main/src/devices/BuildHat
- Adding the basic elements in the existing Build HAT documentation
- Adjusting the TOC and the name of the existing Python page to have a good navigation

image

PR #2402, opened 2022-02-03
Adding .NET support

aallan:
Rough initial rework of #2364, for discussion. Mostly stylistic issues, and removing references to Windows.

Ping @Ellerbach and @topshed.


topshed:
We do have official photos from LEGO of the Spike components that we could use - they are more consistent with each other than the ones in this PR I think. We also use them on the Build HAT Python librarty readthedocs pages.

Other than that, looks good.

aallan:
> We do have official photos from LEGO of the Spike components that we could use - they are more consistent with each other than the ones in this PR I think. We also use them on the Build HAT Python librarty readthedocs pages.

Do you want to checkout the `dotnet-for-lego` branch and commit them on branch? Alternatively you could email them to me if you like? Thanks!

aallan:
Probably merge to `develop` mid-afternoon today if there aren't any substantive objections.

aallan:
Pushed live in #2408

Ellerbach:
> Pushed live in #2408

Great! Happy I'll help .NET developers! Interested in Sense HAT as well?

aallan:
> > Pushed live in #2408
> 
> Great! Happy I'll help .NET developers! Interested in Sense HAT as well?

Looking at the current Sense HAT documentation it's from the "pre-me-doing-documentation" times, so if you look at https://www.raspberrypi.com/documentation/accessories/sense-hat.html#sense-hat-software you can see that what we have is a paragraph (or so) and pointers to external documentation for the Python and RTIMULib libraries.

At some point we need to fix that and document the Sense HAT properly, but that time is not today.

For now, I'd be happy to get a PR that would fit into that scheme. So a section to drop into the "Sense HAT Software" section pointing to the .NET library documentation that is hosted elsewhere (the library's own Github page?).

PR #2407, opened 2022-02-07
UART docs:  Add CM3 and CM3+ default UART assignments

JinShil:


lurch:
If we're adding CM3 and CM3+, it probably makes sense to add the details for CM4 at the same time? ping @dp111

dp111:
I think so. The section is getting a little complex; it might be worth
having a review to simplifying it.

aallan:
There is currently a separate section for the Compute Module, https://www.raspberrypi.com/documentation/computers/configuration.html#cm-cm-3-cm-3-and-cm-4. If we're adding any CM information to the summary table, then this information also needs to go there as well?

aallan:
I agree this needs sorting out in the long term, I'm taking a note, so that's on the list. But I'm merging this change now.

aallan (approved):

Issue #2047, opened 2021-08-11
Suggestion: add automated spell checker to Github workflow


andrum99:
The recent wave of typo fixes has brought into focus the problem of typos in the documentation. I would propose that an automated spell checker be added to the Github workflow. Such a checker would not catch all typos, but would handle those where the entered word was not in the dictionary. It would not catch atomic typos - typos where the entered word **_is_** a dictionary word. Nor would it handle the case where completely the wrong word has been used. Implementing an automated checker will reduce the amount of time developers spend on correcting typos in the documentation.

I had a go a few months ago at implementing such a system on one of my own Github repos, but did not get very far. It is, however, easy to run `aspell` on a local system with a suitable dictionary, and add words to a custom dictionary as you go along. It should therefore be fairly straightforward for someone with suitable skills to integrate spell checking into a Github workflow.

lurch:
Yeah, it's something that we've discussed internally, but it's not yet bubbled up to the top of our priority list :wink: :abc: 

Instead of using `aspell` (which will find a bunch of false positives in the markup etc.), a much better option is [codespell](https://github.com/codespell-project/codespell).

(sidenote: https://github.com/codespell-project/codespell/search?q=author%3Alurch&type=issues :rofl: )

andrum99:
> Yeah, it's something that we've discussed internally, but it's not yet bubbled up to the top of our priority list 😉 🔤
> 
> Instead of using `aspell` (which will find a bunch of false positives in the markup etc.), a much better option is [codespell](https://github.com/codespell-project/codespell).
> 
> (sidenote: https://github.com/codespell-project/codespell/search?q=author%3Alurch&type=issues 🤣 )

I ran aspell on some of the asciidoc files in this repo and the amount of false positives wasn't too bad. The problem with codespell is that unless a word has an entry in its dictionary, the typo will be ignored. (If I've understood correctly from the readme). Whereas what I want it to do is flag ever occurrence of a word that is not in an en_GB dictionary.

aallan:
I'm (mildly) against automating spelling and grammar checks, my experience with doing it in the past hasn't been great. With technical documentation the false postive rate generally is just too high to make it useful. They tend to add to the editing workload rather than reduce it.

andrum99:
> I'm (mildly) against automating spelling and grammar checks, my experience with doing it in the past hasn't been great. With technical documentation the false postive rate generally is just too high to make it useful. They tend to add to the editing workload rather than reduce it.

[...]

PR #2422, opened 2022-02-11
Add codespell

andrewimeson:
This PR adds code spell in a mode where it will annotate what it thinks are
typos on the PR (but not fail the build). It's configured with very conservative
options that I believe should result in a low number of false positives. 

The typo fixes have been split into #2426 by request

This was discussed in #2047 - there was general opposition to the idea _in
principle_. I offer this PR in the hopes that it will illustrate how a slightly
different approach answers the (very valid) objections:

1. Codespell doesn't do a dictionary whitelist based approach (i.e. whatever I
   don't have in my dictionary is wrong), but rather maintains a map of common
   misspellings. In technical documentation, this should cut down on false
   positives greatly
2. Codespell merely annotates the PR when it thinks that a word is wrong, and
   this leaves the option to fix the word (if it's right), add the word to the
   exemption list, or even just ignore the warning and take no action (perhaps
   cleaned up later)
3. If it identified 18 spellings errors (15 distinct words), and had only 3
   false positives (excluding the unorthodox brand spelling "Technic") that had
   to be exempted, I think that suggests a pretty good ratio. If we included the
   words caught in past `codespell`-fixing PRs (#364, 
   #329), the ratio would be even higher. 

Finally, there are **20** merged that mention "spelling," and **171** that
mention "typo." I can't say that codespell could've prevented all of them, but
it's reasonable to expect it could've caught some of them.

### DIFF
--- a/.codespellrc
+++ b/.codespellrc
@@ -1,0 +1,6 @@
+[codespell]
+skip = Gemfile,Gemfile.lock,jekyll-assets,build
+# For a list of available built-in dictionaries, see:
+#  https://github.com/codespell-project/codespell/tree/master/codespell_lib/data
+builtin = clear,rare,informal
+ignore-words-list = technic,ba,hist
--- a/.github/workflows/codespell.yml
+++ b/.github/workflows/codespell.yml
@@ -1,0 +1,21 @@
+---
+name: Check for common misspellings
+
+on:
+  push:
+  pull_request:
+    branches: [develop]
+
+jobs:
+  codespell:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v2
+
+      - name: Check spelling with codespell
+        # Configuration should be handled in .codespellrc as much as possible so
+        # that it runs similarly locally
+        uses: codespell-project/actions-codespell@master
+        with:
+          # Annotate the build, but don't fail it
+          only_warn: 1


documentation/asciidoc/accessories/camera/libcamera_options_vid.adoc
```diff
@@ -90,15 +90,15 @@ Example: `libcamera-vid -t 0 -o test.h264 -k --initial pause`
 	--split				Split multiple recordings into separate files
 ----
 
-This option should be used in conjunction with `--keypress` or `--signal` and causes each recording session (inbetween the pauses) to be written to a separate file.
+This option should be used in conjunction with `--keypress` or `--signal` and causes each recording session (in between the pauses) to be written to a separate file.
```
andrewimeson, line 93: Should this be "between," "in-between," or "in between"?

aallan:
As discussed in #2047, the team here are against adding automated spell and grammar checking to the repository.

aallan:
Can you split out the copyedit into a separate PR to the toolchain changes? Thanks.

andrewimeson:
@aallan I've split it into two PRs.

I think that this approach solves most of the concerns in #2047. However, I'm not a person maintaining the repository (I only have three PRs here) and dealing with whatever benefits or drawbacks this creates.

I think codespell has a unique approach to the problem that results in few false positives (I've used it in a couple repositories, although not in any documentation-heavy ones).

lurch:
@andrewimeson I think it'd be too burdensome to have to keep updating the `ignore-words-list` to have this automatically run on **every** CI-build. Would you mind "refactoring" this PR to instead add a `make codespell` target to the `Makefile`? That way it's something that only (sporadically) gets run manually, by an actual human :slightly_smiling_face:  (who can then PR whatever fixes they deem necessary).

andrewimeson:
@lurch I could do that (might have to brush up on Makefile), but it's worth noting that (with some napkin-quality shell calculations) that this repository has:

- 132,538 words
- 6,388 distinct words
- 4,327 commits

only needed 3 word exempted in the repo-level .codespellrc

helenlynn (commented):
If I were responsible for proofreading everything in this repo, I would lean very slightly towards using this tool, provided that the effort involved in adding and maintaining it were trivial; it would provide a first pass that might occasionally be useful, and that would save some few moments of proofreading work each time I looked at anything. But I'd still need to review everything to catch all the errors that codespell can't, so the saving would be negligible. The kind of typos and spelling errors it catches are the kind that take no perceptible effort to spot and negligible time to fix manually. Since @LizUpton is more strongly opposed to the use of tools like this, as she explained in #2047, her opinion is likely to carry it.

LizUpton:
Afraid this is a no, at least for the time being. We're starting the hiring process for a dedicated tech editor in March, and our strong preference is to have a human doing this work.

PR #2443, opened 2022-02-21
Fix Python sensors sample

Taytay:
The Python sensors code doesn't compile and is incomplete. 
This PR fixes it while staying as true to the original source as possible.

(In particular, the existing source code does not hook up the pressed and released handlers and references `c` instead of `cs`, leading to a compile-time error.)


lurch:
ping @topshed

topshed:
All good.  Thanks for the correction.

aallan (approved):
@topshed said "Yes!"

PR #2452, opened 2022-03-01
Access-point-routed.adoc : add private IP address range

RuffaloLavoisier:
If the external and internal IP ranges are the same, add a Private IP range that can be config.

If it's not suitable for the discussion, I'd appreciate it if you could close it. thanks

### DIFF
--- a/documentation/asciidoc/computers/configuration/access-point-routed.adoc
+++ b/documentation/asciidoc/computers/configuration/access-point-routed.adoc
@@ -152,6 +152,8 @@ address=/gw.wlan/192.168.4.1
 
 The Raspberry Pi will deliver IP addresses between `192.168.4.2` and `192.168.4.20`, with a lease time of 24 hours, to wireless DHCP clients. You should be able to reach the Raspberry Pi under the name `gw.wlan` from wireless clients.
 
+NOTE: In the personal IP network range ,There are three types: `10.0.0.0 to 10.255.255.255`,`172.16.0.0 to 172.31.255.255` and `192.168.0.0 to 192.168.255.255`. Adjust the range accordingly and use it.
+
 There are many more options for `dnsmasq`; see the default configuration file (`/etc/dnsmasq.conf`) or the http://www.thekelleys.org.uk/dnsmasq/doc.html[online documentation] for details.
 
 [[wifi-cc-rfkill]]
--- a/documentation/asciidoc/computers/configuration/access-point-routed.adoc
+++ b/documentation/asciidoc/computers/configuration/access-point-routed.adoc
@@ -152,7 +152,7 @@ address=/gw.wlan/192.168.4.1
 
 The Raspberry Pi will deliver IP addresses between `192.168.4.2` and `192.168.4.20`, with a lease time of 24 hours, to wireless DHCP clients. You should be able to reach the Raspberry Pi under the name `gw.wlan` from wireless clients.
 
-NOTE: In the personal IP network range ,There are three types: `10.0.0.0 to 10.255.255.255`,`172.16.0.0 to 172.31.255.255` and `192.168.0.0 to 192.168.255.255`. Adjust the range accordingly and use it.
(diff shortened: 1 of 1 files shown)


documentation/asciidoc/computers/configuration/access-point-routed.adoc
```diff
@@ -152,7 +152,7 @@ address=/gw.wlan/192.168.4.1
 
 The Raspberry Pi will deliver IP addresses between `192.168.4.2` and `192.168.4.20`, with a lease time of 24 hours, to wireless DHCP clients. You should be able to reach the Raspberry Pi under the name `gw.wlan` from wireless clients.
 
-NOTE: In the personal IP network range ,There are three types: `10.0.0.0 to 10.255.255.255`,`172.16.0.0 to 172.31.255.255` and `192.168.0.0 to 192.168.255.255`. Adjust the range accordingly and use it.
+NOTE: In the personal IP network range, There are three types: `10.0.0.0 to 10.255.255.255`,`172.16.0.0 to 172.31.255.255` and `192.168.0.0 to 192.168.255.255`. Adjust the range accordingly and use it.
```
aallan, line 155: Shouldn't be a comma there in the first place, and there shouldn't be a capital letter after a comma. But putting both of those aside, that isn't even a sentence. Maybe this,

```
NOTE: There are three IP address blocks set aside for private networks. There is the Class A block, from `10.0.0.0` to `10.255.255.255`, a Class B block from `172.16.0.0` to `172.31.255.255`, and probably the most frequently used, a Class C block from `192.168.0.0` to 192.168.255.255`.
```

But that said, I'm not sure how relevant or useful the `NOTE` is anyway? 🤷‍♂️

RuffaloLavoisier: @aallan 
IMHO, I think it's good if it's a guide for the beginner tutorial.

aallan (approved):

documentation/asciidoc/computers/configuration/access-point-routed.adoc
aallan: Fair

RuffaloLavoisier: @aallan 
Thank you for reflecting it :smile:

documentation/asciidoc/computers/configuration/access-point-routed.adoc
```diff
@@ -152,6 +152,8 @@ address=/gw.wlan/192.168.4.1
 
 The Raspberry Pi will deliver IP addresses between `192.168.4.2` and `192.168.4.20`, with a lease time of 24 hours, to wireless DHCP clients. You should be able to reach the Raspberry Pi under the name `gw.wlan` from wireless clients.
 
+NOTE: There are three IP address blocks set aside for private networks. There is the Class A block, from `10.0.0.0` to `10.255.255.255`, a Class B block from `172.16.0.0` to `172.31.255.255`, and probably the most frequently used, a Class C block from `192.168.0.0` to 192.168.255.255`.
```
lurch, line 155: "the Class A block" -> "a Class A block", for consistency with the descriptions of the Class B and Class C blocks?

RuffaloLavoisier: @lurch 
Thanks, https://github.com/raspberrypi/documentation/pull/2453

epoch1970:
- That concern is already addressed in the beginning of the page, in “before you start”.
- The addition would be better inserted near the beginning of “Define the Wireless Interface IP Configuration”
- The notion of network class is outdated https://en.wikipedia.org/wiki/Classful_network

Revert?

lurch:
Meh, this is only a single-line change, IMHO it's not worth reverting.

> The notion of network class is outdated https://en.wikipedia.org/wiki/Classful_network

Again I'm going to say "meh" https://en.wikipedia.org/wiki/Private_network#Private_IPv4_addresses

PR #2456, opened 2022-03-02
Headless.adoc : mention Raspberry Pi Imager advanced option

RuffaloLavoisier:


lurch:
IMHO the fix for this needs some rewording, rather than just adding a link, as the "Adding certain files to this folder..." part is still potentially misleading.

RuffaloLavoisier:
>although we probably do still want to mention the non-Imager customisation, in case people don't want to use Imager for some reason?

@lurch 

presenting various methods in the guidelines, I think users can experience themselves and cope with their situation. People who have already been doing this for a long time without that help will be more so.

PR #2467, opened 2022-03-05
Update the Build HAT .NET instructions

Taytay:
The previous instructions appeared to make assumptions that the reader had created a donet program before. 
This change clarifies:
* How to create a dotnet project
* How to run it and ensure that things are oworking
* That you should be in the project directory in order to add the nuget packages
* That the nuget packages should not point to the nightly IoT build, but to version 2.1.0 (otherwise in its current state there are errors upon installation)

(I'm still working my way through these instructions and have a few more improvements before this PR is ready)

### DIFF
--- a/documentation/asciidoc/accessories/build-hat/net-installing-software.adoc
+++ b/documentation/asciidoc/accessories/build-hat/net-installing-software.adoc
@@ -12,19 +12,21 @@ $ wget -O - https://raw.githubusercontent.com/pjgpetecodes/dotnet5pi/master/inst
 ----
 
 After installing the .NET framework you can create your project:
+
 [.bash]
 ----
 $ dotnet new console --name buildhat
----
+----
 
 This creates a default program in the `buildhat` directory, so we go into that directory and run the program.
 It should print `Hello World!`
+
 [.bash]
----
+----
 $ cd buildhat
 $ dotnet run
 Hello World!
----
+----
 
 You will now need to install the following nuget packages:
 [.bash]
--- a/documentation/asciidoc/accessories/build-hat/net-installing-software.adoc
+++ b/documentation/asciidoc/accessories/build-hat/net-installing-software.adoc
@@ -48,17 +48,4 @@ Hello World!
 === Editing C# Code
 In the instructions below, you will be editing the file `buildhat/Program.cs`, the C# program which was generated when you ran the above commands.
 
-Any text editor will work to edit C# code, including Geany, the IDE/Text Editor that comes pre-installed in Raspbian. However, https://code.visualstudio.com/[Visual Studio Code] (often called "VS Code") is a popular alternative. It can be installed by running:
-
-[.bash]
-----
-$ sudo apt get install code
-----
-
-Once installed, you can open your current folder in VS Code by running:
-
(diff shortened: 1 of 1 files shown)


lurch:
> The previous instructions appeared to make assumptions that the reader had created a donet program before.

If the user _hadn't_ created a dotnet program before, why would they be using the Build HAT .NET instructions? :thinking: 

I believe we recommend that (new) users use the Build HAT Python library (as that's supported by us) rather than the .NET library (as that's supported by a third party: @Ellerbach ). Perhaps that needs clarifying @aallan ? I'm fairly sure that "How to create a dotnet project" falls outside of the Raspberry-Pi focused documentation in this repo? :shrug:

aallan:
The recommended path for using the Build HAT is the Python library. The .NET library was written and is supported by a third party. We should probably add a `NOTE` somewhere near the start to make that clearer. 

I accepted the PR to add documentation around it since it was a solid chunk of content that could help .NET developers using the Build HAT. However, I'm not entirely convinced we're in the business of documenting how to get started with .NET, the documentation isn't about that. That said, I'm not sure how much background documentation is necessary here, there could well be too little here. It's a balancing act.

aallan:
Does anyone have any comments on this, @topshed perhaps?

Taytay:
Perhaps my comment about "assuming the user had created a dotnet program before" was distracting. The bulk of these and other changes are required to make the existing instructions work and not throw an error.

As for the intended audience, you all know more about that than I do, but I'll let you know where I'm coming from: I don't use Python or .NET in my day job, but the build hat is the perfect excuse for me to get to play with a Pi and Legos and some languages I don't otherwise get to use. I was thrilled that there was a dotnet library, because I had been looking for an excuse to get to mess with it again. And this is literally my first .NET program since I used it professionally 15 years ago, so I might have projected my confusion onto other readers.

I thought the audience for these instructions was potentially high schoolers making robots, for whom a Build HAT was their first real project, so I wasn't going to tell them what a command line was, but I endeavoring to make it copy-pastable without errors and followable without knowing off the top of their head what it means to "create a new project" or how/where/when to install a nuget package.

No offense taken if I'm dumbing it down too much or taking the documentation in a direction that you didn't intend!

aallan:
> Perhaps my comment about "assuming the user had created a dotnet program before" was distracting. The bulk of these and other changes are required to make the existing instructions work and not throw an error.

Ah! That's more enlightening. I'm not really awfully familiar with the dotnet environment, and that cuts to the core of it. Thanks.

aallan:
I've marked the PR ready for review, however you mentioned you might have some more tweaks to add before a prospective merge @Taytay?

aallan (approved):
YOLO

lurch:
@Taytay @Ellerbach  I just tried running through these instructions myself https://www.raspberrypi.com/documentation/accessories/build-hat.html#using-the-build-hat-from-net

I _think_ they're missing an instruction about adding
```
using Iot.Device.BuildHat;
```
near the top of `Program.cs` ? Without it, I get an error about "The type or namespace name 'Brick' could not be found (are you missing a using directive or an assembly reference?)" :confused:

PR #2466, opened 2022-03-04
Textify table images

andrum99:
Rebirth of #2462- see #2458.

PR #2470, opened 2022-03-07
Support Asciidoctor table alignment operators

mudge:
GitHub: https://github.com/raspberrypi/documentation/pull/2466

In order to support AsciiDoctor's table alignment operators (see https://docs.asciidoctor.org/asciidoc/latest/tables/align-by-column/), we need to port across the relevant CSS rules from the default stylesheet.

As we already override the default text alignment for table headers, we increase the specificity of each rule so that alignment operators will take precedence.


aallan (approved):
YOLO

PR #2478, opened 2022-03-10
Technical_specification.adoc : inline sentence

RuffaloLavoisier:


lurch:
The first commit seems to be just collapsing entire paragraphs into single lines? I'm not sure we want that (and it has zero affect on the generated HTML).

RuffaloLavoisier:
Ah..like other documents, will follow the rules to remove line changes.

lurch:
I think (for the most part) we only want to include changes where they actually have an affect on the documentation that's visible at https://www.raspberrypi.com/documentation/
We don't want to include changes "just for the sake of making changes", because that's a fairly Sisyphean task.

Issue #2507, opened 2022-04-11
Revision codes for Pi4 rev1.5 are missing


Martin-HiPi:
In the table from [New-style Revision Codes in Use](https://www.raspberrypi.com/documentation/computers/raspberry-pi.html#new-style-revision-codes-in-use), the revision codes for the seen-in-the-wild Pi4B rev1.5 is absent.

Pi4B 2GB rev1.5:
```
Hardware        : BCM2711
Revision        : b03115
Serial          : <16 hex chars>
Model           : Raspberry Pi 4 Model B Rev 1.5
```

PR #2509, opened 2022-04-12
Added the rev1.5 Pi4's

JamesH65:


RuffaloLavoisier:
See #2507

aallan:
Shouldn't the two v1.5 lines be together, rather than one before the v1.1 and one at the 'end' of that block?
Screenshot 2022-04-12 at 10 31 14

lurch (approved):

JamesH65:
I just followed what was already there, seems to be sorted on GB.

PR #2517, opened 2022-04-23
Update installing-from-chromeos.adoc

JeremyJedynak:
As of the 2022-04-07 linked archive, the Raspberry Pi OS images are in XZ format and no longer in Zip format. There appears to be a bug in the Recovery Utility where it recognizes XZ files as image archives, but then the next screen blinks by quickly to the "success" screen without actually writing the image to the device. A workaround until the Recovery Utility is fixed is to first convert the XZ archive to Zip format by 'unxz'ing the original archive and then 'zip'ing a new archive. The Recovery Utility correctly writes the new zip archived image to the device. I tested this workaround with the raspios_full_armhf-2022-04-07 archive.

### DIFF
--- a/documentation/asciidoc/computers/getting-started/installing-from-chromeos.adoc
+++ b/documentation/asciidoc/computers/getting-started/installing-from-chromeos.adoc
@@ -3,11 +3,12 @@
 The easiest way to write images to an SD card and USB drives with Chrome OS is to use the official https://chrome.google.com/webstore/detail/chromebook-recovery-utili/jndclpdbaamdhonoechobihbbiimdgai[Chromebook Recovery Utility]. It can be used to create Chromebook Recovery media, and it will also accept `.zip` files containing images.
 
 * Download the https://chrome.google.com/webstore/detail/chromebook-recovery-utili/pocpnlppkickgojjlmhdmidojbmbodfm[Chromebook Recovery Utility].
-* Download the https://www.raspberrypi.com/software/operating-systems/#raspberry-pi-os-32-bit[Raspberry Pi OS] zip archive.
+* Download the https://www.raspberrypi.com/software/operating-systems/#raspberry-pi-os-32-bit[Raspberry Pi OS] zip or xz archive.
+* If you download a xz archive, you will need to first convert it to a zip archive to proceed with the remainder of these steps.
 * Launch the *Recovery Utility*
 * Click on the *Settings Gears* icon in the upper right-hand corner, next to the window close icon.
 * Select the *Use Local Image* option.
-* Choose the `.zip` file you downloaded.
+* Choose the `.zip` file archive.
 * Insert the SD card and click *Continue*.
 * Read the warning and click the *Create now* button.
 * Wait for the progress bar to complete twice (for unpacking and writing). This might take a few minutes. Once the process is complete, a big green checkmark will be shown.


JeremyJedynak:
I also tried to report the apparent bug in the Chromebook Recovery Utility by emailing chromeos-recovery-tool-admin at google.com, which is the "Contact the developer" link on the chrome web store page for the utility: https://chrome.google.com/webstore/detail/chromebook-recovery-utili/pocpnlppkickgojjlmhdmidojbmbodfm?hl=en

I received the following bounce message:

=======
We're writing to let you know that the group you tried to contact (chromeos-recovery-tool-admin) may not exist, or you may not have permission to post messages to the group. A few more details on why you weren't able to post:

 * You might have spelled or formatted the group name incorrectly.
 * The owner of the group may have removed this group.
 * You may need to join the group before receiving permission to post.
 * This group may not be open to posting.

If you have questions related to this or any other Google Group, visit the Help Center
=======

Without a definitive way to report the apparent bug to the developers of the Chromebook Recovery Utility, the XZ-to-Zip archive conversion workaround might be needed for a while.

lurch:
I suspect that anybody who's using Chromebook Recovery Utility instead of (the recommended) Raspberry Pi Imager, might need more detailed instructions on how to "If you download a xz archive, you will need to first convert it to a zip archive" ? :thinking:  (I've never used ChromeOS myself, so I've no idea how complicated / easy this would be :shrug: )

JeremyJedynak:
@lurch - On a Chromebook, the Chromebook-native Recovery Tool is the preferred imaging tool because it has lower-level access to the USB device to be imaged, and can both image and wipe a USB device. Currently, the only way to run the Raspberry Pi Imager on a Chromebook is via Chrome OS Linux, but the USB device is only made available to the Linux VM via a mount point (/mnt/). This allows reading and writing of files but isn't the same access level as the Chromebook-native Recovery Tool has, and so the Raspberry Pi Imager won't display the mounted USB device in the "Choose Storage" dropdown.

Here are the commands for transcode the archive within Chrome OS Linux with the appropriate packages installed:

```
unxz --keep original_image_archive.xz
zip new_image_archive.zip image.img
```

lurch:
> Here are the commands to transcode the archive within Chrome OS Linux...

Cool, could you add those to your changes in this PR? (and apologies for another stupid question, but how likely is it that a Chromebook user has "Chrome OS Linux" and "the appropriate packages" installed ?)

aallan:
> how likely is it that a Chromebook user has "Chrome OS Linux" and "the appropriate packages" installed ?)

My guess is pretty much zero. If these instructions are to work they should be from whatever default Chrome OS install comes with the laptop. Potentially the best move is to remove this section entirely.

aallan:
I've decided to remove non-Imager installation paths. See #2518.

JeremyJedynak:
@aallan - I'm not sure whether you have tried it, but the official Raspberry Pi Imager is not available for ChromeOS, which is why this documentation regarding the Chromebook Recovery Tool is important to those who want to use a Chromebook to image a Raspberry Pi.  I haven't seen any documentation regarding the Recovery Tool bug or the workaround that I tested and then suggested here. I ask you to please reconsider closing this pull request.

@lurch - Every Chromebook user would benefit from the apparent bug in the Chromebook Recovery Tool being repaired, but unfortunately we don't appear to have a way to report that bug to the Recovery Tool developers. A Chromebook user who has familiarity with a Raspberry Pi linux command prompt has all of the skills necessary to enable Chrome OS Linux and install the same tools via apt. The simple documentation edits suggested in my pull request are minimally sufficient for such a user to understand what they need to do, without my referencing the apparent bug in the Chromebook Recovery Tool.

PR #2530, opened 2022-05-03
Fix sensor name in maximum exposure table

phsilva:
Sensor name is incorrect in the maximum exposure time table.


phsilva:
I also think the value itself could be wrong, as the default value in libcamera[1] is actually 250 but I couldn't test neither find available documentation. The datasheet publicly available is incomplete.

[1] https://git.libcamera.org/libcamera/libcamera.git/tree/src/ipa/raspberrypi/raspberrypi.cpp#n65

lurch:
ping @davidplowman

davidplowman:
Yes, that should be 477.

lurch (approved):

PR #2542, opened 2022-05-17
video.adoc: Document requirement for overclock for certain 4K modes

andrum99:
https://github.com/raspberrypi/linux/pull/5038 makes certain 4K modes (and anything beyond) conditional on the BCM2711 core being overclocked to at least 600MHz - document this requirement.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -46,7 +46,7 @@ html: run_ninja
 
 # Build the html output files and additionally run a small webserver for local previews
 serve_html: run_ninja
-	$(JEKYLL_CMD) serve --host=0.0.0.0
+	$(JEKYLL_CMD) serve
 
 # Delete all the files created by the 'html' target
 clean_html:
--- a/Makefile
+++ b/Makefile
@@ -46,7 +46,7 @@ html: run_ninja
 
 # Build the html output files and additionally run a small webserver for local previews
 serve_html: run_ninja
-	$(JEKYLL_CMD) serve
+	$(JEKYLL_CMD) serve --host=0.0.0.0
 
 # Delete all the files created by the 'html' target
 clean_html:
--- a/documentation/asciidoc/computers/config_txt/video.adoc
+++ b/documentation/asciidoc/computers/config_txt/video.adoc
@@ -826,13 +826,13 @@ These values are valid if `hdmi_group=1` (CEA):
 | 4096x2160
 | 50Hz
 | 256:135
-| Pi 4<<super1>>
+| Pi 4<<needsoverclock>>
 
 | 102
 | 4096x2160
 | 60Hz
 | 256:135
-| Pi 4<<super1>>
+| Pi 4<<needsoverclock>>
 
 | 103
 | 2160p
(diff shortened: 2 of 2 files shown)


andrum99:
This is a draft, since https://github.com/raspberrypi/linux/pull/5038 is also a draft, so may change or disappear.

JamesH65:
Might be Github, but <sup>1</sup> is rendering as "<sup>1</sup>" not the superscript wanted.

JamesH65:
> Might be Github, but 1 is rendering as "1" not the superscript wanted.

Hahahah, so, it actually rendered correctly in the comment. That's funny right there

`<sup>1</sup> is rendering as "<sup>1</sup>"`

andrum99:
> Might be Github, but 1 is rendering as "1" not the superscript wanted.

Thanks - I need to set up a local build environment again - back in the markdown days that worked, but I believe we're sanitising HTML now, or something.

lurch:
https://docs.asciidoctor.org/asciidoc/latest/text/subscript-and-superscript/

andrum99:
OK - try now. I'm open to suggestion as to whether or not the bold is required.

documentation/asciidoc/computers/config_txt/video.adoc
```diff
@@ -826,13 +826,13 @@ These values are valid if `hdmi_group=1` (CEA):
 | 4096x2160
 | 50Hz
 | 256:135
-| Pi 4
+| Pi 4<<super1>>
```
lurch, line 829: `super1` is a bit of a meaningless link-name. `needsoverclock` or similar might be better?

documentation/asciidoc/computers/config_txt/video.adoc
```diff
@@ -865,6 +865,8 @@ These values are valid if `hdmi_group=1` (CEA):
 | Pi 4
 |===
 
+[[super1,^**1**^]] **1.** Only available with an overclocked core:set `core_freq_min=600` and `core_freq=600`. See xref:config_txt.adoc#overclocking[overclocking].
```
lurch, line 868: "core:set" should probably be "core frequency - set"

documentation/asciidoc/computers/config_txt/video.adoc
andrum99: Fixed

andrum99:
Removed draft flag as https://github.com/raspberrypi/linux/pull/5038 is about to integrate.

lurch:
Can you revert the change to `Makefile` please?

Makefile
```diff
@@ -46,7 +46,7 @@ html: run_ninja
 
 # Build the html output files and additionally run a small webserver for local previews
 serve_html: run_ninja
-	$(JEKYLL_CMD) serve
+	$(JEKYLL_CMD) serve --host=0.0.0.0
```
aallan, line 49: Revert needed.

aallan (changes requested):

andrum99:
Apologies folks - I had not realised I had left this in a state where it could not be easily integrated. IT should be ready to roll now.

I guess I need to stop using `git add -A` 😕

aallan (approved):

aallan:
Everyone happy? Ready to merge?

lurch:
I just had a quick scan through https://www.raspberrypi.com/documentation/computers/config_txt.html and there doesn't seem to be any other places (on that page) where we've used a superscript number like this. I _wonder_ if it might be better to just add a "needs overclock" (alongside the "Pi 4" text) in the existing "Notes" column:
Screenshot from 2022-07-05 00-55-41
...and then add a NOTE: block underneath the table explaining the overclocking requirement? :man_shrugging:

andrum99:
> I just had a quick scan through https://www.raspberrypi.com/documentation/computers/config_txt.html and there doesn't seem to be any other places (on that page) where we've used a superscript number like this. I _wonder_ if it might be better to just add a "needs overclock" (alongside the "Pi 4" text) in the existing "Notes" column ...and then add a NOTE: block underneath the table explaining the overclocking requirement? man_shrugging

That sounds a bit like a recipe for disaster - having the explanation split in two. We have ued superscript elsewhere in the docs, so there is precident.

lurch:
Up to @aallan ...

PR #2544, opened 2022-05-19
video.adoc: Pi 4-related fix up

andrum99:
Pi 4 does not enable composite when HDMI is disabled.

### DIFF
--- a/documentation/asciidoc/computers/config_txt/video.adoc
+++ b/documentation/asciidoc/computers/config_txt/video.adoc
@@ -1606,7 +1606,7 @@ Setting `hdmi_force_hotplug` to `1` pretends that the HDMI hotplug signal is ass
 
 ==== `hdmi_ignore_hotplug`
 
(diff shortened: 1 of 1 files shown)


pelwell:
No - it's common behaviour, and the edit doesn't really change the sense, just making clear that the setting is all about HDMI (although it may cause composite to be used instead).

andrum99:
By 'Pi 4' I of course mean 'BCM2711-based Raspberry Pis'. The intent was to correct the documentation, while simplifying. If you would prefer me to specify the products which do not enable composite upon HDMI being disabled then I'm happy to do that.

andrum99:
To clarify: the current page is wrong because it states that composite output will be enabled if the Pi is run without HDMI output enabled. This is not the case for devices based on BCM2711. My edit simply removes reference to composite from this parameter, and clarifies the effect the parameter has on the HDMI output.

Issue #2537, opened 2022-05-11
Update references to the default user/pass


aallan:
Now the default user has been removed we need to update all references to a default user/pass,

```
$ git grep -E "\`(pi|raspberry)\`" "*.adoc" | grep -Ev "default (user|pass)"
documentation/asciidoc/computers/configuration/access-point-routed.adoc:If
SSH is enabled on the Raspberry Pi, it should be possible to connect
to it from your wireless client as follows, assuming the `pi` account
is present: `ssh [pi@192.168.4.1](mailto:pi@192.168.4.1)` or `ssh [pi@gw.wlan](mailto:pi@gw.wlan)`
documentation/asciidoc/computers/configuration/raspi-config.adoc:The
`sudo` is required because you will be changing files that you do not
own as the `pi` user.
documentation/asciidoc/computers/configuration/raspi-config.adoc:From
this submenu you can select whether to boot to console or desktop and
whether you need to log in or not. If you select automatic login, you
will be logged in as the `pi` user.
documentation/asciidoc/computers/configuration/securing-the-raspberry-pi.adoc:Once
you have confirmed that the new account is working, you can delete the
`pi` user. In order to do this, you'll need to first change the
autologin user to your new user `alice`, with the following:
documentation/asciidoc/computers/configuration/securing-the-raspberry-pi.adoc:Please
note that with the current Raspberry Pi OS distribution, there are
some aspects that require the `pi` user to be present. If you are
unsure whether you will be affected by this, then leave the `pi` user
in place. Work is being done to reduce the dependency on the `pi`
user.
documentation/asciidoc/computers/configuration/securing-the-raspberry-pi.adoc:To
delete the `pi` user, type the following:
documentation/asciidoc/computers/configuration/securing-the-raspberry-pi.adoc:This
command will delete the `pi` user but will leave the `/home/pi`
folder. If necessary, you can use the command below to remove the home
folder for the `pi` user at the same time. Note the data in this
folder will be permanently deleted, so make sure any required data is
stored elsewhere.
documentation/asciidoc/computers/configuration/securing-the-raspberry-pi.adoc:This
command will result in a warning that the group `pi` has no more
members. The `deluser` command removes both the `pi` user and the `pi`
group though, so the warning can be safely ignored.
documentation/asciidoc/computers/configuration/securing-the-raspberry-pi.adoc:and
change the `pi` entry (or whichever usernames have superuser rights)
to:
[...]

PR #2552, opened 2022-06-02
add note about the removal of default pi user

digiltd:
this caught me out, been using the default user details for years on all my headless Pi, was stumped when they were not working, especially when the docs made no mention of it. Discovered the news from the forum https://forums.raspberrypi.com/viewtopic.php?t=335261


aallan:
The [getting started](https://www.raspberrypi.com/documentation/computers/getting-started.html#using-raspberry-pi-imager), and [headless](https://www.raspberrypi.com/documentation/computers/configuration.html#setting-up-a-headless-raspberry-pi) have already been updated.

This will get updated as part of #2537.

PR #2557, opened 2022-06-09
Added revision code for CM4S

cr-chsn1:


cr-chsn1:
Quick question: Why won't this be fixed or should be added?
This is the revision code of the new CM4S (https://www.raspberrypi.com/products/compute-module-4s/).

PR #2563, opened 2022-06-20
[doc] Pointed out that default `pi` user is now removed

mfloryan:
Linked to: https://www.raspberrypi.com/news/raspberry-pi-bullseye-update-april-2022/


aallan:
When we're fixing up #2537 we'll be removing and rewording all mentions of a default username/password. There will probably be a note block up front in the getting started section mentioning that in previous releases there was a default username/password, rather than one at each instance.

PR #2578, opened 2022-07-06
Clarify quality setting

sirtet:
Just ran into the file size issue, and it seems other people regularly do so too. The Stack Overflow answer fixed it for me.
I think this issue is worth a correction in the docs.

### DIFF
--- a/documentation/asciidoc/accessories/camera/raspicam.adoc
+++ b/documentation/asciidoc/accessories/camera/raspicam.adoc
@@ -772,7 +772,7 @@ Retrieves the current camera settings and writes them to stdout.
 	--quality,	-q		Set JPEG quality <0 to 100>
 ----
 
-Quality 100 is almost completely uncompressed. 75 is a good all-round value.
+NOTE: The Quality value is NOT corresponding to typical JPEG Compression level values. A good allround value is -q 10. This is roughly equal to a JPEG Compression Level of 85. For details see https://stackoverflow.com/a/22722920/1331544
 
 ----
 	--raw,	-r		Add raw Bayer data to JPEG metadata


JamesH65:
There really isn't a standard JPEG compression level, it's very different across all difference encoders, so "JPEG compression level" isn't really a thing to be compared against.

lurch:
We prefer all our documentation to be self-contained where possible, without links to external sites.

sirtet:
> Additionally that Stack Overflow answer is from 2014 and talks about `raspistill` which is a legacy application deprecated in favour of the libcamera-apps? Not sure how long we'll continue to distribute and document these tools in any case.

Is it documented or hinted anywhere how long raspistill will be carried along?
I did not switch to libcamera because the [docs still kind of recommend raspistill](https://www.raspberrypi.com/documentation/accessories/camera.html#raspistill:~:text=may%20perform%20better%20on%20Raspberry%20Pi%202%20and%20Raspberry%20Pi%20Zero%20devices%2C%20as%20it%20offloads%20more%20to%20the%20GPU) for the zero.

> There really isn't a standard JPEG compression level, it's very different across all different encoders, so "JPEG compression level" isn't really a thing to be compared against.

I am aware. That's why i wrote "...typical..." Any suggestion for a better wording? 
It's obvious the current text recommending 75 as a good compromise value between compression and image quality is completely wrong. To get similar compression like other software achieves with around 75 (in the order of 10% file size, 90% visual quality), raspistill needs to be given something around 10. (See the link, and in it, the link to the raspi forum.)
(Using the suggested -q 75 results in ~90% file size.)
So, i think as the doc is way off from reality, it should be fixed. No?

> We prefer all our documentation to be self-contained where possible, without links to external sites.

Yes, i tried to find a concise wording, but could not get something, so i decided to suggest the link.
Any suggestion for a good text?
A different view: Maybe it's still preferable to have a link in the docs instead of a bug?

andrum99:
> Is it documented or hinted anywhere how long raspistill will be carried along? 

Not raspistill specifically, but since it is part of Raspberry Pi OS Legacy (see https://www.raspberrypi.com/news/new-old-functionality-with-raspberry-pi-os-legacy/), it would be reasonable to assume that it would be supported on that version of the OS for at least as long as that OS is being supported. The current generation of Raspberry Pi OS is a different matter. The best place to ask about this is probably on the forums at https://forums.raspberrypi.com/.

aallan:
No movement in a month. I'm marking this to close unless the OP wants to revise the PR.

sirtet:
>  I'm marking this to close unless the OP wants to revise the PR.

I am more of a casual user, so i don't feel like i could offer a much better alternative text.
I first just wanted to point at the problem, but then thought it would be more constructive to directly suggest an alternative.
I am aware it's not ideal, but can't really come up with something better on my own.

As i understand, the remarks were roughly 
"this is legacy anyways"
"preferably no external links"
So, if there is no dispute about the current text being way off from reality, how else could it be improved? What should be changed in my PR? Just remove the Link?

PR #2608, opened 2022-08-18
Conditional filters: update for CM4S

andrum99:
Recently a forum user (see https://forums.raspberrypi.com/viewtopic.php?t=338978) requested how to use a `config.txt` filter for CM4S boards. The solution is not currently well documented, so add this. Also clarify that the `[cm4]` filter does not apply to the CM4S, and fix a minor grammar fault. I also had to change a word in the paragraph beneath the table, since inserting the footnote to the table means the following paragraph needs the first sentence to be more specific.

### DIFF
--- a/documentation/asciidoc/computers/config_txt/conditional.adoc
+++ b/documentation/asciidoc/computers/config_txt/conditional.adoc
@@ -35,7 +35,7 @@ The conditional model filters are applied according to the following table.
 | Pi 400
 
 | [cm4]
-| Compute Module 4
+| Compute Module 4<<cm4clarification>>
 
 | [pi0]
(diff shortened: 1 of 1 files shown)


lurch:
The linked forum thread also implies that `[pi4]` does apply to CM4S ?

andrum99:
> The linked forum thread also implies that `[pi4]` does apply to CM4S ?

My reading of the thread was that it doesn't, but if you only read timg236's first reply, then it does imply that [pi4] applies to the CM4S. I think I need to ask on that thread for clarification.

aallan:
CM4S is a limited distribution product and won't be included in the general documentation.

PR #2613, opened 2022-09-02
Correct the git repository where we fetch libcamera

davidplowman:
For a while now we've been building from a Raspberry Pi fork of
libcamera, just so we can control extactly when it changes. The
documentation should reflect this location, not the official libcamera
repo.


lurch (approved):
LGTM

PR #2617, opened 2022-09-07
Update Memory.adoc

JmIsACoder:
The `disable_l2cache` section is described in more detail.

modified descriptions：GPUs L2 cache ---> GPUs system L2 cache. I think L2 cache and L2 system cache are two different things, the former refers specifically to what the CPU has, and the latter refers to what the CPU and GPU have in common. 

add notes：(BCM2711 additionally provides a 1MB system L2 cache, which is used primarily by the GPU)

### DIFF
--- a/documentation/asciidoc/computers/config_txt/memory.adoc
+++ b/documentation/asciidoc/computers/config_txt/memory.adoc
@@ -55,4 +55,4 @@ This value will be clamped between a minimum of 128MB, and a maximum of the tota
 
 === `disable_l2cache`
 
-Setting this to `1` disables the CPUs access to the GPUs L2 cache and requires a corresponding L2 disabled kernel. Default value on BCM2835 is `0`. On BCM2836, BCM2837, and BCM2711, the ARMs have their own L2 cache and therefore the default is `1`. The standard Raspberry Pi `kernel.img` and `kernel7.img` builds reflect this difference in cache setting.
+Setting this to `1` disables the CPUs access to the GPUs system L2 cache (BCM2711 additionally provides a 1MB system L2 cache, which is used primarily by the GPU). Default value on BCM2835 is `0`. On BCM2836, BCM2837, and BCM2711, the ARMs have their own L2 cache and therefore the default is `1`. The standard Raspberry Pi `kernel.img` and `kernel7.img` builds reflect this difference in cache setting.


lurch:
ping @pelwell

pelwell:
No, I think the original wording is correct. ARM accesses can be configured to either go via the GPU's L2 cache or to bypass it and go direct to system RAM. In some senses you could say it is then acting as a system cache, but that might make people think it is distinct to the GPU's own L2 cache (which it isn't).

Do you agree, @popcornmix ?

popcornmix:
The original text is far more accurate.

> (BCM2711 additionally provides a 1MB system L2 cache, which is used primarily by the GPU).

Nope.

JmIsACoder:
add link: BCM2711 provides a 1MB system L2 cache, which is used primarily by the GPU. Accesses to memory are routed
either via or around the L2 cache depending on the address range being used.(https://datasheets.raspberrypi.com/bcm2711/bcm2711-peripherals.pdf) @popcornmix 

Is this understanding correct?

Are there two L2s in the system, one shared by CPUs and the other can be shared by GPUs and CPUs?

pelwell:
Yes.

popcornmix:
The GPU L2 cache is 128KB. On Pi0/1 it is shared with the arm (as it doesn't have its own L2 cache),
but on later chips this doesn't occur (technically it could be configured that way, but would always be slower, as the arm has a larger, faster L2 cache so is never run that way).

I'd say the data sheet is confusing. They may have meant 1Mbit by 1MB, but that isn't the usual notation (in fact later on the same page they use 1GB to mean 1Gbyte).

lurch:
Oh! @pelwell is the "system L2 cache" on BCM2711 indeed 128kB rather than 1MB?

@popcornmix we're attempting to standardise our use of size-suffixes according to https://github.com/raspberrypi/style-guide/blob/master/style-guide.md so please feel free to point out any other discrepancies you spot.
Excerpted:
* kb – kilobit. No space between the number and kb.
* kB – kilobyte. No space between the number and kB.
* Mbit – megabit (not Mb, as this is too easily confused with MB)
* MB – megabyte (not MByte). No space between the number and MB. My old Amiga only had 130MB drive space.
* Gb – gigabit. No space between the number and Gb.
* GB – gigabyte. No space between the number and GB. You'll need an 8GB SD card.

JmIsACoder:
If the image below is correct, can we add it to the `disable_l2cache` section? I think it might be easier to understand by adding this image here.

请添加图片描述

popcornmix:
The picture is correct, but I wonder if drawing too much attention to a purely debug setting (`disable_l2_cache`) generates more confusion than it solves.

No one uses `disable_l2_cache=0` on Pi2-4, not should they ever try to. It is guaranteed to make your system go slower, if it doesn't break it completely. The GPU L2 cache is very far away from the arm and using it will cripple performance on anything apart from pi0/pi1 which has no L2 cache of its own.

JmIsACoder:
I have a doubt, why does disabling GPU L2 cache make the system slower? arm can use more cache resources, shouldn't the system performance be better? In other words, if the system uses the CPU intensively and the GPU is not sensitive, will enabling the GPU L2 cache improve system performance? A recent experiment I was doing on the performance of the Raspberry Pi proved this conclusion. Using hackbench to test, when the GPU memory is larger or the GPU L2 cache is enabled, the system performance will be better.

popcornmix:
The GPU L2 cache is far away from the ARM. It takes something like 17 AXI (i.e. core_clock, so slower than arm) clocks to read from it from the arm. And when the cache misses (which it will very frequently do at that size, especially as GPU is also using it), you still have to go to sdram after those 17 AXI cycles.

Give me numbers for a realistic benchmark (e.g. using ffmpeg to software decode a video) with `disable_l2_cache` on and off on a Pi4, and I'll be very surprised if it helps at all.

pelwell:
>  The GPU L2 cache is very far away from the arm

This means that it take many additional clock cycles for ARM accesses to travel via the GPU's L2 cache.

I find it strange that people think that we would choose not to enable a feature that actually improves performance. It's not even as though any code would need to be changed - just the default value.

JmIsACoder:
The reason for the doubt that GPU L2 cache will improve system performance is because of the following experiments.

We built the Xenomai3.2.1 environment on the Raspberry Pi 4B and tested it. 

> Xenomai is a real-time development framework for the Linux kernel. It hopes to provide full, interface-agnostic hard real-time performance to user-space applications by seamlessly integrating into the Linux environment.

> hackbench is a stress test tool for the Linux kernel scheduler. Its main job is to create a specified number of scheduling entity pairs (threads/processes), and let them transmit data through sockets/pipe, and finally count the time overhead of the entire running process .

System Info:

`Raspberry Pi 4 Model B Rev 1.2 Linux raspberrypi 5.15.51-v8 #1 SMP PREEMPT IRQPIPE Thu Aug 4 09:09:10 BST 2022 aarch64 GNU/Linux`

Test Scripts:

```
pi@raspberrypi:~$ cat bash xen3 hackbench.sh
# /bin/bash
export PATH=$PATH: /usr/xenomai/bin
export LD_LIBRARY_PATH=$LD_LIBRARY_PATH:/usr/xenomai/ib
cd rt-tests
nohup hackbench -s 65535 -l 65535 -P 1>/dev/null 2>&1 &
nohup latency -t0 -p 500 -P 90 -h -g ../Xenomai3_pi3_hackbench.txt 1>/dev/null 2>&1 &
```

Results:

**GPU_mem**

modify `config.txt`:
GPU_mem=XXX

```
hackbench  freq1800Mhz     runL2

GPU 16MiB
----lat min|----lat avg|----lat max|-overrun|---msw|
​      3.533|     31.897|     98.681|       0|     0|
GPU 64MiB
----lat min|----lat avg|----lat max|-overrun|---msw|
​      3.862|     22.024|     95.311|       0|     0|
GPU 256MiB
----lat min|----lat avg|----lat max|-overrun|---msw|
​       5.279|     16.767|     86.966|       0|     0|
```

**run or ban GPU L2 cache**

modify `config.txt`:
run GPU L2 cache: `disable_l2cache=0`
ban GPU L2 cache: `disable_l2cache=1`

```
hackbench      GPU 256MiB    freq1800Mhz

run L2 cache,freq=1800
----lat min|----lat avg|----lat max|-overrun|---msw|
​       5.279|     16.767|     86.966|       0|     0|
ban L2 cache,freq=1800
----lat min|----lat avg|----lat max|-overrun|---msw|
​       5.926|     30.591|     89.742|       0|     0|

```

PR #2645, opened 2022-10-14
rp2a0.adoc - grammar nits, correction to meaning

andrum99:


aallan:
Disagree with most of this.

PR #2657, opened 2022-10-18
device-tree.adoc: clarify which built-in USB controller

andrum99:
BCM2711 contains two USB controllers, so be more specific about which one we're talking about.


lurch:
ping @pelwell

pelwell:
What @andrum99 writes is correct. It's the same boot mode as enabled by a `5` in [`BOOT_ORDER`](https://www.raspberrypi.com/documentation/computers/raspberry-pi.html#BOOT_ORDER), so you may want to harmonise the language.

andrum99:
I'm looking at whether you would like this PR resurrected, but in a new PR with just the technical change, not the copy edit. Is this something that you would be interested in Alasdair?

Issue #2314, opened 2021-12-07
Update documentation for ahead of Bookworm


lurch:
~We now [offer both](https://www.raspberrypi.com/news/new-old-functionality-with-raspberry-pi-os-legacy/) Bullseye and Buster (Legacy) releases of Raspberry Pi OS. However some features work in one but not the other, and vice-versa.~

~@JamesH65 has an internal document which is probably a useful starting point for this.~

(edit by @aallan) We need to update the documentation for Bookworm. Online documentation will focus on supporting Bookworm only. There will be NOTE blocks referencing that "things worked differently" in the previous (legacy) version if that's appropriate/needed. We'll try and keep those blocks short.

lurch:
And perhaps https://www.raspberrypi.com/documentation/computers/getting-started.html#installing-the-operating-system and/or https://www.raspberrypi.com/documentation/computers/os.html#introduction ought to mention the different versions too?

URRG:
I want to work on this issue so from where should i start

JamesH65:
This is already underway internally, but it's a private repo.

JamesH65:
Sorry, closed by accident.

lurch:
This should stay open.

aallan:
Should stay open. Still some updating needed. See also #2398.

RuffaloLavoisier:
@aallan 
Are you simply updating the vlc and omx examples according to the version?

aallan:
> @aallan Are you simply updating the vlc and omx examples according to the version?

OMX Player no longer works under Bullseye. All the examples using it in the documentation need to be replaced with an equivalent invocation for VLC (or other relevant software), see #2398 which is the linked issue.

However this issue is the more general "other things that Bullseye broke" issue. There are some other things beyond the OMX Player issue that haven't yet been fixed.

RuffaloLavoisier:
If there is something to be done externally, I would be happy to contribute.

lurch:
This still needs doing, at some point.

JamesH65:
ping

aallan:
Headline things:

 
* X goes away and is replaced by Wayland
   - nominally a big thing, but there has been a lot of work to keep the user experience the same
   - but some change is inevitable
* dhcpcd goes away and is replaced by Network Manager
* random bookworm stuff

aallan:
[...]

PR #2663, opened 2022-10-21
Add note for libcamera2 in GPU mem section

timg236:
Note that libcamera2 doesn't use GPU memory so there really is no point in increasing the GPU memory size.

### DIFF
--- a/documentation/asciidoc/computers/config_txt/memory.adoc
+++ b/documentation/asciidoc/computers/config_txt/memory.adoc
@@ -24,8 +24,10 @@ The recommended maximum values are as follows:
 | `384`
 
 | 1GB or greater
-| `512`, `256` on the Raspberry Pi 4
+| `512`, `76` on the Raspberry Pi 4
 |===
+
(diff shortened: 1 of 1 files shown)


documentation/asciidoc/computers/config_txt/memory.adoc
```diff
@@ -24,9 +24,11 @@ The recommended maximum values are as follows:
 | `384`
 
 | 1GB or greater
-| `512`, `256` on the Raspberry Pi 4
+| `512`, `76` on the Raspberry Pi 4
 |===
 
+IMPORTANT: The default camera stack (libcamera2) on Raspberry Pi OS - Bullseye uses Linux CMA memory to allocate buffers instead of GPU memory so there is no benefit in increasing the GPU memory size.
```
lurch, line 30: I'm not sure if we need the "Bullseye" signifier here, as the assumption is generally that our documentation is just about the latest version of Raspberry Pi OS, which is currently Bullseye? :shrug:
(then again, there's also #2314 )  ping @aallan

aallan (approved):

documentation/asciidoc/computers/config_txt/memory.adoc
aallan: Think it's probably fine in context.

PR #2681, opened 2022-11-01
Update preparing-build-hat.adoc

bwynneHEP:
Based on my own testing, having 1-Wire bus enabled will prevent use of Build HAT. The image of the interfaces tab highlights the Serial settings only, which implies that the other settings can be adjusted independently, but this is not the case.


lurch:
GPIO4 **is** already listed in the [part](https://www.raspberrypi.com/documentation/accessories/build-hat.html#access-the-gpio-pins) that says "The following pins are used by the Build HAT itself and you should not connect anything to them."; and in the "Raspberry Pi Configuration" screenshot just below that table, 1-Wire is shown as disabled.
I'm sure there's probably several things that you could "enable" that would stop the Build HAT working properly? :thinking:

PR #2688, opened 2022-11-06
Add pin pitch info

kr15h:
In wonder why this information is so hard to find on the Raspberry Pi pages. Please add this at least here.


aallan (approved):

aallan:
Seems like a reasonable addition.

PR #2691, opened 2022-11-09
Add Mastodon link to footer

mudge:


aallan (approved):
Thanks Paul. I'll push it to production now!

PR #2697, opened 2022-11-14
Autoboot pseudo code

timg236:


documentation/asciidoc/computers/config_txt/autoboot.adoc
```diff
@@ -24,18 +24,50 @@ Set this property to `1` to load the normal `config.txt` and `boot.img` files in
...
 
 
-=== A/B boot example
-In the following example partition `3` would be contain the pending OS upgrade and would be tested by rebooting in `tryboot` mode (`sudo reboot "0 tryboot"`).
-If the OS determines that the upgrade was successful then it would replace `autoboot.txt` swapping the partition numbers.
-Otherwise, if the system is reboot (e.g. watchdog or cold boot) then system would boot from partition `2` as usual.
+=== Example update flow for A/B booting
 
-Since the `autoboot.txt` file is a single sector it will normally be possible to update this with a single sector update to the SD/EMMC.
+The following pseudo code shows how an hypothetical OS `Update Serivce` could use `tryboot` + `autoboot.txt` to perform an fail-safe OS upgrade.t
```
JamesH65, line 29: Typo. Serivce

timg236: Done + fixed other typo for Update Service

aallan (approved):

documentation/asciidoc/computers/config_txt/autoboot.adoc
```diff
@@ -24,18 +24,50 @@ Set this property to `1` to load the normal `config.txt` and `boot.img` files in
...
 boot_partition=2
 [tryboot]
 boot_partition=3
 ----
+
+**Installing the update**
+
+* System is powered on and boots to partition 2 by default.
+* An `Update Service` downloads the next version of the OS to partition 3.
+* The update is tested by rebooting to partition 3 in `tryboot` mode `reboot "3 tryboot"`
```
lurch, line 44: Does the number `3` in this line need to match the `[tryboot]` section in `autoboot.txt`? Just out of curiosity, why does the number appear in two places? I guess that means that after a successful update (and a correspondingly-updated `autoboot.txt`) the "Update service" would need to run `reboot "2 tryboot"` on the _next_ update cycle?

timg236: You would download the OS update to the inactive partition which in normal mode would be the one inside the [tryboot] section.  So download to 3, then reboot into 3 in "tryboot mode" to test 3. If all is good then autoboot.txt is swapped so 3 is the default.

PR #2703, opened 2022-11-25
cm-emmc-flashing: Update Windows status to reflect reality

timg236:
* Microsoft don't support Windows XP either.
* 32-bit is untested.


lurch (approved):
LGTM

PR #2723, opened 2023-01-02
Change warning to mention bugfix in macOS 13.1

KlausBrunner:
Issue has been fixed in macOS 13.1, so the warning should be adapted.


aallan (approved):
Approved with changes.

PR #2733, opened 2023-01-11
camera: Add notes about high framerate video encode

naushir:
Add some necessary tweaks needed to minimise frame drops during high framerate
video encode.

Signed-off-by: Naushir Patuck <naush@raspberrypi.com>


naushir:
@davidplowman you might want to put something similar into the picamera2 docs?

aallan (approved):

PR #2738, opened 2023-01-13
Remove the claim that libcamera and libcamera-apps can be installed o…

davidplowman:
…n Buster.

It was feasible at one time, but is not realistically true any more.


davidplowman:
@naushir Make sense?

naushir:
LGTM

aallan (approved):
YOLO

PR #2760, opened 2023-01-25
Spelling, punctuation

LizUpton:
If licence/license wasn't a thing, what would be the point of me?


aallan (approved):
Fixed merge conflicts.

Issue #2454, opened 2022-03-02
Add documentation for non-interactive use of raspi-config


aallan:
Need to document non-interactive use of `raspi-config` through using the `nonint` argument. See RPi-Distro/raspi-config#123 for background details.

andrum99:
I don't think there should actually be any documentation for `raspi-config` in the documentation repo: the reason it exists is that previously there was no GUI-based config tool on the Pi, and it was deemed useful to document the text-based `raspi-config`. Non-technical users will generally use the GUI-based tool, and for the rest the tool should be obvious enough. Each menu option is effectively self-documented, so I don't see any need for there to be documentation for `raspi-config` in this repo.

andrum99:
> > I don't think there should actually be any documentation for `raspi-config` in the documentation repo
> 
> Okay. But as the head of documentation, I _do_ think there should be documentation for `raspi-config` in the documentation repo? 🤔

I'm obviously completely fine with that, and respect that it is entirely up to you. I was just adding my tuppence worth: apologies if that was inappropriate.

aallan:
Still on the list.

aallan:
The #2766 pull request addresses this issue. I think it needs reworked before merge however.

aallan:
Alternative PR #2827 which should close this issue.

PR #2766, opened 2023-01-29
Documentation for cli api of `raspi-config`

cinderblock:
I thought it would be nice to document the CLI interface for `raspi-config`. This is a first stab at that. I took the main documentation file and added cli instructions for most of them. Some need some cleanup. All needs to be made consistent with `adoc` format. Keep this as a draft for now.

### DIFF
--- a/documentation/asciidoc/computers/configuration/raspi-config.adoc
+++ b/documentation/asciidoc/computers/configuration/raspi-config.adoc
@@ -9,6 +9,8 @@ sudo raspi-config
 
 The `sudo` is required because you will be changing files that you do not own as the `pi` user.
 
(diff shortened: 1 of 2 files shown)


aallan (commented):
This doesn't integrate with the documentation correctly. There is a new page for the CLI documentation, but it is not inline in the relevant place it's just sitting there, hanging out on it's own and only linked from a single place in the raspi-config.adoc documentation page. I'm not even sure the toolchain would build the asciidoc into HTML for the documentation site as it's not included from one of the top level asciidoc files.

I'm going to close this PR, as it'll be easier to rework something to resolve #2454 in a new PR.

aallan:
Alternative PR #2827 which should close #2454.

cinderblock:
I hadn't see #2454. I was mostly trying to start the conversation. Thanks for pushing this up the priority list :)

PR #2743, opened 2023-01-18
Doxygen build toolchain

nelliemckesson:
Here's a first draft of incorporating the doxygen files. ~~A few prerequisite steps to add the pico-sdk repo and build the html files there:~~ (now done by the `Makefile`)

I added a new make command to build the adoc from the pico-sdk html (note I **did not** add this to the single `make` command, so it needs to be run separately right now):

```
$ make build_doxygen_adoc
```

I also attempted to add some steps to include those new adoc files into the built site, though they don't seem to be working :/ @lurch if you could point me in the right direction to fix whatever I'm missing, I would be very grateful! Also feel free to adjust any of the make/ninja stuff that I added; I did my best but you're the pro.

Once I can get it building into the site, I can do some more QA and add to the json mappings for anything that has yet to be converted.

PR #2791, opened 2023-02-20
Adding pico-sdk box images and adjusting image workflow

nelliemckesson:
@lurch I'm putting this in a separate PR to make it easier to revert if you hate it. This change adds the new box images that Al provided in #2743 , and makes some adjustments to the make/ninja workflow in order to get those images included. Here's an overview:

- In order to make things more scalable (i.e., not hard-code the image names to Grant's docs in case he decides to change the name of anything at any point), I gave each box image a generic name, "pico-sdk_[n].png".
- When the toolchain builds the index.json, for tabs that are "from-json", it will look for any image that has the tab directory in the filename, and then assign those to the subdirs in numeric order. E.g., "pico-sdk_0.png" will get assigned first, then "pico-sdk_1.png", etc etc.
- The ninja build will also grab any images from the "images/full-sized" dir that match the same pattern and copy them over for the served site.

Take a look, let me know what you think or if you want to make any changes. Honestly, it would probably be simpler and make more sense just to include the entire "images/full-sized" dir regardless of whether all of those images are used or not, but I'll defer to you on that.

(Cc: @aallan )


aallan:
@lurch Unless you have objections I'm going to merge this one tomorrow so we only have one branch to hit with a stick.

scripts/create_output_index_json.py
```diff
@@ -16,24 +16,34 @@ def get_global_subitems():
...
         for item in tab_data:
             newsubitem = {}
             newsubitem['title'] = tab_data[item]['name']
             newsubitem['description'] = tab_data[item]['description']
             newsubitem['subpath'] = item + ".adoc"
-            newsubitem['imagepath'] = os.path.join('/images', 'placeholder/placeholder_square.png')
+            if len(available_images) > counter:
+                newsubitem['imagepath'] = os.path.join('/images', 'full-sized', available_images[counter])
+            else:
+                newsubitem['imagepath'] = os.path.join('/images', 'placeholder/placeholder_square.png')
```
lurch, line 37: @nelliemckesson I _think_ this is saying that if we have 6 available_images, but 8 items on the page, then items 7 and 8 will be given the placeholder image? IMHO it'd be better to cycle round and give image1 to item7 and image2 to item8? i.e. something like:
```python
newsubitem['imagepath'] = os.path.join('/images', 'full-sized', available_images[ counter % num_available_images ])
```
:question:

lurch: Ah, but in that case we probably want the "spanner" image to be treated separately to the "hex" images? :thinking:

aallan: The spanner image is for the "getting started" box. The rest of the images are for the other boxes (and it doesn't really matter which one they get)

aallan: @lurch Since @nelliemckesson won't be up for a few more hours, can you go in a tweak this please so you're happy with it. Lets get this merged down.

lurch:
Updated and merged

PR #2800, opened 2023-02-27
Build Doxygen documentation and upgrade dependencies

mudge:
In order to build the Pico SDK documentation, we need the following new dependencies:

* The arm-none-eabi-gcc GNU Arm Embedded Toolchain
* Doxygen 1.9.6

As Doxygen 1.9.6 is not available from apt on Ubuntu 22.04, download and use the compiled version from doxygen.nl directly.

We also take this opportunity to upgrade to the latest versions of all Actions and Ruby dependencies.


aallan (approved):
YOLO

PR #2830, opened 2023-03-06
Fix doxygen titles

### LINKED ISSUE #2810: Doxygen: Use correct top-level title
I've just noticed that every single page under /pico-sdk/ has "Pico C SDK" as its top-level header.

Screenshot from 2023-02-27 12-19-19

Perhaps it'd be more appropriate to put the "page title" there (if we have one), so "High Level APIs" in the screenshot above?

lurch:
Fixes #2810 
ping @nelliemckesson to check that things all look okay from her PoV.


nelliemckesson:
@lurch looks good to me!

nelliemckesson (approved):

Issue #2807, opened 2023-02-27
Doxygen: Add sdk version number in a different location


nelliemckesson:
> It's kinda a shame that the SDK-version-number gets written to _config.yml, which is a file that's under version-control, as it means that git shows _config.yml as changed whenever the SDK-version changes. Would there be any way to do this without modifying a git-controlled file?

aallan:
Related #2811 (may influence how the SDK version number is pulled out or written to?)

nelliemckesson:
Fixed by #2846

PR #2844, opened 2023-03-07
Add release version to all pico-sdk pages

nelliemckesson:

PR #2846, opened 2023-03-07
New data file for things like pico sdk version

nelliemckesson:
Instead of constantly updating _config.yml, we can add the doxygen version to this new data file. Making it generic in case we have any other data in the future that needs a place to go.


nelliemckesson:
(Blocked by #2844 )

nelliemckesson:
@lurch this PR:

1. Adds a new step to the ninja build that...
2. ...runs a new python script that...
3. ...grabs the pico sdk version and creates a new jekyll data file directly in the build dir that...
4. ...is used to populate the tabs.html jekyll template.

nelliemckesson:
Fixes #2807

scripts/create_output_supplemental_data.py
```diff
@@ -0,0 +1,30 @@
+#!/usr/bin/env python3
+
+import os
+import sys
+import json
+import re
+
+def add_release_version(doxyfile_path, data_obj):
```
lurch, line 8: Might be a bit "cleaner" if this was named `get_release_version` and just did `return version`, and then in the "main" you could do `data_obj = { 'pico_sdk_release': get_release_version(input_file) }` ?

lurch:
Should this PR also remove the `doxygen_release: 1.5.0` from `_config.yml` ?

scripts/create_output_supplemental_data.py
nelliemckesson: Good idea!

nelliemckesson:
@lurch adjusted accordingly!

nelliemckesson:
@aallan ready on my end. @lurch any other notes?

aallan (approved):
YOLO

lurch:
@nelliemckesson See #2862

PR #2865, opened 2023-03-17
Fix nav anchor handling

nelliemckesson:
@ lurch this turned out to be a stupidly simple fix: I was marking ids with `[#...]` notation, but create_nav.py was looking for `[[...]]` notation, so I adjusted the adoc conversion script accordingly.


aallan (approved):
YOLO

PR #2868, opened 2023-03-22
Remove Wireframe and Custom PC from the footer

clowder:


aallan (approved):

aallan:
Merged, and pushed to master, so should be up on prod in a few minutes

PR #2881, opened 2023-03-24
Update arm_64bit documentation to match reality

pelwell:
Current firmware defaults Pi 4s to arm_64bit=1, so the documentation needs updating.

### DIFF
--- a/documentation/asciidoc/computers/config_txt/boot.adoc
+++ b/documentation/asciidoc/computers/config_txt/boot.adoc
@@ -41,7 +41,11 @@ Set the `disable_commandline_tags` command to `1` to stop `start.elf` from filli
 
 === `arm_64bit`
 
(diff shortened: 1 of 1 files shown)


aallan (approved):

lurch:
Does this change also mean that the "The default value on the Raspberry Pi 4 and 400, and Raspberry Pi Compute Module 4 is `kernel7l.img`." sentence in the previous paragraph needs tweaking too? Maybe something like:
"The default value on the Raspberry Pi 4 and 400, and Raspberry Pi Compute Module 4 is `kernel8.img`, or `kernel7l.img` if `arm_64bit` is set to 0." ?

pelwell:
Yes.

lurch:
Ok, will submit a suitable PR when GH is working again!

Issue #2719, opened 2022-12-27
USB Ports documentation


cmangla:
See documentation/asciidoc/computers/raspberry-pi/usb-bus-on-raspberry-pi.adoc

For pi4, the document clearly states how many USB ports of which version are available. For pi{zero,1,2,3}, under the paragraph "Raspberry Pi Zero, 1, 2 and 3", the same information is not explicitly noted. It would be useful to note here, for each Pi, the USB ports and versions, and possibly the connected controllers.

lurch:
Sounds similar to #2657 ?
For the benefit of anybody else reading this: [here](https://www.raspberrypi.com/documentation/computers/raspberry-pi.html#raspberry-pi-4) is where the above-mentioned page appears in our online documentation.

aallan:
Stil valid.

aallan:
Still valid.

PR #2882, opened 2023-03-25
USB ports per model, add to usb-bus-on-raspberry-pi.adoc

cmangla:
Attempted partial fix to #2719 .

Information obtained from https://openwrt.org/toh/raspberry_pi_foundation/raspberry_pi#hardware_highlights 

Pi zero 2w is missing in the table, among other issues.


lurch:
I think this could probably be condensed to a couple of sentences, it doesn't need such a big table?

cmangla:
> I think this could probably be condensed to a couple of sentences, it doesn't need such a big table?

I agree that table has a lot of repetition. It was just a way for me to enter that information in this PR. I leave it up to the editors to finalize the writing.

PR #2886, opened 2023-03-31
Add Note for Hooking up SWD Pins on Non-H Raspberry Pi Pico Boards to Debug Probe

### LINKED ISSUE #2875: Mention non-H versions of Pico in Debug Probe documentation
The documentation at https://www.raspberrypi.com/documentation/microcontrollers/debug-probe.html mentions hooking it up to the Pico H. But given the quantity of Pico and Pico W boards that exist (that have the 3-pin-holes SWD connector rather than the JST connector) it might be worth including a NOTE and/or wiring diagram for hooking up the SWD pins on the non-H boards to Debug Probe.

Shiva953:
Added a note for hooking up the SWD pins on the non-H boards to Debug Probe. (Fixes #2875 )


JamesH65:
I wonder if that this should be a complete section for Pico's with no JST connector, with associated images, rather than just a note?

aallan (approved):

PR #2891, opened 2023-04-13
Headless Raspberry Pi Zero W v1.1 with Codec Zero

GuillaumeRossolini:
I am not certain what changes need to be made but the instructions weren't working for me, and this is how I got the hat to work with a recent RaspiOS.
See https://github.com/iqaudio/Pi-Codec/issues/1#issuecomment-1506789157
Note that I haven't tried the mute/unmute instructions at all.

### DIFF
--- a/documentation/asciidoc/accessories/audio/configuration.adoc
+++ b/documentation/asciidoc/accessories/audio/configuration.adoc
@@ -45,6 +45,25 @@ independently. Within the codec itself, other mixers and switches exist to allow
 Both the AUX IN and AUX OUT are 1V RMS. It may be necessary to adjust
 the AUX IN’s mixer to ensure that the input signal doesn’t saturate the ADC. Similarly, the output mixers can be to be adjusted to get the best possible output.
 
+Edit the ``/boot/config.txt`` file again:
+
+----
+$ sudo nano /boot/config.txt
+----
+
+Find the `#dtparam=audio=on` line in the file to add `dtoverlay=iqaudio-codec` on a new line:
+
+```
+#dtparam=audio=on
+dtoverlay=iqaudio-codec
+```
+
+Ctrl X, Y and Enter to save and reboot your device for the settings to take effect:
+
+----
+$ sudo reboot
+----
+
 Preconfigured scripts (loadable ALSA settings) https://github.com/iqaudio/Pi-Codec[are available on GitHub], offering:
  
 * Mono MEMS mic recording, mono speaker playback
@@ -67,7 +86,7 @@ $ sudo apt install git
 The following command will set your device to use the on-board MEMS microphone and output for speaker playback:
 
 ----
-$ sudo alsactl restore -f /home/pi/Pi-Codec/IQaudIO_Codec_OnboardMIC_record_and_SPK_playback.state
+$ sudo alsactl restore -D IQaudIOCODEC -f /home/pi/Pi-Codec/IQaudIO_Codec_OnboardMIC_record_and_SPK_playback.state
 ----
 
 In order for your project to operate with your required settings when it is powered on, edit the `/etc/rc.local` file. The contents of this file are run at the end of every boot process, so it is ideal for this purpose. Edit the file:
@@ -92,7 +111,7 @@ Add the chosen script command above the exit 0 line and then Ctrl X, Y and Enter
 #
 # By default this script does nothing.
 
-sudo alsactl restore -f /home/pi/Pi-Codec/IQaudIO_Codec_OnboardMIC_record_and_SPK_playback.state
+/usr/sbin/alsactl restore -D IQaudIOCODEC -f /home/pi/Pi-Codec/IQaudIO_Codec_OnboardMIC_record_and_SPK_playback.state
 
 exit 0
 ----
@@ -114,7 +133,7 @@ Add the following to the file:
 ----
 pcm.!default {
         type hw
-        card Zero
+        card IQaudIOCODEC
 }
 ----
 
@@ -133,6 +152,7 @@ Firstly a "one-shot" unmute when kernel module loads.
 
 ----
 dtoverlay=iqaudio-dacplus,unmute_amp
+dtoverlay=iqaudio-codec,unmute_amp
 ----
 
 Unmute the amp when an ALSA device is opened by a client. Mute, with a five-second delay
@@ -141,6 +161,7 @@ window will cancel mute.)
 
 ----
 dtoverlay=iqaudio-dacplus,auto_mute_amp
+dtoverlay=iqaudio-codec,auto_mute_amp
 ----
 
 If you do not want to control the mute state through the device tree, you can also script your own


aallan:
@tobybroberts could you take a look at this? Thanks!

aallan:
@GuillaumeRossolini The PR doesn't pass the automated build checks? 

> Exception: documentation/asciidoc/accessories/audio/configuration.adoc contains a Markdown-style header (i.e. '#' rather than '=')

lurch:
AFAIK (ping @pelwell ) the `dtoverlay=` stuff shouldn't be necessary if the HAT eeprom is working properly?

pelwell:
From the first line of the documentation:
> A pre-programmed EEPROM is included on all Raspberry Pi audio boards. Raspberry Pi audio boards are designed to be plug-and-play; Raspberry Pi OS is able to automatically detect and configure itself.

This is true, provided the image (or more specifically its overlays) are at least as recent as September 5 2022 (the corresponding RPiOS release tag is 1.20221028). If not, the addition of the line:
```
dtoverlay=iqaudio-codec
```
is a reasonable workaround, although it will cause the board to have the old IQaudIO branding.

pelwell:
And doing so will have knock-on consequences w.r.t. the ALSA settings, as @GuillaumeRossolini has discovered.

GuillaumeRossolini:
@aallan Thanks for the ping, my bad. Fixed and pushed.

@lurch I only know that without those lines, `aplay -L` doesn't see the device at all.

@pelwell Would you happen to know a better fix, then?

pelwell:
Updating the kernel and its overlays. The absolute minimum that I would expect to work would be to download this file: https://github.com/raspberrypi/rpi-firmware/blob/ea4b8c0e26ad900c991b5f31b8c8256e467527eb/overlays/rpi-codeczero.dtbo
and copy it to /boot/overlays.
e.g.
```
$ wget https://github.com/raspberrypi/rpi-firmware/raw/ea4b8c0e26ad900c991b5f31b8c8256e467527eb/overlays/rpi-codeczero.dtbo
$ sudo cp rpi-codeczero.dtbo /boot/overlays/
```

GuillaumeRossolini:
But this is a fresh install and I did run apt update && apt upgrade?
What am I missing?
(I'm trying your suggestion now)

GuillaumeRossolini:
Perhaps I should note that the hat I have here is pre-rebranding?
cf. https://pip.raspberrypi.com/categories/771-pcn/documents/RP-003867-PC/Codec-Zero-Rebranding.pdf

aallan:
> And doing so will have knock-on consequences w.r.t. the ALSA settings, as @GuillaumeRossolini has discovered.

Okay, so that sounds like it shouldn't be the default route (as per this PR).

However we possibly need an additional *section* talking about this,

> Updating the kernel and its overlays. The absolute minimum that I would expect to work would be to download this file: https://github.com/raspberrypi/rpi-firmware/blob/ea4b8c0e26ad900c991b5f31b8c8256e467527eb/overlays/rpi-codeczero.dtbo and copy it to /boot/overlays. e.g.
> 
> ```
> $ wget https://github.com/raspberrypi/rpi-firmware/raw/ea4b8c0e26ad900c991b5f31b8c8256e467527eb/overlays/rpi-codeczero.dtbo
> $ sudo cp rpi-codeczero.dtbo /boot/overlays/
> ```

and this,

> In that case I would recommend installing the overlay above (for people who don't have it) and putting the following at the top of your config.txt:
> 
> ```
> # Some magic to prevent the normal HAT overlay from being loaded
> dtoverlay=
> # And something less incomprehensible to load the replacement
> dtoverlay=rpi-codeczero
> ```
> 
> You should now be able to follow the usual instructions.

in a "if things go wrong" sort of way. @pelwell is there an obvious way to poke things we the user would know they have to do these additional steps?

pelwell:
Run the following command:
```
$ cat /proc/device-tree/hat/vendor
```
If it says `Raspberry Pi Ltd.` then the most you should need (unless your image is ancient/broken) is to install the overlay.
If it says `IQaudIO Limited www.iqaudio.com` then you will need the addition config.txt settings outlined above.
If it says words to the effect of "file not found" then the HAT is not being detected, but the same config.txt settings _may_ make it work.

pelwell:
> the HAT is not being detected

The most likely explanation for this is that the HAT wasn't programmed properly at the manufacturer, something which unfortunately did occur until it was noticed and the testing regime improved.

GuillaumeRossolini:
@pelwell here you go
```
pi@rpiaudio-cz:~ $ ls -alh /proc/device-tree/hat /proc/device-tree/hat/vendor
ls: cannot access '/proc/device-tree/hat': No such file or directory
ls: cannot access '/proc/device-tree/hat/vendor': No such file or directory
```

All of this is going off of a Raspberry Pi OS Lite (32-bit), release 2023-02-01, using the Raspberry Pi Imager v1.7.4.

aallan:
Ping @tobybroberts. See 👇 

> Run the following command:
> 
> ```
> $ cat /proc/device-tree/hat/vendor
> ```
> 
> If it says `Raspberry Pi Ltd.` then the most you should need (unless your image is ancient/broken) is to install the overlay. If it says `IQaudIO Limited www.iqaudio.com` then you will need the addition config.txt settings outlined above. If it says words to the effect of "file not found" then the HAT is not being detected, but the same config.txt settings _may_ make it work.

We probably need to add this to the instructions!

aallan:
> However, the git repo that you cloned is the old IQaudIO site and contains the old configuration files; these won't work with the rebranded boards. I gave @tobybroberts an updated set of files that do work, but I don't know what happened to them after that.

Do we need to offer the files for download for corner cases like this @pelwell. I can throw them up somewhere and @tobybroberts can link to them from the documentation.

GuillaumeRossolini:
@pelwell I'm not even sure what the alsactl command is used for. I can get audio without calling it.

pelwell:
If your board is actually green then it is an IQaudIO board, and these instructions don't apply to you.

GuillaumeRossolini:
@pelwell no it's the other way around. Green card is post rebranding.

GuillaumeRossolini:
Like I said, mine is black and it says IQaudIO.

aallan:
> > Your card is probably unprogrammed then.
> 
> That's what this was added for, right? https://www.raspberrypi.com/documentation/accessories/audio.html#updating-your-firmware

Ah! Yes. So we don't need to host them. There's stuff in place to handle that case. Good.

pelwell:
Although since the update may require soldering there should be a less hazardous alternative.

JamesH65:
You can get away with an M-M DuPont or other jumper wire, holding it between the two pads. The programming is very quick, so you don't need to hold it there for long. I was surprised at how easy it is to do.

GuillaumeRossolini:
What does it mean if I can see the "0Ω resistor fitted to bridge the write-protect link" on the board, do I need to push it off? Or is it already ready for write access

pelwell:
If the link is fitted then you are good to go - it's a Write Enable signal.

GuillaumeRossolini:
Ok thanks for the confirmation.
I applied the firmware update, it said success.
Change the /boot/config.txt to leave it as verbatim as possible.
Reboot.
```
$ time aplay whatever.wav && echo ok || echo KO
ALSA lib pcm_hw.c:1829:(_snd_pcm_hw_open) Invalid value for card
aplay: main:830: audio open error: No such device

real    0m0.048s
user    0m0.011s
sys     0m0.026s
KO
```
Change /boot/config.txt again to add your `dtoverlay=rpi-codeczero`, reboot.
```
 $ time aplay whatever.wav && echo ok || echo KO
Playing WAVE 'whatever.wav' : Signed 16 bit Little Endian, Rate 44100 Hz, Stereo
aplay: pcm_write:2058: write error: Input/output error

real    0m21.674s
user    0m0.013s
sys     0m0.032s
KO
$ alsactl restore -D "Zero" -f /usr/local/src/com.github/iqaudio/Pi-Codec/IQaudIO_Codec_OnboardMIC_record_and_SPK_playback.state && echo ok || echo KO
No state is present for card Zero
alsa-lib main.c:1014:(snd_use_case_mgr_open) error: failed to import hw:1 use case configuration -2
Found hardware: "RPi_Codec_Zero" "" "" "" ""
Hardware is initialized using a generic method
No state is present for card Zero
KO
$ ls -alh /proc/device-tree/hat
ls: cannot access '/proc/device-tree/hat': No such file or directory
```
(no audio output on the speakers, aplay's delay is 15s too long for this 6s clip)
Change /boot/config.txt yet again to add your `dtoverlay=` above the other, reboot.
Same results.

GuillaumeRossolini:
Since this wasn't the original intent and this thread is closed (no idea if people are going to keep answering here), I'm opening a new issue here:
https://github.com/raspberrypi/documentation/issues/2894

PR #2905, opened 2023-05-03
Add note regarding cut-down firmware

srcshelton:
As per https://github.com/raspberrypi/firmware/issues/1809.

I was wondering whether `gpu_mem` in this change should be a link to https://www.raspberrypi.com/documentation/computers/config_txt.html#gpu_mem, but I couldn't figure out how to represent this is asciidoc markup (and even then, I'm not sure it would work within backticks?).

I've assumed that `start*cd.elf` is preferable to `start?cd.elf` (I think this representation is used elsewhere in the docs).

### DIFF
--- a/documentation/asciidoc/computers/config_txt/boot.adoc
+++ b/documentation/asciidoc/computers/config_txt/boot.adoc
@@ -7,7 +7,7 @@ These options specify the firmware files transferred to the VideoCore GPU prior 
 `start_file` specifies the VideoCore firmware file to use.
 `fixup_file` specifies the file used to fix up memory locations used in the `start_file` to match the GPU memory split. Note that the `start_file` and the `fixup_file` are a matched pair - using unmatched files will stop the board from booting. This is an advanced option, so we advise that you use `start_x` and `start_debug` rather than this option.
 
-Note that the cut-down firmware (`start*cd.elf` and `fixup*cd.dat`) cannot be selected this way - the system will fail to boot.  The only way to enable the cut-down firmware is to specify `gpu_mem=16`.
+NOTE: Cut-down firmware (`start*cd.elf` and `fixup*cd.dat`) cannot be selected this way - the system will fail to boot.  The only way to enable the cut-down firmware is to specify `gpu_mem=16`.
 
 === `start_x`, `start_debug`
 
--- a/documentation/asciidoc/computers/config_txt/boot.adoc
+++ b/documentation/asciidoc/computers/config_txt/boot.adoc
@@ -7,7 +7,7 @@ These options specify the firmware files transferred to the VideoCore GPU prior 
 `start_file` specifies the VideoCore firmware file to use.
 `fixup_file` specifies the file used to fix up memory locations used in the `start_file` to match the GPU memory split. Note that the `start_file` and the `fixup_file` are a matched pair - using unmatched files will stop the board from booting. This is an advanced option, so we advise that you use `start_x` and `start_debug` rather than this option.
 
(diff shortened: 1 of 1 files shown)


lurch:
There's also a little bit of info at https://www.raspberrypi.com/documentation/computers/configuration.html#boot-folder-contents - perhaps the "and for use when `gpu_mem=16` is specified in `config.txt`" there should be changed to "and is used when `gpu_mem=16` is specified in `config.txt`"

documentation/asciidoc/computers/config_txt/boot.adoc
```diff
@@ -7,6 +7,8 @@ These options specify the firmware files transferred to the VideoCore GPU prior
 `start_file` specifies the VideoCore firmware file to use.
 `fixup_file` specifies the file used to fix up memory locations used in the `start_file` to match the GPU memory split. Note that the `start_file` and the `fixup_file` are a matched pair - using unmatched files will stop the board from booting. This is an advanced option, so we advise that you use `start_x` and `start_debug` rather than this option.
 
+NOTE: Cut-down firmware (`start*cd.elf` and `fixup*cd.dat`) cannot be selected this way - the system will fail to boot.  The only way to enable the cut-down firmware is to specify `gpu_mem=16`. The cut-down firmware removes support for cameras, codecs and (deprecated, firmware side) 3D as well as limiting the initial early-boot simple framebuffer to 1080p @  16bpp - although KMS can replace this with up-to 32bpp 4K framebuffer(s) at a later stage as with any firmware.
```
lurch, line 10: I'm not sure if it's worth mentioning the "(deprecated, firmware side) 3D". And I'd probably change "initial early-boot simple framebuffer" to just "initial early-boot framebuffer" ?

srcshelton:
> There's also a little bit of info at https://www.raspberrypi.com/documentation/computers/configuration.html#boot-folder-contents

Ah, I hadn't seen this!  I've also noticed that there are several `config.txt` directives not mentioned in the `config.txt` documentation...

documentation/asciidoc/computers/config_txt/boot.adoc
srcshelton: I believe that "simple framebuffer" is a specific technical name for a type of framebuffer?

I'd suggest that 3D is worth mentioning - although perhaps the comment that it's not the current approach is too much detail?

lurch: > I believe that "simple framebuffer" is a specific technical name for a type of framebuffer?

Perhaps, but given that this early-boot framebuffer is going to get replaced by KMS anyway, I wasn't sure if this was a distinction worth making for most users?

> I'd suggest that 3D is worth mentioning - although perhaps the comment that it's not the current approach is too much detail?

It's the fact that it's deprecated which is why I was suggesting removing the mention of this 3D support. Ping @pelwell and @popcornmix as they know much more about this than I do.

lurch:
> I've also noticed that there are several config.txt directives not mentioned in the config.txt documentation...

More than likely :wink:  Feel free to create appropriate issue(s), and we'll address it when we can.

srcshelton:
> > I've also noticed that there are several config.txt directives not mentioned in the config.txt documentation...
> 
> More than likely 😉 Feel free to create appropriate issue(s), and we'll address it when we can.

Mostly around SecureBoot support, I think?  I'll see if I can gather some specifics...

documentation/asciidoc/computers/config_txt/boot.adoc
srcshelton: Changed in 56cd451c60bc743ead6a78904a341dd73fcebd70

lurch (approved):

lurch:
Thanks Stuart :+1:

PR #2915, opened 2023-05-07
Added a numbered list to first few pgs. of getting started guide

andrewjtech123:
When reading the first few paragraphs of the "Getting started" guide, I thought they could benefit from using a numbered list. This makes it easier for the reader to parse the individual steps involved and scan the page at first glance.


aallan:
In general I'm not in favour of bullet point, or number lists, in documentation unless they're critically necessary. So documenation by bullet point is something we're trying to move away from in the existing documentation.

PR #2917, opened 2023-05-07
building doxygen is a prereq on linux

arkadianriver:
I was unable to run `make` without running `make build_doxygen_adoc` first.
Making this PR from Write the Docs after troubleshooting the issue with Nellie.

### DIFF
--- a/README.md
+++ b/README.md
@@ -137,7 +137,7 @@ As the local server launches, the local URL will be printed in the terminal -- o
 
 ### Cleaning up afterward
 
-You can revert your repositoryt to a pristine state by running,
+You can revert your repository to a pristine state by running,
 ```
 $ make clean_everything
 ```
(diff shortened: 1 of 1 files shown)


arkadianriver:
Note that the `gcc-arm-node-eabi` dependency for the doxygen build was true for me on Pop_OS 22.04. Someone on Ubuntu 18 was unable to install this and didn't need to.

README.md
```diff
@@ -128,29 +135,13 @@ $ make serve_html
...
-
-```
-$ make build_doxygen_doc
-$ make
-```
-
-and clean up afterwards by using,
+### Cleaning up afterward
 
+You can revert your repositoryt to a pristine state by running,
```
cmangla, line 140: Typo `repositoryt`

arkadianriver: 🤦 Thank you!

aallan:
The Linux instructions are intended for Raspberry Pi OS, although they should be valid for most Debian-derived distributions. There are so many distributions that we can't provide instructions for all of them. I'll double-check next week when I get back home from WTD as to whether the instructions as written work on the current version of Raspberry Pi OS.

lurch:
See #2931

PR #2918, opened 2023-05-07
wtd2023-writing-day-contributions

doble-eme:
Write the docs Writing Day contributions


aallan:
In general I'm not in favour of bullet point, or number lists, in documentation unless they're critically necessary. Documentation by bullet point is something we're trying to move away from in the existing documentation.

PR #2975, opened 2023-06-06
Styling for whitepaper info blocks

nelliemckesson:


aallan (approved):
Presuming the CSS might need tweaking — got a feeling that the title and subtitle display on the "cover" graphic might get interesting depending on some of the documentation titles — but otherwise ready to merge @nelliemckesson ?

Looks like the checks are failing but I'm guessing this is because you've added some modules and the initial bundle install isn't catching that?

nelliemckesson:
Ok @aallan all set! I added the gemfile.lock to my commit which should help with the checks, I think?

PR #2976, opened 2023-06-07
update cut cmd to grab last 8 of serial

paulmarston:
Current command is incorrect 'cut -c 8-16' grabs 9 characters. This pull request fixes that.

Or, if you prefer an alternative using sed...

sed -n "s/^Serial.*\(.\{8\}$\)/\1/p" /proc/cpuinfo


aallan:
Ah! Yes, I fixed this when I wrote the [cluster tutorial](https://www.raspberrypi.com/tutorials/cluster-raspberry-pi-tutorial/), but then entirely forgot to go back and actually fix it in the main documentation. Thanks for the PR!

lurch (approved):

PR #3027, opened 2023-07-27
Adding NOTE about building 32-bit kernel

### LINKED ISSUE #3019: Problems following instructions to build Kernel
I carefully followed https://www.raspberrypi.com/documentation/computers/linux_kernel.html#building-the-kernel-locally, to build a 32-bit kernel from scratch for use in developing a device driver for my RPi400. However, building the kernel  (make -j4 zImage modules dtbs) failed very quickly, complaining of an unknown target "zImage".

I eventually found the cause to be that while KERNEL was correctly defined as a shell variable, it was not being exported, so make did not see it in its environment.  The fix is simply to prefix the KERNEL= definition with "export ".

I have created a branch with the relevant lines corrected.

aallan:
If you are attempting to compile a 32-bit kernel on the 32-bit distribution of Raspberry Pi OS you should be aware that while the userland in this distribution is 32-bit, the kernel is a 64-bit kernel. You should therefore follow the instructions for cross-compiling the kernel.

Closes #3019 when merged.

### DIFF
--- a/documentation/asciidoc/computers/linux_kernel/building.adoc
+++ b/documentation/asciidoc/computers/linux_kernel/building.adoc
@@ -12,7 +12,7 @@ The instructions below are divided into native builds and cross-compilation; cho
 
 === Building the Kernel Locally
 
-IMPORTANT: If you are build a 32-bit kernel on the 32-bit distribution of Raspberry Pi OS you should be aware that while the user land in this distribution is 32-bit, the kernel itself is a 64-bit kernel. You therefore need to follow the <<cross-compiling-the-kernel,instructions for cross-compiling the kernel>>.  
+IMPORTANT: Building the 64-bit kernel on the 32-bit distribution of Raspberry Pi OS is a cross-compilation exercise because it requires the installation of the cross-compiler (`gcc-aarch64-linux-gnu`). Building the 32-bit kernel on the 32-bit distribution of Raspberry Pi OS — which has a 32-bit userland, and 64-bit kernel — you should set `ARCH=arm`, although it's likely that this may already be set as `arm_64bit=0` will have been set. Instructions for <<cross-compiling-the-kernel,cross-compiling the kernel>> can be found later on this page.  
 
 On a Raspberry Pi, first install the latest version of https://www.raspberrypi.com/software/operating-systems/#raspberry-pi-os-32-bit[Raspberry Pi OS]. Then boot your Raspberry Pi, log in, and ensure you're connected to the internet to give you access to the sources.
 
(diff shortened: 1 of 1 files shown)


pelwell (approved):

mtaunton:
@aallan Re your changes in commit [0e99ba4]. 

Sorry, I hate to be awkward, but I still find the wording problematic.

To build a 32-bit kernel on 32-bit Raspberry Pi OS, it says to "set `ARCH=arm`", but doesn't explain how / where you do that.  I assume it has to be given at least as an extra argument when making the default configuration, namely as:

> make ARCH=arm bcm2711_defconfig

Is that correct / sufficient? Or does ARCH=arm need to be added also to the make command that builds the kernel itself?

Further, mentioning the possibility that "`arm_64bit=0` will have been set" is misleading. If the user is indeed building a "32-bit kernel on the 32-bit distribution of Raspberry Pi OS — which has a 32-bit userland, and 64-bit kernel", then arm_64bit=1 cannot have been set, since the kernel by default is explicitly a "64-bit kernel".

documentation/asciidoc/computers/linux_kernel/building.adoc
```diff
@@ -12,6 +12,8 @@ The instructions below are divided into native builds and cross-compilation; cho
 
 === Building the Kernel Locally
 
+IMPORTANT: Building the 64-bit kernel on the 32-bit distribution of Raspberry Pi OS is a cross-compilation exercise because it requires the installation of the cross-compiler (`gcc-aarch64-linux-gnu`). Building the 32-bit kernel on the 32-bit distribution of Raspberry Pi OS — which has a 32-bit userland, and 64-bit kernel — you should set `ARCH=arm`, although it's likely that this may already be set as `arm_64bit=0` will have been set. Instructions for <<cross-compiling-the-kernel,cross-compiling the kernel>> can be found later on this page.
```
lurch, line 15: "Building the 32-bit kernel on the 32-bit distribution of Raspberry Pi OS — which has a 32-bit userland, and 64-bit kernel" - that's only the case on Pi 4B, Pi 400, CM4 and CM4S ? https://www.raspberrypi.com/documentation/computers/config_txt.html#arm_64bit

aallan: @lurch Can you nip in and fix that up?

mtaunton:
In the meantime, could someone (@pelwell?) answer my previous question, namely:

To build a 32-bit kernel on 32-bit Raspberry Pi OS, it says to "set ARCH=arm", but doesn't explain how / where you do that. I assume it has to be given at least as an extra argument when making the default configuration, namely as:

make ARCH=arm bcm2711_defconfig

Is that correct / sufficient? Or does ARCH=arm need to be added also to the make command that builds the kernel itself?

pelwell:
You can pass it to each `make` call or make it available to all subsequent make invocations with... `export ARCH=arm`.

mtaunton:
So do both "make" calls need ARCH to be set, ie the setting of various values in .config by "make bcm2711_defconfig" is not sufficient?

PR #3031, opened 2023-07-29
Update headless.adoc

sa0mad:
Added description about HDMI hotplug as needed for smooth reboot without HDMI screen connected to RPi 4.

### DIFF
--- a/documentation/asciidoc/computers/configuration/headless.adoc
+++ b/documentation/asciidoc/computers/configuration/headless.adoc
@@ -1,6 +1,7 @@
 == Setting up a Headless Raspberry Pi
 
 To use your Raspberry Pi without a monitor or keyboard (known as headless), you need to enable wireless networking and SSH.
+Further, you need to configure it such that it boots without a screen.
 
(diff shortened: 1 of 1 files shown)


aallan:
There's something off with your markup, it doesn't render correctly? 

Screenshot 2023-07-29 at 16 20 18

andrum993:
The Pi 4 does **not** need this setting changed in order to boot with no screen attached. The description of this configuration setting is also wrong - it tells the firmware-based  graphics stack to always enable HDMI output, even if there is no monitor detected.

The Pi 4 also does not enable composite video output automatically if there is no screen detected on the HDMI ports - unlike Pi 0-3 the Pi 4 requires an additional setting in config.txt to enable composite video, since it requires some very specific clock timings that result in reduced system performance.

aallan:
@jehlers42 Can you take a look at this, check for veracity and fix up ready to merge. Thanks!

jehlers42:
This appears to be a less-than-preferable way to go about forcing boot without a monitor.

Decided to add info to this article still, but in a different manner. 
https://github.com/raspberrypi/documentation/pull/3102

PR #3032, opened 2023-07-30
pi login no longer works

hanscees:
the pi default login no longer works sinse moire than a year

### DIFF
--- a/documentation/asciidoc/computers/remote-access/secure-shell-from-unix.adoc
+++ b/documentation/asciidoc/computers/remote-access/secure-shell-from-unix.adoc
@@ -2,10 +2,14 @@
 
 You can use SSH to connect to your Raspberry Pi from a Linux desktop, another Raspberry Pi, or from an Apple Mac without installing additional software.
 
-Open a terminal window on your computer replacing `<IP>` with the IP address of the Raspberry Pi you're trying to connect to,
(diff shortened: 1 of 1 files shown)


aallan:
While the core issue here — the fact we've transitioned away from having a default `pi` user — is a problem. This pull request doesn't address it in a way I feel comfortable merging into the documentation. 

There are a few things wrong with it:

* It shouldn't reference a "news" article as a cite. Documentation ages out of relevance, but we shouldn't build in things that cause this to happen automatically.
* It has a call to action, "you need to add a user" but does not link to the section of the documentation (see https://www.raspberrypi.com/documentation/computers/getting-started.html#configuration-on-first-boot or https://www.raspberrypi.com/documentation/computers/configuration.html#configuring-a-user) which walks the user through how to do this task.
* Replacement of "pi" here with "xxxxx" isn't the best choice of stand in username. 

However the other point is that there isn't any suggestion here that "pi" is a default username in the first place, it used to be but it now used throughout the documentation (this is not the only occurrence) as a stand in for whatever username the user has chosen. This is a similar approach to that take by other OS distributions with a similar problem/situation although their choice of standin username is of course different.

We've yet to come up with a consistent, and non-intrusive, way to indicate that we're using a standin username. Although I think the only real source of confusion here is that it used to be the default.

PR #3033, opened 2023-07-31
Fix path for rpcbind defaults and fix rpcbind option

ricardobranco777:
Fix path for rpcbind defaults and fix rpcbind option.

### DIFF
--- a/documentation/asciidoc/computers/remote-access/nfs.adoc
+++ b/documentation/asciidoc/computers/remote-access/nfs.adoc
@@ -267,7 +267,7 @@ You'll want to run this command whenever `/etc/exports` is modified.
 
 ==== Restart services
 
-By default, `rpcbind` only binds to the loopback interface. To enable access to `rpcbind` from remote machines, you need to change `/etc/conf.d/rpcbind` to get rid of either `-l` or `-i 127.0.0.1`.
+By default, `rpcbind` only binds to the loopback interface. To enable access to `rpcbind` from remote machines, you need to change `/etc/default/rpcbind` to get rid of either `-l` or `-h 127.0.0.1`.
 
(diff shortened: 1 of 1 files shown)


lurch:
It looks like this file now says:
```
# /etc/init.d/rpcbind

OPTIONS=""

# Cause rpcbind to do a "warm start" utilizing a state file (default)
OPTIONS="-w"

# Uncomment the following line to restrict rpcbind to localhost only for UDP requests
# OPTIONS="${OPTIONS} -h 127.0.0.1 -h ::1"

# Uncomment the following line to enable libwrap TCP-Wrapper connection logging
# OPTIONS="${OPTIONS} -l "
```
so the bits talking about "to get rid of either `-l` or `-h 127.0.0.1`" don't seem correct?

Ping @pelwell in case he has any NFS-knowledge (or knows who does?)

pelwell:
@popcornmix is the main NFS user here.

pelwell:
At first glance that whole line now appears to be redundant.

popcornmix:
I only run nfs shares on a trusted network so I haven't looked at securing ports with rpcbind.

Typically I just install nfs-kernel-server /  nfs-common, and edit /etc/exports / /etc/fstab and things work for me.
This seems far more complicated (presumably for security reasons).

aallan:
@jehlers42 Can you look at this one, and get it ready to merge, or close as appropriate.

aallan:
Actually @nathan-contino this might be one for you!

nathan-contino:
@aallan removed this line as per discussion with @popcornmix. Ready to merge when you are.

aallan (approved):

aallan:
Merged!

PR #3038, opened 2023-08-03
Note which settings pertain only to the bootloader...

srcshelton:
… and so will only take effect if specified in `config.txt`.

See https://github.com/raspberrypi/firmware/issues/1810 for discussion.


documentation/asciidoc/computers/config_txt/boot.adoc
```diff
@@ -9,6 +9,8 @@ These options specify the firmware files transferred to the VideoCore GPU prior
 
 NOTE: Cut-down firmware (`start*cd.elf` and `fixup*cd.dat`) cannot be selected this way - the system will fail to boot.  The only way to enable the cut-down firmware is to specify `gpu_mem=16`. The cut-down firmware removes support for cameras, codecs and 3D as well as limiting the initial early-boot framebuffer to 1080p @  16bpp - although KMS can replace this with up-to 32bpp 4K framebuffer(s) at a later stage as with any firmware.
 
+NOTE: `start_file` and `fixup_file` are used by the bootloader, and so are only valid in `config.txt`
```
lurch, line 12: Without any other context, I think these standalone NOTEs saying "only valid in `config.txt`" would be too confusing, given that they all appear on https://www.raspberrypi.com/documentation/computers/config_txt.html which is titled "The config.txt file".

documentation/asciidoc/computers/config_txt/what_is_config_txt.adoc
```diff
@@ -44,7 +44,21 @@ Causes the content of the specified file to be inserted into the current file.
 
 For example, adding the line `include extraconfig.txt` to `config.txt` will include the content of `extraconfig.txt` file in the `config.txt` file.
 
-*Include directives are not supported by bootcode.bin or the EEPROM bootloader*
+[NOTE]
+====
+
+*Include directives are not supported by the bootcode.bin or EEPROM bootloaders*.
+
+Settings which are handled by the bootloader and so which will only take effect if they are specified in `config.txt` (rather than any additional included file) are:
```
lurch, line 52: I'd be happy to keep this bit where you document which settings don't work when used in an included file, as I suspect only a _very_ small proportion of users actually use config.txt includes.

documentation/asciidoc/computers/config_txt/boot.adoc
srcshelton: I guess a link could be added to the section discussing includes - or does it just need to be made more specific by changing the end of the line to something along the lines of "… and so are not allowed in any additional `include`d configuration file"?

The aim here is to _avoid_ the confusion of including a file with these settings in, and then finding that they've not taken effect! 😅

lurch: I think it likely that anybody who's using `config.txt` includes will have read the section about config.txt include files, and so will have seen the list of settings that don't work in an included file? I'm reluctant to "pollute" the rest of the config.txt documentation with (potentially confusing) information that's only of benefit to a small proportion of users.

aallan: The page is called "config.txt" makes no sense to add notes like this, adds confusion.

documentation/asciidoc/computers/config_txt/what_is_config_txt.adoc
aallan: I think a small note block about include files is okay. NOTES per individual items, not so much.

documentation/asciidoc/computers/config_txt/boot.adoc
srcshelton: d6900bb65b53e8bb56e11bc8d64080721b3f2e2e removes the individual NOTEs.

documentation/asciidoc/computers/config_txt/what_is_config_txt.adoc
srcshelton: > I think a small note block about include files is okay. NOTES per individual items, not so much.

https://github.com/raspberrypi/documentation/commit/d6900bb65b53e8bb56e11bc8d64080721b3f2e2e removes the individual NOTEs.

aallan (approved):

PR #3045, opened 2023-08-10
Removing ruby version requirement from README

nelliemckesson:
@aallan @lurch I wasn't quite sure how to handle the various references to the ruby version folder. Feel free to throw this away if you have a better way to denote that.

### DIFF
--- a/README.md
+++ b/README.md
@@ -51,7 +51,7 @@ $ /bin/bash -c "$(curl -fsSL https://raw.githubusercontent.com/Homebrew/install/
 Then you need to install Ruby,
 
 ```
-$ brew install ruby@2.7
+$ brew install ruby
 $ gem install bundler -v 2.2.15
 ```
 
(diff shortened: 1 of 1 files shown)


aallan:
So for Homebrew I'm (fairly) sure that the version number for Ruby isn't needed if (and only if) you install the default version. We only needed the version number in there because we were forced to install 2.7. Unfortunately, I can't check right now. Need a clean laptop, which isn't gonna happen till next week.

lurch:
@nelliemckesson I guess you need to edit line 58 too :wink:

jehlers42:
Could possibly because I use ruby-install and chruby instead of doing things the simple way with just Homebrew, but I noticed after installing the toolchain on my work laptop I didn't have to edit my ~/.zshrc file to include any of the PATH or environment variables.

Also does the line 60 callout of Homebrew not liking zsh still true? I've yet to have an issue with it.

aallan:
I can't actually remember what the problem was. Might have been Honebrew rather than Ruby. Or rather a Ruby related Homebrew problem. Guess I'm going to have to get a fresh from the factory Mac and do an install if you've gone off piste and not gone via the Homebrew route.

nathan-contino:
Updated the README based on my installation experience -- as far as I can tell, you don't need to specify a version for the Ruby bundler OR Ruby itself, the latest should be fine for both.

Took the opportunity to clean up some lengthy and difficult-to-verify configuration, instead providing a complete solution for the most common case -- Apple Silicon & the default `zsh` shell -- and providing some pointers for folks with nonstandard environments like `bash` and Intel processors.

And also broke up some multi-command blocks into individual copyable commands, and added links to the packages being manually installed so folks aren't (quite so) blindly installing software on their machine.

aallan (approved):
LGTM

README.md
```diff
@@ -36,103 +36,122 @@ export PATH="$PATH:$GEM_HOME/bin"
...
 
-**IMPORTANT:** Homebrew has problems using `/bin/zsh`, you may have to change your default shell to `/bin/bash`.
+```
+$ gem install bundler
+```
 
 ##### Set up Homebrew Version of Ruby
 
-If you're using `csh` or `tcsh` add the following lines to your `.cshrc` or `.tcshrc`,
+Because macOS provides its own version of Ruby, Homebrew doesn't automatically set up symlinks to access the version you justinstalled version with the `ruby` command. But after a successful install, Homebrew outputs the commands you'll need to run to set up the symlink yourself. If you use the default macOS `zsh` shell on Apple Silicon, you can set up the symlink with the following command:
```
lurch, line 65: "version you justinstalled version" ???

aallan: Fixed in 13ead58

README.md
```diff
@@ -36,103 +36,122 @@ export PATH="$PATH:$GEM_HOME/bin"
...
+
+```
 $ brew install gumbo-parser
+```
+
+And finally, install the [YAML module for Python 3](https://formulae.brew.sh/formula/pyyaml#default):
+
+```
 $ pip3 install pyyaml
 ```
```
lurch, line 114: I _think_ this doesn't need to be here, because `pyyaml` gets installed anyway by the
```bash
pip3 install --user -r requirements.txt
```
command that we run later? (see https://github.com/raspberrypi/documentation/blob/develop/requirements.txt )

aallan: If it works the way it's written, then it's fine.

PR #3052, opened 2023-08-15
update port info for older models

### LINKED ISSUE #2719: USB Ports documentation

jehlers42:
Closes #2719

### DIFF
--- a/documentation/asciidoc/computers/raspberry-pi/usb-bus-on-raspberry-pi.adoc
+++ b/documentation/asciidoc/computers/raspberry-pi/usb-bus-on-raspberry-pi.adoc
@@ -26,6 +26,8 @@ On the Raspberry Pi 4, the USB controller used on previous models is located on 
 
 === Raspberry Pi Zero, 1, 2 and 3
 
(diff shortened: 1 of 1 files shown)


aallan (approved):
LGTM

lurch:
Missing Raspberry Pi 3A+ ? https://www.raspberrypi.com/products/raspberry-pi-3-model-a-plus/

jehlers42:
> Missing Raspberry Pi 3A+ ? https://www.raspberrypi.com/products/raspberry-pi-3-model-a-plus/

All-inclusive docs are always a hard thing to manage as covering every base usually leads to unsightly and fiddly tables down the line.

Is this model seen often enough among our users, especially compared to the normal 3, to warrant mentioning? Same thing goes for the non-Model B+ Pi 1s since those have less ports as well.

I've no problem making a change if so. I just tend to err on the side of "less is more" these days.

PR #3058, opened 2023-08-22
Added a11y to os

### LINKED ISSUE #3036: Add documentation around Orca for visually impaired users
Our general advice to visually impaired users is to check out Raspberry VI (http://www.raspberryvi.org/pages/), a community project dedicated to providing blind and visually impaired Raspberry Pi users with answers, advice, and ideas. 

However, there is also Orca:

> There is a voice prompt when you run the [first installation?] wizard telling you how to install Orca. At any point, in the wizard or in the desktop, hit ctrl-alt-space and Orca will install and the Pi will reboot; when it comes back up, Orca will be running.

We need to find somewhere in the docs to add information about Orca.

jehlers42:
Closes #3036 

Not sure if additional details are required for the magnifier or if the download instructions could be more detailed so feel free to add/suggest if need be.

### DIFF
--- a/documentation/asciidoc/computers/os/accessibility.adoc
+++ b/documentation/asciidoc/computers/os/accessibility.adoc
@@ -1,0 +1,13 @@
+== Accessibility options
+
+=== Visual aids
+
+Visually-impaired users of Raspberry Pi OS can find helpful tools in the Recommend Software menu.
+
+We offer the https://help.gnome.org/users/orca/stable/introduction.html.en[Orca screen reader] to better navigate your way through the Raspberry Pi operating system, along with a screen magnifier to increase the readability of UI and screen elements.
+
+==== Orca screen reader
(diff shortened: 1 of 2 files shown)


documentation/asciidoc/computers/os/accessibility.adoc
```diff
@@ -0,0 +1,13 @@
+== Accessibility options
+
+=== Visual aids
+
+Visually-impaired users of Raspberry Pi OS can find helpful tools in the Recommend Software menu.
```
lurch, line 5: Minor nit-pick, but "Recommend Software" should be "Recommended Software".

lurch:
Hmmm, I just noticed that this has been merged into the `bookworm` branch, so perhaps the screen magnifier shouldn't have been mentioned? :thinking:

Issue #2959, opened 2023-05-18
Screen-blanking section needs to be checked / updated


lurch:
https://www.raspberrypi.com/documentation/computers/configuration.html#configuring-screen-blanking

aallan:
Still needs doing.

lurch:
https://forums.raspberrypi.com/viewtopic.php?p=2139724#p2139847 has new information on adjusting the screen-blanking timeout. Do we want to add that to the documentation @aallan ?

PR #3065, opened 2023-08-22
Update screen blanking

jehlers42:
re: issue #2959 

Removed instructions on adjusting the blank interval via the config file. Internal discussion led to encouraging changes solely through raspi-config or the config GUI.

### DIFF
--- a/documentation/asciidoc/computers/configuration/screensaver.adoc
+++ b/documentation/asciidoc/computers/configuration/screensaver.adoc
@@ -13,16 +13,7 @@ The current setting, in seconds, can be displayed using:
 cat /sys/module/kernel/parameters/consoleblank
 ----
 
-To change the `consoleblank` setting, edit the kernel command line:
-
-[,bash]
-----
-sudo nano /boot/cmdline.txt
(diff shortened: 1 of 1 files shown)


aallan (approved):

aallan:
Closes #2959

aallan:
It's not Monday, but it feels like it sometimes… 🙄 

Screenshot 2023-08-22 at 15 30 18

lurch:
I just noticed that the text just beyond the change in this PR also talks about "sudo apt install xscreensaver" - we probably don't want that, as Bookworm / Wayland comes with a screensaver built in by default? (It displays a rotating-cube of the desktop)
(and I'd guess that a program named `xscreensaver` _might_ not work on Wayland?)

ping @spl237 for thoughts / comments

jehlers42:
Since that section mentions finding the xscreensaver app in Preferences, will the new built-in screensaver have a similar location for options relating to it or should I just remove the entire screensaver bit.

aallan:
I'd lean towards just removing it.

PR #3082, opened 2023-08-24
Custom template for tutorial sections

nelliemckesson:
This PR scrapes the webpage for a given tutorial (e.g., https://www.raspberrypi.com/tutorials/build-your-own-weather-satellite-receiving-station/), and then creates a custom section in the documentation that sources in the tutorial description and header image.

Note that it is looking for these two `meta` elements on the linked page (attn: @mudge to let us know if this markup is prone to frequent changes):

```
<meta name="description" content="In this tutorial, we show you how to build a weather satellite receiving station using Raspberry Pi and a low-cost USB Software Defined Radio">
...
<meta property="og:image" content="https://www.raspberrypi.com/app/uploads/2023/08/Antenna-1024x576.jpg">
```

...though it should degrade gracefully, so if it can't find those it just won't show anything.

@aallan I laid it out card-style, so you should see something like the below, but let me know if you had something else in mind (the card is a clickable link):

Screen Shot 2023-08-24 at 9 39 35 AM

Required adoc markup is as follows:

```
[.tutoriallink, link=https://www.raspberrypi.com/tutorials/build-your-own-weather-satellite-receiving-station/]
==== Your Tutorial Heading

Your text content!
```

Someone else should definitely test this before we merge (by adding a tutorial section somewhere and then `make serve_html`), since I'm doing something a bit fancy and I'd love to be sure I'm not missing any key instructions or dependencies.

### DIFF
--- a/jekyll-assets/_templates/helpers.rb
+++ b/jekyll-assets/_templates/helpers.rb
@@ -1,3 +1,5 @@
+require 'net/http'
+
 module Slim::Helpers
   def book_link
     case (self.attr 'booktype')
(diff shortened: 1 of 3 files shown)


mudge:
If you want something more stable than scraping the HTML, I’m happy to provide a JSON API for this, e.g. https://www.raspberrypi.com/for-home/  currently uses https://www.raspberrypi.com/wp-json/rptl/v1/tutorials to show the 3 most recent tutorials.

nelliemckesson:
@mudge oh nice! More stable is definitely good, but I'll defer to your judgement here -- whatever you think is the best route to go.

nelliemckesson:
@mudge if we go the API route, we'll need a user-friendly way to select the correct record. For example, we'll need to set it up so folks like Al, Josh, etc., just need to include the tutorial url in their adoc files, and then we can get the corresponding tutorial record using just that url... does that make sense? And is it doable?

mudge:
The markup should be stable (I can’t see us ditching the meta tags any time soon) so there’s no rush to change this to an API.

If we do go that route, we could parse the tutorial slug out of a URL and pass that to an API endpoint which would return just the information you need rather than the full HTML.

nelliemckesson:
@mudge ok cool, well since using an API creates more work for you, I'll let you decide if/when we switch over to that ;) If you feel inclined to make one, or if you find yourself tweaking the tutorial page markup, just ping me and I'll switch this handling over as well.

aallan (approved):
This looks great. Thanks @nelliemckesson! 

Agree that doing it the API way is going to be cleaner, but going to merge now and will open another issue to migrate to that when everyone has time!

Issue #2991, opened 2023-06-16
Add cache-busting fingerprints


lurch:
As discussed in #2973 , the caching used for CSS files, JS files etc. means that whenever we change one of these file-types and then "push to production", it can be up to 20 minutes or so before things "sync up again", and during this time people might be seeing "incorrect" content.
This can be fixed by adding a `?ver=1646652613` to the end of the file's URL (with a unique number each time), but as this is a manual process we'll probably forget to do this most of the time :wink:  However we could instead automate this by making the `ver` value e.g. the md5sum of the file's content.

@mudge If/when we do this, what file-types would we need to do this for? Is MD5 a suitable algorithm for this, or do you have another recommendation? (obviously this is only a content-has-changed checksum, not a cryptographically-secure verification! :joy: )

mudge:
@lurch I'd start with CSS and JS files (typically, asset pipelines will do this for image and font filenames too but that would require rewriting any references in CSS files so I don't think this is worth it for now).

Given the use case, MD5 should be fine though another option might be SHA1 (as used by [Rails' Propshaft](https://github.com/rails/propshaft/blob/26082d0b8a5106a79c7e0bbb65ef780cc47bb837/lib/propshaft/asset.rb#L29)).

lurch:
This could/should still be done at some point, but it's currently _very_ low down on the priority-list!

aallan:
This sounds like a combination of @nelliemckesson and @mudge maybe?

PR #3097, opened 2023-09-05
Bust cache for CSS and JS on every build

mudge:
GitHub: https://github.com/raspberrypi/documentation/issues/2991

To prevent stale versions of our stylesheets and JavaScript from being served after a deploy, ensure their URL is unique for each build by appending the time of the build (as seconds since the epoch) in a query string.

This means we will bust the cache even if our CSS and JS don't change (as the query string changes on every build rather than being a function of the contents of the relevant file) but this errs on the side of caution and keeps the implementation very simple compared to introducing an asset pipeline.

We only add the cache busting query string to files we change as opposed to vendored JavaScript (jQuery, etc.) as they shouldn't change without also updating their filename.


aallan (approved):

lurch:
Nice :+1:

PR #3099, opened 2023-09-05
camera: Fix maximum exposure time value for Camera v1

naushir:
This value is 6s only in the legacy firmware stack, which is not functional in Bookworm.


naushir:
@aallan here is the Bookworm change that removes the legacy value.

aallan (approved):

Issue #3089, opened 2023-08-30
/boot is moving to /boot/firmware


lurch:
So all references to `/boot/config.txt` and `/boot/cmdline.txt` will need updating to `/boot/firmware/config.txt`, `/boot/firmware/cmdline.txt`, etc. etc.

EDIT: But only for Bookworm, Bullseye will continue to use `/boot/config.txt`, `/boot/cmdline.txt` etc.

aallan:
@nathan-contino For `bookworm` labelled issues you'll need to check out the `bookworm` branch and work on there. Both @lurch and @jehlers42 are also working on the branch for push/pull as you go rather than waiting to do big commits otherwise you're going to run into merge conflicts.

fti-sfuke:
Hi @lurch @aallan 
Thank you for the update.

I've downloaded the 64-bit RPiOS 'Bookworm' from here: https://raspi.debian.net/tested-images/
Below are the environment details:
```
root@rpi4-20230612:/boot# cat /etc/os-release 
PRETTY_NAME="Debian GNU/Linux 12 (bookworm)"
NAME="Debian GNU/Linux"
VERSION_ID="12"
VERSION="12 (bookworm)"
VERSION_CODENAME=bookworm
ID=debian
HOME_URL="https://www.debian.org/"
SUPPORT_URL="https://www.debian.org/support"
BUG_REPORT_URL="https://bugs.debian.org/"
```
```
root@rpi4-20230612:/boot# uname -a
Linux rpi4-20230612 6.1.0-9-arm64 #1 SMP Debian 6.1.27-1 (2023-05-08) aarch64 GNU/Linux
root@rpi4-20230612:/boot# 
```
I've noticed that there's no configuration-related content in the /boot/firmware/config.txt file. It only contains the following details:
```
root@rpi4-20230612:/boot/firmware# cat config.txt
# Do not modify this file!
#
# It is automatically generated upon install or update of either the
# firmware or the Linux kernel.
#
# If you need to set boot-time parameters, do so via the
# /etc/default/raspi-firmware, /etc/default/raspi-firmware-custom or
# /etc/default/raspi-extra-cmdline files.

# Switch the CPU from ARMv7 into ARMv8 (aarch64) mode
arm_64bit=1

enable_uart=1
upstream_kernel=1

kernel=vmlinuz-6.1.0-9-arm64
# For details on the initramfs directive, see
# https://www.raspberrypi.org/forums/viewtopic.php?f=63&t=10532
[...]

Issue #3096, opened 2023-09-05
Need some meta-documentation


aallan:
After the Bookworm launch we probably need a `Writing.md` to go alongside the `README.md` that talks about how to use some of the macros we now have for the documentation site. So, things that probably need documenting:

* How to reference a white paper, see #2973
* How to reference a Raspberry Pi Press book, see #3035
* How to reference a tutorial, see #3060

Probably also need to document how to `xref:` sections between different pages, and how to reference sections inside the same page. Because of the way we use section titles as default anchors, and other issues it's not as obvious as it could be.

Can't think of anything else that needs documenting though, anyone?

lurch:
> Probably also need to document how to `xref:` sections between different pages, and how to reference sections inside the same page.

Also, because of the way that we build big pages out of multiple include files, it's not always obvious what is the "same page" or not :wink:

lurch:
Might also be worth adding some typographical style notes, i.e. when to use bold text, italic text, backticks, code-blocks, etc.

This would be so that we have consistent formatting for things like menu options (including submenus), program buttons, keypresses, program (executable) names, etc. etc.

aallan:
> Might also be worth adding some typographical style notes, i.e. when to use bold text, italic text, backticks, code-blocks, etc.
> 
> This would be so that we have consistent formatting for things like menu options (including submenus), program buttons, keypresses, program (executable) names, etc. etc.

Maybe. But I honestly don't want this to sprawl. That stuff can be fixed up fairly easily, where as cross references, or inserting content without using macros in some hack'ey sort of way is a pain in the backside.

lurch:
> This would be so that we have consistent formatting for things like menu options (including submenus), program buttons, keypresses, program (executable) names, command-sequences, example program output, etc. etc.

For example, #3152 uses both
```
Navigate to *Preferences* > *Raspberry Pi Configuration*.
```
and
```
Select `Display Options` > `Screen Blanking`.
```

and there's not (yet) any guidance about which of those is the "preferred" markup.

(but obviously this is still fairly low priority)

nathan-contino:
Note that I chose to use `monospace` text for `raspi-config` sort-of-GUI-via-CLI instructions and *bold* text for actual-GUI Raspberry Pi Configuration instructions. Our style guide should cover both!

aallan:
But in the first instance I want documentation on how to use the new macros, and some on how to `xref:` between documents, and how to link between sections in the same document. The rest can wait.

Issue #3044, opened 2023-08-09
Same-page anchors don't get checked


lurch:
Something I noticed while working on #2942 

If there's a link like `xref:legacy_config_txt.adoc#upstream_kernel[upstream_kernel]` then the toolchain will check that the `upstream_kernel` anchor appears _somewhere_ on the `legacy_config_txt.adoc` page (even if that page itself is made up of multiple included files).
However if there's a link like `<<overlay_prefix,overlay_prefix>>` (which creates a link to the `overlay_prefix` anchor on the same page), there's no check that the `overlay_prefix` anchor actually appears on the page.

I (luckily) spotted this because after the config.txt documentation-reorganisation, `documentation/asciidoc/computers/config_txt/boot.adoc` had a `<<upstream_kernel,upstream_kernel>>` link, which linked to a `upstream_kernel` anchor on the same page which no longer existed, the link had to be updated to `xref:legacy_config_txt.adoc#upstream_kernel[upstream_kernel]` instead.

It would be nice if the toolchain could check for `<<...>>` links which link to an anchor that doesn't exist.

aallan:
@nelliemckesson I think this is a you thing?

lurch:
> If there's a link like `xref:legacy_config_txt.adoc#upstream_kernel[upstream_kernel]` then the toolchain will check that the `upstream_kernel` anchor appears somewhere on the `legacy_config_txt.adoc` page (even if that page itself is made up of multiple included files).

@nelliemckesson For reference, see #2345

lurch:
Whilst adding additional link-checking functionality, it might also be worth checking that we don't have any `https://www.raspberrypi.com/documentation/blah` absolute-URLs in use (in the `.adoc` files), since they really ought to be relative-xrefs instead.

lurch:
#3104 didn't solve the absolute-URL thing that I mentioned above, so I created #3106 for that.

PR #3101, opened 2023-09-06
Renamed all boot partition mentions to new firmware name convention

nathan-contino:
Addressing https://github.com/raspberrypi/documentation/issues/3089


nathan-contino:
Unfortunately it looks like this include doesn't work quite right -- I can't seem to reference one include easily across multiple separate files in different directories. Oof. Open to suggestions. I don't think we necessarily *need* the note at every mention of the boot partition, but I suspect there are enough users who won't upgrade to Bookworm immediately that we should make this highly visible.

aallan:
> Unfortunately it looks like this include doesn't work quite right…

Okay there are two things going on here that makes it not work. Firstly you need to include the file so that it's the relative path from the top level include, so you have a file called "foo.adoc" and it is including a file called "foo/bar.adoc" in a subdirectory called "foo", and you're including a file from a file in a different subdirectory "baz" then you need to do the relative path from the top level file, not the file in the subdirectory, because of the order things will get parsed in.

However, there seems to be a mismatch where the toolchain is checking for file existence from the subdirectory level, and not the top level. So if you get the path right so it'll pass the toolchain check, the file is actually in a different place one directory out from that. 

@lurch this is a toolchain problem, I've just done about an hour of testing and it's definitely doing something squirrely here.

The TL;DR is don't include a file from an included file, and then try and reference it from an entirely different chain. It tends to breaks things.

Sorry, but I really don't like this solution in any case. Scattering the same note throughout the text is messy. I think the best solution is to find the place where we talk about this file — and if there isn't one add one somewhere approrpiate — and then add a longer NOTE block there talking about the previous versions.

Yes, that's not perfect either. But we don't have the people to maintain documentation for previous versions.

aallan:
So in summary,

1. rip the NOTE block out from everywhere you've put it
2. find the canonical reference for the `config.txt` file, and add a sightly bigger NOTE there.

documentation/asciidoc/computers/camera/libcamera_vid.adoc
```diff
@@ -106,9 +106,11 @@ Using `libcamera-vid` to capture high framerate video (generally anything over 6
...
-4. It is advisable to set `force_turbo=1` in `/boot/config.txt` to ensure the CPU clock does not get throttled during the video capture. See https://www.raspberrypi.com/documentation/computers/config_txt.html#force_turbo[here] for further details.
+4. It is advisable to set `force_turbo=1` in `/boot/firmware/config.txt` to ensure the CPU clock does not get throttled during the video capture. See https://www.raspberrypi.com/documentation/computers/config_txt.html#force_turbo[here] for further details.
 5. Adjust the ISP output resolution with `--width 1280 --height 720` or something even lower to achieve your framerate target.
-6. On a Pi 4, you can overclock the GPU to improve performance by adding `gpu_freq=550` or higher in `/boot/config.txt`.  See https://www.raspberrypi.com/documentation/computers/config_txt.html#overclocking[here] for further details.
+6. On a Pi 4, you can overclock the GPU to improve performance by adding `gpu_freq=550` or higher in `/boot/firmware/config.txt`.  See https://www.raspberrypi.com/documentation/computers/config_txt.html#overclocking[here] for further details.
```
lurch, line 111: Eek. Not your fault @nathan-contino , but could you convert these https://www.raspberrypi.com/documentation/ absolute-links into the appropriate relative-link xrefs please?

documentation/asciidoc/computers/config_txt/note_moved_bookworm.adoc
```diff
@@ -0,0 +1 @@
+NOTE: Prior to version 12 (_Bookworm_), Raspberry Pi OS stored the boot configuration in a partition named `/boot/`. Versions 12 renamed the boot partition to `/boot/firmware/`.
```
lurch, line 1: In Raspberry Pi OS, we don't really ever refer to the version-numbers (11, 12), we only mention the codenames (Bullseye, Bookworm).

documentation/asciidoc/computers/camera/libcamera_vid.adoc
aallan: Please don't use "here" as a link word. The rule is that text should read properly, and be a real English sentence, without the link. Then insert the link into the sentence. There are a number of instances were "here" is used in the documentation (mostly due to third-party contributions before I started). But we're trying to strip them out.

aallan: So the xref here would be, `xref:config_txt.adoc#overclocking[link text]` (note the `.adoc` there!) as it's linking to another file inside the `computers/` section, if you were linking out to something in `accessories/` it'd be `xref:../accessories/foo.adoc[link text]`.

aallan: ...and this is why #3096 got opened. 😩

aallan:
Opened #3103 about the toolchain bug. Nobody should look at that, it's just there so we remember it exists.

documentation/asciidoc/computers/camera/libcamera_vid.adoc
lurch: > ... into the appropriate relative-link xrefs

I added a note to #3044

nathan-contino:
Closing in favor of https://github.com/raspberrypi/documentation/pull/3105, which I don't have to rebase, and doesn't have this rather messy, broken commit in it 😄

PR #3105, opened 2023-09-08
Updated boot partition mentions to use new path

nathan-contino:
- continuation of https://github.com/raspberrypi/documentation/pull/3101
- addresses https://github.com/raspberrypi/documentation/issues/3089
- replacing `/boot/` mentions with `/boot/firmware/`
- some minor xref fixes and wording fixes
- fixed broken xref from previous PR; instead, linking multiple mentions back to "what is config.txt" and placing a note there


aallan (approved):
LGTM

documentation/asciidoc/computers/config_txt/what_is_config_txt.adoc
```diff
@@ -1,6 +1,8 @@
 == What is `config.txt`?
 
-The Raspberry Pi uses a configuration file instead of the https://en.wikipedia.org/wiki/BIOS[BIOS] you would expect to find on a conventional PC. The system configuration parameters, which would traditionally be edited and stored using a BIOS, are stored instead in a text file named `config.txt`. This is read by the GPU before the ARM CPU and Linux are initialised. It must therefore be located on the first (boot) partition of your SD card, alongside `bootcode.bin` and `start.elf`. This file is normally accessible as `/boot/config.txt` from Linux, and must be edited as the `root` user. From Windows or OS X it is visible as a file in the only accessible part of the card. If you need to apply some of the config settings below, but you don't have a `config.txt` on your boot partition yet, simply create it as a new text file.
+NOTE: Prior to _Bookworm_, Raspberry Pi OS stored the boot partition at `/boot/`. Since _Bookworm_, the boot partition is located at `/boot/firmware/`.
```
lurch, line 3: I'd use "mounted" rather than "stored", but this probably isn't important enough to change.

lurch:
I'm afraid I missed it earlier, but https://github.com/raspberrypi/documentation/blob/bookworm/documentation/asciidoc/computers/configuration/boot_folder.adoc needs a small edit too :wink:

lurch:
With a slightly complex "git grep", I've also found these other instances that still need updating:
```
documentation/asciidoc/computers/compute-module/cm-peri-sw-guide.adoc:Download https://datasheets.raspberrypi.com/cm/minimal-cm-dt-blob.dts[minimal-cm-dt-blob.dts] and copy it to the SD card FAT partition, located in `/boot` when the Compute Module has booted.
documentation/asciidoc/computers/compute-module/cm-peri-sw-guide.adoc:Grab https://datasheets.raspberrypi.com/cm/example1-overlay.dts[example1-overlay.dts] and put it in `/boot` then compile it:
documentation/asciidoc/computers/configuration/boot_folder.adoc:Once Linux has booted, the boot partition is mounted as `/boot`.
documentation/asciidoc/computers/configuration/device-tree.adoc:On a Raspberry Pi it is the job of the loader (one of the `start.elf` images) to combine overlays with an appropriate base device tree, and then to pass a fully resolved Device Tree to the kernel. The base Device Trees are located alongside `start.elf` in the FAT partition (/boot from Linux), named `bcm2711-rpi-4-b.dtb`, `bcm2710-rpi-3-b-plus.dtb`, etc. Note that some models (3A+, A, A+) will use the "b" equivalents (3B+, B, B+), respectively. This selection is automatic, and allows the same SD card image to be used in a variety of devices.
documentation/asciidoc/computers/configuration/device-tree.adoc:This will cause the loader to look for `overlays/acme-board.dtbo` in the firmware partition, which Raspberry Pi OS mounts on `/boot`. It will then search for parameters `foo` and `level`, and assign the indicated values to them.
documentation/asciidoc/computers/configuration/device-tree.adoc:b) copy it from the source DTBs in /boot. This won't include overlays and parameters, but it also won't include any other modifications by the firmware. To allow testing of all overlays, the `dtmerge` utility will create some of the board-specific aliases ("i2c_arm", etc.), but this means that the result of a merge will include more differences from the original DTB than you might expect. The solution to this is to use dtmerge to make the copy:
documentation/asciidoc/computers/configuration/external-storage.adoc:The Raspberry Pi uses mount points `/` and `/boot`. Your storage device will show up in this list, along with any other connected storage.
documentation/asciidoc/computers/raspberry-pi/boot-nvme.adoc:├─nvme0n1p1 259:1    0   256M  0 part /boot
documentation/asciidoc/computers/remote-access/network-boot-raspberry-pi.adoc:echo "10.42.0.211:/tftpboot /boot nfs defaults,vers=4.1,proto=tcp 0 0" | sudo tee -a /nfs/client1/etc/fstab
```
(although I'm not sure about that last one, so maybe leave it for now and deal with it as part of #3033 ?)

PR #3113, opened 2023-09-14
fix memory typo?

ArchemedIan:
seems like someone missed an "8," following the pattern of adding 256, and knowing the max can be set past 76 on a pi4.

### DIFF
--- a/documentation/asciidoc/computers/config_txt/memory.adoc
+++ b/documentation/asciidoc/computers/config_txt/memory.adoc
@@ -24,7 +24,7 @@ The recommended maximum values are as follows:
 | `384`
 
 | 1GB or greater
-| `512`, `76` on the Raspberry Pi 4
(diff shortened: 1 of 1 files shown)


andrum993:
You obviously did not read the fourth paragraph on this page:

> On the Raspberry Pi 4 the 3D component of the GPU has its own memory management unit (MMU), and does not use memory from the gpu_mem allocation. Instead memory is allocated dynamically within Linux. This allows a smaller value to be specified for gpu_mem on the Raspberry Pi 4, compared to previous models.

The page is correct as currently written.

jehlers42:
The live version is definitely correct. A previous version of the page had '256' as the recommended setting for a Pi 4 but that was changed to '76' in a following revision, along with the callout on line 30 re: buffer allocation with CMA memory. It's important to note that '76' is the default value for Pi 4, as mentioned in the opening sentence.

Marking as closed.

Issue #2687, opened 2022-11-05
Tutorial around building a wireless hotspot to bypass captive portals


beta-tester:
i have the actual Raspi OS `2022-09-22-raspios-bullseye-armhf.img` version installed to my RPi Zero W and tried to set it up as an AP+STA.
for the AP part i watched the guide for [Setting up a Routed Wireless Access Point](https://www.raspberrypi.com/documentation/computers/configuration.html#setting-up-a-routed-wireless-access-point).
but it looks this guide is outdated.

there it is written `Buster` is the newest version instead of `Bullseye`.

and they use `iptables` but in the current Raspi OS `iptables` is not installed anymore.
`sudo: iptables: command not found`
and its transformation from `iptables` to `nftables` from the earlier version of RaspiOS Bullseye seems to be gone as well.

aallan:
Honestly somewhat tempted to remove this section, and the following section as well. Feels like something that can be better deal with with the new[ tutorials section](https://www.raspberrypi.com/tutorials/) of the site.

<img width="1505" alt="Screenshot 2022-11-05 at 13 28 46" src="https://user-images.githubusercontent.com/53625/200122345-c3f08f18-cfcb-4716-bc0c-c4260e8bb18b.png">

See also #2314.

lurch:
Also tangentially touches on #2520 (maybe?)

aallan:
Right. I think we should yank this, and the bridge mode section, and rewrite as a tutorial for the main site. Assigning it to @tobybroberts.

aallan:
Primarily using `NetworkManager` 'cause it's the new hotness and that's what people will have (ping @tobybroberts), but potentially both I guess depending on how much overhead of extra effort (and awkwardness) it adds to the tutorials.

aallan:
@tobybroberts wrt our conversation earlier today, we should merge the routed and bridged tutorials together into one "setting up a wireless access point".

tobybroberts:
All noted.

beta-tester:
`NetworkManager` is the new standard then i am fine when there is only a tutorial for that available. thank you in advanced,

when i can add some wishes, please also try to make a tutorial for AP + STA mode.
e.g. for the RPi Zero W it would be nice to have STA and AP simultanuosly on its wlan.
there are some old uncomplete guides out there, but none is using `NetworkManager` to do that yet.

beta-tester:
[...]

PR #3124, opened 2023-09-18
Update access-point-bridged.adoc

camx:
Not sure I have repaired the link. Would someone who actually has a clue please check it.


aallan:
This material is about to be removed, see #2687, and moved off to a [Wordpress tutorial](https://www.raspberrypi.com/tutorials/).

PR #3125, opened 2023-09-18
Update network configuration instructions for Bookworm

nathan-contino:
Closes https://github.com/raspberrypi/documentation/issues/2956

- Updated commands across the network configuration instructions to use `nmcli` instead of older network configuration tools that no longer exist in Bookworm.
- Removed instructions to input an encrypted password via a file; because `nmcli` stores network configuration in a file readable only by admin accounts, this no longer seems necessary. Open to adding encrypted password instructions back in, once I determine if it's even possible with `nmcli`.
- Removed instructions to set a static IP address locally. I'm sure it worked for someone at some point, but given my knowledge of DHCP, I am skeptical that we should endorse voodoo like this officially.
- General ~bug fixes and updates~ word removals, sentence streamlining, etc.

PR #3133, opened 2023-09-25
Added GUI instructions for connecting to a hidden network

nathan-contino:
Follow-up to Bookworm networking improvements inspired by @lurch's comment [in the Bookworm networking PR](https://github.com/raspberrypi/documentation/pull/3125#issuecomment-1729168378).

Unfortunately the GUI actually makes this a little more confusing than the CLI -- with the CLI, it's easy enough to just ask users for an SSID and password and assume that if they use some more complicated security scheme they can ask their admin about it or read the man pages. In the GUI, users have to specifically choose their security scheme. So I've opted to give a strong nudge towards WPA2, which is the most likely choice to work for *most* networks. Something something ask your network administrator if WPA2 enterprise is right for you.


aallan (approved):
Works for me!

PR #3156, opened 2023-10-11
Update the Sense Hat documentation URL

G3zz:
The documentation has moved from pythonhosted to readthedocs. URL updated accordingly.


lurch (approved):
LGTM. This is the same URL that https://github.com/astro-pi/python-sense-hat links to.

aallan:
Confirmed via email with the Foundation.

Issue #3164, opened 2023-10-12
Remove references to using wpa_supplicant.conf


aallan:
Since Bookworm shipped you can no longer use `wpa_supplicant.conf` to configure wireless networking before first boot, you have to use Imager to do this for while setting up the SD Card. This will become more obvious when "new Imager" ships (soon).

For now we need to remove instructions to configure wireless using `wpa_supplicant.conf` in the documentation.

aallan:
@nathan-contino This got missed amongst the many changes. Can you do this as an interrupt task, shouldn't take long its not referenced widely.

```
configuration/boot_folder.adoc:==== wpa_supplicant.conf
configuration/headless.adoc:You will need to define a `wpa_supplicant.conf` file for your particular wireless network. Put this file onto the boot folder of the SD card. When the Raspberry Pi boots for the first time, it will copy that file into the correct location in the Linux root file system and use those settings to start up wireless networking.
configuration/headless.adoc:A xref:configuration.adoc#wireless-networking-command-line[`wpa_supplicant.conf`] file example:
```

vinceskahan:
If possible, I'd suggest that you indicate it 'did' work (for many years) and that it is still supported in previous versions.  There are many places/uses where people will need to stay on older deb-based versions for quite a while until the various apps catch up, which might take a while.

I'd word it with a note box saying "for os based on debian-11 or earlier...." or the like with the old wording, which was fine.

aallan:
> If possible, I'd suggest that you indicate it 'did' work (for many years) and that it is still supported in previous versions. 

Reasonable. See #3169.

PR #3166, opened 2023-10-12
Remove all mentions of wpa_supplicant file

nathan-contino:
Closes https://github.com/raspberrypi/documentation/issues/3164

### DIFF
--- a/documentation/asciidoc/computers/configuration/headless.adoc
+++ b/documentation/asciidoc/computers/configuration/headless.adoc
@@ -26,7 +26,7 @@ IMPORTANT: To use SSH on a fresh Raspberry Pi OS installation, you must create a
 For more information, see xref:remote-access.adoc#ssh[set up an SSH server]. Once you've connected over SSH, you can use `raspi-config` to xref:remote-access.adoc#vnc[enable VNC] if you'd prefer a graphical desktop environment.
 
 [[configuring-a-user]]
-=== Configure a User
+=== Configure a user
 
 In the boot partition of your SD card (`/boot/firmware`), create a file named `userconf.txt`.
 
--- a/documentation/asciidoc/computers/configuration/headless.adoc
+++ b/documentation/asciidoc/computers/configuration/headless.adoc
@@ -1,3 +1,4 @@
+[[setting-up-a-headless-raspberry-pi]]
 == Set up a headless Raspberry Pi
 
 A **headless** Raspberry Pi runs without a monitor, keyboard, or mouse. To run a Raspberry Pi headless, you need a way to access it from a remote machine. To access your Raspberry Pi remotely, you'll need to connect your Raspberry Pi to a network, and a way to access the Raspberry Pi over that network.
--- a/documentation/asciidoc/computers/configuration/headless.adoc
+++ b/documentation/asciidoc/computers/configuration/headless.adoc
@@ -20,14 +20,14 @@ WARNING: Depending on the model and SD card, your Raspberry Pi may require up to
 With no keyboard or monitor, you need a way to xref:remote-access.adoc[remotely control] your headless Raspberry Pi. On first boot, the only option is SSH. To enable SSH on a fresh installation of Raspberry Pi OS, choose one of the following methods:
 
 * enable SSH in the Advanced Menu of Raspberry Pi Imager, then enter a username and password
-* place a file named `ssh` into the root (`/`) of the SD card, then configure a user manually with `userconf.txt` following the instructions below
+* create a file named `ssh` at the root of the SD card, then configure a user manually with `userconf.txt` following the instructions in the section below
 
 For more information, see xref:remote-access.adoc#ssh[set up an SSH server]. Once you've connected over SSH, you can use `raspi-config` to xref:remote-access.adoc#vnc[enable VNC] if you'd prefer a graphical desktop environment.
 
 [[configuring-a-user]]
 === Configure a user manually
 
-Using another computer, at the root (`/`) of your SD card, create a file named `userconf.txt`.
+At the root of your SD card, create a file named `userconf.txt`.
 
 This file should contain a single line of text, consisting of `<username>:<password>`: your desired username, followed immediately by a colon, followed immediately by an *encrypted* representation of the password you want to use.
 
--- a/documentation/asciidoc/computers/configuration/headless.adoc
+++ b/documentation/asciidoc/computers/configuration/headless.adoc
@@ -11,7 +11,7 @@ If you're setting up your Raspberry Pi from scratch, set up wireless networking 
 
 === Connect to a network
 
-To configure network access at first boot in a headless Raspberry Pi, use the advanced settings menu in Rasbperry Pi Imager. Enter the SSID and password of your preferred wireless network. Your Raspberry Pi will use these credentials to connect to the network on first boot. Some wireless adapters and some Raspberry Pi boards do not support 5Ghz networks; check the documentation for your wireless module to ensure compatibility with your preferred network.
+To configure network access at first boot in a headless Raspberry Pi, use the advanced settings menu in Raspberry Pi Imager. Enter the SSID and password of your preferred wireless network. Your Raspberry Pi will use these credentials to connect to the network on first boot. Some wireless adapters and some Raspberry Pi boards do not support 5Ghz networks; check the documentation for your wireless module to ensure compatibility with your preferred network.
 
 WARNING: Depending on the model and SD card, your Raspberry Pi may require up to 5 minutes to boot and connect to your wireless network.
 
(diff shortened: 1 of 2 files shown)


aallan (approved):

aallan:
Works for me!

documentation/asciidoc/computers/configuration/headless.adoc
```diff
@@ -12,54 +12,11 @@ IMPORTANT: If you are installing Raspberry Pi OS, and intend to run it headless,
 
 === Configuring Networking
 
-You will need to define a `wpa_supplicant.conf` file for your particular wireless network. Put this file onto the boot folder of the SD card. When the Raspberry Pi boots for the first time, it will copy that file into the correct location in the Linux root file system and use those settings to start up wireless networking.
+To configure network access at first boot in a headless Raspberry Pi, use the advanced settings menu in the Imager application. Enter a the SSID and password for your preferred wireless network. Your Raspberry Pi will use these credentials to connect to the network on first boot. Some wireless adapters and some Raspberry Pi boards do not support 5Ghz networks; check the documentation for your wireless module to ensure compatibility with your preferred network.
```
lurch, line 15: 'Imager' -> 'Raspberry Pi Imager' ?

"Enter a the SSID" doesn't sound quite right :wink:

documentation/asciidoc/computers/configuration/headless.adoc
```diff
@@ -12,54 +12,11 @@ IMPORTANT: If you are installing Raspberry Pi OS, and intend to run it headless,
...
 
-The Raspberry Pi's IP address will not be visible immediately after power on, so this step is crucial to connect to it headlessly. Depending on the OS and editor you are creating this on, the file could have incorrect newlines or the wrong file extension so make sure you use an editor that accounts for this. Linux expects the line feed (LF) newline character. 
+With no keyboard or monitor, you need some way of xref:remote-access.adoc[remotely accessing] your headless Raspberry Pi. To enable SSH, place a file named `ssh` (no file extension) into the boot folder (`/boot/firmware/`) of the SD Card. For more information, see xref:remote-access.adoc#ssh[set up an SSH server].
```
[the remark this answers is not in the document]
lurch: ...or at least recommend using RPi Imager first, and then give the manual instructions afterwards?
The same thing should probably apply to the "Configuring a User" section too? (i.e. recommend RPi Imager, before giving the manual instructions)

documentation/asciidoc/computers/configuration/headless.adoc
```diff
@@ -8,58 +8,20 @@ If you already have an image on the micro SD card, you can access the xref:confi
...
 
-IMPORTANT: If you are installing Raspberry Pi OS, and intend to run it headless, you will need to create a new user account. Since you will not be able to create the user account xref:getting-started.adoc#configuration-on-first-boot[using the first-boot wizard] as it requires both a monitor and a keyboard, you *MUST* add a `userconf.txt` file to the boot folder to create a user on first boot or configure the OS with a user account using the xref:getting-started.adoc#advanced-options[Advanced Menu] in the Raspberry Pi Imager.
+IMPORTANT: If you are installing Raspberry Pi OS, and intend to run it headless, you will need to create a new user account. Since you will not be able to create the user account xref:getting-started.adoc#configuration-on-first-boot[using the first-boot wizard] as it requires both a monitor and a keyboard, you *MUST* add a `userconf.txt` file to the boot folder to create a user on first boot or configure the OS with a user account using the xref:getting-started.adoc#advanced-options[Advanced Menu] in Raspberry Pi Imager.
```
lurch, line 11: Along the lines of my other comment, I think we should mention RPi Imager more prominently (as the recommeded approach), rather than only mentioning it at the end.

nathan-contino: OK! I just made a minor wording tweak here, but I can update that logic in this section as well.

lurch: The reason it's written the way it is, is that when this feature was first added, RPi Imager initially didn't have the option for configuring the username - that only came later.

documentation/asciidoc/computers/configuration/headless.adoc
```diff
@@ -1,39 +1,41 @@
...
 
-WARNING: Depending on the model and SD card, your Raspberry Pi may require up to 5 minutes to boot and connect to your wireless network.
+For more information, see xref:remote-access.adoc#ssh[set up an SSH server]. Once you've connected over SSH, you can use `raspi-config` to xref:remote-access.adoc#vnc[enable VNC] if you'd prefer a graphical desktop environment.
 
 [[configuring-a-user]]
-=== Configuring a User
+=== Configure a User
 
-You will need to add a `userconf.txt` in the boot partition of the SD card; this is the part of the SD card which can be seen when it is mounted in a Windows or MacOS computer.
+In the boot partition of your SD card (`/boot/firmware`), create a file named `userconf.txt`.
```
lurch, line 32: In contrast to other places where the boot partition is mentioned, I think you _shouldn't_ mention `/boot/firmware` here; because that's only the path that the boot partition appears at once you're actually booted into Raspberry Pi OS, but these instructions are about editing files (from another computer) _before_ you boot Raspberry Pi OS for the first time.

nathan-contino: Aha, excellent point. I now see why the original sentence included the confusing bit about the "part which is seen when it is mounted in a [separate] computer". Fixing now.

documentation/asciidoc/computers/configuration/headless.adoc
```diff
@@ -20,20 +20,18 @@ WARNING: Depending on the model and SD card, your Raspberry Pi may require up to
...
+* place a file named `ssh` into the root (`/`) of the SD card, then configure a user manually with `userconf.txt` following the instructions below
 
 For more information, see xref:remote-access.adoc#ssh[set up an SSH server]. Once you've connected over SSH, you can use `raspi-config` to xref:remote-access.adoc#vnc[enable VNC] if you'd prefer a graphical desktop environment.
 
 [[configuring-a-user]]
-=== Configure a user
+=== Configure a user manually
 
-In the boot partition of your SD card (`/boot/firmware`), create a file named `userconf.txt`.
+Using another computer, at the root (`/`) of your SD card, create a file named `userconf.txt`.
```
lurch, line 30: Sorry to nitpick, but I think even the
```
(`/`)
```
here is inappropriate and potentially confusing. Especially for Windows users!

documentation/asciidoc/computers/configuration/headless.adoc
```diff
@@ -20,20 +20,18 @@ WARNING: Depending on the model and SD card, your Raspberry Pi may require up to
 With no keyboard or monitor, you need a way to xref:remote-access.adoc[remotely control] your headless Raspberry Pi. On first boot, the only option is SSH. To enable SSH on a fresh installation of Raspberry Pi OS, choose one of the following methods:
 
 * enable SSH in the Advanced Menu of Raspberry Pi Imager, then enter a username and password
-* place a file named `ssh` (no file extension) into the boot folder (`/boot/firmware/`) of the SD Card, then configure a user with `userconf.txt`
-
-IMPORTANT: To use SSH on a fresh Raspberry Pi OS installation, you must create a user account in the xref:getting-started.adoc#advanced-options[Advanced Menu] of Raspberry Pi Imager.
+* place a file named `ssh` into the root (`/`) of the SD card, then configure a user manually with `userconf.txt` following the instructions below
```
lurch, line 23: If there's only two options, does it really need a bulleted list? :thinking:  :shrug:

lurch: Nah, in hindsight ignore this :)

documentation/asciidoc/computers/configuration/headless.adoc
```diff
@@ -1,77 +1,40 @@
...
-----
-
-NOTE: Some older Raspberry Pi boards and some USB wireless dongles do not support 5GHz networks.
-
-NOTE: With no keyboard or monitor, you will need some way of xref:remote-access.adoc[remotely accessing] your headless Raspberry Pi. For headless setup, SSH can be enabled by placing a file named `ssh`, without any extension, onto the boot folder of the SD Card. For more information see the section on xref:remote-access.adoc#ssh[setting up an SSH server].
+For more information, see xref:remote-access.adoc#ssh[set up an SSH server]. Once you've connected over SSH, you can use `raspi-config` to xref:remote-access.adoc#vnc[enable VNC] if you'd prefer a graphical desktop environment.
 
 [[configuring-a-user]]
-=== Configuring a User
+=== Configure a user manually
```
lurch, line 28: Is it worth simply folding this entire section into the second bullet-point in the "Remote access" section, rather than making this a whole separate section in its own right?

nathan-contino: hot take: yes! It's too much scope creep for the above section, let alone a bullet point. Breaks up the flow of the above section too much.

I _would_ be willing to add an extra `=` to drop it one header level and nest it within that other section, though. But I don't think it's necessary since it is technically a discrete task.

aallan: Should the configuring a user thing go at the *top of the whole thing* because if you don't configure a user your headless installation is useless.

So "Setting up a headless Raspberry Pi"

First you should configure your user account...

aallan: So just between the lede, and the SSH section. You can't SSH without a user, so the user stuff should go before the SSH stuff?

lurch: I was trying to suggest that we make the `userconf.txt` instructions less prominent, since most people will just be doing this via Raspberry Pi Imager.

nathan-contino: User setup in general? Yes. Manual user setup with userconf? Especially with the latest Bookworm changes for wireless network setup on initial boot, decreasingly important and worthy of pushing down the page.

aallan: Yes, that's true if we have to point users at Imager to configure the wifi anyway, they should also configure the user there at the same time (and SSH!). Then we can demphasis both the `userconf.txt` and `ssh` files.

lurch (approved):
Marvellous. Great work @nathan-contino !

aallan (approved):

nathan-contino:
🥳

PR #3169, opened 2023-10-12
Mention wpa_supplicant.conf in a NOTE

aallan:
Brief mention that wpa_supplicant.conf existed at one point, and doesn't any more.


vinceskahan:
https://github.com/raspberrypi/bookworm-feedback/issues/72 has what to do to get equivalent functionality for headless installations.  Short description is you need to create a firstrun.sh file and drop it into the boot partition (example is in the thread near the bottom) and you need to add some options to cmdline.txt so the firstrun.sh is executed.

I'd suggest you revise your "not available any more" with modified text saying "and do this for the new os and beyond...."

PR #3185, opened 2023-10-17
Wait for element to load before scrolling

### LINKED ISSUE #3176: Anchor links not working in Firefox?
There are reports that the anchor links, e.g. https://rptl.io/venv, for the section titles are not working in Firefox (although they do work fine in Chrome). Need to investigate whether this is happening, and if it is, whether there is anything we can do to make it work on both browsers.

nelliemckesson:
This should fix #3176


nathan-contino (approved):
LGTM, thank you!

PR #3222, opened 2023-10-23
Fix typo + Use connector names

trejan:
Fixed typo.
Use the silkscreened names of the connectors.
Removed the connector names from the config.txt snippets.


aallan:
Looks like you PR'ed against `master` and rebased or something like that?

trejan:
Yeah. Sorry about that. Will do it properly.

PR #3267, opened 2023-11-08
Moved rpi_power_supply + Fixed typo

trejan:
rpi_power_supply is under power/ not bootloader/

```
$ ls /proc/device-tree/chosen/power/rpi_power_supply 
/proc/device-tree/chosen/power/rpi_power_supply
```

I've got a question about rpi_power_supply. The documentation says it is 2 32-bit integers but rpi_power_supply has extra padding? Is this a documentation issue or a firmware issue?

```
$ od -x --endian=big rpi_power_supply
0000000 0000 2e8a 000f 0000

```

### DIFF
--- a/documentation/asciidoc/computers/configuration/device-tree.adoc
+++ b/documentation/asciidoc/computers/configuration/device-tree.adoc
@@ -849,7 +849,7 @@ For bench power supplies (e.g. connected to the GPIO header) define `PSU_MAX_CUR
 
 `rpi_power_supply` - 2 32-bit integers
 
-The USB VID and PID of the official Raspberry Pi 5A power supply (if connected).
+The USB VID and Product VDO of the official Raspberry Pi 5A power supply (if connected).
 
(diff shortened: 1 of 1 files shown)


aallan:
Ping (perhaps?) @pelwell for comment?

pelwell:
Yes, the location is wrong - it belongs in the power section. The first word is the VID, but the second word isn't a PID - it's a PDO. Don't ask me what that means.

pelwell:
Found it: https://chromium.googlesource.com/chromiumos/platform/ec/+/master/include/usb_pd_vdo.h#43

aallan:
> Yes, the location is wrong - it belongs in the power section. The first word is the VID, but the second word isn't a PID - it's a Product VDO. Don't ask me what that means.

Like this? 042aa6f

pelwell (approved):

aallan (approved):

Issue #3237, opened 2023-10-25
Cameras: External Trigger on GS: Please check voltage


njhollinghurst:
Another one like #3146 

https://www.raspberrypi.com/documentation/accessories/camera.html#external-trigger-on-the-gs-camera

According to the IMX296 datasheet, XTRIG should be limited to OVDD+0.3V where OVDD=1.8V; but the documentation suggests it's safe to apply 3.3V to XTRIG.

The GPIO->XTRIG connection on the camera board uses a level shifter, but if R11 is removed as instructed then there will be neither level shifter nor pull-up, so it is hard to see how this can safely be driven from a Pico.

Asked on forum here https://forums.raspberrypi.com/viewtopic.php?t=358276#p2148485

njhollinghurst:
It's definitely dodgy, the question is what is the most elegant circuit to recommend in the doc. Need to decide whether to recommend a level shifter or if another resistor divider will suffice (maybe with smaller resistances, e.g. 470 or 680).

By the way, Pico can't conveniently do PWM + Open Drain (except using PIO), so an OD/Pull-up style is no good here.

@naushir @6by9 might have comments.

aallan:
> [Edit: Not sure why the doc calls for a Pico here rather than a Pi. Just for variety I suppose.]

That's what Simon's summer intern used I think?

simon-martin-rpi:
Speaking to Nick. We keep the docs the same, i.e. remove R11. This guy needs to add a resistor divider to his output to reduce 3.3V to 1.8V. I suggest 18K to GND and 15K input resistor.  Since the whole task needs soldering, I guess they will be Ok with it. I hope to respin this board and put a proper connector on it which accepts any voltage at high impedance one day. We have a lot of units with the wrong circuit to sell first though.

njhollinghurst:
Got it. So we advise removing R11 in all cases., then can use a divider
Will we need to revise the documentation for the new new board version?

njhollinghurst:
Nitpick: The current text is also wrong where it says "This connects GP0 to XTR", since this is GPIO1.

Perhaps the entire paragraph should be something like: "You will need to remove R11, shown in red below, to enable external trigger input. You can then connect an active-low 1.8V logic signal to the XTR pad."

njhollinghurst:
I'll attempt a PR for this.

[...]

PR #3273, opened 2023-11-13
GS Camera external trigger: fix for #3237

njhollinghurst:
Clarify that R11 must always be removed in the current board version (regardless of whether Q2 is fitted). Specify a potential divider of 10K:12K to reduce the high logic level from 3.3V to 1.8V.

Elsewhere, increase the suggested resistor values for the 1.65V divider used to pull up XVS (to avoid the paradoxical situation where we suggest higher impedence for a driver than for a pull-up).

Figures were hand-edited and may need to be regenerated by Fritzing.


njhollinghurst:
In testing. Unfortunately I am seeing an artefact (which probably has nothing to do with resistor values) which needs investigating.

njhollinghurst:
After talking to Simon: I have put back the instruction to remove R11 _only_ if Q2 was fitted, which correct for all existing and foreseeable revisions.

OK I think that is it now. Rien ne va plus.

aallan (approved):

aallan:
Thanks, and merging!

Issue #3288, opened 2023-11-20
Site doesn't build with Doyxgen 1.9.8


aallan:
Need to investigate why the site doesn't build with Doxygen 1.9.8.

aallan:
Linking #3283 which is the branch we'll be using to look at this one!

nelliemckesson:
I'm pretty sure it's because of the file names doxygen is using for 1.9.8 -- the toolchain relies on a file called modules.html which lists all the files in the build. It looks like this file has been renamed files.html. I need to see if files.html exactly matches the former modules.html and if we can get the site to build by just adding that file name as an option, or if more is required.

PR #3293, opened 2023-11-21
Rework TOC generation for doxygen docs

nelliemckesson:
Unfortunately the fix for #3288 wasn't as simple as swapping out a filename, plus if this experience has taught us anything, it's not to rely on arbitrary doxygen filenames. I reworked the TOC build process to read file include info from each content file instead. Required a few tweaks to some other buildscripts, but nothing major. Spot-QA'd left-nav TOC contents and content pages, and all looked good.

### DIFF
--- a/scripts/create_build_adoc_doxygen.py
+++ b/scripts/create_build_adoc_doxygen.py
@@ -44,11 +44,12 @@ if __name__ == "__main__":
         for tab in data['tabs']:
             if 'from_json' in tab and 'directory' in tab and tab['directory'] == output_subdir:
                 filebase = os.path.splitext(adoc_filename)[0]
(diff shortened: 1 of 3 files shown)


aallan:
@nathan-contino Can you stick this branch and the [release site](https://www.raspberrypi.com/documentation/) in different browser windows and do and comparison of the ToC before you head off for Thanksgiving (if you can?). I'm neck deep in something else, unfortunately.

nathan-contino:
Spot-checked the release site vs. this branch; I didn't notice any differences in the left TOC. Let me know if there's any specific sections that have given us issues in the past and I can investigate them a little more carefully.

aallan (approved):
YOLO

PR #3300, opened 2023-11-28
Add missing 50°C fan threshold and mention hysteresis

trejan:
I'm not 100% happy with the new wording but unsure how to phrase it. Please adjust as appropriate. Thanks.

### DIFF
--- a/documentation/asciidoc/computers/raspberry-pi-5/cooling.adoc
+++ b/documentation/asciidoc/computers/raspberry-pi-5/cooling.adoc
@@ -3,7 +3,15 @@
 
 Raspberry Pi 5 has been designed to handle typical client workloads, uncased, with no active cooling. For heavier loads there are two official options when it comes to cooling your Raspberry Pi 5. Both of these plug into the four-pin JST PWM fan connector located in the upper right of the board between the 40-pin GPIO header and the USB 2 ports.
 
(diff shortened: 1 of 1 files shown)


nathan-contino (commented):
LGTM, with a suggestion that should make things easier to parse for our readers. Left the hysteresis term in place, but with a more explicit definition for readers who don't already know the term.

Two small nitpicky questions to think about and possibly incorporate:

* Do we know the speeds for the ranges in RPM? Might be helpful to provide in parens after each item in the list, e.g. ` (0RPM)` after the first item. But if that's subject to change we can omit it, since it could become a maintenance burden.

* Have we measured the noise, preferably in decibels of the fan? Could be helpful to provide a "ceiling" of the maximum amount of noise that the fan can produce, possibly with an example of a common sound at that same decibel rating ("about as loud as a cat purring").

trejan:
> Do we know the speeds for the ranges in RPM? Might be helpful to provide in parens after each item in the list, e.g. ` (0RPM)` after the first item. But if that's subject to change we can omit it, since it could become a maintenance burden.

RPM will depend on the model of fan fitted.

It'd be better to show the PWM duty cycle instead.

* 50°C = 75/255 = 29%
* 60°C = 125/255 = 49%
* 67.5°C = 175/255 = 69%
* 75°C = 250/255 = 98%

Slightly fudging the numbers to use 250 makes it nicer looking.

* 50°C = 75/250 = 30%
* 60°C = 125/250 = 50%
* 67.5°C = 175/250 = 70%
* 75°C = 250/250 = 100%

nathan-contino:
@pelwell can you confirm these fan thresholds? Just want to double check before we merge.

pelwell (approved):

pelwell:
Yes, all good.

aallan (approved):
@pelwell says "Yes!"

PR #3301, opened 2023-11-29
Update installing-tools.adoc to fix ./bootstrap issue

import1bones:
* Add pkg-config in install list. 
To fix: configure.ac:32: error: Macro PKG_PROG_PKG_CONFIG is not available. It is usually defined in file pkg.m4 provided by package pkg-config.


aallan:
Which OS is this? You shouldn't have to install pkg-config on Raspberry Pi OS.

aallan (approved):

Issue #3243, opened 2023-10-26
Instructions for running TigerVNC client don't work on Raspberry Pi OS Bookworm


lurch:
The instructions at https://www.raspberrypi.com/documentation/computers/remote-access.html#connect-to-your-raspberry-pi say to just download the `jar` file and double-click it. But when you try this on Raspberry Pi OS, it seems to open the jar file in `xarchiver` and display its contents.
It _does_ work if I run `java -jar Downloads/VncViewer-1.13.1.jar` from the Terminal. You might also need to provide instructions for installing `java`?

PR #3307, opened 2023-12-07
Update VNC instructions to with detailed instructions for running TigerVNC on Linux

nathan-contino:
* closes https://github.com/raspberrypi/documentation/issues/3243

### DIFF
--- a/documentation/asciidoc/computers/remote-access/vnc.adoc
+++ b/documentation/asciidoc/computers/remote-access/vnc.adoc
@@ -65,7 +65,7 @@ To connect to your Raspberry Pi, you'll need the following:
 If you don't know the IP address of your device, see xref:remote-access.adoc#ip-address[our instructions on finding your IP address].
 
 1. Download https://tigervnc.org/[TigerVNC]. You can install the latest version from the https://github.com/TigerVNC/tigervnc/releases[Releases page of their GitHub repository]. Click on the link in the latest release, and find the binary for your platform. Windows users should download an `exe`; macOS users should download the `dmg`; Linux users should install the `jar`.
-2. On your client device, launch TigerVNC by double-clicking the binary.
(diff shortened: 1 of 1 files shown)


aallan (approved):

andrum993:
It should not longer be necessary to use TigerVNC to connect to wayvnc running on Raspberry Pi OS, since wayvnc has been updated to work with the RealVNC client.

aallan:
TigerVNC will still work though, so we may as well recommend it.

andrum993:
> TigerVNC will still work though, so we may as well recommend it as its cross platform.

I wasn't criticising the PR, it was just a head up to mention that things have changed since the issue was raised.

I would have thought it would be better to stay with the RealVNC client since it is also cross-platform, and it seems sensible to only recommend a single VNC client rather than having to use one for X11, and one for Wayland-based desktops.

Also note that wayvnc is most likely going away soon: RealVNC have stated publicly that their VNC server on Raspberry Pi OS will get Wayland support some time in the first half of 2024, at which point I expect it will become the default VNC server on Raspberry Pi OS's Wayland desktop. We will then be back to the situation on previous releases, where RealVNC was the default VNC client and server for all configurations of Raspberry Pi OS.

aallan:
> at which point I expect it will become the default VNC server on Raspberry Pi OS's Wayland desktop.

Why?

andrum993:
> > at which point I expect it will become the default VNC server on Raspberry Pi OS's Wayland desktop.
> 
> Why?

Because the RealVNC server is, in some ways, a better VNC server than wayvnc, and it has been the default VNC server for many years on Raspberry Pi OS so people will have existing installations that expect to be using it. Anyway, you may want to speak to someone within Raspberry Pi who knows what their actual plans are. I've kind of drifted into "teaching your granny how to suck eggs" territory now, so I'll be quiet. Apologies.

PR #3312, opened 2023-12-07
Tests for doxygen titles and toc content

nelliemckesson:
This adds a couple of tests to help prevent breaking changes:

1. Test the top-level doxygen files to make sure titles are correct
2. Test the nav content to make sure all items are correct. Note that this one only tests the nav.json file content, and not the nav content generated by JS on the live site.


aallan:
Gotta love a unit test!

PR #3315, opened 2023-12-09
Fix scrollToElement infinite loop

### LINKED ISSUE #3258: pico-sdk/hardware.html freezes the tab because of a JS busy loop
Hi,

it seems that [scrollToElement()](https://github.com/raspberrypi/documentation/blob/803c08d739656dd6bad3be2a4afcee910c4467c8/jekyll-assets/_includes/scripts.html#L45) is somewhat broken.

As you can see in the following screenshot, it makes **thousands** of attempts to locate the element in this very long page before giving up. Obviously the whole time, 38 seconds for me, the tab is frozen and Firefox even advises interrupting the script. It happens to me both in Firefox and Chromium on Fedora 39 (beta).

Screenshot from 2023-11-05 17-35-00

nelliemckesson:
fixes #3258

### DIFF
--- a/jekyll-assets/_includes/scripts.html
+++ b/jekyll-assets/_includes/scripts.html
@@ -47,7 +47,6 @@
       // if found, scroll to that id hash
       console.log("Scrolling to ", selector);
       var target = document.querySelector(selector);
-      waitForElementToDisplay(selector, 10, 500);
       if (target) {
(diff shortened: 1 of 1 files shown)


nelliemckesson:
Tested on Chrome, Safari, and Firefox

aallan (approved):

PR #3328, opened 2023-12-20
Fix kernel name for 64-bit

nbuchwitz:


aallan:
Ping @pelwell

lurch:
https://www.raspberrypi.com/documentation/computers/config_txt.html#arm_64bit says
"64-bit kernels may be uncompressed image files or a gzip archive of an image (which can still be called kernel8.img; the bootloader will recognize the archive from the signature bytes at the beginning)."

nbuchwitz:
Sorry for the fuss, muscle memory typed `make Image` instead of `make Image.gz`. Current docs are correct as they use `make Image.gz`

PR #3338, opened 2023-12-31
Update rpicam_apps_packages.adoc - Refer to libcamera-dev

peter-vanpoucke:
Refer to libcamera-dev instead.

### DIFF
--- a/documentation/asciidoc/computers/camera/rpicam_apps_building.adoc
+++ b/documentation/asciidoc/computers/camera/rpicam_apps_building.adoc
@@ -17,13 +17,13 @@ NOTE: When building on a Raspberry Pi with 1GB or less of RAM, there is a risk t
 You can rebuild `rpicam-apps` _without_ first rebuilding the whole of `libcamera` and `libepoxy`. If you do not need support for the GLES/EGL preview window then `libepoxy` can be omitted entirely. Mostly this will include Raspberry Pi OS Lite users, and they must be sure to use `-Denable_egl=false` when running `meson setup` later. These users should run:
 
 ----
-sudo apt install -y libcamera-dev libjpeg-dev libtiff5-dev
+sudo apt install -y libcamera-dev libjpeg-dev libtiff5-dev libpng-dev
 ----
 
 All other users should execute:
 
 ----
-sudo apt install -y libcamera-dev libepoxy-dev libjpeg-dev libtiff5-dev
+sudo apt install -y libcamera-dev libepoxy-dev libjpeg-dev libtiff5-dev libpng-dev
 ----
 
 If you want to use the Qt preview window, please also execute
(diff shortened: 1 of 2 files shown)


peter-vanpoucke:
`rpicam_apps_building.adoc` does *libcamera-dev* mention in *Building `rpicam-apps` without rebuilding `libcamera`*. But

aallan:
Ping @davidplowman

davidplowman:
@naushir Two commits here.

The first one wants to change the `libcamera-dev` package to `rpicam-dev`. This doesn't feel right to me, it's the "dev" version of the "libcamera" package that we want. Would you agree? Is there even an `rpicam-dev` package?

The second commit wants to add some png related packages. I assume these are ok.

naushir:
> The first one wants to change the `libcamera-dev` package to `rpicam-dev`. This doesn't feel right to me, it's the "dev" version of the "libcamera" package that we want. Would you agree? Is there even an `rpicam-dev` package?

Yes, I agree.  I think `libcamera-dev` is the right term here.

aallan:
Okay to merge then @davidplowman @naushir ?

davidplowman:
I think the 2nd commit with the dng change is OK, but I'm not convinced about the first.

peter-vanpoucke:
fyi libpng is required here: <https://github.com/raspberrypi/rpicam-apps/blob/122b835fb927dffa86f0c3437042d3f0048709bc/image/meson.build#L16>

peter-vanpoucke:
> @naushir Two commits here.
> 
> The first one wants to change the `libcamera-dev` package to `rpicam-dev`. This doesn't feel right to me, it's the "dev" version of the "libcamera" package that we want. Would you agree? Is there even an `rpicam-dev` package?
> 
> The second commit wants to add some png related packages. I assume these are ok.

hi @davidplowman. Just to be sure we're on the same page, I'm trying to change the package to `libcamera-dev`. `rpicam-dev` was originally there.

davidplowman:
> fyi libpng seems to be required [here](https://github.com/raspberrypi/rpicam-apps/blob/122b835fb927dffa86f0c3437042d3f0048709bc/image/meson.build#L16).

Thanks for the clarification, and indeed thanks for posting the documentation correction!

@aallan can we merge the png (I meant "png" earlier where I said "dng") commit and leave the other one? I'd want to double-check on that other commit before deciding on it.

davidplowman:
> > @naushir Two commits here.
> > The first one wants to change the `libcamera-dev` package to `rpicam-dev`. This doesn't feel right to me, it's the "dev" version of the "libcamera" package that we want. Would you agree? Is there even an `rpicam-dev` package?
> > The second commit wants to add some png related packages. I assume these are ok.
> 
> hi @davidplowman. Just to be sure we're on the same page, I'm trying to change the package to `libcamera-dev`. `rpicam-dev` was originally there.

Ah, sorry messages crossed in the post. Indeed, rpicam-dev would be wrong. libcamera-dev would be correct. In which case, yes - we can merge both commits. Sorry for not noticing, haven't really got back in the driving seat after New Year. Thanks everyone!

aallan (approved):

PR #3339, opened 2023-12-31
Update microcontroller_docs.adoc

horner:
added warning about the incompatibility between the pico and the pico W.

### DIFF
--- a/documentation/asciidoc/microcontrollers/microcontroller_docs.adoc
+++ b/documentation/asciidoc/microcontrollers/microcontroller_docs.adoc
@@ -22,7 +22,7 @@ https://datasheets.raspberrypi.com/picow/pico-w-datasheet.pdf[Raspberry Pi Pico 
 
 https://datasheets.raspberrypi.com/picow/connecting-to-the-internet-with-pico-w.pdf[Connecting to the Internet with Raspberry Pi Pico W]:: Getting Raspberry Pi Pico W online with C/{cpp} or MicroPython
 
-NOTE: Documentation introducing working with Wi-Fi and Bluetooth on Raspberry Pi Pico W with C/{cpp} or MicroPython is presented in the https://datasheets.raspberrypi.com/picow/connecting-to-the-internet-with-pico-w.pdf[Connecting to the Internet with Raspberry Pi Pico W] book.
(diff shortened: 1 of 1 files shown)


horner:
how can we update the PDF documentation that this thing points to? I’d like to add documentation regarding getting it to work on windows.

aallan:
> added warning about the incompatibility between the pico and the pico W.

There are few incompatibilities between the Pico and Pico W, the onboard LED is pretty much the only major point, and only if you're using C. If you're using MicroPython you should used the `LED` pragma which work on both the Pico and Pico W. What other incompatibilities are you concerned with?

> how can we update the PDF documentation that this thing points to? I’d like to add documentation regarding getting it to work on windows.

While our online documentation is published under a CC BY-SA, our PDF documentation is published under a CC BY-ND license. The source for the documents is not publically available. You can however post issues and feedback around the PDF documentation on the https://github.com/raspberrypi/pico-feedback/issues[raspberrypi/pico-feedback] repo.

horner:
I have an intern who is working on the pico with a windows machine.  The PDF really does not help someone on a windows machine from banging their head for hours.  In fact, the PDF only mentions Windows twice.

I would like more explicit instruction on setting up a windows machine.  We are going to author some of it, but not sure where to contribute it.  


Regarding LED pragma:

This code: https://github.com/raspberrypi/pico-examples/blob/master/blink/blink.c will not work on the pico_w.  Should we make a PR to fix that or is it by design that the example should not work for the pico_w??  ;-)

horner:
lol.  I just saw your issue: https://github.com/raspberrypi/pico-feedback/issues/369

Thank you.

aallan:
> I have an intern who is working on the pico with a windows machine. The PDF really does not help someone on a windows machine from banging their head for hours. In fact, the PDF only mentions Windows twice.
>
> I would like more explicit instruction on setting up a windows machine. 

This PDF?  https://datasheets.raspberrypi.com/pico/getting-started-with-pico.pdf 

Which has 9 pages talking about how to manually install on Windows? But start out with the recommendation not to do that, because it's painful and instead install the toolchain using https://github.com/raspberrypi/pico-setup-windows, which comes with its own documentation?

> Regarding LED pragma:
> 
> This code: https://github.com/raspberrypi/pico-examples/blob/master/blink/blink.c will not work on the pico_w. Should we make a PR to fix that or is it by design that the example should not work for the pico_w?? ;-)

The blink code example for Pico W lives at https://github.com/raspberrypi/pico-examples/tree/master/pico_w/wifi/blink. There is an upcoming documentation update (next couple of weeks) that makes this clearer.

horner:
Ok, but build date is off.  I have an older document which I will put on my next comment.  The new PDF shows this:
image

aallan:
Yes that was the last documentation release. There is another release coming soon. In a few weeks time.

PR #3348, opened 2024-01-04
Revert "Bump lxml from 4.9.3 to 5.0.0"

nelliemckesson:
Reverts raspberrypi/documentation#3343

In making some updates to the books toolchain, I noticed a couple major changes in lxml 5.0.0. I didn't notice any effects of those when I tested this PR originally, but I'd like to revert the change so I can test a bit more heavily.


aallan (approved):
We play the Uno Reverse card!

Issue #3324, opened 2023-12-16
Compute Module documentation incluedes only legacy display configuration


nbuchwitz:
The firmware section  [1] of the compute module documentation explains how to use a dsi display with the compute module. Unfortunately it only mentions the legacy method with a `dt-blob.bin` (and don't mention that this has been replaced by the kms driver autodetection / display overlay).

It should at least mention that this is only needed for the legacy fkms driver and not with the kms driver (which is default). Ideally it should link to the official display documentation [2]

I'm not really sure, what would be the best approach, without duplicating too much content. Happy to provide a PR if we can find a suitable way.

[1] https://www.raspberrypi.com/documentation/computers/compute-module.html#firmware-configuration
[2] https://www.raspberrypi.com/documentation/accessories/display.html

aallan:
@nathan-contino Another one for you to try and squeeze in before the Christmas break! :)

PR #3375, opened 2024-01-15
Update compute module display doc to use 'new' KMS driver

nathan-contino:
* closes https://github.com/raspberrypi/documentation/issues/3324
* general cleanup, parallel wording to better match the camera doc
* removed "display and both cameras" example with poor images and content that is duplicated by the camera documentation
* updated the display doc to use the KMS drivers instead of the old FKMS stack


aallan (approved):
YOLO

PR #3376, opened 2024-01-15
CE

LizUpton:
Pass for style and punctuation (spelling was an understated thing of beauty). Also swapped out the weird ellipsis character "…" and replaced it with "...", which is functionally identical, because we're not using the single character anywhere else. Nate says we should blame him.


nathan-contino (approved):
LGTM

Issue #3153, opened 2023-10-10
Document Boot Folder changes


aallan:
With the changes coming with Bookworm, is the content of the [Boot Folder](https://www.raspberrypi.com/documentation/computers/configuration.html#the-boot-folder) still correct?

For instance, it mentions `wpa_supplicant`, with Network Manager running the end user shouldn't be touching this file at all. 

Also, how have the binary blobs changed for Pi 5?

lurch:
IIRC the EEPROM on the Pi 4 means that the Pi 4 doesn't need a `bootcode.bin`, and the EEPROM on the Pi 5 means that the Pi 5 doesn't need a `bootcode.bin` nor a `start.elf` file. I believe the Pi 5 will load `kernel_2712.img` if it exists, but fall back to `kernel8.img` if it doesn't.

Contents of the boot partition on a 64-bit Bookworm image:
```
bcm2710-rpi-2-b.dtb       bcm2711-rpi-cm4s.dtb  fixup.dat         start4cd.elf
bcm2710-rpi-3-b.dtb       bcm2712-rpi-5-b.dtb   fixup_db.dat      start4db.elf
bcm2710-rpi-3-b-plus.dtb  bootcode.bin          fixup_x.dat       start4.elf
bcm2710-rpi-cm3.dtb       cmdline.txt           initramfs_2712    start4x.elf
bcm2710-rpi-zero-2.dtb    config.txt            initramfs8        start_cd.elf
bcm2710-rpi-zero-2-w.dtb  fixup4cd.dat          issue.txt         start_db.elf
bcm2711-rpi-400.dtb       fixup4.dat            kernel_2712.img   start.elf
bcm2711-rpi-4-b.dtb       fixup4db.dat          kernel8.img       start_x.elf
bcm2711-rpi-cm4.dtb       fixup4x.dat           LICENCE.broadcom
bcm2711-rpi-cm4-io.dtb    fixup_cd.dat          overlays/
```

Oh yeah, one of the other changes is that Bookworm now uses an initramfs file by default, by usage of the [auto_initramfs](https://github.com/raspberrypi/documentation/blob/develop/documentation/asciidoc/computers/config_txt/boot.adoc#auto_initramfs) keyword in `config.txt`.

[...]

PR #3389, opened 2024-01-16
Update boot folder documentation to match latest changes for bookworm…

nathan-contino:
* closes https://github.com/raspberrypi/documentation/issues/3153
* miscellaneous cleanup while I'm in this part of the docs, including eliminating an entire header in this section that didn't contribute any useful hierarchy

### DIFF
--- a/documentation/asciidoc/computers/config_txt/boot.adoc
+++ b/documentation/asciidoc/computers/config_txt/boot.adoc
@@ -51,6 +51,7 @@ The `initramfs` command specifies both the ramfs filename *and* the memory addre
 
 NOTE: This option uses different syntax from all the other options, and you should not use a `=` character here.
 
+[[auto_initramfs]]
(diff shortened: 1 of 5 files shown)


aallan (approved):

vinceskahan:
I understand what you did and did read the linked page re: headless setup via the imaging process at https://www.raspberrypi.com/documentation/computers/configuration.html#set-up-a-headless-raspberry-pi but still feel this is extremely inadequate versus how things were set up in previous deb-11 and earlier os.

But I'm going to stop asking because it is never going to happen. 

Your project clearly is not going to make it easy for folks to set up a wireless-only network-enabled headless pi without using your rpi imager tool.   Very unhappy with that decision and it's a big step backward, but your project seems set in its ways.

Thank you for at least documenting the as-is FWIW.

Issue #2215, opened 2021-10-15
broken scrip not in repo so can't open issue 


CarlFK:
https://github.com/raspberrypi/documentation/blob/develop/documentation/asciidoc/computers/remote-access/network-boot-raspberry-pi.adoc
says:

`wget https://datasheets.raspberrypi.com/soft/prepare_pxetools.sh`

That file errors as this line doesn't do what it should do:

`IP=$(ifconfig eth0 | grep "inet addr" | cut -d " " -f 12 | cut -d ":" -f 2)`

there are more errors, but as there is no place to report the errors I give up.

PR #3406, opened 2024-01-17
Remove references to broken pxetools

nathan-contino:
* `pxetools` no longer works with the latest releases of Raspberry Pi OS Lite. Removing links from the docs so users aren't misled.
* We can always bring this back if we fix this in the future. I'm just not sure who to bother about it.
* ...and one small nitpick style fix -- monospacing a literal string.


aallan (approved):

lurch:
Fixes #2215 :slightly_smiling_face:

PR #3407, opened 2024-01-17
Add missing column specifier

trejan:


nathan-contino (approved):
LGTM. D'oh.

aallan:
Numbers in that place indicates the column width, not the column number. So needs to be "1,2,2" not "1,2,3"

PR #3495, opened 2024-01-19
0 in 'reboot "0 tryboot"' means nothing

slowpeek:
According to the sources, there is only a check for " tryboot" substring. Everything else is ignored.

https://github.com/raspberrypi/linux/blob/3bb5880ab3dd31f75c07c3c33bf29c5d469b28f3/drivers/firmware/raspberrypi.c#L217

### DIFF
--- a/documentation/asciidoc/computers/config_txt/autoboot.adoc
+++ b/documentation/asciidoc/computers/config_txt/autoboot.adoc
@@ -43,7 +43,7 @@ boot_partition=3
 
 * System is powered on and boots to partition 2 by default
 * An `Update service` downloads the next version of the OS to partition 3
-* The update is tested by rebooting to `tryboot` mode `reboot "0 tryboot"` where `0` means the default partition
+* The update is tested by rebooting to `tryboot` mode `reboot "0 tryboot"`
 
 **Committing or cancelling the update**
 


lurch:
I dunno how it all plugs together, but see also https://www.raspberrypi.com/documentation/computers/raspberry-pi.html#PARTITION and https://github.com/raspberrypi/linux/blob/rpi-6.1.y/drivers/firmware/raspberrypi.c#L205 and https://github.com/raspberrypi/linux/blob/rpi-6.1.y/drivers/watchdog/bcm2835_wdt.c#L95
(i.e. AFAIK the number does "mean something")

slowpeek:
@lurch Sry for late response. You're right. Lukring around the sources I noticed [this](https://github.com/raspberrypi/linux/blob/3bb5880ab3dd31f75c07c3c33bf29c5d469b28f3/drivers/firmware/raspberrypi.c#L535). There is no more  hooks attached to the reboot event with `register_reboot_notifier()` anywhere. That's why I thought the number is not used at all. But the watchdog stuff indeed takes care of putting the partition number into RSTS.

Just to see if it works, I rolled a test:
- sample armbian install with extra unused space on the disk
- with fdisk create a new partition (=BOOT2), type `0b` (W95 FAT32), same size as the original boot partition (=BOOT1, 256M in case of armbian). "Boot flag" doesnt matter but the partition type does
- format BOOT2 as fat
- copy contents of BOOT1 to BOOT2
- add some distinctive kernel parameter in `cmdline.txt` on BOOT2

With that I run `reboot 3` (as root) and after logging in I can see the marker in `/proc/cmdline`. Following plain `reboot` boots into BOOT1, as expected.

"Funny" thing: even though `reboot` just works for a regular user, `reboot X` requires root perms:

    user@rpi4b:~$ reboot 3
    Failed to write reboot parameter file: Permission denied

The "parameter file" is [/run/systemd/reboot-param](https://github.com/systemd/systemd/blob/7ebbaaa7f426b5bb11b4a697898844bf6d8b6ec9/src/shared/reboot-util.c#L28-L53)

lurch:
I don't use Armbian myself, I only use Raspberry Pi OS, but I've noticed that if you're typing directly on the Pi with a USB keyboard (either at a text-mode console session, or in a Terminal window in a GUI) then you can `reboot` without the need for `sudo`, but if you're connecting to the Pi remotely (e.g. over SSH), then you _do_ need to use `sudo reboot` even when using the same user.
(I've not experimented with any of the reboot-partition stuff since I was using NOOBS, which was a long long time ago :wink: )

slowpeek:
In my case it was a headless pi, without any network. The console is over a usb-ttl adapter.

PR #3521, opened 2024-01-30
Update rpicam_apps_intro.adoc Added clarification for debian version and rpicam-apps vs libcamera-app.

arsenicks:
Added clarification for debian version and rpicam-apps vs libcamera-app.

### DIFF
--- a/documentation/asciidoc/computers/camera/rpicam_apps_intro.adoc
+++ b/documentation/asciidoc/computers/camera/rpicam_apps_intro.adoc
@@ -3,6 +3,8 @@
 [WARNING]
 ====
 `rpicam-apps` applications have been renamed from `libcamera-\*` to `rpicam-*`. Symbolic links are installed to allow users to keep using the old application names, but these will be deprecated in the future. Users are encouraged to adopt the new application names as soon as possible.
+
+There's no current plan to build `rpicam-apps` packages for Debian version prior to Bookworm. If you are using Bullseye, you'll still have `libcamera-\*` apps and would not be able to install `rpicam-apps` using the package manager.
 ====
 
 === Introduction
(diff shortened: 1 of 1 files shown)


arsenicks:
Might not be the best wording but we need to add something in the documentation, this is confusing as hell.

documentation/asciidoc/computers/camera/rpicam_apps_intro.adoc
```diff
@@ -3,6 +3,8 @@
 [WARNING]
 ====
 `rpicam-apps` applications have been renamed from `libcamera-\*` to `rpicam-*`. Symbolic links are installed to allow users to keep using the old application names, but these will be deprecated in the future. Users are encouraged to adopt the new application names as soon as possible.
+
+There's no current plan to build `rpicam-apps` packages for Debian version prior to Bookworm. If you are using Bullseye, you'll still have `libcamera-\*` apps and would not be able to install `rpicam-apps` using the package manager.
```
nathan-contino, line 7:
```suggestion
Raspberry Pi OS _Bookworm_ renamed the camera capture applications from `libcamera-\*` to `rpicam-*`. Symbolic links currently allow users to keep using the old application names, but those links will eventually be deprecated. Adopt the new application names as soon as possible. Raspberry Pi OS versions prior to _Bookworm_ only provide access to older versions of the suite using the `libcamera-\*` name prefix.
```

Suggestion: wording that better matches our usual admonition text.

nathan-contino (commented):
Good point, I can see how there's currently ambiguity here since packages often span multiple OS versions. If you incorporate my wording suggestion (the GitHub UI allows you to do so with a single button press), we might be able to incorporate this.

arsenicks:
Good for me, here's the new commit, thanks for helping!

aallan (approved):

PR #3526, opened 2024-02-04
Added --verbose flag to Command Line Options

arsenicks:
Added --verbose(-v) to the online documentation

### DIFF
--- a/documentation/asciidoc/computers/camera/rpicam_options_common.adoc
+++ b/documentation/asciidoc/computers/camera/rpicam_options_common.adoc
@@ -521,6 +521,13 @@ Moves the lens to a fixed focal distance, normally given in dioptres (units of 1
 
 It should be noted that lenses can only be expected to be calibrated approximately, and there may well be variation between different camera modules.
 
+----
+	--verbose,	-v		Set verbosity level.
+----
+
+Level 0 is no output, 1 is default, 2 is verbose.
+
+Example: `rpicam-vid -o test.h264 --verbose 2`
 
 
 ==== Output file options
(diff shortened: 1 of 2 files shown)


lurch:
https://github.com/raspberrypi/rpicam-apps/blob/main/core/options.hpp#L111 suggests that this should be in `rpicam_options_common.adoc` rather than in `rpicam_options_vid.adoc` ?

arsenicks:
> https://github.com/raspberrypi/rpicam-apps/blob/main/core/options.hpp#L111 suggests that this should be in `rpicam_options_common.adoc` rather than in `rpicam_options_vid.adoc` ?

Yes if it's a common flag that make sense, I didn't looked it up if it was supported on all the other app but I can move the modification to the `rpicam_options_common.adoc` is that ok with you to keep the same PR and simply add another commit ?

There was simply no reference to this options and I find it pretty usefull when having problems! Thanks

nathan-contino:
Yep, keep the same PR and add another commit moving the text. I checked the source code and it looks like the description is correct.

arsenicks:
Here you go, I hope the two more commit will not cause any issue, I was away from my desktop so I had to do it on the webui and wasn't able to rebase.

Have a good day!

nathan-contino (approved):
LGTM. Thank you for documenting an option that was previously not covered in our doc!

arsenicks:
My pleasure, I contribute when I can and I felt this was more than needed, the move from 'libcamera-app' to 'rpicam-apps' is really confusing if you are running bullseye for example I've made another PR to help clarify this I hope the merge from develop will happen soon! Have a great day

arsenicks:
@nathan-contino Do you have the permission to merge the PR or I should try to ping someone else ? Thanks!

nathan-contino:
The folks with permission internally are all out on vacation right now. Should be able to merge this early next week when they return.

PR #3529, opened 2024-02-08
Added warning about correct storage device when saying safe to write

sipos:
The documentation tells people it is safe to proceed when they are using admin credentials to write to the target storage device. This is true, but only if they have followed the previous warning/instructions to be sure they have the correct storage device selected, and not ignored the on-screen warning. Since the result of proceeding having ignored these warnings is so disasterous (destruction of data), when explicitly saying something is safe, it should be clear what the caveats to that are, as people often skip earlier sections.

### DIFF
--- a/documentation/asciidoc/computers/getting-started/install.adoc
+++ b/documentation/asciidoc/computers/getting-started/install.adoc
@@ -106,7 +106,7 @@ Finally, respond **Yes** to the "Are you sure you want to continue?" popup to be
 
 image::images/imager/are-you-sure.png[alt="Confirming a reimage of a storage device in Imager."]
 
-If you see an admin prompt asking for permissions to read and write to your storage medium, it's safe to proceed (provided that you have definitely selected the correct storage device earlier).
+If you see an admin prompt asking for permissions to read and write to your storage medium, grant Imager the permissions to proceed.
 
 .Grab a cup of coffee or go for a walk. This could take a few minutes.
 image::images/imager/writing.png[alt="Writing an image to a device in Imager."]
--- a/documentation/asciidoc/computers/getting-started/install.adoc
+++ b/documentation/asciidoc/computers/getting-started/install.adoc
@@ -106,7 +106,7 @@ Finally, respond **Yes** to the "Are you sure you want to continue?" popup to be
 
 image::images/imager/are-you-sure.png[alt="Confirming a reimage of a storage device in Imager."]
 
-If you see an admin prompt asking for permissions to read and write to your storage medium, it's safe to proceed.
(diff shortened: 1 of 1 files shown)


tdewey-rpi:
While I appreciate the sentiment, I'm not sure I understand the rationale here. In order for this change to make a tangible difference, I believe the following needs to happen:

1. They have to select an unintended storage device
2. They have to go through a confirmation prompt explicitly warning them that data on the chosen storage device may be deleted
3. They have to confirm the choice regarding OS Customisation

And this same individual, who ignored two major interjection points (choosing the device, confirmation dialog) is expected to consult the Raspberry Pi Documentation?

Perhaps the better change would be to make explicit that the prompt has been triggered by Raspberry Pi Imager, and is expected on certain systems.

documentation/asciidoc/computers/getting-started/install.adoc
```diff
@@ -106,7 +106,7 @@ Finally, respond **Yes** to the "Are you sure you want to continue?" popup to be
 
 image::images/imager/are-you-sure.png[alt="Confirming a reimage of a storage device in Imager."]
 
-If you see an admin prompt asking for permissions to read and write to your storage medium, it's safe to proceed.
+If you see an admin prompt asking for permissions to read and write to your storage medium, it's safe to proceed (provided that you have definitely selected the correct storage device earlier).
```
nathan-contino, line 109:
```suggestion
If you see an admin prompt asking for permissions to read and write to your storage medium, grant Imager the permissions to proceed.
```

I see now. I think you're correct that "it's safe to proceed" is too strong -- and potentially misleading. But we've already dropped a lot of warnings about overwriting storage devices earlier in this passage.

Consider this rewrite instead, which satisfies your core concern (that we're telling folks an irreversible action is _safe_) without polluting the narrative with additional asides and parentheticals.

nathan-contino (commented):
Thanks for elaborating on your reasoning. I think I've found a way to improve this piece of documentation from both perspectives. See my code suggestion.

documentation/asciidoc/computers/getting-started/install.adoc
tdewey-rpi: > I see now. I think you're correct that "it's safe to proceed" is too strong -- and potentially misleading. But we've already dropped a lot of warnings about overwriting storage devices earlier in this passage.
> 
> Consider this rewrite instead, which satisfies your core concern (that we're telling folks an irreversible action is _safe_) without polluting the narrative with additional asides and parentheticals.

This would satisfy my concerns around conciseness and effectiveness.

sipos: This seems like better wording than my suggestion. I think someone should adopt this suggestion. Thanks for the suggestion.

nathan-contino (approved):
LGTM. I'll run this by TPTB when they return from vacation.

PR #3531, opened 2024-02-09
Add information about maximum SD card size

nathan-contino:
* added inline instead of in a note because there was a natural fit (no need to disrupt the flow with a note if you can avoid it)
* added a sentence about performance because that always seems to be the next question people ask, so we might as well give some guidance

### DIFF
--- a/documentation/asciidoc/computers/getting-started/setting-up.adoc
+++ b/documentation/asciidoc/computers/getting-started/setting-up.adoc
@@ -68,7 +68,7 @@ image::images/peripherals/sd-card.png[alt="Inserting a microSD card into a Raspb
 
 [[recommended-capacity]]
 
-We recommend using an SD card with at least 16GB of storage for Raspberry Pi OS installations. For Raspberry Pi OS Lite, we recommend at least 4GB. You can use any SD card whose capacity exceeds this minimum size; there is no maximum size. As with any other boot media, you'll see improved performance on SD cards with faster read and write speeds.
+We recommend using an SD card with at least 32GB of storage for Raspberry Pi OS installations. For Raspberry Pi OS Lite, we recommend at least 16GB. You can use any SD card with a capacity of less than 1.5TB. As with any other boot media, you'll see improved performance on SD cards with faster read and write speeds.
 
 Because of a hardware limitation, the following devices will only boot from a boot partition of 256GB or less:
 
--- a/documentation/asciidoc/computers/getting-started/setting-up.adoc
+++ b/documentation/asciidoc/computers/getting-started/setting-up.adoc
@@ -68,7 +68,7 @@ image::images/peripherals/sd-card.png[alt="Inserting a microSD card into a Raspb
 
 [[recommended-capacity]]
 
-We recommend using an SD card with at least 32GB of storage for Raspberry Pi OS installations. For Raspberry Pi OS Lite, we recommend at least 16GB. You can use any SD card with a capacity of less than 1.5TB. As with any other boot media, you'll see improved performance on SD cards with faster read and write speeds.
+We recommend using an SD card with at least 32GB of storage for Raspberry Pi OS installations. For Raspberry Pi OS Lite, we recommend at least 16GB. You can use any SD card with a capacity of less than 2TB. Capacities above 2TB are currently not supported due to limitations in the https://en.wikipedia.org/wiki/Master_boot_record[MBR]. As with any other boot media, you'll see improved performance on SD cards with faster read and write speeds.
 
 Because of a hardware limitation, the following devices will only boot from a boot partition of 256GB or less:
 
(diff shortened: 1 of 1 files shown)


documentation/asciidoc/computers/getting-started/setting-up.adoc
```diff
@@ -68,7 +68,7 @@ image::images/peripherals/sd-card.png[alt="Inserting a microSD card into a Raspb
 
 [[recommended-capacity]]
 
-We recommend using an SD card with at least 16GB of storage with Raspberry Pi OS. If you use Raspberry Pi OS Lite, we recommend at least 4GB.
+We recommend using an SD card with at least 16GB of storage for Raspberry Pi OS installations. For Raspberry Pi OS Lite, we recommend at least 4GB. You can use any SD card whose capacity exceeds this minimum size; there is no maximum size. As with any other boot media, you'll see improved performance on SD cards with faster read and write speeds.
```
aallan, line 71: Probably want to bump that to 32GB and 16GB, that was written a long time ago. Also we might want to be a bit less direct that "there is no maximum size" because when 2D cards hit 2TB (if they can) then there will be, but it'll be a software limit rather than a hardware limit.

tdewey-rpi, line 71: While strictly true regarding the maximum size, there's an implementation wrinkle we should probably make people aware of.

Today, the boot process expects an MBR partition table. MBR partition tables can support partitions up to 2TB in size. While I've yet to see a micoSD card larger than 1.5TB, we're close enough that it might be worth capturing that.

There are some workarounds for larger still drives, such as using a Hybrid GPT/MBR partitioning scheme: https://github.com/raspberrypi/firmware/issues/1465#issuecomment-814005398

However, directly using GPT partition tables is, to my understanding, not available today. So the naive way of using >2TB drives would not work. Paging @timg236 for visibility.

nathan-contino: [Should we be worried?](https://www.tomsguide.com/news/the-worlds-first-2tb-microsd-card-is-here-what-you-need-to-know)

aallan:
Right. Stick a sentence in along the lines of "In theory there is no maximum limit, however the practical size is 2TB due to the boot process expecting a MBR partition table."

documentation/asciidoc/computers/getting-started/setting-up.adoc
nathan-contino:
```suggestion
We recommend using an SD card with at least 32GB of storage for Raspberry Pi OS installations. For Raspberry Pi OS Lite, we recommend at least 16GB. You can use any SD card with a capacity of less than 1.5TB. As with any other boot media, you'll see improved performance on SD cards with faster read and write speeds.
```
Added your minimum size tweaks and a simple version of @tdewey-rpi's comment below. I'm not sure if the Getting Started page is the right place to discuss partition tables, but if others feel the information should live here, I'm open to it.

timg236: 4TB ... too rich for me!

The bootloader uses 32-bit logical block addresses so 2TB I think is the limit for now. No promises for when this will happen but I think NVMe would be the highest priority for very large drives and 4K native sectors.

Hybrid/GPT that have MBR / FAT at the start of the disk should work, but I'm fairly sure that A/B booting won't work unless all the FAT partitions appear in the "MBR"

documentation/asciidoc/computers/getting-started/setting-up.adoc
```diff
@@ -68,7 +68,7 @@ image::images/peripherals/sd-card.png[alt="Inserting a microSD card into a Raspb
 
 [[recommended-capacity]]
 
-We recommend using an SD card with at least 16GB of storage with Raspberry Pi OS. If you use Raspberry Pi OS Lite, we recommend at least 4GB.
+We recommend using an SD card with at least 32GB of storage for Raspberry Pi OS installations. For Raspberry Pi OS Lite, we recommend at least 16GB. You can use any SD card with a capacity of less than 1.5TB. As with any other boot media, you'll see improved performance on SD cards with faster read and write speeds.
```
nathan-contino, line 71:
```suggestion
We recommend using an SD card with at least 32GB of storage for Raspberry Pi OS installations. For Raspberry Pi OS Lite, we recommend at least 16GB. You can use any SD card with a capacity of less than 2TB. Capacities above 2TB are currently not supported due to limitations in the https://en.wikipedia.org/wiki/Master_boot_record[MBR]. As with any other boot media, you'll see improved performance on SD cards with faster read and write speeds.
```

nathan-contino:
Updated according to your feedback -- in my experience, this is a reasonable place to link to Wikipedia for folks to learn more about MBR. All we need to do is explain that this limitation stems from MBR and point folks somewhere where they can learn more.

PR #3532, opened 2024-02-12
Fix incorrect reference for SPI loopback test

edrutte:
Currently links to general troubleshooting page by mistake, instead of the SPI troubleshooting section with the loopback test.


nathan-contino (approved):
LGTM, thank you!

aallan (approved):

PR #3523, opened 2024-02-01
Rename Pico Probe -> Debug Probe

nathan-contino:
* Closes https://github.com/raspberrypi/openocd/issues/94

Documentation rename in tandem with https://github.com/raspberrypi/picoprobe/tree/debugprobe_rename

Turns out the only docs mentions are this and the Getting Started guide section. I'm working on that fix in parallel.

TODO fixes elsewhere:

* Change the "Picoprobe" link on the [product page](https://www.raspberrypi.com/products/debug-probe/) to "debugprobe" (with an updated link pointing to https://github.com/raspberrypi/debugprobe).

* Update the product brief (https://datasheets.raspberrypi.com/debug/raspberry-pi-debug-probe-product-brief.pdf), which contains just one mention of Picoprobe and one adjacent link to the repo. (`The Debug Probe is based on Raspberry Pi Pico hardware and runs the open source Raspberry Pi Picoprobe software (https://github.com/raspberrypi/picoprobe)` -> `The Debug Probe is based on Raspberry Pi Pico hardware and runs the open source Raspberry Pi debugprobe software (https://github.com/raspberrypi/debugprobe)`)

* Update the [Raspberry Pi 3-pin Debug Connector Specification](https://datasheets.raspberrypi.com/debug/debug-connector-specification.pdf), which contains one "Picoprobe" term, but no link. (`the debug dongle, e.g., RP2040 running Picoprobe` -> `the debug dongle, e.g., RP2040 running debugprobe`)

There's also a whole lot of forum posts. Nothing we can do about that.

PR #3535, opened 2024-02-15
Add udev rule for running openocd

peterharperuk:
I can run openocd without sudo on Raspberry Pi OS and Ubuntu when running on rpi, but my virtual box seems to need it. This seems to cause people to worry and raise support issues, so I suggest we add some details about the udev rule required.

### DIFF
--- a/documentation/asciidoc/microcontrollers/debug-probe/swd-connection.adoc
+++ b/documentation/asciidoc/microcontrollers/debug-probe/swd-connection.adoc
@@ -14,6 +14,12 @@ $  sudo openocd -f interface/cmsis-dap.cfg -f target/rp2040.cfg -c "adapter spee
 
 NOTE: When you use the Debug Probe to upload a binary the ELF version of the file is used, not the UF2 file that you would use when you drag-and-drop.
 
+On linux you might be able to run `openocd` without `sudo` depending on your distribution. If it doesn't work without `sudo` you can add a udev rule, e.g. add the following to `/etc/udev/rules.d/51-usb-perms.rules` and reload your udev rules with `sudo udevadm control --reload ; sudo udevadm trigger`
+
+----
+SUBSYSTEM=="usb", ATTR{idVendor}=="0a5c", ATTR{idProduct}=="000c", GROUP="plugdev", MODE="0660"
+----
+
 === Debugging with SWD
 
 It’ll also let you use `openocd` in server mode, and connect GDB, which gives you break points and “proper” debugging.


peterharperuk:
@lurch

documentation/asciidoc/microcontrollers/debug-probe/swd-connection.adoc
```diff
@@ -14,6 +14,12 @@ $  sudo openocd -f interface/cmsis-dap.cfg -f target/rp2040.cfg -c "adapter spee
 
 NOTE: When you use the Debug Probe to upload a binary the ELF version of the file is used, not the UF2 file that you would use when you drag-and-drop.
 
+On linux you might be able to run `openocd` without `sudo` depending on your distribution. If it doesn't work without `sudo` you can add a udev rule, e.g. add the following to `/etc/udev/rules.d/51-usb-perms.rules` and reload your udev rules with `sudo udevadm control --reload ; sudo udevadm trigger`
```
JamesH65, line 17: linux->Linux

documentation/asciidoc/microcontrollers/debug-probe/swd-connection.adoc
```diff
@@ -14,6 +14,12 @@ $  sudo openocd -f interface/cmsis-dap.cfg -f target/rp2040.cfg -c "adapter spee
...
+
+----
+SUBSYSTEM=="usb", ATTR{idVendor}=="0a5c", ATTR{idProduct}=="000c", GROUP="plugdev", MODE="0660"
+----
```
nathan-contino, line 21:
```suggestion
INFO
====
If this command requires `sudo`, you can remove the `sudo` requirement by adding the following udev rule to `/etc/udev/rules.d/51-usb-perms.rules`:

----
SUBSYSTEM=="usb", ATTR{idVendor}=="0a5c", ATTR{idProduct}=="000c", GROUP="plugdev", MODE="0660"
----

Reload your udev rules with `sudo udevadm control --reload ; sudo udevadm trigger`. You should now be able to run `openocd` without `sudo`.
====
```

Suggestion: move this to an INFO admonition block, since it's not part of the standard instruction flow. Streamline the wording a bit. Some minor changes to follow the documentation style guide, e.g. introducing code blocks with a colon and a clear description.

NOTE: we should probably merge https://github.com/raspberrypi/documentation/pull/3523 first because it's big, important, and this PR could introduce merge conflicts since there's lot of scattered renaming in the same file edited in this PR.

peterharperuk:
I'm going to close this and reopen if I can make sense of what's going on

peterharperuk:
Note: on Linux it shouldn't be necessary to run openocd as root or add another udev file. When you install openocd with apt it installs /lib/udev/rules.d.60-openocd.rules. One of the last rules is looking for a product that contains the text "CMSIS-DAP", which the debug probe uses. This gives permission if you're in the plugdev group. This should also work for a debug probe on a pico.

PR #3545, opened 2024-02-26
Update gstreamer.adoc

dronevdl:
Without this addition, the following error will occur.

ERROR: from element /GstPipeline:pipeline0/GstLibcameraSrc:libcamerasrc0: Internal data stream error. Additional debug info:
../src/gstreamer/gstlibcamerasrc.cpp(310): processRequest (): /GstPipeline:pipeline0/GstLibcameraSrc:libcamerasrc0: streaming stopped, reason error (-5)

and there will be an exception in /var/log/kern.log

### DIFF
--- a/documentation/asciidoc/computers/camera/gstreamer.adoc
+++ b/documentation/asciidoc/computers/camera/gstreamer.adoc
@@ -52,7 +52,7 @@ gst-launch-1.0 udpsrc address=192.168.0.3 port=5000 caps=application/x-rtp ! rtp
 
 [,bash]
 ----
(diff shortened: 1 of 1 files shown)


aallan:
Ping @davidplowman for comment

davidplowman:
Don't know too much about gstreamer. Maybe @6by9 can comment whether this looks good?

6by9 (approved):
Yes, since GStreamer 1.18 v4l2h264enc has taken the profile and level from the caps, not from the extra-controls. Not setting it defaults to level 1.0, which is max 128×96@30.9fps!

aallan (approved):

PR #3551, opened 2024-03-06
Add fallback to handle new doxygen helper files

nelliemckesson:
Newer versions of doxygen add some helper files that we don't want to process as part of the docs. This PR adds a more graceful fallback to handle those files.


nelliemckesson:
@aallan ready for review/merge

aallan (approved):

PR #3564, opened 2024-03-20
Update regex to use valid escape sequences

nelliemckesson:
Python wants a specific syntax when using escape sequences in `re.` commands. We used to be able to get away with it, but they switched the type of warning they use for these, which means we can't really ignore it anymore. The updates the offending regex accordingly to use raw strings.

### DIFF
--- a/README.md
+++ b/README.md
@@ -60,6 +60,8 @@ And install the [Ruby bundler](https://bundler.io/):
 $ gem install bundler
 ```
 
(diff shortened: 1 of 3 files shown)


aallan (approved):

nelliemckesson:
(See description here for reference: https://stackoverflow.com/a/77531416)

aallan:
> (See description here for reference: https://stackoverflow.com/a/77531416)

Sometimes I genuinely hate progress. It was working just fine before. 🙄 

Thanks Nellie.

PR #3576, opened 2024-04-02
Clarify microSD/SD in setting-up.adoc

JohnVeness:


nathan-contino (approved):
LGTM

documentation/asciidoc/computers/getting-started/setting-up.adoc
```diff
@@ -60,7 +60,9 @@ Plug your power supply into the port marked "POWER IN", "PWR IN", or "PWR". Some
 
 Raspberry Pi models lack onboard storage, so you have to supply it. You can boot your Raspberry Pi from an operating system image installed on any supported media: microSD cards are used commonly, but USB storage, network storage, and storage connected via a PCIe HAT are also available. However, only recent Raspberry Pi models support all of these media types.
 
-All Raspberry Pi consumer models since the Raspberry Pi 1 Model A+ feature a microSD slot. Your Raspberry Pi automatically boots from the microSD slot when the slot contains a card.
+All Raspberry Pi consumer models since the Raspberry Pi 1 Model A+ feature a microSD slot. Models prior to that feature a full-sized SD slot. This documentation will usually say "SD" to refer to either size.
```
nathan-contino, line 63:
```suggestion
All Raspberry Pi consumer models since the Raspberry Pi 1 Model B+ feature a microSD slot. Models prior to that feature a full-sized SD slot. This documentation will usually say "SD" to refer to either size.
```
The B+ was launched before the A+, and the + is the important part here.

documentation/asciidoc/computers/getting-started/setting-up.adoc
```diff
@@ -60,7 +60,9 @@ Plug your power supply into the port marked "POWER IN", "PWR IN", or "PWR". Some
 
 Raspberry Pi models lack onboard storage, so you have to supply it. You can boot your Raspberry Pi from an operating system image installed on any supported media: microSD cards are used commonly, but USB storage, network storage, and storage connected via a PCIe HAT are also available. However, only recent Raspberry Pi models support all of these media types.
 
-All Raspberry Pi consumer models since the Raspberry Pi 1 Model A+ feature a microSD slot. Your Raspberry Pi automatically boots from the microSD slot when the slot contains a card.
+All Raspberry Pi consumer models since the Raspberry Pi 1 Model B+ feature a microSD slot. Models prior to that feature a full-sized SD slot. This documentation will usually say "SD" to refer to either size.
```
aallan, line 63: I don't think it's called "Raspberry Pi 1 Model B+", the B+ was just called the "Raspberry Pi, Model B+".
I think we're retcon'ing the original Raspberry Pi as the "Raspberry Pi 1" (so we don't have to keep saying "except for the original Raspberry Pi") but I'd baulk at retconning the B+.
@LizUpton Was there ever an "original Raspberry Pi, Model A" ..?

lurch: Yup https://raspi.tv/2013/raspberry-pi-model-a-first-impressions-and-photos

aallan:
> How about a table…

The only two models with a full sized SD card slot is the original Model B and Model A. Those are also the only models we've ever discontinued. It's not an issue that justifies any actual space in the documentation, certainly not a table!

aallan:
> The table could also say which models support booting from USB storage, network storage, and storage connected via a PCIe HAT.

The only model that supports a PCIe HAT is the 5, so that's not a useful table either.

documentation/asciidoc/computers/getting-started/setting-up.adoc
aallan: Huh. I'd entirely forgotten that existed.

lurch: Yeah, IIRC the "original Raspberry Pi, Model A" used the same PCB as "original Raspberry Pi, Model B", just with slightly different components soldered to it.

JohnVeness:
> The only two models with a full sized SD card slot is the original Model B and Model A. Those are also the only models we've ever discontinued.

I was actually wondering about that - should the documentation still cover the original model A and B?

LizUpton:
> I was actually wondering about that - should the documentation still cover the original model A and B?

No, it should not. They're discontinued products, we sold a few thousand of each (compared to millions of customers for each of the modern products), and we have significantly more important parts of the documentation to work on. 

In general, we try to discourage sub- and copy-edit-level PRs to the documentation stack, especially when they come in piecemeal. There is an internal process (starting with the newest additions to the corpus; the documentation was only properly professionalised and brought in-house in 2020, so you'll observe that material from before then can lag) for sub and copy, but little PRs like this are very time-consuming by comparison, require the attention of far more people, and are not very helpful for us.

JohnVeness:
Woah, OK. That's very discouraging.

PR #3580, opened 2024-04-02
Move dual-HDMI note in setting-up.adoc

JohnVeness:
When reading from top to bottom and if you knew nothing about Raspberry Pi, it is jarring to read about dual HDMI connections when you haven't yet been introduced to what connections the Pi has!

Also, should this paragraph be a "NOTE:"?


nathan-contino (approved):
LGTM

PR #3581, opened 2024-04-02
Change WiFi to Wi-Fi throughout

JohnVeness:


nathan-contino (approved):
LGTM

aallan:
I still think we should use "wireless" rather than "wi-fi" but the style guide says otherwise. 🤷‍♂️

PR #3582, opened 2024-04-02
Use British English for "tick"

JohnVeness:


aallan:
I think checkbox is the commonly accepted term, even here in the UK.

JohnVeness:
> I think checkbox is the commonly accepted term, even here in the UK.

I don't think that "check" as a verb (which is what I changed in this PR) is commonly used here.

aallan:
It is in computing circles. I don't know a single programmer than calls it a tick box, it's a check box.

PR #3583, opened 2024-04-03
Fix series-1 plus-model confusion in getting started section

nathan-contino:
Turns out both + models had TRRS! Only the original model 1s had RCA. And those aren't mentioned here because that would just bloat the table unnecessarily 😄

### DIFF
--- a/documentation/asciidoc/computers/getting-started/setting-up.adoc
+++ b/documentation/asciidoc/computers/getting-started/setting-up.adoc
@@ -115,7 +115,7 @@ Raspberry Pi models have the following display connectivity:
 |HDMI, audio and composite out via 3.5mm http://en.wikipedia.org/wiki/Phone_connector_(audio)#TRRS_standards[TRRS] jack
 
 |Raspberry Pi 1 Model A+
-|HDMI, RCA connector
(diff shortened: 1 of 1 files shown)


documentation/asciidoc/computers/getting-started/setting-up.adoc
```diff
@@ -115,7 +115,7 @@ Raspberry Pi models have the following display connectivity:
 |HDMI, audio and composite out via 3.5mm http://en.wikipedia.org/wiki/Phone_connector_(audio)#TRRS_standards[TRRS] jack
 
 |Raspberry Pi 1 Model A+
```
aallan, line 117: Are we really retcon'ing the B+ (and the A+) to be "Raspberry Pi 1, Model B+"..? Is this a thing, did I miss a thing?

nathan-contino: Good point. I suppose the 1 doesn't _hurt_ anything, and if anything helps identify the device (a la Apple Watch Series 0). But if you feel strongly about it I can remove the 1 since it is not strictly required and might be technically incorrect.

aallan (approved):

PR #3594, opened 2024-04-18
Update lxml version and fix some syntax warnings

nelliemckesson:


nathan-contino (approved):
LGTM!

nathan-contino:
Expected; Nellie's fixing it tomorrow!

tests/run_documentation_tests.sh
mudge, line 1: Do you want to add these to the GitHub Actions configuration so we only ever deploy when the tests are passing?

nelliemckesson: Ooh yes! Should I just add a step near the end to run the .sh script?

nelliemckesson: Ok cool, done, seems to be working!

nelliemckesson: (It has to run relatively late in the process bc it depends on the files generated during make :/ )

.github/workflows/build.yml
```diff
@@ -40,13 +40,17 @@ jobs:
...
+          wget https://www.doxygen.nl/files/doxygen-1.10.0.linux.bin.tar.gz
+          tar xf doxygen-1.10.0.linux.bin.tar.gz -C "$HOME"
+          echo "$HOME/doxygen-1.10.0/bin" >> $GITHUB_PATH
       - name: Build Doxygen documentation
         run: make build_doxygen_adoc
       - name: Build documentation
         run: make -j 2
+      - name: Run tests
+        run: |
+          chmod 777 ./tests/run_documentation_tests.sh
```
mudge, line 52: Rather than changing the mode of this file only for the tests, git will respect the mode of a committed file so maybe commit it with a mode of `755`?

nelliemckesson: Nice, updated accordingly, still works!

mudge:
If you're happy with the output after the Doxygen and lxml upgrade, no objections from me!

PR #3614, opened 2024-05-08
Update led_blink_warnings.adoc

JamesH65:
These errors are also applicable to the Pi 5


JamesH65:
@timg236 Is this now correct for the Pi 5?

lurch:
Presumably the "RP1 not found" is also specific to just the Pi 5?

timg236:
> Presumably the "RP1 not found" is also specific to just the Pi 5?

RP1 is Pi5 family only

timg236:
Changes look good though

nathan-contino (approved):
LGTM!

Issue #3573, opened 2024-03-27
Link from voltage problem lightning indicator is broken?


sandyol55:
I guess this may not be a documentation issue but not sure where else to flag it?

On a 2G Pi4 running latest Bookworm.
Yellow Lightning power indicator came up with a link to :-
raspberrypi.com/documentation/computers/raspberry-pi-5.html#powering-raspberry-pi-5
which gave a 404 message.

aallan:
@nathan-contino Where did you move that content to when we tore down the Pi 5 specific page? I'll set up a redirect for now, and mail Simon to fix it for the next OS update.

PR #3616, opened 2024-05-08
Add anchor point for Raspberry Pi 5 power OS tooltip

nathan-contino:
* Closes https://github.com/raspberrypi/documentation/issues/3573
* OS links to https://www.raspberrypi.com/documentation/computers/raspberry-pi-5.html#powering-raspberry-pi-5
* currently redirects to https://www.raspberrypi.com/documentation/computers/raspberry-pi.html#powering-raspberry-pi-5 (note that the base page redirects, but the anchor point remains the same)
* This anchor point doesn't exist, so you just get dumped at the top of the page
* should go to https://www.raspberrypi.com/documentation/computers/raspberry-pi.html#recommended-power-supplies

* Adds an explicit additional anchor point a la https://docs.asciidoctor.org/asciidoc/latest/attributes/id/#add-additional-anchors-to-a-section

### DIFF
--- a/documentation/asciidoc/computers/raspberry-pi/power-supplies.adoc
+++ b/documentation/asciidoc/computers/raspberry-pi/power-supplies.adoc
@@ -4,7 +4,7 @@ The power supply requirements differ by Raspberry Pi model. All models require a
 
 The current consumed by each Raspberry Pi depends on the peripherals connected.
 
-=== Recommended power supplies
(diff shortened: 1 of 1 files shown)


nathan-contino:
Sadly, it looks like our toolchain doesn't support multiple explicit anchor points in the way that AsciiDoctor ought to. I've moved the anchor point to a nearby relevant space because Works OK and It is Fine.

PR #3640, opened 2024-05-15
Relocate kernel build files to /firmware also in the cross-compiling section

aidatorajiro:
In the recent change, kernel images and other build files are relocated to /boot/firmware, however, the description in cross-compiling section have not been updated accordingly.


aidatorajiro:
Well, I realized that what was relocated was just a mount point. Closing issue.

nathan-contino:
Thanks! I _believe_ I updated all of the relevant documentation prose already, but if you notice anywhere that hasn't been updated, let me know.

There are a couple of lingering `/boot/config.txt` usages, but they are restricted to older OS instructions.

Issue #3556, opened 2024-03-11
Make the `$ ` at the beginning of interactive code/console snippets uncopyable.


nathan-contino:
As per the suggestion here: https://github.com/raspberrypi/documentation/pull/3549#discussion_r1517429422

There are advantages to prefixing interactive CLI snippets with the `$` symbol. However, it can decrease copy/paste friendliness and present an obstacle to new users. We should:

* omit the `$ ` prefix when you click the the "copy to clipboard" button on code snippets
* [stretch goal] make the `$ ` prefix unselectable in CSS styling such that users physically cannot copy and paste it when they grab the snippet manually (without clicking the 'copy to clipboard' button)

Right now we manually add the `$ ` prefix at the beginning of these snippets. We should consider *automatically* adding the prefix to the beginning of `console` type snippets.

nathan-contino:
For example, see the code snippets in this section: https://www.raspberrypi.com/documentation/computers/configuration.html#change-a-users-password

Note that interactive code snippets are sometimes multi-line, but those lines should almost always get copy/pasted one at a time. For multiline snippets, we want a `$ ` prefix on every line, and either _no_ "copy to clipboard" button (because you would never want to copy all the lines at once) or a "copy to clipboard" button for each line individually (which sounds slightly awkward). You could possible break up multiline snippets into multiple one-line snippets either automatically or manually (and then throw a warning if someone creates a multiline `console` snippet in the future).

aallan:
> omit the $  prefix when you click the the "copy to clipboard" button on code snippets

This is something I want.

> We should consider automatically adding the prefix to the beginning of console type snippets.

This sounds awfully fragile, and means combing the documentation for every code block and deciding whether it is or isn't a command line example.

nathan-contino:
Too fast!

PR #3656, opened 2024-05-17
Add syntax highlighting for labelled code snippets

nathan-contino:
Not sure why we didn't have this before. Opening this because it makes longer snippets infinitely more readable.

Couple of examples:

Smol:

Screenshot 2024-05-17 at 09 53 16

Large:

Screenshot 2024-05-17 at 09 55 29

Minor cost (70 lines of CSS). Not particularly breakable (worst case, snippets just end up not-highlighted). Uses rouge, which we already use to highlight syntax in our PDF documentation.

Not all of our snippets are correctly marked, but this makes it easier to spot unmarked snippets and fix them.

Also fixes https://github.com/raspberrypi/documentation/issues/3556 by rendering the '$ ' at the beginning of `console`-marked snippets as uncopyable and trimming it from the text returned by the 'copy' button (in a safe non-explodey manner).

### DIFF
--- a/jekyll-assets/css/syntax-highlighting.css
+++ b/jekyll-assets/css/syntax-highlighting.css
@@ -1,0 +1,70 @@
+.highlight pre { background-color: #404040 }
+.highlight .hll { background-color: #404040 }
+.highlight .c { color: #999999; font-style: italic } /* Comment */
+.highlight .err { color: #a61717; background-color: #e3d2d2 } /* Error */
+.highlight .g { color: #d0d0d0 } /* Generic */
+.highlight .k { color: #6ab825; font-weight: bold } /* Keyword */
+.highlight .l { color: #d0d0d0 } /* Literal */
+.highlight .n { color: #d0d0d0 } /* Name */
+.highlight .o { color: #d0d0d0 } /* Operator */
+.highlight .x { color: #d0d0d0 } /* Other */
+.highlight .p { color: #d0d0d0 } /* Punctuation */
+.highlight .cm { color: #999999; font-style: italic } /* Comment.Multiline */
+.highlight .cp { color: #cd2828; font-weight: bold } /* Comment.Preproc */
+.highlight .c1 { color: #999999; font-style: italic } /* Comment.Single */
+.highlight .cs { color: #e50808; font-weight: bold; background-color: #520000 } /* Comment.Special */
+.highlight .gd { color: #d22323 } /* Generic.Deleted */
+.highlight .ge { color: #d0d0d0; font-style: italic } /* Generic.Emph */
+.highlight .gr { color: #d22323 } /* Generic.Error */
+.highlight .gh { color: #ffffff; font-weight: bold } /* Generic.Heading */
+.highlight .gi { color: #589819 } /* Generic.Inserted */
+.highlight .go { color: #cccccc } /* Generic.Output */
+.highlight .gp { color: #aaaaaa } /* Generic.Prompt */
+.highlight .gs { color: #d0d0d0; font-weight: bold } /* Generic.Strong */
+.highlight .gu { color: #ffffff; text-decoration: underline } /* Generic.Subheading */
+.highlight .gt { color: #d22323 } /* Generic.Traceback */
+.highlight .kc { color: #6ab825; font-weight: bold } /* Keyword.Constant */
+.highlight .kd { color: #6ab825; font-weight: bold } /* Keyword.Declaration */
+.highlight .kn { color: #6ab825; font-weight: bold } /* Keyword.Namespace */
+.highlight .kp { color: #6ab825 } /* Keyword.Pseudo */
+.highlight .kr { color: #6ab825; font-weight: bold } /* Keyword.Reserved */
+.highlight .kt { color: #6ab825; font-weight: bold } /* Keyword.Type */
+.highlight .ld { color: #d0d0d0 } /* Literal.Date */
+.highlight .m { color: #3677a9 } /* Literal.Number */
+.highlight .s { color: #ed9d13 } /* Literal.String */
+.highlight .na { color: #bbbbbb } /* Name.Attribute */
+.highlight .nb { color: #24909d } /* Name.Builtin */
+.highlight .nc { color: #447fcf; text-decoration: underline } /* Name.Class */
+.highlight .no { color: #40ffff } /* Name.Constant */
+.highlight .nd { color: #ffa500 } /* Name.Decorator */
+.highlight .ni { color: #d0d0d0 } /* Name.Entity */
+.highlight .ne { color: #bbbbbb } /* Name.Exception */
+.highlight .nf { color: #447fcf } /* Name.Function */
+.highlight .nl { color: #d0d0d0 } /* Name.Label */
+.highlight .nn { color: #447fcf; text-decoration: underline } /* Name.Namespace */
+.highlight .nx { color: #d0d0d0 } /* Name.Other */
+.highlight .py { color: #d0d0d0 } /* Name.Property */
+.highlight .nt { color: #6ab825; font-weight: bold } /* Name.Tag */
+.highlight .nv { color: #40ffff } /* Name.Variable */
+.highlight .ow { color: #6ab825; font-weight: bold } /* Operator.Word */
+.highlight .w { color: #666666 } /* Text.Whitespace */
+.highlight .mf { color: #3677a9 } /* Literal.Number.Float */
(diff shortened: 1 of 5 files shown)


jekyll-assets/scripts/copy-to-clipboard.js
```diff
@@ -48,7 +48,24 @@ window.addEventListener('load', function() {
       if (trigger.parentNode.querySelector('div.line')) {
         var text = extractDoxygenCode(trigger.parentNode);
       } else {
-        var text = trigger.parentNode.querySelector('pre').textContent;
+        var text = trigger.parentNode.querySelector('pre').textContent
```
nelliemckesson, line 51: @nathan-contino removed the closing `;` ?

jekyll-assets/scripts/copy-to-clipboard.js
```diff
@@ -48,7 +48,24 @@ window.addEventListener('load', function() {
...
+          // trim the '$ ' from the clipboard copy so users don't have to trim it themselves
+          text_split_into_lines = text_split_into_lines.map(function(x) {
+            if (x.startsWith('$ ')) {
+              return x.substring(2);
+            }
+            else return x;
+          })
+
+          // re-assemble the snippet into multiple lines
+          text = text_split_into_lines.join('\n');
```
nelliemckesson, line 67: @nathan-contino curious why you chose to do it this way rather than with a regex one-liner, e.g., `text.replaceAll(/^$\s/, "").replaceAll(/\n$\s/, "\n");` ?

nelliemckesson:
@nathan-contino I have a couple of comments for you here. Also, I generally prefer to use unique feature branches, in case we need to roll something back (easier to disentangle if unique features are maintained in separate branches).

nelliemckesson:
(^ Also fyi @mudge , for the purposes of future reviews...)

jekyll-assets/scripts/copy-to-clipboard.js
```diff
@@ -48,7 +48,24 @@ window.addEventListener('load', function() {
...
+        var text = trigger.parentNode.querySelector('pre').textContent
+        
+        // if the code snippet represents a console snippet
+        if (trigger.parentNode.querySelector('pre').querySelector('code') != null && trigger.parentNode.querySelector('pre').querySelector('code').getAttribute('data-lang') == 'console') {
+          // for each line of the code snippet
+          var text_split_into_lines = text.split('\n');
+
+          // trim the '$ ' from the clipboard copy so users don't have to trim it themselves
+          text_split_into_lines = text_split_into_lines.map(function(x) {
+            if (x.startsWith('$ ')) {
```
nelliemckesson, line 60: @nathan-contino Will the `$` always be followed by a simple space, or do we need to check for any space char, e.g., tabs, etc... ?

scripts/create_build_adoc.py
```diff
@@ -82,6 +82,7 @@ def check_no_markdown(filename):
 :page-sub_title: {}
 :sectanchors:
 :figure-caption!:
+:source-highlighter: rouge
```
nelliemckesson, line 85: @nathan-contino  This should probably also be added to create_build_adoc_doxygen.py, for consistency.

jekyll-assets/scripts/copy-to-clipboard.js
mudge: We use a JavaScript formatter (e.g. [Prettier](https://prettier.io) or, more recently, [Biome](https://biomejs.dev)) to check for consistency on other projects, might it be useful here too?

jekyll-assets/scripts/copy-to-clipboard.js
```diff
@@ -48,7 +48,24 @@ window.addEventListener('load', function() {
       if (trigger.parentNode.querySelector('div.line')) {
         var text = extractDoxygenCode(trigger.parentNode);
       } else {
-        var text = trigger.parentNode.querySelector('pre').textContent;
+        var text = trigger.parentNode.querySelector('pre').textContent
+        
+        // if the code snippet represents a console snippet
+        if (trigger.parentNode.querySelector('pre').querySelector('code') != null && trigger.parentNode.querySelector('pre').querySelector('code').getAttribute('data-lang') == 'console') {
```
mudge, line 54: Could you reduce some repetition by doing a single query with a more powerful CSS selector here? e.g.

```javascript
if (trigger.parentNode.querySelector('pre > code[data-lang="console"]')) {
```

jekyll-assets/scripts/copy-to-clipboard.js
mudge: Agreed with @nelliemckesson, could you potentially do this with a one-liner (note the global `g` and multiline `m` flags), e.g.

```javascript
text.replaceAll(/^\$\s+/gm, "")
```

jekyll-assets/css/syntax-highlighting.css
```diff
@@ -0,0 +1,82 @@
...
+.highlight .gp, .highlight .gp + .w {
+	/* Disable text selection highlighting for the '$ ' prefix at the beginning of console snippets (if it exists -- otherwise has no effect)
+   * Credit: https://stackoverflow.com/a/4407335/1558022 with modifications by Nate Contino */
+  -webkit-touch-callout: none; /* iOS Safari */
+    -webkit-user-select: none; /* Safari */
+     -khtml-user-select: none; /* Konqueror HTML */
+       -moz-user-select: none; /* Firefox */
+        -ms-user-select: none; /* Internet Explorer/Edge */
+            user-select: none; /* Non-prefixed version, currently
+                                  supported by Chrome and Opera */
```
mudge, line 81: Looking at the [Can I Use data for `user-select`](https://caniuse.com/mdn-css_properties_user-select), you might be able to get away with the non-prefixed, `-webkit-` and `-ms-` prefixes only. Might it also be worth offloading this kind of thing to a tool such as [autoprefixer](https://github.com/postcss/autoprefixer)?

jekyll-assets/scripts/copy-to-clipboard.js
nathan-contino: Thanks, I didn't realize `querySelector` could handle something that complex! That is much cleaner.

jekyll-assets/css/syntax-highlighting.css
nathan-contino: Personally speaking I'd rather use all of the possibilities since it's a very small amount of code and I'd rather try to be compliant with each of the different browsers.

I'm skeptical of adding a library for this since it's a one-time use case, but if we do need prefixing elsewhere, definitely worth considering.

Nellie should call this one since she calls the shots when it comes to the toolchain anyway.

nathan-contino:
Note, for an upcoming follow-up PR: I found one degenerate case so far where this parsing falls apart (https://www.raspberrypi.com/documentation/computers/os.html#install-python-libraries-using-pip), specifically when we use the following venv prefixes:

```
(<env-name>) $
```

```
(venv) $
```

```
(.venv) $
```

The first one seems to be the trickiest, since it includes a number of special characters. But ideally we should match any amount of text in between parens in a console-marked snippet that occurs on a line before the `$ prompt`. Fortunately, I haven't found anything else that breaks the current parsing, but I will try to improve this case in the follow-up changes.

mudge:
@nathan-contino perhaps the following would cover that case (adding an optional non-capturing group to match `(...) ` before the dollar sign):

```javascript
text.replaceAll(/^(?:\(.+?\)\s+)?\$\s+/gm, "")
```

nathan-contino:
Nice, I'll have to give it a try when I create my follow-up PR! That looks good to me, but especially with RegEx the proof is always in the pudding. I suspect I can make this quite a bit shorter with these suggestions!

nathan-contino:
Gave some of these suggestions a go this morning, but it seems that the suggested `querySelector` and regex don't quite work correctly for me locally -- the console snippet copy button winds up including the prefix. need to do more digging to see where it's breaking. I'll take another pass at this later.

Additionally, the venv snippets break *syntax highlighting*, since rouge doesn't support the venv syntax, so a true fix would also extend console syntax highlighting to properly handle the venv prefix. But since those snippets aren't meant to be copied (aside from the instruction on `deactivate`), I already just mark those snippets as text. Probably not worth the effort to implement since we only talk about virtual envs in a single part of our docs.

jekyll-assets/scripts/copy-to-clipboard.js
nelliemckesson: Yesssss I will add that to the list

PR #3663, opened 2024-05-18
Clean up wording, bulleted lists, and code snippets in raspi-config and noninteractive section

nathan-contino:
* Brings syntax highlighting to these snippets, which were previously unmarked.
* Adds information about NVMe boot to the noninteractive boot order options list.
* Clarifies previously confusing noninteractive wifi connection instruction.

### DIFF
--- a/documentation/asciidoc/computers/configuration/raspi-config.adoc
+++ b/documentation/asciidoc/computers/configuration/raspi-config.adoc
@@ -1,6 +1,8 @@
 [[raspi-config]]
 == `raspi-config`
 
+TIP: Raspberry Pi Desktop users can also access a graphical version of this application at **Preferences** > **Raspberry Pi Configuration**.
+
 `raspi-config` helps you configure your Raspberry Pi. To open the configuration tool, run the following command:
 
 [source,console]
@@ -8,26 +10,22 @@
 $ sudo raspi-config
 ----
 
-NOTE: If you are using the Raspberry Pi desktop, you can use the graphical `Raspberry Pi Configuration` application from the `Preferences` menu to configure your Raspberry Pi.
-
-You should then see a blue screen with options in a grey box:
+You should see a blue screen with options in a grey box:
 
 image::images/raspi-config.png[raspi-config main screen]
 
-NOTE: The menu shown may differ slightly.
+=== Usage
 
-Use the `up` and `down` arrow keys to move the highlighted selection between the options available. Pressing the `right` arrow key will jump out of the Options menu and take you to the `<Select>` and `<Finish>` buttons. Pressing `left` will take you back to the options. Alternatively, you can use the `Tab` key to switch between these.
+Use the **Up** and **Down** arrow keys to move the highlighted selection between the options available.
 
-Generally speaking, `raspi-config` aims to enable the user to make the most common configuration changes. This may result in automated edits to xref:config_txt.adoc#what-is-config-txt[`/boot/firmware/config.txt`] and various standard Linux configuration files. Some options require a reboot to take effect. If you changed any of these, `raspi-config` will ask if you wish to reboot when you select the `<Finish>` button.
+Press the **Right** arrow key or press **Tab** to access the `<Select>` and `<Finish>` buttons. Press **Left** or press **Tab** to return to the options.
 
-NOTE: In long lists of option values (like the list of timezone cities), you can also type a letter to skip to that section of the list. For example, entering `L` will skip you to Lisbon, just two options away from London, to save you scrolling all the way through the alphabet.
+`raspi-config` automates edits to xref:config_txt.adoc#what-is-config-txt[`/boot/firmware/config.txt`] and various Linux configuration files. Some options require a reboot to take effect. If you changed any of these, `raspi-config` asks you to reboot when you exit.
+
+TIP: In long lists of option values (like the list of timezone cities), type a letter to skip to that section of the list. For example, enter `L` to skip to Lisbon.
 
 [[menu-options]]
 === List of options
-
-NOTE: Due to the continuous development of the `raspi-config` tool, the list of options below may not be completely up-to-date. Also please be aware that different models of Raspberry Pi may have different options available.
-
-NOTE: All options are available via a non-interactive command line interface. See the section on the <<raspi-config-cli,`raspi-config` command line interface>> for more information.
 
 ==== System options
 
@@ -405,6 +403,8 @@ $ sudo raspi-config nonint do_browser <chromium-browser/firefox>
 
 [[underscan-nonint]]
 ===== Underscan
+
+NOTE: Not available when running Wayland.
 
 If the initial text shown on the screen disappears off the edge, enable overscan to adjust the border. On some displays, particularly monitors, disabling overscan will make the picture fill the whole screen and remove the black border.
 
--- a/documentation/asciidoc/computers/configuration/raspi-config.adoc
+++ b/documentation/asciidoc/computers/configuration/raspi-config.adoc
@@ -1,13 +1,12 @@
 [[raspi-config]]
-== The `raspi-config` Tool
+== `raspi-config`
 
-`raspi-config` is the Raspberry Pi configuration tool, originally written by https://github.com/asb[Alex Bradbury]. To open the configuration tool, type the following on the command line:
+`raspi-config` helps you configure your Raspberry Pi. To open the configuration tool, run the following command:
 
+[source,console]
 ----
-sudo raspi-config
+$ sudo raspi-config
 ----
-
-The `sudo` is required because you will be changing files that you do not own as the regular user.
 
 NOTE: If you are using the Raspberry Pi desktop, you can use the graphical `Raspberry Pi Configuration` application from the `Preferences` menu to configure your Raspberry Pi.
 
@@ -47,9 +46,7 @@ Specify the audio output destination.
 [[change-user-password]]
 ===== Password
 
-You can change the 'default' user password.
-
-NOTE: Originally, the default user on early versions of Raspberry Pi OS was `pi` with the password `raspberry`. The default user is now set while configuring the OS image or on first boot using a configuration wizard.
+Change your password.
 
 [[hostname]]
 ===== Hostname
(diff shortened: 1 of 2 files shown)


documentation/asciidoc/computers/configuration/raspi-config.adoc
```diff
@@ -1,35 +1,32 @@
 [[raspi-config]]
-== The `raspi-config` Tool
+== `raspi-config`
 
-`raspi-config` is the Raspberry Pi configuration tool, originally written by https://github.com/asb[Alex Bradbury]. To open the configuration tool, type the following on the command line:
+TIP: Raspberry Pi Desktop users can also access a graphical version of this application at **Preferences** > **Raspberry Pi Configuration**.
```
lurch, line 4: (although there are some "advanced options" which are only in the CLI tool and not the GUI tool)

documentation/asciidoc/computers/configuration/raspi-config.adoc
```diff
@@ -170,12 +165,12 @@ Enable or disable a read-only filesystem.
 
 ===== Fan
 
-Set the behaviour of a GPIO connected fan.
+Customise the behaviour of a GPIO connected fan.
```
lurch, line 168: Should "GPIO connected fan" be ""GPIO-connected fan" ? Might be worth clarifying that (AFAIK) this is only for use by https://www.raspberrypi.com/products/raspberry-pi-4-case-fan/

documentation/asciidoc/computers/configuration/raspi-config.adoc
```diff
@@ -203,14 +198,12 @@ This option sets the country code for your wireless network.
 
 image::images/raspi-adv.png[raspi-config advanced options]
 
-NOTE: The options documented here will change depending on the model of Raspberry Pi that you're using, and whether you're using the Wayland or X11 backend. In the current version of Raspberry Pi OS Bookworm, the Raspberry Pi 4, Pi 400 and Pi 5 use Wayland by default; other models of Raspberry Pi use X11 by default.
-
 [[expand-filesystem]]
 ===== Expand filesystem
 
-This option will expand your installation to fill the whole SD card, giving you more space to use for files. You will need to reboot the Raspberry Pi to make this available. 
+Expands your installation to fill the entire storage device, giving you more space to use for files. You must reboot the Raspberry Pi to make this available.
```
lurch, line 204: Raspberry Pi OS _should_ do this automatically, during its first-boot process. I think the option is only in `raspi-config` as a fail-safe fallback? (or if you e.g. use the "SD Card Copier" to clone your OS to a larger storage-medium)

documentation/asciidoc/computers/configuration/raspi-config.adoc
```diff
@@ -247,364 +239,404 @@ Update this tool to the latest version.
...
 
-The `sudo` is required because you will be changing files that are not owned by the default user.
-
-NOTE: There is no consistent meaning for `0` and `1` in arguments. Each function will document what `0` and `1` mean for that particular function.
+NOTE: The meaning of `0` and `1` varies between options. Always check the documentation before passing a value to an option.
 
 [[raspi-config-cli-commands]]
 === List of options
 
-NOTE: The `raspi-config` tool is under continuous development, and the list of options below may not be completely up to date. Please be aware that different models of Raspberry Pi may have different options available.
```
lurch, line 278: Might be worth retaining this warning? Or it least adding the "Please be aware that different models of Raspberry Pi may have different options available." to the previous NOTE?

documentation/asciidoc/computers/configuration/raspi-config.adoc
```diff
@@ -247,364 +239,404 @@ Update this tool to the latest version.
...
 ==== System options
 
 ===== Wireless LAN
 
 Allows setting of the wireless LAN SSID and passphrase.
 
+[source,console]
 ----
-sudo raspi-config nonint do_wifi_ssid_passphrase <ssid> <passphrase> [hidden] [plain]
+$ sudo raspi-config nonint do_wifi_ssid_passphrase <ssid> <passphrase> [hidden] [plain]
```
lurch, line 272: Perhaps we should remove the `raspi-config nonint do_wifi_ssid_passphrase` documentation, and just refer people to the `nmcli` docs instead? https://www.raspberrypi.com/documentation/computers/configuration.html#using-the-command-line :shrug:

documentation/asciidoc/computers/configuration/raspi-config.adoc
```diff
@@ -247,364 +239,404 @@ Update this tool to the latest version.
...
 
-- `0` - bcm2835 Headphones
-- `1` - vc4-hdmi-0
-- `2` - vc4-hdmi-1
+* `0`: bcm2835 headphone jack
+* `1`: vc4-hdmi-0
+* `2`: vc4-hdmi-1
 
-NOTE: You may need to run the interactive version of `raspi-config` to determine the appropriate numbers to use with this option.
+For a full list of possible `<N>` values, see the abbreviations used in the interactive `raspi-config` version of this option.
```
lurch, line 318: abbreviations -> numbers ?

documentation/asciidoc/computers/configuration/raspi-config.adoc
nathan-contino: I was trying to find the best way to bring this up. Probably an extra sentence here will do!

documentation/asciidoc/computers/configuration/raspi-config.adoc
nathan-contino: Ahhh excellent point, looks like the Active Cooler and Pi 5 case fan use a separate fan header. Thanks!

documentation/asciidoc/computers/configuration/raspi-config.adoc
nathan-contino: Worth mentioning for sure.

documentation/asciidoc/computers/configuration/raspi-config.adoc
nathan-contino: Continuous development is so common at this point that it's no longer a great excuse for outdated doc -- if we're missing something, I hope someone opens an issue and calls it out!

As for the model differences, that's worth calling out in both the int and non-int sections. But probably not in an admonition :)

documentation/asciidoc/computers/configuration/raspi-config.adoc
nathan-contino: I tested it today and it does work. Some of the errors are a bit wonky when you use nmcli, though -- I suspect some code paths might assume `wpa_supplicant` exists even when using nmcli.

I'm tempted to just point users to nmcli because of these issues, but as long as the option exists and works, some users might prefer it.

@spl237 , should we continue supporting this option? I'd be happy to open an issue on the `raspi-config` repo to either remove it or fix it. Your call.

documentation/asciidoc/computers/configuration/raspi-config.adoc
nathan-contino: Good call 😆 . Classic copy/paste mistake!

documentation/asciidoc/computers/configuration/raspi-config.adoc
spl237: It's not something I ever use in the GUI; I think it is there for the benefit of people who use raspi-config from the command-line. So @XECDesign's call, not mine!

documentation/asciidoc/computers/configuration/raspi-config.adoc
```diff
@@ -25,231 +23,229 @@ Press the **Right** arrow key or press **Tab** to access the `<Select>` and `<Fi
...
-===== Overlay file system
+==== Overlay file system
 
 Enable or disable a read-only filesystem.
 
-===== Fan
+==== Fan
 
-Customise the behaviour of a GPIO connected fan.
+Customise the behaviour of the GPIO-connected https://www.raspberrypi.com/products/raspberry-pi-4-case-fan/[Raspberry Pi 4 Case Fan]. Does not effect the fans in the https://www.raspberrypi.com/products/raspberry-pi-5-case/[Raspberry Pi 5 Case for Raspberry Pi 5] or https://www.raspberrypi.com/products/active-cooler/[Raspberry Pi 5 Active Cooler], which connect using a special four-pin fan header.
```
lurch, line 165: Another "effect" -> "affect" needed here.

documentation/asciidoc/computers/configuration/raspi-config.adoc
```diff
@@ -25,231 +23,229 @@ Press the **Right** arrow key or press **Tab** to access the `<Select>` and `<Fi
...
+=== Advanced options
 
 image::images/raspi-adv.png[raspi-config advanced options]
 
 [[expand-filesystem]]
-===== Expand filesystem
+==== Expand filesystem
 
-Expands your installation to fill the entire storage device, giving you more space to use for files. You must reboot the Raspberry Pi to make this available.
+Expands your OS partition to fill the whole storage device, giving you more space to use for files. Reboot your Raspberry Pi to complete this action. Normally, Raspberry Pi OS runs this action on first boot. This option can be useful if you clone your OS to a separate storage device with more storage than the original.
```
lurch, line 201: Could change the last "storage" to "capacity" ?

documentation/asciidoc/computers/configuration/raspi-config.adoc
```diff
@@ -622,9 +618,9 @@ $ sudo raspi-config nonint do_overlayfs <0/1>
 * `0`: enable overlay filesystem
 * `1`: disable overlay filesystem
 
-===== Fan
+==== Fan
 
-Set the behaviour of a GPIO connected fan.
+Customise the behaviour of the GPIO-connected https://www.raspberrypi.com/products/raspberry-pi-4-case-fan/[Raspberry Pi 4 Case Fan]. Does not effect the fans in the https://www.raspberrypi.com/products/raspberry-pi-5-case/[Raspberry Pi 5 Case for Raspberry Pi 5] or https://www.raspberrypi.com/products/active-cooler/[Raspberry Pi 5 Active Cooler], which connect using a special four-pin fan header.
```
lurch, line 626: effect -> affect

documentation/asciidoc/computers/configuration/raspi-config.adoc
XECDesign: What wonky error messages are you seeing and what's not working? If there's an issue I can reproduce, I'm happy to fix it.

However, I've always said that the `nonint` functionality is not intended for the user, so it doesn't make much sense documenting it. The idea is there is that it's used by other Raspberry Pi tools so that if we switch from dhcpcd to nm or make any other change, we only need to support it in one place. That way other tools won't need to be rebuilt and they won't diverge in behaviour. However, individual nonint commands can change or be removed without notice.

The short version - in my opinion, yes it absolutely makes sense for the user to use nmcli directly.

PR #3674, opened 2024-05-29
Grant ToC the ability to display monospace text

nathan-contino:
Our section headers already have the ability to display monospace text. This is especially useful when referencing file names, CLI commands, and CLI options, because it makes it clearer when we're referring to one of those things as opposed to a normal English word.

Unfortunately, our table of contents (on the right side of the page) and page titles don't have the ability to display monospace text. Right now, that information is lost and all text is displayed in the same font. This reduces the amount of information we can convey in the ToC -- for instance, the ToC title "non-interactive `raspi-config`" clearly conveys the idea that this is a normal English modifier (non-interactive) followed by the name of a CLI command. Without the monospacing, it's hard to tell where the command begins and ends. (for instance: is `non-interactive` a command in the above title? Easy to identify with monospacing, harder otherwise.)

Example:

Screenshot 2024-05-29 at 12 46 01

I've tested to make sure that this doesn't clobber tab titles in browsers and has no effect on searchability. Code changes are minimal, including a few Jekyll liquid template modifications, a one-line Python script modification, and a couple of changes to `index.json` to take advantage of this new ability.


nathan-contino:
Also fixed that test, because it's failing and doesn't seem worth a separate PR since I should have fixed it in my last toolchain visit. I hear ya on the test running -- we should definitely update our GH Actions to properly fail the checks if that test fails.

nelliemckesson (approved):

nelliemckesson:
Thanks, looks good!

nathan-contino:
Fantastic! Good for me to merge it in?

nelliemckesson:
@nathan-contino yep!

lurch:
> our table of contents (on the right side of the page)

The table of contents is on the left side of the page? :stuck_out_tongue_winking_eye:

PR #3678, opened 2024-05-29
Fix wonky anchor points generated by tocify

nathan-contino:
Modifying minimised JS leads to madness. I noticed that on the production site with our latest changes, tocify generates strange URL slugs in the interactive toc menu only. Likely because we're feeding HTML into a part of the script that expects text. This solves the issue and switches us from a `min.js` version of the script which, let's be honest, wasn't really saving that much space.

### DIFF
--- a/jekyll-assets/scripts/jquery.tocify.min.js
+++ b/jekyll-assets/scripts/jquery.tocify.min.js
@@ -1,4 +1,983 @@
-/*! jquery.tocify - v1.9.0 - 2013-10-01 
-* http://gregfranko.com/jquery.tocify.js/
-* Copyright (c) 2013 Greg Franko; Licensed MIT*/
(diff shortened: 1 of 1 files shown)


jekyll-assets/scripts/jquery.tocify.min.js
```diff
@@ -1,4 +1,983 @@
...
+
+            // Appends a list item HTML element to the last unordered list HTML element found within the HTML element calling the plugin
+            item = $("<li/>", {
+
+                // Sets a common class name to the list item
+                "class": itemClassName,
+
+                "data-unique": hashValue
+
+            }).append(self.html());
```
nathan-contino, line 388: Here's the key spot where I modified the script. Instead of tossing text into a link HTML element, we can just reuse the existing element entirely -- display text, link, etc.

nathan-contino:
Fixes an issue where tocify's local toc tree (generated only for the current page) creates awful links like `https://www.raspberrypi.com/documentation/computers/legacy_config_txt.html#a-classanchor-hrefenable_gic-raspberry-pi-4-onlyacodeenable_giccode-raspberry-pi-4-only`. The original anchor points are still usable, but if you navigate by the toc sidebar, you'll see this URL. Not nice, merging so we can fix this quick.

nelliemckesson:
@nathan-contino understood that this was a relatively urgent change, though it's always best practice to get a review if at all possible (sorry, I'm a stickler for process!). Would probably be good to rename the file to a non-min filename, for the sake of clarity in the code base. I was impressed that you edited the min code but I guess I should have trusted my innate fear of doing that haha (I did split the original changed min file up and do a diff to see the changes, which seemed theoretically ok, but I didn't test it).

nathan-contino:
Sorry about that -- I didn't notice the modified slugs because they only appeared after tocify modified the ToC, which only happened after loading the section completely. I suspect that the lack of a cache-busting query param also contributed to the invisible wonky slugs issue when developing locally. I will open a PR changing the filename to a non-minified version!

Issue #3692, opened 2024-06-05
Linked documentation for python's pip install  externally-managed-environment has missing Anchor tag


DeeJay:
Linked documentation for python's pip install  externally-managed-environment has missing Anchor tag

Text emitted by `pip install fred` includes 

> For more information visit http://rptl.io/venv

That short URL expands to 

[https://www.raspberrypi.com/documentation/computers/os.html#python-on-raspberry-pi](https://www.raspberrypi.com/documentation/computers/os.html#python-on-raspberry-pi)

But the **#python-on-raspberry-pi** anchor tag has been removed from the documentation.

So the link unhelpfully rests at [https://www.raspberrypi.com/documentation/computers/os.html#introduction](https://www.raspberrypi.com/documentation/computers/os.html#introduction)

The correct target URL should be [https://www.raspberrypi.com/documentation/computers/os.html#install-python-libraries-using-pip](https://www.raspberrypi.com/documentation/computers/os.html#install-python-libraries-using-pip)

lurch:
Looks like this was inadvertently broken by a header-rename in #3665

nathan-contino:
Thanks for bringing this up! It's tough to keep implicit anchor points consistent when I'm cleaning up sections of the docs. Fixing this shortly.

PR #3693, opened 2024-06-05
Fix broken external anchor points

nathan-contino:
Uses special AsciiDoc anchors that we can't link to internally in the docs. However, these points WILL work if you manually enter the URL. This means these sections technically have two anchor points -- the implicit one generated by the section title, and this new anchor point. This prevents us from linking to the old version of a page name.

Fixes https://github.com/raspberrypi/documentation/issues/3692
Annoyingly, I already added a similar anchor for this same purpose, but I used the wrong one and couldn't find the rptl link. D'oh. That's my bad.

Also fixes a broken link for Compute Module eMMC flashing.

This _shouldn't_ happen too often if I remember to manually check external links when I refactor. Internal links get checked automatically by AsciiDoc, but external ones are tough to check. Honestly I should just get check for requests that frequently 404 so when a lot of users try to use a link that I break, I can fix it. That'll also help with, say, docs sections linked from Stack Overflow, blog posts, etc that we don't even know about.

### DIFF
--- a/documentation/asciidoc/computers/compute-module/cm-emmc-flashing.adoc
+++ b/documentation/asciidoc/computers/compute-module/cm-emmc-flashing.adoc
@@ -1,5 +1,5 @@
 [[flash-compute-module-emmc]]
-== Flash an image to a Compute Module
+== [[flashing-the-compute-module-emmc]]Flash an image to a Compute Module
 
 The Compute Module has an on-board eMMC device connected to the primary SD card interface. This guide explains how to flash (write) an operating system image to the eMMC storage of a single Compute Module.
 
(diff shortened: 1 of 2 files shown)


lurch:
> This means these sections technically have _two_ anchor points

Does this method allow us to have _multiple_ anchors that all link to the same place? We obviously have no way of knowing which anchors do or don't have external bookmarks set on them, so maybe we should just make it a rule-of-thumb that whenever you change the name of a header, you always add the old-anchor as an alternate link for backwards-compatibility? :thinking: 

> This prevents us from linking to the old version of a page name.

...but more importantly, it means that bookmarks to old anchors **continue** to work (which is how / why this fixes #3692 ).

> Honestly I should just get check for requests that frequently 404 so when a lot of users try to use a link that I break, I can fix it.

Except you don't get a 404 if it's only the anchor on a page that is missing :wink:  (which is what happened in #3692 )

nathan-contino:
I believe you can chain these anchors, but I have to test it. Might not work in our specific toolchain.

Good point that missing anchors don't produce 404s. However, we could still index the, say, top 500 most popular URL requests and check for the header.

It's far too tedious to ALWAYS add these anchors, unfortunately. And during refactors I frequently combine and split content in ways that would break the semantic meaning of those links anyway. When there's a high-profile section, I already try to keep an explicit anchor in place; I'll just keep doing that. If I miss one, folks are free to open an issue and I can fix it afterwards, though hopefully I won't miss any.

One nice side-effect of our long-page docs system: ctrl+F is a great way to find content when your anchor points break!

PR #3698, opened 2024-06-05
Improve Compute Module section based on feedback

nathan-contino:
* Minor copy-edit of bootloader section.
* Added note about Lite devices to Flash section.
* Updates to the `BOOT_ORDER` documentation because I added a link and it could use some improvement.

### DIFF
--- a/documentation/asciidoc/computers/compute-module/cm-bootloader.adoc
+++ b/documentation/asciidoc/computers/compute-module/cm-bootloader.adoc
@@ -1,41 +1,55 @@
-[[cm4bootloader]]
-=== Compute Module 4 bootloader
+== Compute Module EEPROM bootloader
 
-The default bootloader configuration on CM4 is designed to support bringup and development on a https://www.raspberrypi.com/products/compute-module-4-io-board/[Compute Module 4 IO board], and the software version flashed at manufacture may be older than the latest release. For final products please consider:
+Since Compute Module 4, Compute Modules use an EEPROM bootloader. This bootloader lives in a small segment of on-board storage instead of the boot partition. As a result, it requires different procedures to update. Before using CM4 in production, always follow these best practices:
 
-* Selecting and verifying a specific bootloader release. The version in the `usbboot` repo is always a recent stable release.
-* Configuring the boot device (e.g. network boot). See `BOOT_ORDER` section in the xref:raspberry-pi.adoc#raspberry-pi-bootloader-configuration[bootloader configuration] guide.
-* Enabling hardware write-protection on the bootloader EEPROM to ensure that the bootloader can't be modified on inaccessible products (such as remote or embedded devices).
+* Select a specific bootloader release. Verify that every Compute Module you use has that release. The version in the `usbboot` repo is always a recent stable release.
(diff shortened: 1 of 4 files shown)


documentation/asciidoc/computers/compute-module/cm-bootloader.adoc
```diff
@@ -1,41 +1,55 @@
...
+=== Flash Compute Module bootloader EEPROM
 
-==== Flash storage devices other than SD cards
+To flash the bootloader EEPROM:
 
-The Linux-based https://github.com/raspberrypi/usbboot/blob/master/mass-storage-gadget/README.md[mass-storage gadget] supports flashing of NVMe, eMMC and USB block devices. This is normally faster than using the `rpiboot` firmware driver and also provides a UART console to the device for easier debug.
+. Set up the hardware as you would when xref:../computers/compute-module.adoc#flash-compute-module-emmc[flashing the eMMC], but ensure `EEPROM_nWP` is _not_ pulled low.
+. Once complete, `EEPROM_nWP` may be pulled low again.
+. Run the following command to write `recovery/pieeprom.bin` to the bootloader EEPROM:
```
lurch, line 15: Feels like these two bullet-points are the wrong way around? :thinking:

documentation/asciidoc/computers/compute-module/cm-bootloader.adoc
```diff
@@ -1,41 +1,55 @@
...
+[source,console]
+----
+$ ./rpiboot -d recovery
+----
 
-See also: https://github.com/raspberrypi/usbboot/blob/master/Readme.md#compute-module-4-extensions[CM4 rpiboot extensions].
+=== Flash storage devices other than SD cards
 
-==== Modify the bootloader configuration
+The Linux-based https://github.com/raspberrypi/usbboot/blob/master/mass-storage-gadget/README.md[`mass-storage-gadget`] supports flashing of NVMe, eMMC and USB block devices. `mass-storage-gadget` works faster than `rpiboot` and also provides a UART console to the device for debugging.
```
lurch, line 24: I believe that `rpiboot` is _also_ used to load `mass-storage-gadget` onto the CM4, so perhaps this should be phrased something like:
"`mass-storage-gadget` writes devices faster than the firmware-based `rpiboot` mechanism, and also provides a UART console to the device for debugging."
?

documentation/asciidoc/computers/compute-module/cm-bootloader.adoc
```diff
@@ -1,41 +1,55 @@
...
 
-To modify the CM4 bootloader configuration:
+`usbboot` also includes a number of https://github.com/raspberrypi/usbboot/blob/master/Readme.md#compute-module-4-extensions[`rpiboot` extensions] that enable you to interact with the EEPROM bootloader on a Compute Module.
```
lurch, line 26: I _think_ `usbboot` and `rpiboot` are basically two different names for the same thing?

nathan-contino: My understanding is that `rpiboot` is a utility within the large bucket of utilities that is `usbboot`. But it's also used for damn near every part of `usbboot`, so... pretty much the same, yeah. Removed!

Issue #3353, opened 2024-01-09
Rotating 7-inch Touch Display is a mess -- needs proper research then update of config-tools and then new instructions in doc


jdg71nl:
(this is my "first create-issue", so please help if forget something)

I am using the **7-inch Touch Display accessory** for many years now, for both business and personal projects, and I am always fighting with the "change-orientation" settings.

The two most popular display housing options (Multicomp and Pibow) both "force" you to install the display upside-down, relative to the housing if sitting on a flat surface, or desk.
The best solution would be for the housing to allow hardware installation/mounting in 2 positions (0 and 180 degrees).
But that is not the case, so we are stuck with **software-rotation**.

I have seen and used the rotation settings in the official website:
https://www.raspberrypi.com/documentation/accessories/display.html
https://github.com/raspberrypi/documentation/blob/develop/documentation/asciidoc/accessories/display/display_intro.adoc

But I always end-up with either the text/console being 180-rotated, and the GUI not.
Or reverse.
And then there is the issue of the touch-interface rotation being independent from the display-rotation.
And finally, the VNC-server does not "follow" the display-rotation.

I realize that fixing this issue, is foremost the addition/update of the driver config (in cmdline.txt or config.txt), more that just an update of this doc.
But how do I file an github-issue for the correct "7-inch Touch Display" driver (-config) ?

This is a screen-recording of VNC with the pointer being reversed:

https://github.com/raspberrypi/documentation/assets/12169650/3e2176b6-7f6d-4b8d-b2cf-1f59b1a9cb3c

6by9:
Setting `video=DSI-1:800x480@60,rotate=180` in cmdline.txt will invert the console by setting rotation on the primary plane for that output. However currently Wayfire doesn't reset that option, so you end up with the background plane rotated when the cursor plane isn't.
There is a one line fix for that to a package called wlroots that will resolve that. I've created the [patch in the mainline project](https://gitlab.freedesktop.org/wlroots/wlroots/-/merge_requests/4508), and that should ripple through to Raspberry Pi OS fairly quickly. With that you should be able to use both the cmdline.txt setting and the Screen Configuration app in Wayfire and get consistent results.

Wayfire will want the touch matching the default orientation of the display itself. When assigned to a display, transforms on the display will also be reflected in the touch input as well. I believe you can insert a custom transform in libinput/Wayfire, but it just becomes more complex.
[...]

PR #3705, opened 2024-06-06
Add labwc workaround

nathan-contino:
As discussed in https://github.com/raspberrypi/documentation/issues/3353

### DIFF
--- a/documentation/asciidoc/accessories/display/display_intro.adoc
+++ b/documentation/asciidoc/accessories/display/display_intro.adoc
@@ -150,7 +150,7 @@ Then, use the Raspberry Pi Configuration CLI to switch to the `labwc` window com
 $ sudo raspi-config
 ----
 
-Go to **Advanced Options**. Select **Wayland**, then select **labwc**. Exit `raspi-config` and reboot with `sudo reboot`. Switching to `labwc` may reset some configuration, including keyboard layout.
+Go to **Advanced Options**. Select **Wayland**, then select **labwc**. Exit `raspi-config` and reboot with `sudo reboot`. Switching to `labwc` may reset some configuration, including keyboard layout and rotation settings.
 ====
 
 ==== Touch Display device tree option reference
--- a/documentation/asciidoc/accessories/display/display_intro.adoc
+++ b/documentation/asciidoc/accessories/display/display_intro.adoc
@@ -132,6 +132,27 @@ Then, disable automatic display detection by removing the following line from `c
 display_auto_detect=1
 ----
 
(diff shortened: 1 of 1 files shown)


documentation/asciidoc/accessories/display/display_intro.adoc
```diff
@@ -132,6 +132,27 @@ Then, disable automatic display detection by removing the following line from `c
...
+----
+
+Then, use the Raspberry Pi Configuration CLI to switch to the `labwc` window compositor. Run the following command to open `raspi-config`:
+
+[source,console]
+----
+$ sudo raspi-config
+----
+
+Go to **Advanced Options**. Select **Wayland**, then select **labwc**. Exit `raspi-config` and reboot with `sudo reboot`. Switching to `labwc` may reset some configuration, including keyboard layout.
```
steveiliop56, line 153: You can also add that it will likely reset rotation settings.

nathan-contino:
```suggestion
Go to **Advanced Options**. Select **Wayland**, then select **labwc**. Exit `raspi-config` and reboot with `sudo reboot`. Switching to `labwc` may reset some configuration, including keyboard layout and rotation settings.
```

documentation/asciidoc/accessories/display/display_intro.adoc
```diff
@@ -132,6 +132,27 @@ Then, disable automatic display detection by removing the following line from `c
 display_auto_detect=1
 ----
 
+[NOTE]
+====
+In Raspberry Pi OS _Bookworm_, a bug in https://github.com/WayfireWM/wayfire[Wayfire] currently prevents cursor movement rotation. There is no workaround for this bug in Wayfire. Instead, you can use https://github.com/labwc/labwc[`labwc`] (currently in beta).
```
6by9, line 137: It incorrectly rotates the cursor when using WayVNC. Locally it is fine.

AIUI we aren't be looking for a fix in Wayfire.
Labwc is the route we are taking forward as it should be work reasonably on the earlier Pis, and hope to adopt it very soon.

steveiliop56: Yeah I agree but what I am saying is that the problem of the inverted mouse in wayvnc is a wayfire problem of not assigning something correctly. I am just recommending labwc to people that just want to make wayvnc with rotation work.

nathan-contino: Fixing this in a separate PR. Thanks!

PR #3717, opened 2024-06-10
Additional kernel flow improvements

nathan-contino:


documentation/asciidoc/computers/linux_kernel/building.adoc
```diff
@@ -179,28 +179,47 @@ TIP: On multi-core Raspberry Pi models, the `make -j<n>` option distributes work
...
+----
+$ sudo make -j6 modules_install
+----
+
+* For 32-bit kernels:
++
+[source,console]
+----
+$ sudo make -j6 modules_install
+----
```
lurch, line 196: Should this be collapsed to "For both 64-bit and 32-bit kernels, run:" ? Seems a bit weird to label them separately, if the command in both cases is identical?

nathan-contino: Good point. I strayed a little too close to dogma here when it came to parallel structure!

lurch: If only there was some kind of Tab system... :laughing:

PR #3724, opened 2024-06-14
Update patching.adoc use of `head` to show correct output

cHemingway:
Linux (unsure as of when) puts `# SPDX-License-Identifier: GPL-2.0`  as the first line of Makefile, so having head print only 3 lines would not show the sublevel. Changing to 4 lines fixes this.

Left the rest of the output unchanged to avoid breaking other parts of the example.


nathan-contino (approved):
LGTM! Thanks so much for catching this. The sublevel is pretty important here, so this should help out a lot of users.

PR #3749, opened 2024-06-26
video: Correct limitations of odd pixel timings for Pi 5

popcornmix:
Odd timings have been supported on Pi 5 since
https://github.com/raspberrypi/linux/pull/5679


nathan-contino (approved):
LGTM, thank you!

PR #3769, opened 2024-07-13
Update building.adoc

ehorohorin:
path to broadcom/*.dtb fixed for 64-bit (6.5+), could you check and validate this?
native rpi5 build

### DIFF
--- a/documentation/asciidoc/computers/linux_kernel/building.adoc
+++ b/documentation/asciidoc/computers/linux_kernel/building.adoc
@@ -223,7 +223,7 @@ $ sudo cp arch/arm/boot/dts/*.dtb /boot/firmware/
 +
 [source,console]
 ----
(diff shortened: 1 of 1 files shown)


nathan-contino:
This is in the 32-bi section, so it doesn't seem right to me that it should use `amd64`. That being said, I don't know as much about this as I'd like, so I may be missing something (maybe the Pi 5 doesn't even have an `amd` dir?) I'll close this in a day or two if I don't see any concrete evidence that it should change.

nathan-contino:
@pelwell small PR, do you know if this is correct?

pelwell (changes requested):
@ehorohorin I think you've missed that the section you are modifying is for installing the 32-bit kernel. The 64-bit install is earlier in the document and already includes the correct path.

ehorohorin:
Sorry, my bad.

ehorohorin:
No need to apply it.

PR #3770, opened 2024-07-15
Update rpicam_apps_building.adoc

gftrobots:
I took me a while to get rpicam running. It is helpful to provide the hint to remove the pre-installed libraries and load the config afterwards.

### DIFF
--- a/documentation/asciidoc/computers/camera/rpicam_apps_building.adoc
+++ b/documentation/asciidoc/computers/camera/rpicam_apps_building.adoc
@@ -178,6 +178,23 @@ Finally, run the following command to install your freshly-built `rpicam-apps` b
 $ sudo meson install -C build
 ----
 
+[NOTE]
+======
+rpicam-apps is pre-installed on current 64-bit OS distributions. Before the installation steps above you should remove the pre-installed version.
+
+[source,console]
+----
+$ sudo apt-get remove --purge rpicam-apps
+----
+
+After meson install the ldconfig cache is updated automatically. Running ldconfig without any options will update the cache.
+
+[source,console]
+----
+$ sudo ldconfig
+----
+======
+
 Open a new terminal window after installation to ensure that you use the new binary.
 
 Finally, follow the `dtoverlay` and display driver instructions in the  xref:camera_software.adoc#configuration[Configuration section].


documentation/asciidoc/computers/camera/rpicam_apps_building.adoc
```diff
@@ -178,6 +178,23 @@ Finally, run the following command to install your freshly-built `rpicam-apps` b
 $ sudo meson install -C build
 ----
 
+[NOTE]
+======
+rpicam-apps is pre-installed on current 64-bit OS distributions. Before the installation steps above you should remove the pre-installed version.
```
lurch, line 183: I believe it's pre-installed on all current distributions, not just 64-bit?

documentation/asciidoc/computers/camera/rpicam_apps_building.adoc
```diff
@@ -178,6 +178,23 @@ Finally, run the following command to install your freshly-built `rpicam-apps` b
...
+
+[source,console]
+----
+$ sudo apt-get remove --purge rpicam-apps
+----
+
+After meson install the ldconfig cache is updated automatically. Running ldconfig without any options will update the cache.
```
lurch, line 190: If it's updated automatically, why does it also need to be run manually?

gftrobots: sorry, 
After meson install the ldconfig is updated, but the cache not.

documentation/asciidoc/computers/camera/rpicam_apps_building.adoc
gftrobots: It could be that it is installed on all current installations, I cannot very it. It it is the case, than:

rpicam-apps is pre-installed on current OS distributions. Before the installation steps above you should remove the pre-installed version.

gftrobots (commented):
I replied to your comment in the code and suggest minor adjustments.

nathan-contino:
Addressing this in https://github.com/raspberrypi/documentation/pull/3774, closing this PR. Sorry I didn't use your suggestion exactly, but thank you for making it in the first place. If my changes didn't split the content into two separate parts of the documentation flow, I would have iterated on this PR instead.

PR #3772, opened 2024-07-16
Document Connect support in the boot wizard and raspi-config

nathan-contino:
* Added documentation for some enhancements to raspi-config and the boot wizard
* Updated screenshots of raspi-config across the board; they were very out of date. Using screenshots from a Pi 5, so some categories are hidden (IMO that's OK, since the sections still exist in our text)


nathan-contino:
@mudge can you double-check that this all looks good? Thanks!

mudge (approved):
The new Connect documentation looks good to me.

PR #3779, opened 2024-07-16
Update install.adoc

pawelzwronek:
Default username is 'pi' as a login via ssh. Username and password is optional.

### DIFF
--- a/documentation/asciidoc/computers/getting-started/install.adoc
+++ b/documentation/asciidoc/computers/getting-started/install.adoc
@@ -88,7 +88,7 @@ If you plan to use your Raspberry Pi remotely over your network, check the box n
 
 * Choose the **password authentication** option to SSH into your Raspberry Pi over the network using the username and password you provided in the general tab of OS customisation.
 
-* Choose **Allow public-key authentication only** to preconfigure your Raspberry Pi for passwordless public-key SSH authentication using a private key from the computer you're currently using. If already have an RSA key in your SSH configuration, Imager uses that public key. If you don't, you can click **Run SSH-keygen** to generate a public/private key pair. Imager will use the newly-generated public key.
+* Choose **Allow public-key authentication only** to preconfigure your Raspberry Pi for passwordless public-key SSH authentication using a private key from the computer you're currently using. If already have an RSA key in your SSH configuration, Imager uses that public key. If you don't, you can click **Run SSH-keygen** to generate a public/private key pair. Imager will use the newly-generated public key. Default username to log in via ssh is **pi** `ssh -l pi <here raspberrypi ip address>`. When this option is enabled setting *username* and *password* in *General* tab is *optional*.
 
 image::images/imager/os-customisation-services.png[alt="Services settings in the OS customisation menu."]
 


nathan-contino:
There is no longer a default username and password, so this is misleading. Instead, users must either preconfigure a user account or set one up on first boot.

pawelzwronek:
Today I installed Raspberry Pi OS Lite (64-bit) on my new RPi 5. After not setting up username and password, and enabling only ssh with public key I managed to login over ssh to my headless RPi with 'pi' as a username with is super useful and this is what I was expecting but didn't know at first what was default username, I tried admin and root but no luck. I don't need other username and password. I think public key is safer than password.

nathan-contino:
Ahh, I see what you're suggesting, now. I can reproduce this issue when I don't configure a username and password, though the `pi` user disappears after I go through the boot wizard. Haven't tried this on a Lite setup, but I'm not sure what would happen there, since the boot wizard requires a desktop.

On one hand, it might be worth documenting this for completeness' sake.

On the other hand: what possible workflow justifies not pre-configuring a username and password for a Lite installation? Maybe if I understand the use case I can better explain the tradeoffs in the documentation.

Thanks!

pawelzwronek:
My use case is to log in only with the public/private key generated when I clicked RUN SSH-KEYGEN, as I did before with my PRi 1 where I started sshd and configured private key. Then I don't need to enter password to log in. I need to only setup a private key and a login in Putty or just run `ssh -l pi <ip>` from command line to get to the shell. Generated keys sits here `C:\Users\<username>\.ssh` I've never connected a monitor or a keyboard to my RPi. To my RPi 1 I think I connected over UART/USB for the first time but not sure. Also, when my RPi is configured like this, connecting with VSCode to my rpi is super easy. I need only to select **Remote-SSH: Connect to Host** enter an ip address and the magic starts :) VSCode server is getting installed on my rpi and I can open any folder, even '/' with my VSCode on my desktop. Also an option to uncheck a checkbox **Set username and password** would suggest that there must be some default username.

pawelzwronek:
Also I noticed when RPi booted for the first time I think it restarted couple times, because I saw red light for 2,3 or 4 times, don't remember exactly.

lurch:
> Also an option to uncheck a checkbox **Set username and password** would suggest that there must be some default username.

The "normal scenario" with this is that if you don't set a username and password in RPi Imager, then when you power on the Pi for the first time the first-boot wizard pops up on the locally-attached HDMI monitor, where you can then configure the username and password using a locally-attached USB keyboard and mouse.
I asked a colleague about this, and they said "I'd always recommend setting a password. What happens if they don't is not defined and may change."

> Also I noticed when RPi booted for the first time I think it restarted couple times

Yes, that's part of the "expanding the root filesystem to use up all of the available space on the SD card" process.

pawelzwronek:
I think using private key is superior to passwords, as users likes very much easy passwords like '123456' or so. On the other hand I need only to start Putty/ssh to get access to the shell, but I treat my PC as super secure, it's my in house. Also I think a headless configuration for a RPi is quite common so I'm happy I managed to get it working with ease.

PR #3781, opened 2024-07-17
Update software.adoc

JellySwims:

PR #3782, opened 2024-07-18
Update Sense HAT flow to avoid confusing users

nathan-contino:
https://github.com/raspberrypi/documentation/pull/3781 revealed some problems with our Sense HAT documentation -- namely that it's hard for users to follow. This fixes some of that.

Also included some random touch-ups to other sections of the docs because... why not?


nathan-contino:
Thank you @JellySwims for bringing this to my attention.

PR #3791, opened 2024-07-29
fix "ffmpeg -codec" to "ffmpeg -codecs"

arsenicks:
the "s" was missing in the flag to list available codec

using ffmpeg -codec returns:
Missing argument for option 'codec'.
Error splitting the argument list: Invalid argument


nathan-contino (approved):
Right you are. Thanks!

PR #3793, opened 2024-07-31
Update internal heading link

saeedjohar:
Corrected a typo in the documentation which seems to be mismatch


lurch:
Yes, it probably makes more sense to link to the higher-level https://www.raspberrypi.com/documentation/computers/linux_kernel.html#cross-compile-the-kernel section, than it does the lower-level https://www.raspberrypi.com/documentation/computers/linux_kernel.html#cross-compiled-build section.

nathan-contino (approved):
LGTM, thank you.

PR #3794, opened 2024-08-05
Improved adoc compilation process

nelliemckesson:
This replaces the old `transform_doxygen_html.py` file with the doxygentoasciidoc library, and then adds some improved functionality to the adoc compilation processes.


nathan-contino (approved):
LGTM!

PR #3797, opened 2024-08-06
Update rpicam_options_common.adoc

Jacob-Bai:
Correct 'autofocus-range' to 'autofocus-window'


nathan-contino (approved):
LGTM, thank you!

PR #3798, opened 2024-08-07
Changed misleading video option for single HDMI models

Gemba:
With fkms/kms it is `HDMI-A-1` on a Rpi3 b+ and not `HDMI` as stated before.

### DIFF
--- a/documentation/asciidoc/computers/configuration/kernel-command-line-config.adoc
+++ b/documentation/asciidoc/computers/configuration/kernel-command-line-config.adoc
@@ -62,7 +62,7 @@ Possible options for the display type - the first part of the `video=` entry - i
 | Video Option | Display
 
 | HDMI-A-1
-| HDMI 1 (HDMI 0 on silkscreen of Raspberry Pi 4B, HDMI on single HDMI boards)
+| HDMI 1 (HDMI 0 on silkscreen of Raspberry Pi 4B, use HDMI-A-1 on single HDMI boards)
 
 | HDMI-A-2
 | HDMI 2 (HDMI 1 on silkscreen of Raspberry Pi 4B)


documentation/asciidoc/computers/configuration/kernel-command-line-config.adoc
```diff
@@ -62,7 +62,7 @@ Possible options for the display type - the first part of the `video=` entry - i
 | Video Option | Display
 
 | HDMI-A-1
-| HDMI 1 (HDMI 0 on silkscreen of Raspberry Pi 4B, HDMI on single HDMI boards)
```
lurch, line 65: I believe the wording here is referring to the silkscreen-label?

lurch: Although I guess this _does_ probably need expanding from "Raspberry Pi 4B" to "Raspberry Pi 4B or Raspberry Pi 5" ? :thinking:  (and might also be worth mentioning which HDMI port is which on the Pi 400?)

Gemba: Well, at least it is also confusing for others. Maybe the table should list primary which outputname should be used per which model. The silkscreen label is secondary IMHO, as the majority of users will know which output port(s) they use.

lurch: > Maybe the table should list primary which outputname should be used per which model.

It already does? `HDMI-A-1`, `DSI-1` and `Composite-1` for all models, and additionally `HDMI-A-2` for those models with two HDMI ports.

andrum993: As lurch has pointed out, the page is correct as currently written, although it needs updated for Pi 5.

Having said that, this PR does flag up a potential issue with the table, in that the wording is perhaps not the easiest to interpret, in that it relies on the reader understanding what the term `silkscreen` means. Looking at the page history, this has been through copy edit by Liz, so I'm rather reluctant to chime in with any suggestion for improvement here, but perhaps something like this would be of benefit?:

| Video Option | Display Output |
---------------------------
| HDMI-A-1 | HDMI 1 (labelled HDMI 0 on top of Raspberry Pi circuit board; labelled HDMI on computers with single HDMI output) |
| HDMI-A-2 | HDMI 2 (computers with two HDMI outputs only - labelled HDMI 1 on top of circuit board) |
| DSI-1 | DSI or DPI display |
| Composite-1 | Composite |

Also, the paragraphs above the table are not 100% correct:

`The following example defines a display named HDMI-A- at a resolution...`

Should actually read something like:

`The following example configures the display named HDMI-A-1 at a resolution...`

since the display is not _defined_ by the line, rather it is configured by it, and there is a typo where `-1` is missing.

And:

`Possible options for the display type - the first part of the video= entry - include:`

Should read something like:

`Possible options for the display output - the first part of the video= entry - include:`

since the part of the video= line being specified is a specific display output port, not a display type.

nathan-contino:
Seems like a misunderstanding. Thanks anyway!

PR #3817, opened 2024-08-14
Fix rpicam demo

zhu-hongwei:
image


nathan-contino (approved):
Thank you! Good catch.

PR #3828, opened 2024-08-21
Update doxygentoasciidoc to latest version

mudge:
The latest version replaces the "-f" argument with a single positional argument so update our Makefile accordingly.


lurch:
Tested this locally, still builds okay :+1:

PR #3876, opened 2024-10-09
Update getting-started.adoc

selvi-p:
Fixed few typos, abbreviations, article usage


nathan-contino (commented):
Thanks for noticing this subject-verb agreement issue. I've fixed it elsewhere.

Unfortunately we don't need the other changes.

PR #3878, opened 2024-10-09
Use arm-none-eabi-gdb on macOS to support x86 and aarch64 macOS devices

jackwilsdon:
`arm-none-eabi-gdb` works to debug the RP2040 on both x86 and aarch64 macOS devices.


jackwilsdon:
Yeah it looks like the formula was only added August last year: https://github.com/Homebrew/homebrew-core/pull/138370

PR #3881, opened 2024-10-09
Update getting-started

selvi-ponnusamy:
Changed headings to sentence case


selvi-ponnusamy:
Separate PR has been raised to change the casing in geaders.

PR #3901, opened 2024-11-04
Roll out TD2

nathan-contino:


documentation/asciidoc/accessories/touch-display-2/about.adoc
```diff
@@ -0,0 +1,129 @@
...
+To connect a Touch Display 2 to a Raspberry Pi, use a Flat Flexible Cable (FFC) and a GPIO connector. The FFC you'll use depends upon your Raspberry Pi model:
+
+* for Raspberry Pi 5, use the included 22-way to 15-way FFC
+* for any other Raspberry Pi model, use the included 15-way to 15-way FFC
+
+Once you have determined the correct FFC for your Raspberry Pi model, complete the following steps to connect your Touch Display 2 to your Raspberry Pi:
+
+. Disconnect your Raspberry Pi from power.
+. Lift the retaining clips on either side of the FFC connector on the Touch Display 2.
+. Insert one 15-way end of your FFC into the Touch Display 2 FFC connector, with the copper contacts facing upwards, away from the Touch Display 2.
```
lurch, line 45: s/copper/metal/ ? (just to be on the safe side :wink: )

documentation/asciidoc/accessories/touch-display-2/about.adoc
```diff
@@ -0,0 +1,129 @@
...
+. Insert the other end of your FFC into the Raspberry Pi DSI connector, with the copper contacts facing towards the Ethernet and USB-A ports.
+. While holding the FFC firmly in place, simultaneously push both retaining clips down on the DSI connector of the Raspberry Pi.
+. Plug the GPIO connector cable into the port marked `J1` on the Touch Display 2.
+. Connect the other (three-pin) end of the GPIO connector cable to pins 2, 4, and 6 of the xref:../computers/raspberry-pi.adoc#gpio[Raspberry Pi's GPIO]. Connect the red cable (5V power) to pin 2, and the black cable (ground) to pin 6. Viewed from above, with the Ethernet and USB-A ports facing down, these pins are located at the top right of the board, with pin 2 in the top right-most position.
++
+.The GPIO connection to the Touch Display 2
+image::images/touch-display-2-gpio-connection.png[The GPIO connection to the Touch Display 2, width="40%"]
++
+TIP: If pin 6 isn't available, you can use any other open `GND` pin to connect the black wire. If pin 2 isn't available, you can use any other 5V pin to connect the red wire, such as pin 4.
```
lurch, line 58: I'm not sure if this TIP makes sense for the Touch Display 2, given that the red and black wires are in a fixed 3-pin block, and (unlike Touch Display 1) can't be relocated to arbitrary other pins.

documentation/asciidoc/accessories/touch-display-2/about.adoc
```diff
@@ -0,0 +1,129 @@
...
+
+TIP: In Raspberry Pi OS releases prior to _Bookworm_, use `matchbox-keyboard` instead. If you use the wayfire desktop compositor, use `wvkbd` instead.
+
+=== Change screen orientation
+
+If you want to physically rotate the display, or mount it in a specific position, select **Screen Configuration** from the **Preferences** menu. Right-click on the touch display rectangle (likely DSI-1) in the layout editor, select **Orientation**, then pick the best option to fit your needs.
+
+==== Rotate screen without a desktop
+
+To set the screen orientation on a device that lacks a desktop environment, edit the `/boot/firmware/cmdline.txt` configuration file to pass an orientation to the system. Add the following line to `cmdline.txt`:
```
lurch, line 80: I'm not sure if "Add the following line" is the right wording here - `/boot/firmware/cmdline.txt` has to be [just a single line](https://www.raspberrypi.com/documentation/computers/configuration.html#kernel-command-line-cmdline-txt), so perhaps this should say "Add the following entry to the end of `cmdline.txt`" ?

documentation/asciidoc/accessories/touch-display-2/about.adoc
nathan-contino: Very true! But theoretically someone could yank them out, or they could connect the 3-pin block to another set of pins that fulfills these criteria. Perhaps I should change this to instead point at another valid set of pins where you could possibly attach the 3-pin block?

nathan-contino: Ah, i see that there's no other physical place where the 3-pin block satisfies this requirement. I'll leave this alone for now but I'm open to idea about how to improve it -- if you still feel like it's worth removing, happy to do that, but I feel like we should probably make it clear that 5V/GND are the only requirements.

PR #3903, opened 2024-11-04
configtxt/boot: Update arm_64bit flag

tdewey-rpi:
A casual reader might have previously assumed Raspberry Pi 5 requires this flag to boot in 64-bit mode. That is incorrect - Raspberry Pi 5 follows Raspberry Pi 4 in assuming 64-bit mode by default.

Edit the defaults line to reflect this, and further edit it to include more complete platform names.

### DIFF
--- a/documentation/asciidoc/computers/config_txt/boot.adoc
+++ b/documentation/asciidoc/computers/config_txt/boot.adoc
@@ -50,6 +50,7 @@ The following Raspberry Pi models support this flag:
 * Compute Module 3
 * Compute Module 3+
 * Compute Module 4
+* Compute Module 4S
 
 Later models, such as Raspberry Pi 5, _only_ support the 64-bit kernel. Such models do not support this flag.
 
--- a/documentation/asciidoc/computers/config_txt/boot.adoc
+++ b/documentation/asciidoc/computers/config_txt/boot.adoc
@@ -25,7 +25,6 @@ The Raspberry Pi 5 firmware defaults to loading `kernel_2712.img` because this i
 
 === `arm_64bit`
 
-NOTE: This flag has been removed from Raspberry Pi 5, as Raspberry Pi 5 only supports a 64-bit kernel.
 
 If set to 1, the kernel will be started in 64-bit mode. Setting to 0 selects 32-bit mode.
 
--- a/documentation/asciidoc/computers/config_txt/boot.adoc
+++ b/documentation/asciidoc/computers/config_txt/boot.adoc
@@ -29,7 +29,7 @@ If set to 1, the kernel will be started in 64-bit mode. Setting to 0 selects 32-
 
 In 64-bit mode, the firmware will choose an appropriate kernel (e.g. `kernel8.img`), unless there is an explicit `kernel` option defined, in which case that is used instead.
 
-Defaults to 1 on Pi 4s (Pi 4B, Pi 400, CM4 and CM4S), and 0 on all other platforms. However, if the name given in an explicit `kernel` option matches one of the known kernels then `arm_64bit` will be set accordingly.
+Defaults to 1 on Raspberry Pi 4, 5, 400 and Compute Module 4, 4S platforms. Defaults to 0 on all other platforms. However, if the name given in an explicit `kernel` option matches one of the known kernels then `arm_64bit` will be set accordingly.
 
 NOTE: 64-bit kernels may be uncompressed image files or a gzip archive of an image (which can still be called kernel8.img; the bootloader will recognize the archive from the signature bytes at the beginning). The 64-bit kernel will only work on the Raspberry Pi 3, 3+, 4, 400, Zero 2 W and 2B rev 1.2, and Raspberry Pi Compute Modules 3, 3+ and 4. Raspberry Pi 5 only supports a 64-bit kernel, so this parameter has been removed for that device.
 
--- a/documentation/asciidoc/computers/config_txt/boot.adoc
+++ b/documentation/asciidoc/computers/config_txt/boot.adoc
@@ -33,7 +33,26 @@ In 64-bit mode, the firmware will choose an appropriate kernel (e.g. `kernel8.im
 
 Defaults to 1 on Raspberry Pi 4, 400 and Compute Module 4, 4S platforms. Defaults to 0 on all other platforms. However, if the name given in an explicit `kernel` option matches one of the known kernels then `arm_64bit` will be set accordingly.
 
-NOTE: 64-bit kernels may be uncompressed image files or a gzip archive of an image (which can still be called kernel8.img; the bootloader will recognize the archive from the signature bytes at the beginning). The 64-bit kernel will only work on the Raspberry Pi 3, 3+, 4, 400, Zero 2 W and 2B rev 1.2, and Raspberry Pi Compute Modules 3, 3+ and 4.
+64-bit kernels come in the following forms:
+
+* uncompressed image files
+* gzip archives of an image
+
+Both forms may use the `img` file extension; the bootloader recognizes archives using signature bytes at the start of the file.
+
+The following Raspberry Pi models support this flag:
+
+* 2B rev 1.2
+* 3
+* 3+
+* 4
+* 400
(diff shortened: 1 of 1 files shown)


documentation/asciidoc/computers/config_txt/boot.adoc
```diff
@@ -29,7 +29,7 @@ If set to 1, the kernel will be started in 64-bit mode. Setting to 0 selects 32-
 
 In 64-bit mode, the firmware will choose an appropriate kernel (e.g. `kernel8.img`), unless there is an explicit `kernel` option defined, in which case that is used instead.
 
-Defaults to 1 on Pi 4s (Pi 4B, Pi 400, CM4 and CM4S), and 0 on all other platforms. However, if the name given in an explicit `kernel` option matches one of the known kernels then `arm_64bit` will be set accordingly.
+Defaults to 1 on Raspberry Pi 4, 5, 400 and Compute Module 4, 4S platforms. Defaults to 0 on all other platforms. However, if the name given in an explicit `kernel` option matches one of the known kernels then `arm_64bit` will be set accordingly.
```
JamesH65, line 32: I'd add CM5 as well - everyone knows its coming.

tdewey-rpi: I try to avoid mentioning any potential products ahead of a formal announcement, especially if I've structured the text such that any new products could be trivially included in the line.

lurch:
The NOTE block underneath this text also says "Raspberry Pi 5 only supports a 64-bit kernel, so this parameter has been removed for that device."

tdewey-rpi:
> The NOTE block underneath this text also says "Raspberry Pi 5 only supports a 64-bit kernel, so this parameter has been removed for that device."

A fair note.

I'm not sure the best way to handle this, then - I'm fairly certain I don't like the disjointed-mentions around this flag, but perhaps it would be better served by trimming 5 from this change, and moving the _note_ block above this paragraph?

lurch:
> I'm not sure the best way to handle this, then - I'm fairly certain I don't like the disjointed-mentions around this flag, but perhaps it would be better served by trimming 5 from this change, and moving the _note_ block above this paragraph?

Might be worth splitting the Pi 5 info out of the current long NOTE block, and adding it as a separate NOTE block which appears earlier? :thinking:

documentation/asciidoc/computers/config_txt/boot.adoc
```diff
@@ -25,13 +25,15 @@ The Raspberry Pi 5 firmware defaults to loading `kernel_2712.img` because this i
...
+Defaults to 1 on Raspberry Pi 4, 400 and Compute Module 4, 4S platforms. Defaults to 0 on all other platforms. However, if the name given in an explicit `kernel` option matches one of the known kernels then `arm_64bit` will be set accordingly.
 
-NOTE: 64-bit kernels may be uncompressed image files or a gzip archive of an image (which can still be called kernel8.img; the bootloader will recognize the archive from the signature bytes at the beginning). The 64-bit kernel will only work on the Raspberry Pi 3, 3+, 4, 400, Zero 2 W and 2B rev 1.2, and Raspberry Pi Compute Modules 3, 3+ and 4. Raspberry Pi 5 only supports a 64-bit kernel, so this parameter has been removed for that device.
+NOTE: 64-bit kernels may be uncompressed image files or a gzip archive of an image (which can still be called kernel8.img; the bootloader will recognize the archive from the signature bytes at the beginning). The 64-bit kernel will only work on the Raspberry Pi 3, 3+, 4, 400, Zero 2 W and 2B rev 1.2, and Raspberry Pi Compute Modules 3, 3+ and 4.
```
nathan-contino, line 36:
```suggestion
64-bit kernels come in the following forms:

* uncompressed image files
* gzip archives of an image

Both forms may use the `img` file extension; the bootloader recognizes archives using signature bytes at the start of the file.

The following Raspberry Pi models support this flag:

* 2B rev 1.2
* 3
* 3+
* 4
* 400
* Zero 2 W
* Compute Module 3
* Compute Module 3+
* Compute Module 4

Later models, such as Raspberry Pi 5, _only_ support the 64-bit kernel. Such models do not support this flag.
```

documentation/asciidoc/computers/config_txt/boot.adoc
```diff
@@ -25,13 +25,15 @@ The Raspberry Pi 5 firmware defaults to loading `kernel_2712.img` because this i
 
 === `arm_64bit`
 
+NOTE: This flag has been removed from Raspberry Pi 5, as Raspberry Pi 5 only supports a 64-bit kernel.
```
nathan-contino, line 28:
```suggestion
```

I suggest removing this because notes are disruptive to section flow and I think I've found a better way to convey this information in the text below. In particular, while calling out this specific model _might_ be helpful right now, it's not very futureproof: as we release more models, we'll have to expand this note to a massive and unhelpful size. Better to clarify in the meat and potatoes of this section, I think.

documentation/asciidoc/computers/config_txt/boot.adoc
```diff
@@ -25,13 +25,34 @@ The Raspberry Pi 5 firmware defaults to loading `kernel_2712.img` because this i
...
+
+* 2B rev 1.2
+* 3
+* 3+
+* 4
+* 400
+* Zero 2 W
+* Compute Module 3
+* Compute Module 3+
+* Compute Module 4
```
nathan-contino, line 52:
```suggestion
* Compute Module 4
* Compute Module 4S
```
I think the original writer missed this, or forgot to update it when we released 4S and added it above.

nathan-contino (approved):
LGTM! Thanks for waiting a bit on this. I knew there was a better way to present the info, but I had to chew on it for a little while. I think we're in a much better place now with this change.

Now, to propagate these ideas into the other config.txt entries 😉

documentation/asciidoc/computers/config_txt/boot.adoc
nathan-contino: Agree with Tom in principle. But for my own edification: I assumed that CM5 wouldn't support 32-bit mode at all, just like Pi 5. So I don't think we need to add it here at release. Is that correct, @JamesH65 ? Or will CM5 support 32-bit mode?

documentation/asciidoc/computers/config_txt/boot.adoc
```diff
@@ -25,13 +25,34 @@ The Raspberry Pi 5 firmware defaults to loading `kernel_2712.img` because this i
...
 
 In 64-bit mode, the firmware will choose an appropriate kernel (e.g. `kernel8.img`), unless there is an explicit `kernel` option defined, in which case that is used instead.
 
-Defaults to 1 on Pi 4s (Pi 4B, Pi 400, CM4 and CM4S), and 0 on all other platforms. However, if the name given in an explicit `kernel` option matches one of the known kernels then `arm_64bit` will be set accordingly.
+Defaults to 1 on Raspberry Pi 4, 400 and Compute Module 4, 4S platforms. Defaults to 0 on all other platforms. However, if the name given in an explicit `kernel` option matches one of the known kernels then `arm_64bit` will be set accordingly.
+
+64-bit kernels come in the following forms:
+
+* uncompressed image files
+* gzip archives of an image
```
lurch, line 38: I wonder if we should remove the plural from these two lines, given that you'll only ever be using one kernel at any given time? And IMHO "archive" usually implies that there's more than one file inside the archive, so perhaps this could be simplified to? (so "uncompressed image file" and "gzip-compressed image file" ?)

documentation/asciidoc/computers/config_txt/boot.adoc
lurch: I believe it's actually the SoC (and not the specific Model) that determines if it's capable of booting 32-bit kernel only (BCM2835, BCM2836), either 32- or 64-bit kernel (BCM2837, RP3A0, BCM2711), or 64-bit kernel only (BCM2712) ?

But (in general) we explicitly list the Models in documentation rather than SoCs, because people find it easier to identify their Model than to identify their SoC :socks: :wink:

Issue #3566, opened 2024-03-21
Need to mention the Raspberry Pi 4 Case Fan in the thermals section


aallan:
The section on thermals and official products, see https://www.raspberrypi.com/documentation/computers/raspberry-pi.html#fan-cases, doesn't mention the Raspberry Pi 4 Case Fan, https://www.raspberrypi.com/products/raspberry-pi-4-case-fan/.

lurch:
Cool! :rofl:

JamesH65:
Please fetch coat.

PR #3929, opened 2024-11-21
Update Fan Cases with mention of the Pi 4 Case Fan

nllho:
Resolves https://github.com/raspberrypi/documentation/issues/3566.

Keeping it minimal here! My thinking (for these types of tickets in general) is that if the current/latest model is already documented, we only need to include previous models for completeness. That means that we won't have the finer details of Pi 4 fan behaviour in the doc (at least until there's a good use case to go around and dig them up!) The assembly and usage instructions for both fans can be found on their respective product pages and product briefs.

Other notes (just general tech writing thoughts):
* Since I chose to keep the Pi 4 fan mention quite short, I didn't create a new section for it. 
* I put the mention at the top of the Fan Cases section, rather than at the bottom. That particular section ends with a table, and in general I'll avoid placing short pieces of information right below tables or graphics, since that makes them somewhat easy to miss.
* I made some small readability updates to reinforce the fact that the detailed information refers to the Pi 5 fan only.

### DIFF
--- a/documentation/asciidoc/computers/raspberry-pi/frequency-management.adoc
+++ b/documentation/asciidoc/computers/raspberry-pi/frequency-management.adoc
@@ -66,9 +66,9 @@ Raspberry Pi 5 has two official fan options to assist with cooling:
 * https://www.raspberrypi.com/products/active-cooler/[Active Cooler]
 * https://www.raspberrypi.com/products/raspberry-pi-5-case/[Case for Raspberry Pi 5]
 
-Both of these plug into the four-pin JST-SH PWM fan connector located in the upper right of the board between the 40-pin GPIO header and the USB 2 ports. The fan connector pulls from the same current limit as USB peripherals. We recommend the Active Cooler case for overclockers, since it provides better cooling performance.
+Both of the Raspberry Pi 5 fan options plug into the four-pin JST-SH PWM fan connector located in the upper right of the board between the 40-pin GPIO header and the USB 2 ports. The fan connector pulls from the same current limit as USB peripherals. We recommend the Active Cooler case for overclockers, since it provides better cooling performance.
 
-Both of the available official accessories are actively managed by Raspberry Pi firmware. As the temperature of the Raspberry Pi increases, the fan reacts in the following way:
+The fans of the Active Cooler and the Raspberry Pi 5 Case are both actively managed by Raspberry Pi firmware. As the temperature of the Raspberry Pi 5 increases, the fan reacts in the following way:
 
 * below 50°C, the fan does not spin at all (0% speed)
 * at 50°C, the fan turns on at a low speed (30% speed)
@@ -82,7 +82,7 @@ At boot the fan is turned on, and the tachometer input is checked to see if the 
 
 ==== Fan connector pinout
 
-The fan connector is a 1mm pitch JST-SH socket containing the following four pins:
+The Raspberry Pi 5 fan connector is a 1mm pitch JST-SH socket containing the following four pins:
 
 [cols="1,2,2",width="50"%"]
 |===
--- a/documentation/asciidoc/computers/raspberry-pi/frequency-management.adoc
+++ b/documentation/asciidoc/computers/raspberry-pi/frequency-management.adoc
@@ -59,6 +59,8 @@ Thanks to built-in throttling, heatsinks are not necessary to prevent overheatin
 
 === Fan cases
 
+For the Raspberry Pi 4, you have the option of adding the https://www.raspberrypi.com/products/raspberry-pi-4-case-fan/[Raspberry Pi 4 Case Fan] to the lid of the Raspberry Pi 4 case.
+
 Raspberry Pi 5 has two official fan options to assist with cooling:
 
 * https://www.raspberrypi.com/products/active-cooler/[Active Cooler]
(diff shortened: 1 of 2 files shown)


documentation/asciidoc/computers/raspberry-pi/frequency-management.adoc
```diff
@@ -59,14 +59,16 @@ Thanks to built-in throttling, heatsinks are not necessary to prevent overheatin
 
 === Fan cases
 
+For the Raspberry Pi 4, you have the option of adding the https://www.raspberrypi.com/products/raspberry-pi-4-case-fan/[Raspberry Pi 4 Case Fan] to the lid of the Raspberry Pi 4 case.
```
nathan-contino, line 62: Can we get an introduction sentence, instead of jumping straight into Raspberry Pi 4 info? Maybe something like: "To ensure the best performance for your Raspberry Pi, use an active cooling solution such as a fan."

Then, we could either break this up into "Raspberry Pi 4" and "Raspberry Pi 5" sections (slightly preferred), or perhaps provide info about each model in an unordered list. Either way, we should probably restructure this sentence introduction to better match the Raspberry Pi 5 equivalent:

```suggestion
For Raspberry Pi 4, add the https://www.raspberrypi.com/products/raspberry-pi-4-case-fan/[Raspberry Pi 4 Case Fan] to the lid of the Raspberry Pi 4 case.
```

And I'd probably modify the line below to something like "For Raspberry Pi 5, use one of the official fan options:"

documentation/asciidoc/computers/raspberry-pi/frequency-management.adoc
```diff
@@ -59,14 +59,16 @@ Thanks to built-in throttling, heatsinks are not necessary to prevent overheatin
...
 * https://www.raspberrypi.com/products/raspberry-pi-5-case/[Case for Raspberry Pi 5]
 
-Both of these plug into the four-pin JST-SH PWM fan connector located in the upper right of the board between the 40-pin GPIO header and the USB 2 ports. The fan connector pulls from the same current limit as USB peripherals. We recommend the Active Cooler case for overclockers, since it provides better cooling performance.
+Both of the Raspberry Pi 5 fan options plug into the four-pin JST-SH PWM fan connector located in the upper right of the board between the 40-pin GPIO header and the USB 2 ports. The fan connector pulls from the same current limit as USB peripherals. We recommend the Active Cooler case for overclockers, since it provides better cooling performance.
 
-Both of the available official accessories are actively managed by Raspberry Pi firmware. As the temperature of the Raspberry Pi increases, the fan reacts in the following way:
+The fans of the Active Cooler and the Raspberry Pi 5 Case are both actively managed by Raspberry Pi firmware. As the temperature of the Raspberry Pi 5 increases, the fan reacts in the following way:
```
nathan-contino, line 71: I believe the Raspberry Pi 4 Case Fan firmware also actively manages fan speed. Perhaps we could trim the first sentence here, and instead mention that idea in the introduction?
```suggestion
As the temperature of the Raspberry Pi 5 increases, official fans react in the following way:
```

documentation/asciidoc/computers/raspberry-pi/frequency-management.adoc
nathan-contino: Building on the hysteresis comment below, perhaps add a sentence like this to the introduction:

"Raspberry Pi firmware manages fan speeds for all official fans."

That way, it's clear that the Raspberry Pi 4 fan doesn't just blow at 100% all of the time :-)

nathan-contino (commented):
Nice improvement. Your additions made me think a little more critically about the section layout, so I made some suggestions. Address my nitpicky suggestions and we'll merge this in!

nathan-contino (approved):
lgtm!

documentation/asciidoc/computers/raspberry-pi/frequency-management.adoc
```diff
@@ -78,9 +86,9 @@ Temperature decreases use the same mapping with a 5°C **hysteresis**; fan speed
 
 At boot the fan is turned on, and the tachometer input is checked to see if the fan is spinning. If it is, then the `cooling_fan` device tree overlay is enabled. This overlay is in `bcm2712-rpi-5-b.dtb` by default, but with `status=disabled`.
 
-==== Fan connector pinout
+==== Raspberry Pi 5 fan connector pinout
```
nllho, line 89: One more nit 😁 Thoughts on this header? It's at the same level as the other "fan" headers (I didn't want to make it too small). Normally I don't think I would separate this information out using a header, but in this case, the connector info is linked in the introduction.

nllho (commented):
Thanks so much @nathan-contino! The nits are very helpful, especially early on -- I can start to see how much editing and rewriting we want to aim for in these types of PRs.

Have reviewed the style guide (and will reread it often, I'm sure). For future issues, in addition to making changes addressed in the issue, I'll think about style and structure edits at roughly the same "level" as you're suggesting here :)

documentation/asciidoc/computers/raspberry-pi/frequency-management.adoc
nathan-contino: Seems like a reasonable compromise. Ideally we'd nest this within the Pi 5 section, but... that might go too deep. This seems to be the best place for this right now.

PR #3040, opened 2023-08-03
Custom template for integrating book links

nelliemckesson:
This creates some special handling for sections that are intended to link to a book. You should be able to use this same format for sections of any level. You can include any type of content inside your section, as usual, and a final paragraph will automatically be appended to the section, linking to the book with text that corresponds to the type of book. The following attributes are required:

- `.booklink`
- `booktype` (options are: `free`, `buy`, or `donate`)
- `link` to the book, buy page, or donate page
- `image` of the book cover (can be either included in the repo or a url). Note that in order for it to be picked up by the toolchain, the `image` attribute must be formatted the same way you'd format a regular asciidoc image, including both the `image::` and trailing `[]`; see below for an example.

```
[.booklink, booktype="free", link=https://pip.raspberrypi.com/categories/685-whitepapers-app-notes/documents/RP-003471-WP/Using-a-DPI-display.pdf, image=image::images/simple-electronics-with-gpio-zero.jpg[]]
==== Going further

You can find more information on how to program electronics connected to your Raspberry Pi with the GPIO Zero Python library in the Raspberry Pi Press book https://github.com/raspberrypipress/released-pdfs/raw/main/simple-electronics-with-gpio-zero.pdf[Simple Electronics with GPIO Zero]. Written by Phil King, it is part of the MagPi Essentials series published by Raspberry Pi Press. The book gets you started with the GPIO Zero library, and walks you through how to use it by building a series of projects. 
```

See #3035

PR #3941, opened 2024-11-26
Remove book callouts

nllho:
These were all I could find with a cursory search, let me know if you know of any others! After this, we should remove the unused custom template code (https://github.com/raspberrypi/documentation/pull/3040).

Context for posterity:
We're not going to continue with book callouts (booklinks) in this format because it doesn't meet our doc standards for maintainability or style. In the future we'll find a better way to gather and present related resources in the docs :)


nathan-contino (approved):
lgtm!

Issue #3094, opened 2023-09-04
Document changing menus of raspi-config


aallan:
The options on `raspi-config` will now change depending on which Raspberry Pi hardware it's running on post-Bookworm. This needs documenting. The decision I've made for #2955 is to document the Pi 4 case only right now with a `NOTE` to say that there will be changes depending on what hardware you're on.

We probably need to document this properly at some point, but I don't see it as urgent.

aallan:
Low priority one @nathan-contino

Issue #2955, opened 2023-05-18
raspi-config options have changed


lurch:
(and may still be changing)
https://www.raspberrypi.com/documentation/computers/configuration.html#list-of-options

lurch:
@spl237 tells me that the contents of the "Advanced" menu will vary depending on which model of Pi is being used, and also on which graphics stack is being used. He says that "most users" shouldn't be changing these settings, and therefore we shouldn't mention them in our documentation. The only options of raspi-config that we ought to be documenting are those which are also displayed in the "GUI version" i.e. the "Raspberry Pi Configuration" program.

aallan:
I use the Advanced Menu pretty much every time I set up a Pi. It will be documented.

spl237:
You shouldn't be using it. What are you doing in the Advanced Menu that needs doing every time you set up a Pi?

aallan:
Still a thing.

jehlers42:
@spl237 My PR'd changes to the getting started guide has a quick mention of the menu, basically calling it out before the user would hit "Write": "Optionally, you can click the cog wheel icon to open the <<Advanced options>> menu."

We also call out a few times about the need to use the advanced menu to create users for headless setups. Granted I'm still new to this whole Pi thing so I've mostly been going with what already existed in the guide so I'm not sure how needed that info truly is.

I do tend to agree that "service only" type situations shouldn't be fully documented, but this one seems easily-exposed to our users since it's a simple cogwheel icon on the installation GUI. Curious users are likely to click it and potentially get themselves in trouble. Ideally the UX would be the first line of defence here, with a "are you sure you want to do this?" confirmation dialog or tooltip.

Otherwise my suggestion would be a warning box like you recommend with a link to a doc that does explain the menu, but also makes it very clear the risks associated with changing those settings.

aallan:
@jehlers42 I think you're talking about the Imager, we were talking about the raspi-config command line tool which is used post-installation.

spl237:
Yes, apologies for that - we hopefully won't be making many more changes, but the image is still displaying more churn than I would like...

aallan:
Okay. So the decision is to document `raspi-config` as it works on Pi 4, with a `NOTE` block to say that options in the Advanced menu will depend on the underlying hardware you're running things on and will change between boards.

@lurch assigning to you.

lurch:
Fixed on `bookworm` branch.

PR #3945, opened 2024-11-29
Update raspi-config

nllho:
Some minor updates that came up from poking around in https://github.com/raspberrypi/documentation/issues/3094. Think we can close this issue after addressing the main feedback below :)

Incorporate Simon's feedback from https://github.com/raspberrypi/documentation/issues/2955#issuecomment-1561344049:
- Add warnings to the Advanced Options sections. I borrowed the warning wording from a different part of the docs, if we prefer something else just let me know!
- Specify that it is the Advanced Options may vary across models (GUI options should stay consistent).

Other bits of cleanup for readability:
- Simplify the introduction.
- Make negations easier to read, e.g. 'avoid' instead of 'do not'.

Also proposing that we remove images from this section. To me, they don't seem to be a huge value add, (except for maybe the first image, which could help people confirm they're in the right menu). But let me know if the images here are useful for other reasons :)


documentation/asciidoc/computers/configuration/raspi-config.adoc
```diff
@@ -217,7 +210,7 @@ Configure the network's proxy settings.
 
 ==== Boot order
 
-On Raspberry Pi 4 and later, specify whether to boot from USB or network if the SD card isn't inserted. For more information, see xref:raspberry-pi.adoc#raspberry-pi-bootloader-configuration[bootloader configuration].
+On Raspberry Pi 4 and later, specify whether to boot from USB or network in absence of an SD card. For more information, see xref:raspberry-pi.adoc#raspberry-pi-bootloader-configuration[bootloader configuration].
```
lurch, line 213: Or SSD, on a Pi 5 with an M.2 HAT?

documentation/asciidoc/computers/configuration/raspi-config.adoc
```diff
@@ -1,33 +1,32 @@
...
 
-TIP: Raspberry Pi Desktop users can access a graphical version of this application at **Preferences** > **Raspberry Pi Configuration**. However, some advanced configuration is only available in `raspi-config`.
+`raspi-config` helps you configure your Raspberry Pi. Changes to `raspi-config` will affect xref:config_txt.adoc#what-is-config-txt[`/boot/firmware/config.txt`] and various Linux configuration files.
 
-`raspi-config` helps you configure your Raspberry Pi.  Available options may differ between Raspberry Pi models. To open the configuration tool, run the following command:
+=== Getting started
+
+To open the configuration tool from the desktop GUI, go to **Preferences** > **Raspberry Pi Configuration**.
+
+Alternatively, to access the configuration tool via the terminal:
```
lurch, line 10: I'm not sure this is "correct" as it implies that the GUI version and the terminal version are equivalent, when they're not. The GUI version is much easier to use, but as the now-deleted TIP used to say, there are several advanced options which are available in the CLI tool which aren't available in the GUI tool.
IMHO :slightly_smiling_face:

documentation/asciidoc/computers/configuration/raspi-config.adoc
```diff
@@ -1,33 +1,32 @@
 [[raspi-config]]
 == `raspi-config`
 
-TIP: Raspberry Pi Desktop users can access a graphical version of this application at **Preferences** > **Raspberry Pi Configuration**. However, some advanced configuration is only available in `raspi-config`.
+`raspi-config` helps you configure your Raspberry Pi. Changes to `raspi-config` will affect xref:config_txt.adoc#what-is-config-txt[`/boot/firmware/config.txt`] and various Linux configuration files.
```
lurch, line 4: Could change "various Linux configuration files" to "various other configuration files" ? :shrug:

documentation/asciidoc/computers/configuration/raspi-config.adoc
```diff
@@ -1,33 +1,32 @@
...
-
-image::images/raspi-config.png[raspi-config main screen]
-
-Use the **Up** and **Down** arrow keys to move the highlighted selection between the options available.
-
-Press the **Right** arrow key or press **Tab** to access the `<Select>` and `<Finish>` buttons. Press **Left** or press **Tab** to return to the options.
+To navigate the configuration tool from the terminal:
 
-`raspi-config` automates edits to xref:config_txt.adoc#what-is-config-txt[`/boot/firmware/config.txt`] and various Linux configuration files. Some options require a reboot to take effect. If you changed any of these, `raspi-config` asks you to reboot when you exit.
+* Use the arrow keys to scroll through the settings list.
```
lurch, line 19: "Use the Up and Down arrow keys..." ?

documentation/asciidoc/computers/configuration/raspi-config.adoc
```diff
@@ -167,15 +160,13 @@ Enable or disable a read-only filesystem.
 
 ==== Fan
 
-Customise the behaviour of the GPIO-connected https://www.raspberrypi.com/products/raspberry-pi-4-case-fan/[Raspberry Pi 4 Case Fan]. Does not affect the fans in the https://www.raspberrypi.com/products/raspberry-pi-5-case/[Raspberry Pi 5 Case for Raspberry Pi 5] or https://www.raspberrypi.com/products/active-cooler/[Raspberry Pi 5 Active Cooler], which connect using a special four-pin fan header.
+Customise the behaviour of the GPIO-connected https://www.raspberrypi.com/products/raspberry-pi-4-case-fan/[Raspberry Pi 4 Case Fan]. This setting is inapplicable to other fan models.
```
lurch, line 163: It's only a "gut feeling", but I think "not applicable" feels more natural than "inapplicable" ? :shrug:

documentation/asciidoc/computers/configuration/raspi-config.adoc
nathan-contino: Maybe even simpler will do?

```suggestion
`raspi-config` helps you configure your Raspberry Pi. Changes to `raspi-config` will affect xref:config_txt.adoc#what-is-config-txt[`/boot/firmware/config.txt`] and other configuration files.
```

documentation/asciidoc/computers/configuration/raspi-config.adoc
nathan-contino: While I generally agree that we should reduce the amount of negated terms in the doc, we can make an exception when the negated version reduces the wordiness:

```suggestion
Customise the behaviour of the GPIO-connected https://www.raspberrypi.com/products/raspberry-pi-4-case-fan/[Raspberry Pi 4 Case Fan]. Not applicable to other fan models.
```

documentation/asciidoc/computers/configuration/raspi-config.adoc
nathan-contino, line 10:
```suggestion
Alternatively, run the following command to access the configuration tool via the terminal:
```
Suggestion: Introducing command line snippets. This also helps contrast example output, bash that users should write into a file, and interactive snippets.

nathan-contino: After the code snippet, I would add something like this reduced-size tip admonition:

```
TIP: Some advanced configuration, including <common example>, is only available in `raspi-config`.
```

Use your best judgment to determine if there's a decent example worth tossing in there. Personally I think the statement feels a little empty and meaningless without an example -- _what_ kind of 'advanced' config?!?! -- so I lean towards including it if we can come up with one that's short and sweet.

documentation/asciidoc/computers/configuration/raspi-config.adoc
```diff
@@ -102,14 +99,12 @@ Enable or disable 4Kp60 resolution for HDMI outputs.
...
 
-image::images/raspi-interface.png[raspi-config interface options]
-
 [[ssh]]
 ==== SSH
 
 Enable or disable remote terminal access to your Raspberry Pi using SSH.
 
-SSH allows you to remotely access the command line of the Raspberry Pi from another computer. SSH is disabled by default. Read more about using SSH on the xref:remote-access.adoc#ssh[SSH documentation page]. If connecting your Raspberry Pi directly to a public network, you should not enable SSH unless you have set up secure passwords for all users.
+SSH allows you to remotely access the command line of the Raspberry Pi from another computer. SSH is disabled by default. If connecting your Raspberry Pi directly to a public network, set up secure passwords for all users before enabling SSH. Learn more about using SSH on the xref:remote-access.adoc#ssh[SSH documentation page].
```
nathan-contino, line 107:
```suggestion
SSH allows you to remotely access the command line of the Raspberry Pi from another computer. SSH is disabled by default. If connecting your Raspberry Pi directly to a public network, set up passwords for all users before enabling SSH. Learn more about using SSH on the xref:remote-access.adoc#ssh[SSH documentation page].
```

While users absolutely should use a secure password, I think it goes without saying. Let's trim it here.

documentation/asciidoc/computers/configuration/raspi-config.adoc
```diff
@@ -246,7 +239,7 @@ Display a description of `raspi-config`.
 [[finish]]
 === Finish
 
-Exits `raspi-config`. If you made changes that require a reboot, `raspi-config` prompts you to reboot. When implementing changes for the first time, it's best to reboot. If you chose to resize your SD card, rebooting may take longer than usual.
+Exit `raspi-config`. If necessary, raspi-config will ask you to reboot when you exit. When implementing changes for the first time, reboot to ensure your changes take effect.
```
nathan-contino, line 242:
```suggestion
Exit `raspi-config`. If necessary, `raspi-config` will ask you to reboot when you exit. When implementing changes for the first time, reboot to ensure your changes take effect.
```
Suggestion: consistently refer to raspi-config in monospace text, since it's both a program and a command.

documentation/asciidoc/computers/configuration/raspi-config.adoc
```diff
@@ -344,9 +337,9 @@ $ sudo raspi-config nonint do_hostname <hostname>
 ----
 
 [[boot-options-nonint]]
-==== Boot/auto login
+==== Boot/Auto login
 
-Select whether to boot to console or desktop, and whether or not your Raspberry Pi automatically logs into your current user account when powered on.
+Boot to console or desktop with the option of an automatic login to your current user account.
```
nathan-contino, line 342:
```suggestion
Select the following behaviour at boot time:

* whether to boot to console or desktop
* whether your Raspberry Pi automatically logs into your current user account when powered on
```

Suggestion: I think your streamlining makes it less clear that automatic login is a separate option here, and introduces some ambiguity about whether you can use automatic login if you boot to console. I think an unordered list would help?

Maybe someday we can get the engineers to split this up into two separate options :-)

documentation/asciidoc/computers/configuration/raspi-config.adoc
```diff
@@ -708,6 +701,10 @@ For a full list of possible `<country>` values, see the abbreviations used in th
 [[advanced-options-nonint]]
 === Advanced options
 
+Advanced options are intended for use during troubleshooting. Available options may differ between Raspberry Pi models.
```
nathan-contino, line 704: Suggestions:

* remove "may" from "may differ" -- since we make the software, I think it's safe to say that options _DO_ differ between models!
* the new introductory sentence sounds a bit clunky and isn't quite right; there are a lot of advanced options users might use for purposes other than troubleshooting. unfortunately i don't have a better suggestion, but I'm open to ideas. Even if those ideas are 'remove this entire paragraph' :-)

If we borrowed that introductory sentence from elsewhere in the docs, I think we should consider rewriting it in its original home as well. Unless it's more true in that setting?

nathan-contino (commented):
almost LGTM. Just a few minor issues. But I love the fact that we're getting rid of unneeded, hard-to-maintain images, and making this text more consistent. Nice job so far!

documentation/asciidoc/computers/configuration/raspi-config.adoc
lurch: I dunno if that's the right wording, as USB is also a "valid boot media" :thinking:

documentation/asciidoc/computers/configuration/raspi-config.adoc
lurch: I guess you _could_ even change "affect" to "modify" ? :shrug:

documentation/asciidoc/computers/configuration/raspi-config.adoc
lurch: ping @tdewey-rpi for his thoughts here, just to be on the safe side. I guess at the very least we ought to link to https://www.raspberrypi.com/documentation/computers/configuration.html#secure-your-raspberry-pi ?

tdewey-rpi: You summoned me, @lurch?

I agree with both of you.

1. I don't think we need to say 'secure password' - what does that even mean, and how can you be sure something that is 'secure' today is considered 'secure' tomorrow, etc.
2. We can, however, provide a link to the securing page - and then change this to a later security centre.

lurch: :wave: 

I've just re-read Nate's actual suggested wording - "If connecting your Raspberry Pi directly to a public network, set up passwords for all users before enabling SSH." - AFAIK it's not possible to have a user on Raspberry Pi OS _without_ a password... :thinking:

nathan-contino: In that case, we can probably just remove the sentence entirely!

I think I have managed to create a user without a password before. But it may have been an older version of the OS.

documentation/asciidoc/computers/configuration/raspi-config.adoc
nathan-contino:
```suggestion
* Use the up and down arrow keys to scroll through the settings list.
```

documentation/asciidoc/computers/configuration/raspi-config.adoc
nathan-contino:
```suggestion
```

documentation/asciidoc/computers/configuration/raspi-config.adoc
```diff
@@ -184,26 +175,28 @@ Select a locale, for example `en_GB.UTF-8 UTF-8`.
...
 ==== WLAN country
 
-Sets the country code for your wireless network.
+Set the country code for your wireless network.
 
 [[advanced-options]]
 === Advanced options
 
-image::images/raspi-adv.png[raspi-config advanced options]
+Advanced options are intended for use during troubleshooting. Available options may differ between Raspberry Pi models.
```
nathan-contino, line 192:
```suggestion
```

documentation/asciidoc/computers/configuration/raspi-config.adoc
```diff
@@ -1,33 +1,32 @@
...
+=== Getting started
+
+To open the configuration tool from the desktop GUI, go to **Preferences** > **Raspberry Pi Configuration**.
+
+Alternatively, run the following command to access the configuration tool via the terminal:
 
 [source,console]
 ----
 $ sudo raspi-config
 ----
```
nathan-contino, line 15:
```suggestion
----
TIP: Some advanced configuration is available in the `raspi-config` CLI, but not the Raspberry Pi Configuration GUI.
```

documentation/asciidoc/computers/configuration/raspi-config.adoc
```diff
@@ -673,7 +670,7 @@ For a full list of possible `<locale>` values, see the abbreviations used in the
 [[change-timezone-nonint]]
 ==== Time zone
 
-Set your local time zone, starting with the region then selecting a city, e.g. "Europe/London".
+Set your local time zone in the format `Region/City`, for example 'Europe/London'. Type a letter to jump to that letter in the list.
```
lurch, line 673: Huh? This is the documentation for the **non-interactive** usage of `raspi-config`, so I don't think you'll be able to "Type a letter to jump to that letter in the list." :wink:

nathan-contino: Thanks, retconned that sentence.

PR #3950, opened 2024-12-03
Actual mobile toc; dark mode; speed up site loading

nathan-contino:
Cleaned up a LOT of HTML and CSS.

Added a dark theme for the entire docs site.

Improved responsiveness and load times for the site in general.

Uses space more effectively on desktop and mobile.

Sample images:

Dark desktop:
Screenshot 2024-12-03 at 14 45 52

Light desktop:
Screenshot 2024-12-03 at 14 45 43

Light mobile:
Screenshot 2024-12-03 at 14 46 13

Dark mobile:
Screenshot 2024-12-03 at 14 47 02

Light mobile menu:
Screenshot 2024-12-03 at 14 46 36

Dark mobile menu:
Screenshot 2024-12-03 at 14 46 49


jekyll-assets/_includes/footer.html
mudge, line 1: This file (as with `header.html`) is designed to be shared and updated across the multiple sites that make up raspberrypi.com (e.g. Raspberry Pi Events, PIP, Raspberry Pi ID, the homepage, the news pages, etc.).

If you make changes here, they'll be lost whenever we need to update it for changes to the header or footer or are you wanting to break away from the site-wide versions and have a documentation-only version?

jekyll-assets/_includes/head.html
```diff
@@ -1,5 +1,6 @@
 <meta charset="utf-8">
 <meta http-equiv="X-UA-Compatible" content="IE=edge">
+<link rel="shortcut icon" href="{{ site.baseurl }}/favicon.ico">
```
mudge, line 3: Is this different to the default behaviour of using https://www.raspberrypi.com/favicon.ico?

jekyll-assets/_includes/theme.html
```diff
@@ -0,0 +1,3 @@
+<div id="theme-toggle-container">
+	<a id="theme-toggle" alt="toggle site theme" title="toggle site theme" onclick="toggleTheme()"></a>
```
mudge, line 2: Would it be better to be unobtrusive here and [add a `click` event listener](https://developer.mozilla.org/en-US/docs/Web/API/EventTarget/addEventListener) on the element (identifying it by its ID of `theme-toggle`)?

e.g.

```javascript
document.getElementById("theme-toggle").addEventListener("click", (e) => {
  e.preventDefault();

  toggleTheme();
});
```

jekyll-assets/_layouts/boxes.html
```diff
@@ -1,19 +1,20 @@
...
-    {% include header.html %}      
-    {% if page.name == "404.adoc" %}
-      {% include toptitle404.html %}
-      {% include subtitle404.html %}
-    {% else %}
-      {% include toptitle.html %}
-      {% include subtitle.html %}
-    {% endif %}
-    {% include tabs.html %}
+  <body onload="makeToc()">
```
mudge, line 6: As above re being unobtrusive: could you fire `makeToc()` on `DOMContentLoaded`?

```javascript
document.addEventListener("DOMContentLoaded", () => {
  makeToc();
});
```

jekyll-assets/_layouts/docs.html
```diff
@@ -1,25 +1,181 @@
...
+                            {% if entry.subsections[0] %}
+                            for="{{ item.path }}#{{ entry.anchor }}"
+                            {% else %}
+                            {% endif %}>
+                            <div class="toc-toggle-container">
+                              {% if entry.subsections[0] %}
+                              <a class="level3" href="{{ site.baseurl }}{{ item.path }}#{{ entry.anchor }}" >
+                                <a id="{{ item.path }}#{{entry.anchor}}-toc" class="toc-item" alt="expand documentation category" title="expand documentation category"
+                                >
+                                <style onload="if (window.location.hash.includes('{{ entry.anchor }}')) { document.getElementById('{{ item.path }}#{{entry.anchor}}-toc').setAttribute('style', 'font-weight:bold'); document.getElementById('{{ item.path }}#{{ entry.anchor }}').checked = true; }"></style>
```
mudge, line 102: Does this bit of JS belong only here or could it be extracted elsewhere given the associated `style` element is empty?

jekyll-assets/_layouts/docs.html
```diff
@@ -1,25 +1,181 @@
...
+                                </a>
+                              </a>
+                              {% else %}
+                              <a class="level3">
+                                <a id="{{ item.path }}#{{entry.anchor}}-toc" alt="expand documentation category" title="expand documentation category"
+                                  href="{{ site.baseurl }}{{ item.path }}#{{ entry.anchor }}"
+                                  onclick="document.getElementById('mobile-toggle').checked = false;"
+                                  class="toc-item no-dropdown"
+                                  >
+                                  <style onload="if (window.location.hash.includes('{{ entry.anchor }}')) { document.getElementById('{{ item.path }}#{{entry.anchor}}-toc').setAttribute('style', 'font-weight:bold'); document.getElementById('{{ item.path }}#{{ entry.anchor }}').checked = true; }"> </style>
```
mudge, line 113: As above re using an empty `style` tag purely to execute JavaScript.

jekyll-assets/_includes/head.html
nathan-contino: Currently, local builds of the docs yell at you on the CLI about a missing favicon and lack a favicon entirely. I added this to fix the issue locally, and I _think_ it shouldn't break anything. Though your sanity check is appreciated. Hopefully that makes sense now.

jekyll-assets/_includes/footer.html
nathan-contino: To support dark mode, we do need to tweak the header and footer files, as otherwise they'll remain in light mode at all times. I've kept the tweaks minimal (mostly colour based, with some padding trimmed from the header to reduce excess space usage), so we _could_ try to mostly share. But unless we add dark mode to the main site as well, this will require deviation.

jekyll-assets/_includes/theme.html
nathan-contino: My understanding is that this ultimately does the exact same thing, but with an element lookup. Is there a performance benefit that I'm not aware of, or some penalty to using onclick?

jekyll-assets/_layouts/boxes.html
nathan-contino: Similar question as the toggle comment above -- happy to move the call, but I'd like to know if there's a performance or organisational benefit before I do that!

jekyll-assets/_layouts/docs.html
nathan-contino: as above, so below!

jekyll-assets/_layouts/boxes.html
mudge: If you can extract all your JavaScript into a separate file then it can be cached across page loads and served from our CDN while the HTML is typically always loaded from the origin servers.

nathan-contino: Ah, gotcha. I can pull the makeToc and theme stuff into a separate file, but the inline toc manipulation will have to stay here as it's part of the static site generation process.

lurch:
I had a play with this on Nate's laptop this afternoon - it looks gorgeous! :heart_eyes:

jekyll-assets/_includes/footer.html
lurch: Given that this PR is modifying **both** the navigation and also adding a dark-mode, I wonder if it _might_ be worth splitting that into two separate PRs? I.e. get the navigation changes merged in, and then add dark-mode theme as separate (later) step? :thinking:  Or would it be far too much work to disentangle the two things?

Also, we'll probably want @jackbenwillis to design us some "light mode" and "dark mode" icons which would integrate with the design better than the generic sun and moon icons that you're currently using?

documentation/asciidoc/microcontrollers/c_sdk/official_sdk.adoc
```diff
@@ -1,4 +1,4 @@
-== Raspberry Pi Pico C/{cpp} SDK
+== Raspberry Pi Pico C/C++ SDK
```
lurch, line 1: Bit of historical context - in some situations we **needed** to use `{cpp}` instead of `C++` because of the way asciidoc uses `++` to mark https://docs.asciidoctor.org/asciidoc/latest/pass/pass-macro/ , so if `C++` was used twice within one paragraph then things went badly wrong :rofl: 

But it's possible that past-lurch was a bit overzealous in swapping `C++` to `{cpp}` in places that it wasn't actually needed? :wink:

Makefile
```diff
@@ -118,7 +118,7 @@ html: run_ninja
 
 # Build the html output files and additionally run a small webserver for local previews
 serve_html: run_ninja
-	$(JEKYLL_CMD) serve
+	$(JEKYLL_CMD) serve --watch
```
lurch, line 121: I _think_ the reason that I didn't do this previously, was because a lot of the files that you edit need to be "converted" by some of the other Python scripts first, and so you typically need to run `make html` before running `make serve_html`. And so you might get confused if you were using `serve --watch` and you edited a file which needed to be "converted", which Jekyll isn't able to do automatically? (because you wouldn't see any changes when you refreshed your browser)

But of course this is fine if you actually understand what's going on! :slightly_smiling_face:

_config.yml
```diff
@@ -17,8 +17,9 @@ title: Raspberry Pi Documentation
 description: >- # this means to ignore newlines until "baseurl:"
   Raspberry Pi Documentation.
 baseurl: "/documentation" # the subpath of your site, e.g. /blog
-url: "" # the base hostname & protocol for your site, e.g. http://example.com
```
lurch, line 20: I'm sure that there must have been some reason why this was left empty; but that reason is probably now lost to the mists of time...

jekyll-assets/_includes/contribute.html
```diff
@@ -1,5 +0,0 @@
-<div class="contribute">
```
lurch, line 1: Interestingly, it looks like "contribute" used to _only_ be at the bottom of pages like https://www.raspberrypi.com/documentation/accessories/ but not at the bottom of pages like https://www.raspberrypi.com/documentation/accessories/sd-cards.html
I can't remember why! :wink: 

```
$ git grep -C4 footer.html
jekyll-assets/_layouts/boxes.html-    </div>
jekyll-assets/_layouts/boxes.html-
jekyll-assets/_layouts/boxes.html-    {% include contribute.html %} 
jekyll-assets/_layouts/boxes.html-    {% include copyright.html %} 
jekyll-assets/_layouts/boxes.html:    {% include footer.html %}  
jekyll-assets/_layouts/boxes.html-    {% include search.html %}
jekyll-assets/_layouts/boxes.html-
jekyll-assets/_layouts/boxes.html-  </body>
jekyll-assets/_layouts/boxes.html-</html>
--
jekyll-assets/_layouts/docs.html-      </section>
jekyll-assets/_layouts/docs.html-    </div>
jekyll-assets/_layouts/docs.html-
jekyll-assets/_layouts/docs.html-    {% include copyright.html %} 
jekyll-assets/_layouts/docs.html:    {% include footer.html %}  
jekyll-assets/_layouts/docs.html-    {% include scripts.html %}
jekyll-assets/_layouts/docs.html-    {% include search.html %}
jekyll-assets/_layouts/docs.html-
jekyll-assets/_layouts/docs.html-  </body>
```

jekyll-assets/_includes/footer.html
```diff
@@ -75,11 +75,11 @@
 }
 #__rptl-footer .__rptl-footer-icon {
   margin-right: 0.5em;
-  fill: #333;
+  fill: var(--subtle-text);
 }
 #__rptl-footer .__rptl-footer-link, #__rptl-footer .__rptl-footer-link--with-icon {
   font-weight: 400;
-  color: #333;
+  color: var(--subtle);
```
lurch, line 82: Is it correct that you've got both `var(--subtle-text);` and `var(--subtle);` ?

jekyll-assets/css/style.css
```diff
@@ -1,35 +1,71 @@
...
   --red-tint: #d75a64;
-  --docsearch-primary-color: var(--red);
+  --brand-colour: #cd2355;
+  --copy-button-colour: #d8d8d8;
+  --code-bg-colour: #151515;
+  --toc-hover-colour: #252525;
+  --subtle: #707070;
+  --less-subtle: #333;
+  --home: #cd2355;
+  --rp2040-context-tag: #50C878;
```
lurch, line 14: Does there also need to be a `--rp2350-context-tag` ? :shrug:

jekyll-assets/css/style.css
```diff
@@ -1,35 +1,71 @@
...
+  --accent: #cd2355;
```
lurch, line 15: I know **nothing** about how all this works, but I see that you have:
```
--brand-colour: #cd2355;
--home: #cd2355;
--accent: #cd2355;
```
Is that intentional?

jekyll-assets/css/style.css
```diff
@@ -1,35 +1,71 @@
...
+  --expanded-label: "▼";
+  color: var(--textcolor);
+}
+.light {
+  --bg: #fff;
+  --footer-bg: #f5f6f9;
+  --near-bg: #f6f6f6;
+  --red-tint: #d75a64;
+  --top-title-colour: #343434;
+  --brand-colour: #cd2355;
```
lurch, line 31: Please forgive me if this is a stupid question, but if this `--brand-colour` value is identical in both the `:root` and `.light` sections, does it really need to be specified in two separate places? (and obviously also any other colours that are the "same" in both sections)

jekyll-assets/_layouts/docs.html
```diff
@@ -1,26 +1,181 @@
...
-      <section id="content">
-        <h1>{{ page.sub_title | markdownify | remove: '<p>' | remove: '</p>'}}</h1>
-        {{ content }}
-      </section>
+  <body onload="makeToc()">
+    <script type="text/javascript">initTheme()</script>
+    {% include header.html %}
+    <input type="checkbox" id="mobile-toggle" />
+    <div id="docs-content">
+      <div id="toc-container-container">
```
lurch, line 11: Are we going to be using Docker? :rofl: :ship: :ship:

jekyll-assets/scripts/theme.js
```diff
@@ -0,0 +1,32 @@
+// toggles the site theme between dark and light
+function toggleTheme() {
+  // fetch the theme from local storage (if it exists)
+  var theme = localStorage.getItem('theme');
+  // if theme them has never been set, or is light, set the theme to the dark symbol in local storage to change it
```
lurch, line 5: "theme them" ?

jekyll-assets/_includes/footer.html
mudge: Rather than modifying `header.html` and `footer.html` directly and thereby preventing us from easily updating them across the sites, would a middle ground be for you to add your own separate CSS overrides in your own stylesheet until we can upstream dark mode support in the header across the whole site?

jekyll-assets/_includes/theme.html
mudge: See [my other comment](https://github.com/raspberrypi/documentation/pull/3950#discussion_r1868086493) re separating the HTML from the JavaScript but also the broader concept of [unobtrusive JavaScript](https://en.wikipedia.org/wiki/Unobtrusive_JavaScript) to encourage [progressive enhancement](https://www.gov.uk/service-manual/technology/using-progressive-enhancement).

mudge:
There's a big picture question here whether the documentation site wants to keep the standard raspberrypi.com header and footer such that clicking between the [homepage](https://www.raspberrypi.com), [news](https://www.raspberrypi.com/news/), and [documentation](https://www.raspberrypi.com/documentation/) sites retains an _identical_ header at all browser sizes (especially as documentation is hosted in a subdirectory rather than its own subdomain) or whether it wants its own look and feel.

I would prefer we stick with the standard header and instead upstream any changes you want to accommodate in your new design. Even though that is more effort, it would stop us being inconsistent across www.raspberrypi.com. Given the amount of changes here, as @lurch suggests, perhaps it is worth separating the cross-site changes from the documentation-specific ones?

Makefile
nathan-contino: Yes, this doesn't actually result in a live update of the site, presumably because of those other 'conversions'. I made this change as part of a slow-but-steady effort to eventually get live previews working (it gets some skin in the game).

_config.yml
nathan-contino: I _think_ it might have an impact on the URLs used for the rptl header. Need to double-check the behaviour on a staging build before I release!

jekyll-assets/_includes/contribute.html
nathan-contino: Weird. Because in the old system, only pages that have a GitHub link _lack_ the contribute disclaimer.

Perhaps we should just get rid of it entirely. I suspect we should instead toss that text into the README.md or CONTRIBUTING.md for this repo (if we need it at all?)

jekyll-assets/_includes/footer.html
nathan-contino: One is for subtle text. The other is for subtle UI elements :)

I'll unleash a round of comments and CSS organisation today to note some of my trials and errors for the maintainers of the future.

jekyll-assets/_includes/header.html
```diff
@@ -229,15 +229,23 @@
...
-  color: #333;
+  color: var(--textcolor);
   font-size: 1.65em;
   padding: 16px;
 }
 #__rptl-header .__rptl-header-nav-link:hover {
-  color: #fff;
-  background-color: #333;
+  color: var(--near-bg);
+  background-color: var(--less-subtle);
```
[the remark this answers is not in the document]
nathan-contino: `--yet-even-more-subtler`

jekyll-assets/css/style.css
nathan-contino: I'll take a look at the SDK doc today to check, great point. There probably SHOULD be!

jekyll-assets/css/style.css
nathan-contino: Not a stupid question. Actually a stupid implementation from me. Part of my aforementioned consolidation of variables will be a 'common' non-theme-dependent set of vars.

jekyll-assets/_layouts/docs.html
nathan-contino: yo dawg, i heard you like containers, so i put a container in your container

jekyll-assets/scripts/theme.js
nathan-contino: lol nice catch. Will fix.

nathan-contino:
I can respect that we want a consistent header across the sites. We should chat about how the shared header can support dark mode on the docs site -- maybe you can use CSS variables for colours, explicitly set those variables on each of the subdirectories, and docs can handle the overrides (as I do now) for dark and light?

mudge:
I've added you to the repository for the raspberrypi.com header and footer if you want to look at adding CSS variables with fallbacks for colours you want to override.

jekyll-assets/_includes/contribute.html
lurch: :shrug: It's entirely up to you...

jekyll-assets/_includes/footer.html
nathan-contino: Great point. Opened https://github.com/raspberrypi/raspberrypi.com-header-and-footer/pull/81 to address this -- once we've merged that, I'll update this PR to use that new approach (and likely some new header and footer-specific variables).

jekyll-assets/scripts/theme.js
```diff
@@ -0,0 +1,32 @@
...
+  var theme = localStorage.getItem('theme');
+  // if the theme has never been set, or is light, set the theme to the dark symbol in local storage to change it
+  if (localStorage.getItem('theme') !== null && theme == '🌝' ) {
+    localStorage.setItem('theme', '🌚');
+  } else {
+    // otherwise, the theme is currently set to dark, so set the theme to the light symbol in local storage to change it
+    localStorage.setItem('theme', '🌝');
+  }
+  // finally, toggle the light option off or on the body to change the display of the theme
+  document.body.classList.toggle('light');
```
mudge, line 13: Did you try setting `color-scheme: light` (and `color-scheme: dark`) and then relying solely on the `prefers-color-scheme` media query? If that works, I wonder if it’d make it simpler to implement dark mode support in the header and footer without relying on specific class names.

mudge:
Another thing to consider is that any changes to the page markup may require a change to our Algolia DocSearch Crawler which currently uses the following selectors to extract the various "levels" of documentation for search results:

* Level 0: `#tab-menu li.selected`
* Level 1: `#content h1`
* Level 2: `#content h2`
* Level 3: `#content h3`
* Level 4: `#content h4`
* Level 5: `#content h5`
* Level 6: `#content h6`
* Content: `.sect2, .sect3, .sect4, .sect5`

(Note the Pico SDK only uses levels 0-2 from above and uses `code.memname` for content.)

nathan-contino:
Good reminder. I think this should all be OK though -- I have left the `#content` ID in place, and the headers all remain the same as well. Most of this update revolves around changes to elements around the content, instead

jekyll-assets/_layouts/docs.html
```diff
@@ -1,27 +1,187 @@
...
+          {% if subdir.toc[0] %}
+            <ul class="sectlevel1">
+              <input class="toc-toggle-box" type="checkbox" id="{{ subdir.path }}"
+                {% if page.url contains subdir.path %}
+                checked=true
+                {% endif %}
+              />
+              <li>
+                <span>
+                  <label class="toc-toggle" for="{{ subdir.path }}" onclick="event.stopPropagation()">
```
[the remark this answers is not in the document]
nathan-contino: I have played around with this, but the label/checkbox pattern worked better for my purposes. details elements don't allow you to customise transitions.

mudge:
> Good reminder. I think this should all be OK though -- I have left the `#content` ID in place, and the headers all remain the same as well. Most of this update revolves around changes to elements around the content, instead

Aren't the tabs gone? Replaced with a full TOC instead? Perhaps there is an updated selector we can use to inform the crawler which top-level section they're in?

jekyll-assets/scripts/theme.js
nathan-contino: Unfortunately a lot of browsers don't support the prefers-color-scheme query for anti-fingerprinting purposes. As a result, I'm much more comfortable with an explicit toggle persisted via a cookie or local storage. Using that query also has the unfortunate downside of assuming light mode as the default, which often results in that 'light flash' for dark mode users that I refer to in the comments!

nathan-contino:
Hmmm. Might be that I'm misunderstanding how search indexing works with the tabs. I'll take a look.

jekyll-assets/_layouts/docs.html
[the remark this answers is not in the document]
mudge: What transition are you trying to customise? The rotating disclosure arrow?

It looks like you're already rendering the arrow with a `::before` pseudo-element and then animating its transition with a sibling selector on `input:checked`. You should be able to do the same with `<details>` by [removing the default icon](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/details#customizing_the_disclosure_widget) and using `details[open]` to differentiate opened and closed elements.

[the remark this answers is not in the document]
nathan-contino: I'm not sure there's much of an advantage to using details here instead. For better or for worse, the label approach is what I'm more familiar with. Maybe we can address this in a follow-up PR later on, if we realise there's a significant advantage?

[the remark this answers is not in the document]
nathan-contino: Makes perfect sense, thanks for the explanation!

Reader mode seems to ignore the sidebar entirely (fair, since it isn't core content), but at screen widths where the on-this-page element displays, we do wind up with a big list of headers at the end of the reader mode document.

In a future update, I'll explore switching to `details` elements for the sidebar and also look into javascript-free alternatives to the current on-this-page widget. IMO it's OK for such a ancillary element to dispense accessibility, but preferably I'd find a better way that's more compatible with reader mode as well. Ultimately it would be best to eliminate tabs from the landing page and simplify the site rendering process, but I think we're iterating in the right direction for now. Just don't ask me about duplicate header names on the same (long) page -- there doesn't seem to be a great way to handle those right now.

PR #3955, opened 2024-12-08
Update streaming.adoc for RTSP streaming using Raspberry Pi 5 as a server.

GitHubGoody:
The suggested commands do not work on a Raspberry Pi 5.

See the discussion here - https://github.com/raspberrypi/rpicam-apps/issues/747

PR #3960, opened 2024-12-09
Add (required for Pi 5?) libav command to rpicam-vid RTSC streaming instructions

nathan-contino:
Adapted from https://github.com/raspberrypi/documentation/pull/3955

Moved the suggested text into a better position in the instruction flow.


nathan-contino:
@naushir will merge this when you confirm the argument we should pass here. Is h264 truly the best option?

naushir (approved):

naushir:
LGTM

PR #3964, opened 2024-12-09
Revision code bitfield

timg236:


documentation/asciidoc/computers/raspberry-pi/revision-codes.adoc
```diff
@@ -342,6 +342,18 @@ NOQuuuWuFMMMCCCCPPPPTTTTTTTTRRRR
...
+|
+| 18: CM5
+
+|
+|
+| 19: 500
+
+|
+|
+| 1a: CM5 Lite
```
nathan-contino, line 355: I don't think we have any other Lite models (CM4 Lite, CM3 Lite, CM4S Lite, or CM3+ Lite...) in this chart. Can we either add them or omit CM5 Lite?

documentation/asciidoc/computers/raspberry-pi/revision-codes.adoc
```diff
@@ -628,6 +640,48 @@ NOTE: This list is not exhaustive - there may be codes in use that are not in th
...
+| Sony UK
+
+| c041a0
+| CM5 Lite
+| 1.0
+| 4GB
+| Sony UK
+
+| d041a0
+| CM5 Lite
```
nathan-contino, line 680: Same for this chart. Unless the other Lite CMs don't have different revision codes?

nathan-contino (commented):
One question before we can merge.

documentation/asciidoc/computers/raspberry-pi/revision-codes.adoc
timg236: CM5 is different - it has different numbers in OTP for EMMC and LITE. This is because it loads different device-tree files since CM5 EMMC can use the fast HS400 EMMC speed and LITE can't.
It also helps with support requests because we can easily spot the difference in logs / screenshots.

documentation/asciidoc/computers/raspberry-pi/revision-codes.adoc
timg236: Previous CMs had the same revision code for LITE / EMMC

lurch: Correct. Looks like CM5 is the first time that there's been a different revision code for the CM and CM-Lite variants.
Perhaps this also means that CM5 Lite will have its own entry in https://www.raspberrypi.com/documentation/computers/raspberry-pi.html#check-raspberry-pi-model-and-cpu-across-distributions ?

timg236:
I believe CM5 LITE is "Compute Module 5 Lite"

lurch:
> > Perhaps this also means that CM5 Lite will have its own entry in https://www.raspberrypi.com/documentation/computers/raspberry-pi.html#check-raspberry-pi-model-and-cpu-across-distributions ?

> I believe CM5 LITE is "Compute Module 5 Lite"

Yeah, diving into the DTS files, it looks like [bcm2712-rpi-cm5l.dtsi](https://github.com/raspberrypi/linux/blob/rpi-6.6.y/arch/arm64/boot/dts/broadcom/bcm2712-rpi-cm5l.dtsi#L6) sets the `model` string to "Raspberry Pi Compute Module 5 Lite", but it presumably picks up the `compatible` string from [bcm2712-rpi-cm5.dtsi](https://github.com/raspberrypi/linux/blob/rpi-6.6.y/arch/arm64/boot/dts/broadcom/bcm2712-rpi-cm5.dtsi#L51) ; which means that as far as the table at https://www.raspberrypi.com/documentation/computers/raspberry-pi.html#check-raspberry-pi-model-and-cpu-across-distributions is concerned, "Compute Module 5 Lite" displays the same values as "Compute Module 5".
Sorry, I should have thought to check this myself before asking my earlier question!

PR #3988, opened 2024-12-15
Update host-wireless-network.adoc

GuillaumeRossolini:
This change makes it more obvious where the SSID is configured in this one-liner (removed ambiguity with con-name).
Matches the style of the other example a few paragraphs above (see "Enable hotspot"),


nathan-contino (approved):
LGTM. Thanks for this improvement, I agree!

PR #4008, opened 2025-01-20
Slight DeviceTree boot-mode clarification

lurch:
...in relation to https://github.com/raspberrypi/rpi-eeprom/issues/654#issuecomment-2602305569
ping @pelwell


pelwell (approved):
That's not my documentation, but the change looks OK.

PR #4009, opened 2025-01-20
Update install.adoc

lemielek:
as stated here, not using sudo, it is not clear that it should be in context of normal user without elevated privileges. Other responses for problem does not contain this resolution.

### DIFF
--- a/documentation/asciidoc/services/connect/install.adoc
+++ b/documentation/asciidoc/services/connect/install.adoc
@@ -40,7 +40,7 @@ Run the following command on your Raspberry Pi to install Connect Lite:
 $ sudo apt install rpi-connect-lite
 ----
 
-After installation, use the `rpi-connect` command line interface to start Connect for your current user:
+After installation, use the `rpi-connect` command line interface to start Connect for your current user (do not use sudo):
 
 [source,console]
 ----


lemielek:
https://forums.raspberrypi.com/viewtopic.php?p=2219886#p2219886
missing link where it is stated

lurch:
I'm not sure if this change is necessary? If `sudo` **was** needed, then we'd have included that in the example commands; but `sudo` isn't needed, and so we've not included it in the examples.

lemielek:
I think is needed. Little more experienced users does not type 'sudo something' every time but 'sudo su' and then following commands, Here are three of them. Then they are still in su mode trying to do 'rpi-connect on ' and they get 
✗ Cannot start Raspberry Pi Connect: exit status 1
Failed to connect to bus: No medium found

Of course then many of us tries to do 'sudo rpi-connect on' and get the same error. 
And nobody knows why,

It is not clear that sudo is not needed.
That is why many topics in this area on forums. Not easy to find answer I linked yesterday.

So I will even extend 
'(not use sudo)'
to
'(not use sudo, you will get error)'

Or maybe it will be better to patch error message in rpi-connect sources to something like that:

'✗ Cannot start Raspberry Pi Connect: exit status 1
Failed to connect to bus: No medium found
If you use sudo than do not use it'

 but I'am not sure if it is possible and I am not  a developer.

nathan-contino:
I like the idea of patching the error message. I'll raise this with the developers internally. Personally I would rather not warn users about sudo when a lot of people are very much not thinking about sudo for this command -- feels like a bit of a non sequitur. But I can understand the `su` usage pattern... so I'll either try to get this fixed, or add an admonition pushing users in the right direction if they trip this error.

nathan-contino:
Case closed, then! If you run this command as root in 2.3.0, you'll see a helpful error message. No need to confuse most users with extra text in the explanation above.

Thanks for raising this issue, though. If our developers weren't so speedy, we might have needed a docs fix 😆

lurch:
> If our developers weren't so speedy, we might have needed a docs fix 😆

@mudge is the hero here! :muscle:  :clap:

lemielek:
Thank you very much.

PR #4013, opened 2025-01-23
Update revision-codes.adoc

JamesH65:
Add the 1.1 board revisions of the Pi 5.

PR #4015, opened 2025-01-24
Re-order big table in revision-codes.adoc

lurch:
The current order is semi-random because some people have added new Pis by alphabetical revision-code, and some people have added new Pis to the bottom of the list. So to solve any conflicts, make it explicitly newest-Pi-at-bottom.

The revision-codes table is now sorted by Type (from the preceding table), then Revision, then RAM, and finally Code. This is the most likely order-of-manufacture, which means we'll normally just add new revision-codes to the very bottom of the table, without having to worry about re-ordering entries.

See comments in #4013 

Ping @JamesH65 and @nathan-contino  (I won't be upset if you decide to reject this).


JamesH65:
There was never any implicit ordering in the past so I have no preferences one way or another.

nathan-contino (approved):
LGTM, ty

PR #4019, opened 2025-01-31
Explain the new software camera synchronisation feature

davidplowman:
I've also added some other rather overdue updates, notably to the streaming section.

Also a warning banner at the top that this has nothing to do with the legacy stack. I still get folks complaining about this, all these years later...

@naushir You might want to read through the camera sync stuff too!

### DIFF
--- a/documentation/asciidoc/computers/camera/rpicam_apps_multicam.adoc
+++ b/documentation/asciidoc/computers/camera/rpicam_apps_multicam.adoc
@@ -12,7 +12,7 @@ NOTE: `libcamera` does not yet provide stereoscopic camera support. When running
 
 ==== Software Camera Synchronisation
 
-Raspberry Pi's _libcamera_ implementation has the ability to synchronise the frames of different cameras using only software. This will cause one camera to adjust it's frame timing so as to coincide as closely as possible with the frames of another camera. No soldering or hardware connections are required, and it will work with all Raspberry Pi's camera modules, and even third party ones so long as their drivers implement frame duration control correctly.
+Raspberry Pi's _libcamera_ implementation has the ability to synchronise the frames of different cameras using only software. This will cause one camera to adjust it's frame timing so as to coincide as closely as possible with the frames of another camera. No soldering or hardware connections are required, and it will work with all of Raspberry Pi's camera modules, and even third party ones so long as their drivers implement frame duration control correctly.
 
 **How it works**
 
@@ -28,7 +28,7 @@ When cameras are on different devices, the system clocks should be synchronised 
 
 **The Server**
 
-The server, as previously explained, broadcasts timing messages onto the network, by default every second. The server will run for a fixed number of frames, by default 100, after which it will inform the camera application on the device that the "sychronisation point" has been reached. At this moment, the application will start using the frames, so in the case of `rpicam-vid`, they will start being encoded and recorded. Recall that the behaviour and even existence of clients has no bearing on this.
+The server, as previously explained, broadcasts timing messages onto the network, by default every second. The server will run for a fixed number of frames, by default 100, after which it will inform the camera application on the device that the "synchronisation point" has been reached. At this moment, the application will start using the frames, so in the case of `rpicam-vid`, they will start being encoded and recorded. Recall that the behaviour and even existence of clients has no bearing on this.
 
 If required, there can be several servers on the same network so long as they are broadcasting timing messages to different network addresses. Clients, of course, will have to be configured to listen for the correct address.
 
@@ -38,11 +38,11 @@ Clients listen out for server timing messages and, when they receive one, will s
 
 The clients learn the correct "synchronisation point" from the server's messages, and just like the server, will signal the camera application at the same moment that it should start using the frames. So in the case of `rpicam-vid`, this is once again the moment at which frames will start being recorded.
 
-Normally it makes sense to start clients _before_ the server, as the clients will simply wait (the "syncrhonisation point" has not been reached) until a server is seen broadcasting onto the network. This obviously avoids timing problems where a server might reach its "synchronisation point" even before all the clients have been started!
+Normally it makes sense to start clients _before_ the server, as the clients will simply wait (the "synchronisation point" has not been reached) until a server is seen broadcasting onto the network. This obviously avoids timing problems where a server might reach its "synchronisation point" even before all the clients have been started!
 
 **Usage in `rpicam-vid`**
 
-We can use software camera synchronisation with `rpicam-vid` to record videos that are sychronised frame-by-frame. We're going to assume we have two cameras attached, and we're going to use camera 0 as the server, and camera 1 as the client. `rpicam-vid` defaults to a fixed 30 frames per second, which will be fine for us.
+We can use software camera synchronisation with `rpicam-vid` to record videos that are synchronised frame-by-frame. We're going to assume we have two cameras attached, and we're going to use camera 0 as the server, and camera 1 as the client. `rpicam-vid` defaults to a fixed 30 frames per second, which will be fine for us.
 
 First we should start the client:
 [source,console]
@@ -50,7 +50,7 @@ First we should start the client:
 $ rpicam-vid -n -t 20s --camera 1 --codec libav -o client.mp4 --sync client
 ----
 
-Note the `--sync client` parameter. This will record for 20 seconds in total but note that this _includes_ the time to start the server and achieve synchronisation. So while the start of the recordings, and all the frames, will be synchronised, the end of the recordings is not.
+Note the `--sync client` parameter. This will record for 20 seconds but _only_ once the synchronisation point has been reached. If necessary, it will wait indefinitely for the first server message.
 
 To start the server:
 [source,console]
@@ -58,11 +58,11 @@ To start the server:
 $ rpicam-vid -n -t 20s --camera 0 --codec libav -o server.mp4 --sync server
 ----
 
-This will run for 20 seconds but with the default settings (100 frames at 30fps) will give clients just over 3 seconds to get synchronised before anything is recorded. So the final video file will contain slightly under 17 seconds of video.
+This too will run for 20 seconds counting from when the synchronisation point is reached and the recording starts. With the default synchronisation settings (100 frames at 30fps) this means there will be just over 3 seconds for clients to get synchronised.
 
 The server's broadcast address and port, the frequency of the timing messages and the number of frames to wait for clients to synchronise, can all be changed in the camera tuning file. Clients only pay attention to the broadcast address here which should match the server's; the other information will be ignored. Please refer to the https://datasheets.raspberrypi.com/camera/raspberry-pi-camera-guide.pdf[Raspberry Pi Camera tuning guide] for more information.
 
 In practical operation there are a few final points to be aware of:
 
 * The fixed framerate needs to be below the maximum framerate at which the camera can operate (in the camera mode that is being used). This is because the synchronisation algorithm may need to _shorten_ camera frames so that clients can catch up with the server, and this will fail if it is already running as fast as it can.
-* Whilst cameras frames should be correctly synchronised, at higher framerates, or depending on system load, it is possible for frames, either on the clients or server, to be dropped. In these cases the frame timestamps will help an application to work out what has happened, though it's usually easier simply to try and avoid frame drops - perhaps by lowering the framerate, increasing the number of buffers being allocated to the camera queues, or reducing system load (see the xref:camera_software.adoc#buffer-count[`--buffer-count` option].)
+* Whilst camera frames should be correctly synchronised, at higher framerates or depending on system load, it is possible for frames, either on the clients or server, to be dropped. In these cases the frame timestamps will help an application to work out what has happened, though it's usually simpler to try and avoid frame drops - perhaps by lowering the framerate, increasing the number of buffers being allocated to the camera queues (see the xref:camera_software.adoc#buffer-count[`--buffer-count` option]), or reducing system load.
--- a/documentation/asciidoc/computers/camera/rpicam_apps_multicam.adoc
+++ b/documentation/asciidoc/computers/camera/rpicam_apps_multicam.adoc
@@ -8,4 +8,61 @@
 
 To list all the cameras available on your platform, use the xref:camera_software.adoc#list-cameras[`list-cameras`] option. To choose which camera to use, pass the camera index to the xref:camera_software.adoc#camera[`camera`] option.
 
-NOTE: `libcamera` does not yet provide stereoscopic camera support. When running two cameras simultaneously, they must be run in separate processes. This means there is no way to synchronise sensor framing or 3A operation between them.  As a workaround, you could synchronise the cameras through an external sync signal for the HQ (IMX477) camera, and switch the 3A to manual mode if necessary.
+NOTE: `libcamera` does not yet provide stereoscopic camera support. When running two cameras simultaneously, they must be run in separate processes, meaning there is no way to synchronise 3A operation between them.  As a workaround, you could synchronise the cameras through an external sync signal for the HQ (IMX477) camera or use the software camera synchronisation support that is described below, switching the 3A to manual mode if necessary.
+
+==== Software Camera Synchronisation
+
+Raspberry Pi's _libcamera_ implementation has the ability to synchronise the frames of different cameras using only software. This will cause one camera to adjust it's frame timing so as to coincide as closely as possible with the frames of another camera. No soldering or hardware connections are required, and it will work with all Raspberry Pi's camera modules, and even third party ones so long as their drivers implement frame duration control correctly.
+
+**How it works**
+
+The scheme works by designating one camera to be the _server_. The server will broadcast timing messages onto the network at regular intervals, such as once a second. Meanwhile other cameras, known as _clients_, can listen to these messages whereupon they may lengthen or shorten frame times slightly so as to pull them into sync with the server. This process is continual, though after the first adjustment, subsequent adjustments are normally small.
+
+The client cameras may be attached to the same Raspberry Pi device as the server, or they may be attached to different Raspberry Pis on the same network. The camera model on the clients may match the server, or they may be different.
+
+Clients and servers need to be set running at the same nominal framerate (such as 30fps). Note that there is no back-channel from the clients back to the server. It is solely the clients' responsibility to be up and running in time to match the server, and the server is completely unaware whether clients have synchronised successfully, or indeed whether there are any clients at all.
+
+In normal operation, running the same make of camera on the same Raspberry Pi, we would expect the frame start times of the camera images to match within "several tens of microseconds". When the camera models are different this could be significantly larger as the cameras will probably not be able to match framerates exactly and will therefore be continually drifting apart (and brought back together with every timing message).
+
+When cameras are on different devices, the system clocks should be synchronised using NTP (normally the case by default for Raspberry Pi OS), or if this is insufficiently precise, another protocol like PTP might be used. Any discrepancy between system clocks will feed directly into extra error in frame start times - even though the advertised timestamps on the frames will not tell you.
+
+**The Server**
+
+The server, as previously explained, broadcasts timing messages onto the network, by default every second. The server will run for a fixed number of frames, by default 100, after which it will inform the camera application on the device that the "sychronisation point" has been reached. At this moment, the application will start using the frames, so in the case of `rpicam-vid`, they will start being encoded and recorded. Recall that the behaviour and even existence of clients has no bearing on this.
+
+If required, there can be several servers on the same network so long as they are broadcasting timing messages to different network addresses. Clients, of course, will have to be configured to listen for the correct address.
+
+**Clients**
+
+Clients listen out for server timing messages and, when they receive one, will shorten or lengthen a camera frame duration by the required amount so that subsequent frames will start, as far as possible, at the same moment as the server's.
+
+The clients learn the correct "synchronisation point" from the server's messages, and just like the server, will signal the camera application at the same moment that it should start using the frames. So in the case of `rpicam-vid`, this is once again the moment at which frames will start being recorded.
+
+Normally it makes sense to start clients _before_ the server, as the clients will simply wait (the "syncrhonisation point" has not been reached) until a server is seen broadcasting onto the network. This obviously avoids timing problems where a server might reach its "synchronisation point" even before all the clients have been started!
+
+**Usage in `rpicam-vid`**
+
+We can use software camera synchronisation with `rpicam-vid` to record videos that are sychronised frame-by-frame. We're going to assume we have two cameras attached, and we're going to use camera 0 as the server, and camera 1 as the client. `rpicam-vid` defaults to a fixed 30 frames per second, which will be fine for us.
+
+First we should start the client:
+[source,console]
+----
+$ rpicam-vid -n -t 20s --camera 1 --codec libav -o client.mp4 --sync client
+----
+
+Note the `--sync client` parameter. This will record for 20 seconds in total but note that this _includes_ the time to start the server and achieve synchronisation. So while the start of the recordings, and all the frames, will be synchronised, the end of the recordings is not.
+
+To start the server:
+[source,console]
+----
+$ rpicam-vid -n -t 20s --camera 0 --codec libav -o server.mp4 --sync server
+----
+
+This will run for 20 seconds but with the default settings (100 frames at 30fps) will give clients just over 3 seconds to get synchronised before anything is recorded. So the final video file will contain slightly under 17 seconds of video.
+
+The server's broadcast address and port, the frequency of the timing messages and the number of frames to wait for clients to synchronise, can all be changed in the camera tuning file. Clients only pay attention to the broadcast address here which should match the server's; the other information will be ignored. Please refer to the https://datasheets.raspberrypi.com/camera/raspberry-pi-camera-guide.pdf[Raspberry Pi Camera tuning guide] for more information.
+
+In practical operation there are a few final points to be aware of:
+
+* The fixed framerate needs to be below the maximum framerate at which the camera can operate (in the camera mode that is being used). This is because the synchronisation algorithm may need to _shorten_ camera frames so that clients can catch up with the server, and this will fail if it is already running as fast as it can.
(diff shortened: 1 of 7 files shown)


documentation/asciidoc/computers/camera/rpicam_apps_multicam.adoc
```diff
@@ -8,4 +8,61 @@
 
 To list all the cameras available on your platform, use the xref:camera_software.adoc#list-cameras[`list-cameras`] option. To choose which camera to use, pass the camera index to the xref:camera_software.adoc#camera[`camera`] option.
 
-NOTE: `libcamera` does not yet provide stereoscopic camera support. When running two cameras simultaneously, they must be run in separate processes. This means there is no way to synchronise sensor framing or 3A operation between them.  As a workaround, you could synchronise the cameras through an external sync signal for the HQ (IMX477) camera, and switch the 3A to manual mode if necessary.
+NOTE: `libcamera` does not yet provide stereoscopic camera support. When running two cameras simultaneously, they must be run in separate processes. This means there is no way to fully synchronise sensor framing or 3A operation between them.  As a workaround, you could synchronise the cameras through an external sync signal for the HQ (IMX477) camera, and switch the 3A to manual mode if necessary, or you could try the software camera synchronisation support that is described below.
```
naushir, line 11: > no way to fully synchronise sensor framing

Maybe this needs to be omitted from the sentence above?

documentation/asciidoc/computers/camera/rpicam_apps_multicam.adoc
```diff
@@ -8,4 +8,61 @@
...
+
+Raspberry Pi's _libcamera_ implementation has the ability to synchronise the frames of different cameras using only software. This will cause one camera to adjust it's frame timing so as to coincide as closely as possible with the frames of another camera. No soldering or hardware connections are required, and it will work with all Raspberry Pi's camera modules, and even third party ones so long as their drivers implement frame duration control correctly.
+
+**How it works**
+
+The scheme works by designating one camera to be the _server_. The server will broadcast timing messages onto the network at regular intervals, such as once a second. Meanwhile other cameras, known as _clients_, can listen to these messages whereupon they may lengthen or shorten frame times slightly so as to pull them into sync with the server. This process is continual, though after the first adjustment, subsequent adjustments are normally small.
+
+The client cameras may be attached to the same Raspberry Pi device as the server, or they may be attached to different Raspberry Pis on the same network. The model of camera on the clients may match the server, or they may be different.
```
naushir, line 21: Minor nitpick:
s/model of camera/camera model/

documentation/asciidoc/computers/camera/rpicam_apps_multicam.adoc
```diff
@@ -8,4 +8,61 @@
...
+
+**The Server**
+
+The server, as previously explained, broadcasts timing messages onto the network, by default every second. The server will run for a fixed number of frames, by default 100, after which it will inform the camera application on the device that the "sychronisation point" has been reached. At this moment, the application will start using the frames, so in the case of `rpicam-vid`, they will start being encoded and recorded. Recall that the behaviour and even existence of clients has no bearing on this.
+
+If required, there can be several servers on the same network so long as they are broadcasting timing messages to different network addresses. Clients, of course, will have to be configured to listen for the correct address.
+
+**Clients**
+
+Clients listen out for server timing messages and, when they receive one, will shorten or lengthen a camera frame by the required amount so that subsequent frames will start, as far as possible, at the same moment as the server's.
```
naushir, line 37: s/camera frame/camera frame duration/

naushir (approved):
Very minor comments.  Looks good otherwise!

documentation/asciidoc/computers/camera/rpicam_apps_multicam.adoc
davidplowman: Yes, will rephrase slightly!

documentation/asciidoc/computers/camera/rpicam_apps_multicam.adoc
davidplowman: Yep.

documentation/asciidoc/computers/camera/rpicam_apps_multicam.adoc
davidplowman: All done!

naushir:
@nathan-contino are we ok to merge this update?

documentation/asciidoc/computers/camera/rpicam_apps_multicam.adoc
```diff
@@ -8,4 +8,61 @@
...
+NOTE: `libcamera` does not yet provide stereoscopic camera support. When running two cameras simultaneously, they must be run in separate processes, meaning there is no way to synchronise 3A operation between them.  As a workaround, you could synchronise the cameras through an external sync signal for the HQ (IMX477) camera or use the software camera synchronisation support that is described below, switching the 3A to manual mode if necessary.
+
+==== Software Camera Synchronisation
+
+Raspberry Pi's _libcamera_ implementation has the ability to synchronise the frames of different cameras using only software. This will cause one camera to adjust it's frame timing so as to coincide as closely as possible with the frames of another camera. No soldering or hardware connections are required, and it will work with all Raspberry Pi's camera modules, and even third party ones so long as their drivers implement frame duration control correctly.
```
lurch, line 15: minor nitpick: "work with all" -> "work with all of" ?

documentation/asciidoc/computers/camera/rpicam_apps_multicam.adoc
```diff
@@ -8,4 +8,61 @@
...
+
+Note the `--sync client` parameter. This will record for 20 seconds in total but note that this _includes_ the time to start the server and achieve synchronisation. So while the start of the recordings, and all the frames, will be synchronised, the end of the recordings is not.
+
+To start the server:
+[source,console]
+----
+$ rpicam-vid -n -t 20s --camera 0 --codec libav -o server.mp4 --sync server
+----
+
+This will run for 20 seconds but with the default settings (100 frames at 30fps) will give clients just over 3 seconds to get synchronised before anything is recorded. So the final video file will contain slightly under 17 seconds of video.
```
lurch, line 61: "default settings" -> "default synchronisation settings" ?

documentation/asciidoc/computers/camera/rpicam_apps_multicam.adoc
```diff
@@ -8,4 +8,61 @@
...
+
+The server's broadcast address and port, the frequency of the timing messages and the number of frames to wait for clients to synchronise, can all be changed in the camera tuning file. Clients only pay attention to the broadcast address here which should match the server's; the other information will be ignored. Please refer to the https://datasheets.raspberrypi.com/camera/raspberry-pi-camera-guide.pdf[Raspberry Pi Camera tuning guide] for more information.
```
lurch, line 63: I'm sure that this is probably a stupid question, but does it make sense for these settings to be in the camera tuning file, rather than being command-line arguments?

documentation/asciidoc/computers/camera/rpicam_apps_multicam.adoc
```diff
@@ -8,4 +8,61 @@
...
+
+In practical operation there are a few final points to be aware of:
+
+* The fixed framerate needs to be below the maximum framerate at which the camera can operate (in the camera mode that is being used). This is because the synchronisation algorithm may need to _shorten_ camera frames so that clients can catch up with the server, and this will fail if it is already running as fast as it can.
+* Whilst cameras frames should be correctly synchronised, at higher framerates, or depending on system load, it is possible for frames, either on the clients or server, to be dropped. In these cases the frame timestamps will help an application to work out what has happened, though it's usually easier simply to try and avoid frame drops - perhaps by lowering the framerate, increasing the number of buffers being allocated to the camera queues, or reducing system load (see the xref:camera_software.adoc#buffer-count[`--buffer-count` option].)
```
lurch, line 68: Should "cameras frames" be "camera's frames" or "camera frames"?

lurch, line 68: "Whilst cameras frames should be correctly synchronised, at higher framerates, or depending on system load, it is possible for frames, either on the clients or server, to be dropped." - comma overload? :wink: 

Perhaps "Whilst camera frames should be correctly synchronised, at higher framerates or under high system load, it is possible for frames on either the clients or server to be dropped."

lurch, line 68: "usually easier simply to try" -> "usually easier to try" ?

documentation/asciidoc/computers/camera/rpicam_configuration.adoc
```diff
@@ -38,6 +38,8 @@ Raspberry Pi OS recognises the following overlays in `/boot/firmware/config.txt`
 
 To use one of these overlays, you must disable automatic camera detection. To disable automatic detection, set `camera_auto_detect=0` in `/boot/firmware/config.txt`. If `config.txt` already contains a line assigning an `camera_auto_detect` value, change the value to `0`. Reboot your Raspberry Pi with `sudo reboot` to load your changes.
 
+If your Raspberry Pi has two camera connectors (Raspberry Pi 5 or CM4, for example), then you can specify which one you are referring to by adding `,cam0` or `,cam1` (don't add any spaces) to the `dtoverlay` that you used from the table above. If you do not add either of these, it will default to checking  camera connector 1 (`cam1`). But note that for official Raspberry PI camera modules, auto-detection will correctly identify all the cameras connected to your device.
```
lurch, line 41: "PI" -> "Pi"

lurch, line 41: I think that all (most?) models of Compute Module include two camera ports, so I guess you could replace "or CM4" with "or one of the Compute Modules" ? :shrug:

documentation/asciidoc/computers/camera/rpicam_options_common.adoc
```diff
@@ -553,3 +553,11 @@ Flushes output files to disk as soon as a frame finishes writing, instead of wai
 Specifies a JSON file that configures the post-processing applied by the imaging pipeline. This applies to camera images _before_ they reach the application. This works similarly to the legacy `raspicam` "image effects". Accepts a file name path as input.
 
 Post-processing is a large topic and admits the use of third-party software like OpenCV and TensorFlowLite to analyse and manipulate images. For more information, see xref:camera_software.adoc#post-processing-with-rpicam-apps[post-processing].
+
+==== `buffer-count`
+
+The number of buffers to allocate for still image capture or for video recording. The default value of zero lets each application choose a value for itself (1 for still image capture, and 6 for video recording). Increasing the number can sometimes help to reduce the number of frame drops, particularly at higher framerates.
```
lurch, line 559: "choose a value for itself" -> "use it's default value" ? Or perhaps having "default value" twice in the same sentence is too confusing? @nathan-contino

documentation/asciidoc/computers/camera/streaming.adoc
```diff
@@ -1,74 +1,82 @@
...
 
 [source,console]
 ----
-$ vlc udp://@:<port> :demux=h264
+$ ffplay udp://@:<port> -fflags nobuffer -flags low_delay -framedrop
 ----
+As noted previously, `vlc` no longer handles unencapsulated h264 streams.
 
-Alternatively, use the following command on a client to stream using `ffplay`:
+In fact, support for unencapsulated h264 can generally be quite poor so it is often better to send an MPEG-2 Transport Stream instead. Making use of `libav`, this can be accomplished with:
```
lurch, line 22: Should the use of "h264" here actually be "H.264"? Or does it not matter?

documentation/asciidoc/computers/camera/rpicam_apps_multicam.adoc
davidplowman: Yes.

documentation/asciidoc/computers/camera/rpicam_apps_multicam.adoc
```diff
@@ -8,4 +8,61 @@
...
+
+If required, there can be several servers on the same network so long as they are broadcasting timing messages to different network addresses. Clients, of course, will have to be configured to listen for the correct address.
+
+**Clients**
+
+Clients listen out for server timing messages and, when they receive one, will shorten or lengthen a camera frame duration by the required amount so that subsequent frames will start, as far as possible, at the same moment as the server's.
+
+The clients learn the correct "synchronisation point" from the server's messages, and just like the server, will signal the camera application at the same moment that it should start using the frames. So in the case of `rpicam-vid`, this is once again the moment at which frames will start being recorded.
+
+Normally it makes sense to start clients _before_ the server, as the clients will simply wait (the "syncrhonisation point" has not been reached) until a server is seen broadcasting onto the network. This obviously avoids timing problems where a server might reach its "synchronisation point" even before all the clients have been started!
```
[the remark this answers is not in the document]
davidplowman: Thanks!

documentation/asciidoc/computers/camera/rpicam_apps_multicam.adoc
davidplowman: Yep.

documentation/asciidoc/computers/camera/rpicam_apps_multicam.adoc
davidplowman: This kind of thing is always debatable, so it's a fair question. I'm slightly inclined to the view that there's no reason I would ever change this sort of thing once a session is running, so leaving it in the tuning file seems OK. You could, I suppose, argue that someone might prefer to tweak this stuff when they start the camera, and then start everything running. But they have to have some external means of knowing what these numbers should be, so why not put them in the tuning file? Though on the other hand... Argh!

I think I'll probably leave things as they are unless we can think of a clear reason to change things, just because that's less work at this point!

documentation/asciidoc/computers/camera/rpicam_apps_multicam.adoc
davidplowman: Indeed, "camera frames", I think!

davidplowman: Yes, I can de-comma that a bit!

documentation/asciidoc/computers/camera/rpicam_configuration.adoc
davidplowman: Thanks.

davidplowman: Will do.

documentation/asciidoc/computers/camera/rpicam_options_common.adoc
davidplowman: Have gone with "The default value of zero lets each application choose a reasonable number for its own use case..."

documentation/asciidoc/computers/camera/rpicam_vid.adoc
```diff
@@ -11,20 +11,32 @@ For example, the following command writes a ten-second video to a file named `te
...
 
 [source,console]
 ----
-$ vlc test.h264
+$ ffplay test.h264
 ----
 
+[WARNING]
+====
+Older versions of vlc used to play H.264 files correctly, but recent versions do not - displaying only a few, or possibly garbled, frames. You should either use a different media player, or save your files in a more widely supported container format - such as MP4 (see below).
```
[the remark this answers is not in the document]
davidplowman: Will change. Maybe "were able to play".

documentation/asciidoc/computers/camera/streaming.adoc
davidplowman: h264 or H264 or h.264 or H.264? I think folks understand they're all the same, but I think H.264 is probably most common, so will use that.

documentation/asciidoc/computers/camera/streaming.adoc
```diff
@@ -101,56 +108,99 @@ $ rpicam-vid -t 0 --codec libav --libav-format mpegts --libav-audio  -o "udp://<
 
 === GStreamer
 
-https://gstreamer.freedesktop.org/[GStreamer] is a Linux framework for reading, processing and playing multimedia files. This section shows how to use `rpicam-vid` to stream video over a network.
+https://gstreamer.freedesktop.org/[GStreamer] is a Linux framework for reading, processing and playing multimedia files. We can also use it in conjunction with `rpicam-vid` for network streaming.
+
+This setup uses `rpicam-vid` to output an encoded h.264 bitstream to stdout. As we've done previously, we're going to encapsulate this in an MPEG-2 Transport Stream for better downstream compatibility.
```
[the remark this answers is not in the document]
davidplowman: It's not wrong, in that it is outputting an H.264 bitstream, though it's also been encapsulated (as explained in the next sentence). I'll run the two together, maybe:
"This setup uses `rpicam-vid` to output an  H.264 bitstream to stdout, though as we've done previously, we're going to encapsulate it in an MPEG-2 Transport Stream for better downstream compatibility."

davidplowman:
I'm just going to convert this to "draft" for a bit. I'd like to fix the timer issue with rpicam-vid synchronisation (that the timer doesn't count from when the sync happens) first, then I'll update the PR again.

davidplowman:
> I'm just going to convert this to "draft" for a bit. I'd like to fix the timer issue with rpicam-vid synchronisation (that the timer doesn't count from when the sync happens) first, then I'll update the PR again.

All done now.

documentation/asciidoc/computers/camera/rpicam_apps_multicam.adoc
lurch: Fair enough! :slightly_smiling_face:

documentation/asciidoc/computers/camera/rpicam_apps_multicam.adoc
```diff
@@ -50,19 +50,19 @@ First we should start the client:
 $ rpicam-vid -n -t 20s --camera 1 --codec libav -o client.mp4 --sync client
 ----
 
-Note the `--sync client` parameter. This will record for 20 seconds in total but note that this _includes_ the time to start the server and achieve synchronisation. So while the start of the recordings, and all the frames, will be synchronised, the end of the recordings is not.
+Note the `--sync client` parameter. This will record for 20 seconds but _only_ once the sychronisation point has been reached. If necessary, it will wait indefinitely for the first server message.
```
lurch, line 53: "sychronisation" typo

documentation/asciidoc/computers/camera/rpicam_apps_multicam.adoc
```diff
@@ -50,19 +50,19 @@ First we should start the client:
...
-* Whilst cameras frames should be correctly synchronised, at higher framerates, or depending on system load, it is possible for frames, either on the clients or server, to be dropped. In these cases the frame timestamps will help an application to work out what has happened, though it's usually easier simply to try and avoid frame drops - perhaps by lowering the framerate, increasing the number of buffers being allocated to the camera queues, or reducing system load (see the xref:camera_software.adoc#buffer-count[`--buffer-count` option].)
\ No newline at end of file
+* Whilst camera frames should be correctly synchronised, at higher framerates or depending on system load, it is possible for frames, either on the clients or server, to be dropped. In these cases the frame timestamps will help an application to work out what has happened, though it's usually simpler to try and avoid frame drops - perhaps by lowering the framerate, increasing the number of buffers being allocated to the camera queues, or reducing system load (see the xref:camera_software.adoc#buffer-count[`--buffer-count` option].)
```
lurch, line 68: Would it make sense to move the "(see the xref:camera_software.adoc#buffer-count[`--buffer-count` option])" to appear just after "increasing the number of buffers being allocated to the camera queues", rather than after "reducing system load" ?

documentation/asciidoc/computers/camera/rpicam_configuration.adoc
```diff
@@ -38,7 +38,7 @@ Raspberry Pi OS recognises the following overlays in `/boot/firmware/config.txt`
...
 
-If your Raspberry Pi has two camera connectors (Raspberry Pi 5 or CM4, for example), then you can specify which one you are referring to by adding `,cam0` or `,cam1` (don't add any spaces) to the `dtoverlay` that you used from the table above. If you do not add either of these, it will default to checking  camera connector 1 (`cam1`). But note that for official Raspberry PI camera modules, auto-detection will correctly identify all the cameras connected to your device.
+If your Raspberry Pi has two camera connectors (Raspberry Pi 5 or one of the compute modules, for example), then you can specify which one you are referring to by adding `,cam0` or `,cam1` (don't add any spaces) to the `dtoverlay` that you used from the table above. If you do not add either of these, it will default to checking  camera connector 1 (`cam1`). But note that for official Raspberry Pi camera modules, auto-detection will correctly identify all the cameras connected to your device.
```
lurch, line 41: Minor nitpick, but I think we tend to capitalise them as "Compute Modules" rather than "compute modules"?

documentation/asciidoc/computers/camera/rpicam_apps_multicam.adoc
davidplowman: Agree.

documentation/asciidoc/computers/camera/rpicam_configuration.adoc
davidplowman: Done!

documentation/asciidoc/computers/camera/rpicam_apps_multicam.adoc
lurch: Sick! :laughing:

naushir:
@nathan-contino I think this is ready to be merged now.

documentation/asciidoc/computers/camera/rpicam_apps_multicam.adoc
```diff
@@ -8,4 +8,61 @@
...
+
+In practical operation there are a few final points to be aware of:
+
+* The fixed framerate needs to be below the maximum framerate at which the camera can operate (in the camera mode that is being used). This is because the synchronisation algorithm may need to _shorten_ camera frames so that clients can catch up with the server, and this will fail if it is already running as fast as it can.
+* Whilst camera frames should be correctly synchronised, at higher framerates or depending on system load, it is possible for frames, either on the clients or server, to be dropped. In these cases the frame timestamps will help an application to work out what has happened, though it's usually simpler to try and avoid frame drops - perhaps by lowering the framerate, increasing the number of buffers being allocated to the camera queues (see the xref:camera_software.adoc#buffer-count[`--buffer-count` option]), or reducing system load.
```
nathan-contino, line 68:
```suggestion
==== Software Camera Synchronisation

Raspberry Pi's `libcamera` implementation can synchronise the frames of different cameras using only software. Using this synchronisation, one camera adjusts its frame timing to coincide as closely as possible with the frames of another camera. This technique requires no soldering or additional hardware connections, and works with all official Raspberry Pi camera modules as well as third party modules, as long as their drivers correctly implement frame duration control.

The scheme works by designating one camera to be the **server**. The server broadcasts timing messages onto the network at regular intervals (e.g. once a second). Other cameras, known as **clients**, listen to these messages. Clients lengthen or shorten frame times slightly to gradually synchronise with the server.

Clients may be attached to the same Raspberry Pi device as the server, or they may be attached to separate Raspberry Pis on the same network. Clients can use different camera module hardware than the server.

Clients and servers must run at the same nominal framerate (e.g. 30 FPS). Clients do not communicate back to the server; the server is completely unaware whether clients have synchronised successfully, or indeed whether there are any clients at all.

In normal operation, running the same make of camera on the same Raspberry Pi, we would expect the frame start times of the camera images to match within several tens of microseconds. When the camera models differ, this could be significantly larger, since the cameras may not be able to match framerates exactly and will therefore continually drift apart and re-synchronise.

When using cameras connected to separate Raspberry Pis on the same network, the system clocks should be synchronised using NTP. Raspberry Pi uses NTP to set the system time by default. If NTP is insufficiently precise, you could use another protocol, like PTP. Any discrepancy between system clocks feeds directly into extra error in frame start times - even though the advertised timestamps on the frames will not tell you.

===== The Server

By default, the server broadcasts one timing message every second. The server runs for a fixed number of frames, by default 100, after which it informs the camera application that the **synchronisation point** has been reached. Once the server reaches the synchronisation point, the application starts consuming frames (e.g. `rpicam-vid` will start encoding and recording).

If required, you can run multiple servers on the same network as long as they broadcast timing messages to different network addresses. When running multiple servers, you must configure each client to listen to the correct address.

===== Clients

Clients listen for server timing messages. When a client receives a message, the client shortens or lengthens a camera frame duration by the required amount so that subsequent frames start, as close as possible, at the same moment as the subsequent frame on the server.

The clients learn the correct synchronisation point from the server's messages. Just like the server, clients signal the camera application at the same moment that it can consume frames.

For the best results, start clients before the server. Clients will wait until a server broadcasts onto the network. This avoids timing problems where a server might reach its synchronisation point before the clients have even started.

===== Usage in `rpicam-vid`

We can use software camera synchronisation with `rpicam-vid` to record videos that are synchronised frame-by-frame. Consider the following:

* we have two cameras attached
* camera 0 is the server
* camera 1 is the client
* `rpicam-vid` defaults to a fixed 30 frames per second

First, run the following command to start the client:

[source,console]
----
$ rpicam-vid -n -t 20s --camera 1 --codec libav -o client.mp4 --sync client
----

Note the `--sync client` parameter. This records for 20 seconds _once the synchronisation point has been reached_. If necessary, this client will wait indefinitely for the first server message.

To start the server], run the following command:

[source,console]
----
$ rpicam-vid -n -t 20s --camera 0 --codec libav -o server.mp4 --sync server
----

This too will run for 20 seconds _once the synchronisation point has been reached_. The default synchronisation settings (100 frames at 30 FPS) provide just over 3 seconds for clients to synchronise.

You can customise the following settings in the camera tuning file:

* server broadcast address
* server broadcast port
* frequency of the timing messages
* the number of frames to wait for clients to synchronise

Clients only pay attention to the broadcast address specified in the tuning file, which should match the server's. For more information about tuning files, see https://datasheets.raspberrypi.com/camera/raspberry-pi-camera-guide.pdf[Raspberry Pi Camera tuning guide].

When configuring your tuning file, keep the following tips in mind:

* The fixed framerate must not exceed the maximum framerate at which the camera can operate in the camera mode used. The synchronisation algorithm may need to _shorten_ camera frames so that clients can catch up with the server, and this will fail if it is already running as fast as it can.
* Whilst camera frames should be correctly synchronised, at higher framerates or depending on system load, the clients or server could drop frames. In these cases the frame timestamps will help an application to work out what has happened, though it's usually simpler to try and avoid frame drops - perhaps by lowering the framerate, increasing the number of buffers being allocated to the camera queues (see the xref:camera_software.adoc#buffer-count[`--buffer-count` option]), or reducing system load.
```

PR #4021, opened 2025-02-02
Correct explanation of raspi-config do_wifi_ssid_passphrase "plain" option

ninepints:
I was looking through the documentation for something else and noticed that the `raspi-config nonint do_wifi_ssid_passphrase` command has a `plain` option that doesn't really make sense. Digging into the documentation history and `raspi-config` implementation, I think the option was inadvertently mis-documented in commit 8c52b7b (understandably - the documentation prior to that commit was very curt). I am pretty new to Raspberry Pi stuff, so let me know if I missed anything.

### DIFF
--- a/documentation/asciidoc/computers/configuration/raspi-config.adoc
+++ b/documentation/asciidoc/computers/configuration/raspi-config.adoc
@@ -271,7 +271,7 @@ Pass a wireless network name (SSID) and passphrase, if required. The following f
 
 The `<hidden>` option indicates the visibility of the SSID. If the network broadcasts an open SSID, pass `0` or omit the option. If your SSID is hidden, pass `1`. Defaults to `0`.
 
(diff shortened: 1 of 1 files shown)


nathan-contino:
Thanks for this fix. I'll have to give this a test before we can merge anything. The wording changes were actually the result of some testing on my part -- I had trouble connecting to my personal wifi network, whose SSID and password both include special characters. So those changes reflect my own experience trying to connect in an edge case. (for example, `!` is a shortcut for history expansion in bash, so network names that include that character need special handling)

However, I suspect the correct answer is somewhere in between. Let me chew on this a little and see if I can come up with something that blends together your explanation with own personal experiences with special characters.

ninepints:
Sounds good! If it helps, the place where the `plain` flag is used is on line 3151 [here](https://github.com/RPi-Distro/raspi-config/blob/d6b4a8227a3b1a70cdbb24c79e1aff8a1fea562a/raspi-config). It looks like it naively wraps the given value in double-quotes if `plain` is set, so if the value itself contained a double quote that might still cause problems?

nathan-contino (approved):
LGTM after thinking it over. Thanks!

PR #4025, opened 2025-02-07
Explicitly prefix hexadecimal values with 0x in New Style Revision Codes table

petemoore:
This totally threw me for a while, because I on my rpi 400 I was retrieving the value 19 for the "Type" field, which appeared to be for the rpi 500. It was only later I noticed that the field values were in hex (so 19 in the table for the rpi 500 was really 0x19, i.e. 25, and the raspberry pi value listed as 13 was actually 0x13, i.e. 19). Since other people could easily be confused by this, I figured it was better to be explicit and prefix the hex values with `0x` which is reasonably conventional. The way I eventually noticed the numbers in the table weren't decimal was when I noticed a value `f` above. Thanks!


nathan-contino (approved):
lgtm. One of those little usability issues that I love to see reported. Thanks for making our docs better!

PR #4041, opened 2025-02-28
Update boot order config process

aaronaxvig:
The tool calls itself the Raspberry Pi Software Configuration Tool so I updated that.

The menu option is "Boot Order", not "Bootloader Order".

It appears that after choosing a boot order it immediately drops back to the CLI, writes the bootloader changes (effectively the same as running `sudo rpi-eeprom-update -a`?), and returns back into the Config Tool.  It appears that the manual bootloader write is not necessary.

It is not strictly necessary to boot from NVMe first.  It seems reasonable to me that the average user will know to remove the SD card if they choose the first option: SD Card and then NVMe and then USB.


nathan-contino (approved):
lgtm

timg236:
Sorry, missed the notification. Change looks good.

PR #4059, opened 2025-03-27
Only respect system colour schemes and add spacing to header

mudge:
Screenshot 2025-03-27 at 15 44 00

*   Rather than squashing the "Raspberry Pi Documentation" header, search bar and tab menu altogether, add some spacing between each element to be more consistent with our other sites.

*   Rather than implementing a JavaScript-based override on top of the default user preferences for colour schemes, rely entirely on the user preferences reported by the browser. This way, we can implement dark mode support in pure CSS and remove the dependency on JavaScript (without which, the site mistakenly would render in dark mode by default).


lurch:
Perhaps some users liked manually clicking on the moon icon and flipping it into dark mode? :shrug:  At the very least, I guess it offers a good visual hint that light/dark modes are available? (although I've not browsed enough other sites to know what "common practice" is here)

> (without which, the site mistakenly would render in dark mode by default).

I asked Nate about that once, and IIRC he said that it's because a light-mode user temporarily seeing a dark-page, is much less annoying than a dark-mode user temporarily seeing a light-page.

mudge:
> Perhaps some users liked manually clicking on the moon icon and flipping it into dark mode? 🤷 At the very least, I guess it offers a good visual hint that light/dark modes are available? (although I've not browsed enough other sites to know what "common practice" is here)
Ultimately, this is a feature removal (the ability to toggle a light and dark "theme" above and beyond the user's browser/system preferences) so is going to frustrate users that relied on the toggle but is part of a bigger piece of work to simplify the documentation site to make it easier to maintain. My real aim is to warm up to combatting an issue @JamesH65 reported on the [Hardware page](https://www.raspberrypi.com/documentation/pico-sdk/hardware.html) which currently has over 50,000 DOM elements and is grinding some users' browsers to a halt.

> I asked Nate about that once, and IIRC he said that it's because a light-mode user temporarily seeing a dark-page, is much less annoying than a dark-mode user temporarily seeing a light-page.
This way, there will be no flash as the browser will render with the appropriate colour scheme on initial page load (there's no JavaScript logic to change colours after initial rendering).

lurch:
> part of a bigger piece of work to simplify the documentation site to make it easier to maintain.

:+1: :+1:

Issue #4063, opened 2025-03-31
Excessive number of DOM elements in table of contents


mudge:
@JamesH65 flagged that the [Hardware page](https://www.raspberrypi.com/documentation/pico-sdk/hardware.html) is extremely slow in his browser even going so far as to crash it.

Looking at the structure of the page, it has over 50,000 DOM elements (for reference: Google Lighthouse will error if you have more than 1,400 elements on a page). While we'd expect a lot of elements as the Hardware section of the SDK is very long, looking at the table of contents alone reveals 12,858 elements and this appears on every single page. Digging a bit more, the table of contents seems to have 1,242 entries when fully expanded so that's the absolute minimum number of elements we would need to render it.

We can look at simplifying the markup for the table of contents but can we actually reduce the depth the table of contents goes to, e.g. to top-level sections only (e.g. "Getting started with your Raspberry Pi") rather than sub-sections (e.g. "Getting started with your Raspberry Pi - Power supply")? The theory being that would drastically cut down the size of the table of contents without also rewriting its markup.

Wren6991:
It runs fine for me on Firefox on my laptop, but really it needs to be readable on a Pi 3 etc.

I agree this doesn't need to be a single flat page, and we could split the hierarchy across multiple pages to reduce the number of active DOM elements.

Wren6991:
Is there a relationship between the displayed TOC depth and keys reachable from the search box?

mudge:
Looking at a built `build/jekyll/_data/nav.json` (including the Pico SDK), including only top-level titles and their sections would reduce the list to 359 entries:

* Computers
    * Getting started
        * Getting started with your Raspberry Pi
        * Install an operating system
        * Set up your Raspberry Pi
        * Configuration on first boot
        * Next steps
    * Raspberry Pi OS
        * Introduction
        * Update software
        * Play audio and video
        * Utilities
        * Accessibility options
        * Use Python on a Raspberry Pi
        * Use GPIO from Python
    * Configuration
        * `raspi-config`
        * non-interactive `raspi-config`
        * Displays
        * Audio
        * Networking
        * Screen blanking
        * Users
        * External storage
        * Kernel command line (`cmdline.txt`)
        * Localise your Raspberry Pi
        * Secure your Raspberry Pi
        * Set up a headless Raspberry Pi
        * Host a wireless network from your Raspberry Pi
        * Use a proxy server
        * `boot` folder contents
        * LED warning flash codes
        * Configure UARTs
[...]

PR #4065, opened 2025-04-01
Don't include subsections in site-wide table of contents

mudge:
GitHub: https://github.com/raspberrypi/documentation/issues/4063

To reduce the number of elements included on every single page of the documentation site, no longer include subsections of every section (e.g. terminate at Pico C SDK > Hardware APIs > hardware_adc, rather than also including the subsections of "Detailed Description", "Functions", etc.) as these are available in the right-hand "On this page" TOC shown on wide-enough screens.

Note we're not yet cutting down the amount of markup to render the TOC but starting by cutting out the source data.

This removes the JavaScript logic to render the currently clicked section in bold.


clowder (approved):
Seems like a quick win. Out of curiosity, how many nodes does this trim?

mudge:
> Seems like a quick win. Out of curiosity, how many nodes does this trim?

In development, it comes down from 12,878 elements under `<div id="toc">` to 3,339.

lurch:
> This removes the JavaScript logic to render the currently clicked section in bold.

I was about to reply "That seems like a shame", but then I noticed that the RHS menu **does** highlight the currently-being-viewed section of the page in bold :+1:  E.g. 
Screenshot from 2025-04-02 11-12-51

So I just need to train my eyes to look at the RHS menu instead of the LHS menu... :joy:

mudge:
I thought it was safe to remove as it is a little misleading: it highlights the anchor on the page you clicked but as soon as you scroll, it is no longer accurate. The "On this page" section doesn't have this problem as it live highlights and expands the page structure as you scroll.

PR #4067, opened 2025-04-02
Upgrade to Tocbot 4.35.3 and simplify setup

mudge:
Previously, the "On this page" table of contents would contain a server-side rendered table of contents generated by the Jekyll AsciiDoc plugin but this was always set to be hidden (even to screen readers). We would then replace this element with the client-side rendered table of contents generated by Tocbot. The latter is only visible on screens wider than 1460px so there is no need for us to have logic to refresh the Tocbot whenever the window is resized under 768px (as it will be hidden anyway).


lurch:
For bears with little brains: is the TLDR here that there's now less processing being done by the client-side javascript?

mudge:
There are two savings here:

1. We no longer serve a server-side rendered but always hidden version of the "On this page" table of contents
2. We no longer refresh the "On this page" table of contents on every resize of the browser window

mudge:
> When you say "server-side rendered" do you mean "build-time created"?
Yes: here the "server" doing the rendering of static HTML is the GitHub Actions runner.

mbaird (approved):

PR #4071, opened 2025-04-14
Update overclocking.adoc

JamesH65:
Add core_freq_fixed documentation.

### DIFF
--- a/documentation/asciidoc/computers/config_txt/overclocking.adoc
+++ b/documentation/asciidoc/computers/config_txt/overclocking.adoc
@@ -1,4 +1,4 @@
-== Overclocking options
+== Clocking options
 
 The kernel has a https://www.kernel.org/doc/html/latest/admin-guide/pm/cpufreq.html[CPUFreq] driver with the powersave governor enabled by default, switched to ondemand during boot, when xref:configuration.adoc#raspi-config[raspi-config] is installed. With the ondemand governor, CPU frequency will vary with processor load. You can adjust the minimum values with the `*_min` config options, or disable dynamic clocking by applying a static scaling governor (powersave or performance) or with `force_turbo=1`.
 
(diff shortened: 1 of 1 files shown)


JamesH65:
@lurch Not sure who to get approvals from nowadays!

documentation/asciidoc/computers/config_txt/overclocking.adoc
```diff
@@ -1,4 +1,4 @@
-== Overclocking options
```
lurch, line 1: This header shouldn't be renamed, as there are other places in the documentation linking to https://www.raspberrypi.com/documentation/computers/config_txt.html#overclocking-options

documentation/asciidoc/computers/config_txt/overclocking.adoc
```diff
@@ -101,6 +101,10 @@ WARNING: Setting any overclocking parameters to values other than those used by
 
 | temp_soft_limit
 | *3A+/3B+ only*. CPU speed throttle control. This sets the temperature at which the CPU clock speed throttling system activates. At this temperature, the clock speed is reduced from 1400MHz to 1200MHz.  Defaults to `60`, can be raised to a maximum of `70`, but this may cause instability.
+
+| core_freq_fixed
+| Setting to 1 disables active scaling of the core clock frequency and ensures that any peripherals that use the core clock will maintain a consistent speed. The fixed clock speed is the higher/turbo frequency for the platform in use. Use this in preference to setting specific core_clock frequencies as it provides portability of config files between platforms.
```
lurch, line 106: Is this the same thing as `force_turbo=1` which is mentioned in the initial paragraph of this page?

documentation/asciidoc/computers/config_txt/overclocking.adoc
lurch: (and this is causing the CI to fail)

JamesH65: Bugger, it's the wrong title for the section really, thoguht I'd try and sneak it in!

documentation/asciidoc/computers/config_txt/overclocking.adoc
lurch: Okay, so does the "or disable dynamic clocking by applying a static scaling governor (powersave or performance) or with `force_turbo=1`." wording near the start of this page need to be updated to mention `core_freq_fixed=1` instead?

JamesH65: More of a @pelwell question I think. My understanding is that force_turbo affects the ARM clock as well, whilst this option just locks the GPU core clock and leaves the ARM to be scaled as necessary.

pelwell: Correct - I don't think any further changes are required.

lurch (approved):

PR #4078, opened 2025-04-22
Make the Connect upgrade caveat more specific

clowder:
This issue only effects the remote shell sessions, because they're subprocesses of `rpi-connectd` and terminated when it's restarted. Also it's possible to work around the issue using `screen`.

### DIFF
--- a/documentation/asciidoc/services/connect/troubleshooting.adoc
+++ b/documentation/asciidoc/services/connect/troubleshooting.adoc
@@ -10,7 +10,7 @@
 
 * Browsers intercept certain keys and key combinations. As a result, you can't type these keys into your Connect window. Screen sharing includes a toolbar to simulate some of the most popular intercepted keys.
 
-* Upgrading `rpi-connect` and `rpi-connect-lite` using Connect is not supported. The upgrade process will terminate all remote shell sessions and drop all connections.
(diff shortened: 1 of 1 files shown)


documentation/asciidoc/services/connect/troubleshooting.adoc
```diff
@@ -10,7 +10,7 @@
 
 * Browsers intercept certain keys and key combinations. As a result, you can't type these keys into your Connect window. Screen sharing includes a toolbar to simulate some of the most popular intercepted keys.
 
-* Upgrading `rpi-connect` and `rpi-connect-lite` using Connect is not supported. The upgrade process will terminate all remote shell sessions and drop all connections.
+* Upgrading `rpi-connect` and `rpi-connect-lite` using Connect's remote shell is not supported. The upgrade process will terminate all remote shell sessions and drop all connections. To upgrade in a remote shell session the upgrade process must be launched using a tool like `screen` to ensure the process continues uninterrupted after the session is terminated.
```
lurch, line 13: "To upgrade in a" -> "To upgrade Connect in a" ?

"after the session is terminated." -> "after the parent session is terminated." ?

mudge, line 13: Given @lurch's feedback and the fact we mention both screen and tmux elsewhere in the docs, how about:

```suggestion
* Upgrading `rpi-connect` and `rpi-connect-lite` using Connect's remote shell is not supported. The upgrade process will terminate all remote shell sessions and drop all connections. To upgrade Connect in a remote shell session, use a tool like `screen` or `tmux` to ensure the process continues uninterrupted after your connection is closed.
```

clowder: Seems good to me. Happy with this suggestion @lurch?

lurch (approved):

PR #4089, opened 2025-04-30
Updates for synchronous and externally triggered cameras

njhollinghurst:
Update synchronized cameras documentation to:

- Deprecate the XVS sync method for GS Cam, in favour of XTRIG

- Recommend "always-on" property instead of using pull-ups

- Recommend "trigger-mode" for IMX477 be set using the new OF property (better for source and sink on same Pi) while noting that the module parameter still works and that "trigger-mode" means something quite different for IMX296!

- Document the need to edit the pipeline config file for timeout

- Some attempt to tidy the structure while covering all the options

- Remove $ prompts from scripts/commands, so they can be copy-pasted

### DIFF
--- a/documentation/asciidoc/accessories/camera/external_trigger.adoc
+++ b/documentation/asciidoc/accessories/camera/external_trigger.adoc
@@ -21,17 +21,6 @@ We can use a Raspberry Pi Pico to provide the trigger. Connect any Pico GPIO pin
 
 image::images/pico_wiring.jpg[alt="Image showing Raspberry Pi Pico wiring",width="50%"]
 
-==== Boot up the Raspberry Pi with the camera connected.
-
-Enable external triggering through superuser mode:
-
-[source,console]
-----
-$ sudo su
-$ echo 1 > /sys/module/imx296/parameters/trigger_mode
-$ exit
-----
-
 ==== Raspberry Pi Pico MicroPython Code
 
 [source,python]
@@ -55,15 +44,37 @@ The low pulse width is equal to the shutter time, and the frequency of the PWM e
 
 NOTE: In this example, Pin 28 connects to the XTR touchpoint on the GS camera board.
 
-=== Operation
+=== Camera driver configuration
 
-Run the code on the Pico, and set the camera running:
+This step is only necessary if you have more than one camera with XTR wired in parallel.
+
+Edit `/boot/firmware/config.txt`. Change `camera_auto_detect=1` to `camera_auto_detect=0`.
+
+Append this line:
+[source]
+----
+dtoverlay=imx296,always-on
+----
+When using the CAM0 port on a Raspberry Pi 5, CM4 or CM5, append `,cam0` to that line without a space. If both cameras are on the same Raspberry Pi you will need two dtoverlay lines, only one of them ending with `,cam0`.
+
+If the external trigger will not be started right away, you also need to increase the libcamera timeout xref:camera.adoc#libcamera-configuration[as above].
+
+=== Starting the camera
+
+Enable external triggering:
+
+[source,console]
+----
+$ echo 1 | sudo tee /sys/module/imx296/parameters/trigger_mode
+----
+
+Run the code on the Pico, then set the camera running:
 
 [source,console]
 ----
 $ rpicam-hello -t 0 --qt-preview --shutter 3000
 ----
 
-Every time that the Pico pulses the pin, it should generate a frame. To control the framerate, vary the duration between pulses.
+Every time the Pico pulses the pin, it should capture a frame. However, if `--gain` and `--awbgains` are not set, some frames will be dropped to allow AGC and AWB algorithms to settle.
 
-NOTE: When running `rpicam-apps`, always specify a fixed shutter duration to ensure the AGC does not adjust the camera's shutter speed. The duration does not matter, since it is actually controlled by the external trigger pulse.
+NOTE: When running `rpicam-apps`, always specify a fixed shutter duration, to ensure the AGC does not try to adjust the camera's shutter speed. The value is not important, since it is actually controlled by the external trigger pulse.
--- a/documentation/asciidoc/accessories/camera/synchronous_cameras.adoc
+++ b/documentation/asciidoc/accessories/camera/synchronous_cameras.adoc
@@ -1,102 +1,108 @@
 == Synchronous Captures
 
-Both the HQ Camera and the Global Shutter Camera, have support for synchronous captures.
-Making use of the XVS pin (Vertical Sync) allows one camera to pulse when a frame capture is initiated.
-The other camera can then listen for this sync pulse, and capture a frame at the same time as the other camera.
+The High Quality (HQ) Camera supports synchronous captures.
+One camera (the "source") can be configured to generate a pulse on its XVS (Vertical Sync) pin when a frame capture is initiated.
+Other ("sink") cameras can listen for this pulse, and capture a frame at the same time as the source camera.
 
-=== Using the HQ Camera
+This method is largely superseded by xref:../computers/camera_software.adoc#software-camera-synchronisation[software camera synchronisation] which can operate over long distances without additional wires and has sub-millisecond accuracy. But when cameras are physically close, wired synchronisation may be used.
 
-For correct operation, both cameras require a 1.65V pull up voltage on the XVS line, which is created by a potential divider through the 3.3V and GND pins on the Raspberry Pi.
+NOTE: Global Shutter (GS) Cameras can also be operated in a synchronous mode. However, the source camera will record one extra frame. Instead, for GS Cameras we recommend using an xref:camera.adoc#external-trigger-on-the-gs-camera[external trigger source]. You cannot synchronise a GS Camera and an HQ Camera.
 
-image::images/synchronous_camera_wiring.jpg[alt="Image showing potential divider setup",width="50%"]
+=== Connecting the cameras
 
-Create a potential divider from two 10kΩ resistors to 3.3V and ground (to make 1.65V with an effective source impedance of 5kΩ). This can be connected to either Raspberry Pi.
+Solder a wire to the XVS test point of each camera, and connect them together.
 
-Solder the GND and XVS test points of each HQ Camera board to each other.
+Solder a wire to the GND test point of each camera, and connect them together.
 
-Connect the XVS wires to the 1.65V potential divider pull-up.
+*For GS Cameras only,* you will also need to connect the XHS (Horizontal Sync) test point of each camera together. On any GS Camera that you wish to act as a sink, bridge the two halves of the MAS pad with solder.
 
-==== Boot up both Raspberry Pis
+NOTE: An earlier version of this document recommended an external pull-up for XVS. This is no longer recommended. Instead, ensure you have the latest version of Raspberry Pi OS and set the `always-on` property for all connected cameras.
 
-The file `/sys/module/imx477/parameters/trigger_mode` determines which board outputs pulses, or waits to receive pulses (source and sink).
-This parameter can only be altered in superuser mode.
+=== Driver configuration
 
-Run the following commands to configure the sink:
+You will need to configure the camera drivers to keep their 1.8V power supplies on when not streaming, and optionally to select the source and sink roles.
+
+==== For the HQ Camera
+
+Edit `/boot/firmware/config.txt`. Change `camera_auto_detect=1` to `camera_auto_detect=0`.
+
+Append this line for a source camera:
+[source]
+----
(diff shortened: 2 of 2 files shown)


lurch:
> * Remove $ prompts from scripts/commands, so they can be copy-pasted

If you use the `[source,console]` above a code-block, then the `$` is automatically non-copyable, see e.g. https://www.raspberrypi.com/documentation/computers/camera_software.html#rpicam-jpeg

njhollinghurst:
It's a bit messy, but hard to make it simpler without omitting required information (indeed it _does_ omit repeating the  "timeout-defeat" instructions for GS external trigger, but presumably people will infer that it needs to be done as in the previous section).

I feel this stuff would be better as a separate Application Note rather than part of the generic camera documentation?

Conspicuous by is absence is any mention of the new S/W synchronization method!

documentation/asciidoc/accessories/camera/external_trigger.adoc
```diff
@@ -55,15 +44,35 @@ The low pulse width is equal to the shutter time, and the frequency of the PWM e
 
 NOTE: In this example, Pin 28 connects to the XTR touchpoint on the GS camera board.
 
-=== Operation
+=== Camera driver configuration
+
+This step is only necessary if you have more than one camera with XTR wired in parallel.
+
+As the superuser, edit `/boot/firmware/config.txt`. Change `camera_auto_detect=1` to `camera_auto_detect=0`.
```
lurch, line 51: We generally don't use phrasing like "As the superuser..." - I think I'd probably just go with "Edit `/boot/firmware/config.txt` and change..."
You could link to https://www.raspberrypi.com/documentation/computers/config_txt.html#common-hardware-configuration-options if you really want to.

documentation/asciidoc/accessories/camera/external_trigger.adoc
```diff
@@ -55,15 +44,35 @@ The low pulse width is equal to the shutter time, and the frequency of the PWM e
...
+For either a source or a sink, append this line:
+[source]
+----
+dtoverlay=imx296,always-on
+----
+When using the CAM0 port on a Raspberry Pi 5, CM4 or CM5, append `,cam0` to that line without a space. If both cameras are on the same Raspberry Pi you will need two dtoverlay lines, only one of them with `cam0`.
+
+=== Starting the camera
+
+Enable external triggering through superuser mode:
```
lurch, line 62: I think I'd shorten this to just "Enable external triggering:"

documentation/asciidoc/accessories/camera/synchronous_cameras.adoc
```diff
@@ -1,102 +1,97 @@
 == Synchronous Captures
 
-Both the HQ Camera and the Global Shutter Camera, have support for synchronous captures.
-Making use of the XVS pin (Vertical Sync) allows one camera to pulse when a frame capture is initiated.
-The other camera can then listen for this sync pulse, and capture a frame at the same time as the other camera.
+The High Quality (HQ) Camera supports synchronous captures.
+Using the XVS pin (Vertical Sync) allows one camera (the "source") to pulse when a frame capture is initiated.
```
lurch, line 4: "to pulse" -> "to send a pulse" ?

documentation/asciidoc/accessories/camera/external_trigger.adoc
njhollinghurst: I'll remove it. I suppose it was a sort of echo of a previous use of that [word](https://github.com/raspberrypi/documentation/blob/07dc7fa977899f14f9706916669acd96016c556d/documentation/asciidoc/accessories/camera/synchronous_cameras.adoc?plain=1#L22).

documentation/asciidoc/accessories/camera/synchronous_cameras.adoc
```diff
@@ -1,102 +1,97 @@
...
+Other ("sink") cameras can listen for this pulse, and capture a frame at the same time as the source camera.
 
-=== Using the HQ Camera
+[NOTE]
+====
+Global Shutter (GS) Cameras can also be operated in a synchronous mode. However, the source camera will record one extra frame. Instead, for the GS Camera we recommend using an xref:camera.adoc#external-trigger-on-the-gs-camera[external trigger source]. You cannot synchronise a GS Camera to an HQ Camera.
```
lurch, line 9: People might misinterpret "You cannot synchronise a GS Camera to an HQ Camera." to mean that you _can_ synchronise an HQ Camera to a GS Camera"? :wink:  Perhaps changing "to" to "with" is safer?

documentation/asciidoc/accessories/camera/synchronous_cameras.adoc
njhollinghurst: Yes. And I probably should re-check the datasheet and whether or not we're using "fast trigger mode" or whatever it is... to check that the final part of the sentence is accurate too.

documentation/asciidoc/accessories/camera/synchronous_cameras.adoc
```diff
@@ -1,102 +1,97 @@
...
+=== Connecting the cameras
 
-image::images/synchronous_camera_wiring.jpg[alt="Image showing potential divider setup",width="50%"]
+Solder a wire to the XVS test point of each camera, and connect them together.
 
-Create a potential divider from two 10kΩ resistors to 3.3V and ground (to make 1.65V with an effective source impedance of 5kΩ). This can be connected to either Raspberry Pi.
+Solder a wire to the GND test point of each camera, and connect them together.
 
-Solder the GND and XVS test points of each HQ Camera board to each other.
+*For GS Cameras only,* you will also need to connect the XHS (Horizontal Sync) test point of each camera. On any GS Camera that you wish to act as a sink, bridge the two halves of the MAS pad with solder.
```
lurch, line 18: "test point of each camera." -> "test point of each camera together." ?

documentation/asciidoc/accessories/camera/synchronous_cameras.adoc
njhollinghurst: "and"

documentation/asciidoc/accessories/camera/synchronous_cameras.adoc
```diff
@@ -1,102 +1,97 @@
...
 
-The file `/sys/module/imx477/parameters/trigger_mode` determines which board outputs pulses, or waits to receive pulses (source and sink).
-This parameter can only be altered in superuser mode.
+You will need to configure the camera drivers to keep their 1.8V power supplies on when not streaming, and optionally to select the source and sink roles.
 
-Run the following commands to configure the sink:
+==== For the HQ Camera
 
-[source,console]
+As the superuser, edit `/boot/firmware/config.txt`. Change `camera_auto_detect=1` to `camera_auto_detect=0`.
```
lurch, line 28: See earlier comment...

documentation/asciidoc/accessories/camera/synchronous_cameras.adoc
```diff
@@ -1,102 +1,97 @@
...
+Frames should be synchronised. Use `--frames` to ensure the same number of frames are captured, and that the recordings are exactly the same length.
+Running the sink first ensures that no frames are missed.
 
-Because the sink and source record a different number of frames, use `ffmpeg` to resync the videos. By dropping the first frame from the source, we then get two recordings with the same starting point and frame length:
-
-[source,console]
-----
-$ ffmpeg -i source.h264 -vf select="gte(n\, 1)" source.h264
-----
+NOTE: When using the GS camera in synchronous mode, the sink will not record exactly the same number of frames as the source. **The source records one extra frame before the sink starts recording**. Because of this, you need to specify that the sink records one less frame with the `--frames` option. You could use `ffmpeg` to delete the first frame from the source camera's video recording.
```
lurch, line 97: "You could use" -> "Alternatively you could use"

njhollinghurst:
Ah, I've belatedly noticed that we do already document [Software Camera Synchronisation](https://www.raspberrypi.com/documentation/computers/camera_software.html#software-camera-synchronisation), although it's well hidden under [rpicam-apps](https://www.raspberrypi.com/documentation/computers/camera_software.html#software-camera-synchronisation).

I should definitely link there from here.

documentation/asciidoc/accessories/camera/synchronous_cameras.adoc
njhollinghurst: Not only but also!

I need to test this again because getting it wrong leads to a (soft) lockup if the sink doesn't get the right number of syncs.

documentation/asciidoc/accessories/camera/external_trigger.adoc
njhollinghurst: Inherited from the original at L26, but can be shortened.

But now I'm wondering if I strictly need to edit  `external_trigger.adoc`?  The original was nice and succinct. The "always-on" parameter is perhaps not essential in this case because the driver is quite low impedance.

documentation/asciidoc/accessories/camera/synchronous_cameras.adoc
lurch: > Not alternatively, but as well as!

Ahhh, in which case perhaps the "could" ought to be "should then" ? (and perhaps worth bringing back the example ffmpeg command line)

njhollinghurst: Yes. I was trying to corral the GS-Cam-specific bits into NOTEs, but the ffmpeg command can't really go inside one of those, so might have to re-jig the structure.

It would be nice if we could obliterate GS Camera from this part, but somebody is bound to be using it already...

njhollinghurst: Well it would be nice to steer GS Cam users towards the alternative "external trigger" method (which requires fewer wires, is software-configurable, and doesn't have the asymmetry / off-by-one-frame issues)

Sadly, now that we've documented this method we can't really un-document it.

lurch: Ahhh, now I see what you're saying :slightly_smiling_face:  Yeah, I guess all we can do is strongly-discourage it.

documentation/asciidoc/accessories/camera/synchronous_cameras.adoc
```diff
@@ -1,102 +1,97 @@
...
-[source,console]
+As the superuser, edit `/boot/firmware/config.txt`. Change `camera_auto_detect=1` to `camera_auto_detect=0`.
+
+Append this line for a source camera:
+[source]
 ----
-$ sudo su
-$ echo 2 > /sys/module/imx477/parameters/trigger_mode
-$ exit
+dtoverlay=imx477,always-on,trigger-mode=1
```
njhollinghurst, line 33: Whoops. I've just spotted this is an error. Although the OF property and module parameter are called `trigger-mode`, the dtparam is called `sync-source` or `sync-sink`!

Will fix in next revision.

documentation/asciidoc/accessories/camera/synchronous_cameras.adoc
```diff
@@ -1,102 +1,97 @@
...
 
-Create a potential divider from two 10kΩ resistors to 3.3V and ground (to make 1.65V with an effective source impedance of 5kΩ). This can be connected to either Raspberry Pi.
+On the GS Camera, the sink role is enabled by the MAS pin and cannot be configured by software ("trigger_mode" has a different use, relating to the xref:camera.adoc#external-trigger-on-the-gs-camera[external trigger method], and should _not_ be set for this method).
 
-Solder 2 wires to the XVS test points on each board and connect both of these wires together to the 1.65V potential divider.
+=== Libcamera configuration
 
-Solder the GND of each Camera board to each other. Also solder 2 wires to the XHS test points on each board and connect these. No pullup is needed for XHS pin.
+If the cameras are not all started within 1 second, the `rpicam-vid` application can time out. To prevent this, you must edit a configuration file on any Raspberry Pi(s) with sink cameras. On Raspberry Pi 5 or CM5, make a copy of the file `/usr/share/libcamera/pipeline/rpi/pisp/example.yaml`. On other Raspberry Pi models, make a copy of the file `/usr/share/libcamera/pipeline/rpi/vc4/rpi_apps.yaml`.
```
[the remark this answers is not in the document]
njhollinghurst: Updated - is that better?

documentation/asciidoc/accessories/camera/synchronous_cameras.adoc
njhollinghurst: I've tentatively removed the `ffmpeg` bit...

naushir (approved):
LGTM

PR #4091, opened 2025-05-01
docs: include info on installing network-manager package

gh0st:
After doing a regular `sudo apt update && sudo apt upgrade -y` I couldn't `sudo nmcli ...` so if another user runs into the same issue I put a little note in here on how to continue forward.

### DIFF
--- a/documentation/asciidoc/computers/configuration/host-wireless-network.adoc
+++ b/documentation/asciidoc/computers/configuration/host-wireless-network.adoc
@@ -22,6 +22,8 @@ To create a hosted wireless network on the command line, run the following comma
 $ sudo nmcli device wifi hotspot ssid <example-network-name> password <example-password>
 ----
 
+> Note: If the `sudo nmcli` command fails for you, you must install the network-manager package with `sudo apt install network-manager`.
+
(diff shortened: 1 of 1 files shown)


lurch:
Huh, this should be installed by default?! Which OS (and version) are you using @gh0st ?
Note that the docs at https://www.raspberrypi.com/documentation/ specifically only cover the Bookworm version of Raspberry Pi OS, as downloaded from https://www.raspberrypi.com/software/operating-systems/ (or installed via Raspberry Pi Imager).

gh0st:
```sh
$ cat /etc/os-release 
PRETTY_NAME="Debian GNU/Linux 11 (bullseye)"
NAME="Debian GNU/Linux"
VERSION_ID="11"
VERSION="11 (bullseye)"
VERSION_CODENAME=bullseye
ID=debian
HOME_URL="https://www.debian.org/"
SUPPORT_URL="https://www.debian.org/support"
BUG_REPORT_URL="https://bugs.debian.org/"
```

gh0st:
Ah I missed the fact that the docs specifically only cover the bookworm version. Shall we just close this MR then?

PR #4098, opened 2025-05-12
Fix display of bullet-points inside a table cell

lurch:
See https://www.raspberrypi.com/documentation/computers/raspberry-pi.html#pico-microcontrollers for where it's currently being rendered incorrectly.


mudge (approved):
For posterity: this works by instructing AsciiDoc to format the cell contents as AsciiDoc, see https://docs.asciidoctor.org/asciidoc/latest/tables/format-cell-content/#a-operator

PR #4099, opened 2025-05-12
Make kernel -> model table more exhaustive

lurch:
So that it better matches https://www.raspberrypi.com/documentation/computers/raspberry-pi.html

### DIFF
--- a/documentation/asciidoc/computers/configuration/boot_folder.adoc
+++ b/documentation/asciidoc/computers/configuration/boot_folder.adoc
@@ -70,12 +70,12 @@ Various xref:linux_kernel.adoc#kernel[kernel] image files that correspond to Ras
 
 | `kernel.img`
 | BCM2835
-| Pi Zero, Pi 1
+| Pi Zero, Pi 1, CM1
 |
 
 | `kernel7.img`
 | BCM2836, BCM2837
-| Pi Zero 2 W, Pi 2, Pi 3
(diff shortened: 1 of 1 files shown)


mudge (approved):

learmj:
[https://www.raspberrypi.com/documentation/computers/configuration.html#kernel-files-img](https://www.raspberrypi.com/documentation/computers/configuration.html#kernel-files-img) doesn't map v7l to pi3,cm3 etc. It's kernel8 which is associated to those devices.

mudge (commented):
I'll defer to @learmj on the correctness of this change.

lurch:
> https://www.raspberrypi.com/documentation/computers/configuration.html#kernel-files-img doesn't map v7l to pi3,cm3 etc. It's kernel8 which is associated to those devices.

AIUI the BCM2837 in the Pi 3, CM3, Zero 2 W, etc. can either run the 32-bit `v7` kernel _or_ the 64-bit `v8` kernel?
ping @pelwell as he'll know the definitive answer.

learmj:
That's correct as I understand it. I think my typo above hasn't helped what I was trying to convey in my comment. If the table is essentially listing compatibility then the change looks fine to me. Sorry.

pelwell:
@learmj I don't understand your point. What you say is correct, but how is it relevant?

lurch:
> If the table is essentially listing compatibility then the change looks fine to me.

Yes, that's exactly what it is :+1:

PR #4101, opened 2025-05-13
Update Connect screenshots to latest version

mudge:


lurch:
Did anything change in `disallow-remote-shell.png`, or did it just get re-compressed?

mudge:
> Did anything change in `disallow-remote-shell.png`, or did it just get re-compressed?

I set the pixel density of all images (as the new screenshots all had a "Retina" density of 144 DPI) and recompressed them and I suspect that was caught up in it (apologies for the extra 21 bytes).

PR #4107, opened 2025-05-19
revoke-devkey: Note that this is now set automatically

timg236:


documentation/asciidoc/computers/config_txt/boot.adoc
```diff
@@ -255,7 +255,12 @@ Default: `0`
 [[revoke_devkey]]
 ==== `revoke_devkey`
 
-If this property is set to `1` then `recovery.bin` will write a value to OTP that prevents the ROM from loading old versions of the second stage bootloader which do not support `secure-boot`. This prevents `secure-boot` from being turned off by reverting to an older release of the bootloader.
+Rasberry Pi 4 only.
```
lurch, line 258: Typo :wink:

documentation/asciidoc/computers/config_txt/boot.adoc
```diff
@@ -255,7 +255,12 @@ Default: `0`
 [[revoke_devkey]]
 ==== `revoke_devkey`
 
-If this property is set to `1` then `recovery.bin` will write a value to OTP that prevents the ROM from loading old versions of the second stage bootloader which do not support `secure-boot`. This prevents `secure-boot` from being turned off by reverting to an older release of the bootloader.
+Raspberry Pi 4 only.
+
+If this property is set to `1` then `recovery.bin` will write a value to OTP that prevents the ROM from loading old versions of the second stage bootloader which do not support `secure-boot`. This prevents `secure-boot` from being turned off by reverting to an older release of the bootloader. Therefore, this property must be set if `secure-boot` is enabled on production devices.
```
lurch, line 260: I guess "this property must be set" implies that it _always_ needs to be present in `config.txt`, whereas if this is modifying OTP values then it (technically) only needs to be present in `config.txt` one time?
Would "Therefore, this feature must be enabled if `secure-boot` is enabled on production devices." be a suitable replacement? (or perhaps "this OTP modification must be performed"?)

Oh - I've just seen the section above which says "These changes are irreversible and can only be programmed via `RPIBOOT` when flashing the bootloader EEPROM image." so perhaps that invalidates my comment :wink:

lurch (approved):

documentation/asciidoc/computers/config_txt/boot.adoc
timg236: All of the secure-boot properties only need to exists once for that device i.e. at the time of programming.

revoke_devkey is now redundant but the documentation needs to remain for existing configurations.

Issue #4003, opened 2025-01-11
Update needed to "Raspberry Pi Debug Probe" for PICO 2


cpottle9:
I see two issues that need to be addressed.

1) There is no PICO2 H. Need to describe manually installing headers.

2) The openocd version available on the PI 4 bookworm does not support the 2350.
    I eventually found the repo https://github.com/raspberrypi/openocd/tree/sdk-2.0.0 and built openocd.
    Not trivial. 
    You need to either provide a detailed recipe or provide downloadable binaries.

cpottle9:
Someone pointed out on Raspberry PI Forums that instructions for building openocd for PICO 2 is are included in Appendix A of https://datasheets.raspberrypi.com/pico/getting-started-with-pico.pdf.

I think that information belongs with the rest of the Debug Probe documentation here: https://www.raspberrypi.com/documentation/microcontrollers/debug-probe.html.

XECDesign:
I haven't heard anything from any maintainer of our OpenOCD fork asking for the package to be updated, so I assume it's not ready yet.

cpottle9:
I guess the process for updating openocd for raspberry pi os would include upstreaming the RP2350 changes to the main openocd repo and then waiting for an official version.
That is the only way to ensure all the other hardware targest are supported.

In the meantime I think it would be sufficient to update the debug probe documentation providing a link to appendix A of "Getting started with Raspberry Pi".

I wasted a couple of  finding raspberry pi openocd repo and figuring out how to build it.
This small change would have helped me immensely.

Lastly, I agree with issue https://github.com/raspberrypi/documentation/issues/3831.
A photo of a PICO 2 and PICO W with the 3 pin header soldered in place and the cable connected.

lurch:
https://github.com/raspberrypi/debugprobe/releases/ contains [`debugprobe_on_pico2.uf2`](https://github.com/raspberrypi/debugprobe/releases/download/debugprobe-v2.2.2/debugprobe_on_pico2.uf2) and https://github.com/raspberrypi/pico-sdk-tools/releases contains [`openocd-0.12.0+dev-aarch64-lin.tar.gz`](https://github.com/raspberrypi/pico-sdk-tools/releases/download/v2.1.0-0/openocd-0.12.0+dev-aarch64-lin.tar.gz) - does that meet your requirements?

@P33M and @will-v-pi - is there any reason why the version of openocd in our APT repos can't / shouldn't be updated?

cpottle9:
[...]

PR #4109, opened 2025-05-20
Rationalise debugprobe setup

P33M:
There's little point in duplicating instructions. I strongly prefer a single source of truth wherever possible, and a monolithic setup guide is better.

There's more to be done in the PDFs that link back to the documentation website (they shouldn't), but the website should at least call out the right chapters...

### DIFF
--- a/documentation/asciidoc/microcontrollers/debug-probe/installing-tools.adoc
+++ b/documentation/asciidoc/microcontrollers/debug-probe/installing-tools.adoc
@@ -1,66 +1,12 @@
 == Install tools
 
-To use the Debug Probe, install the following tools.
+To use the Debug Probe, OpenOCD and the GNU Project Debugger (GDB) are required. An Integrated Development Environment (IDE) may also be useful.
 
-=== Install OpenOCD
+On Raspberry Pi OS, most Linux variants, macOS, and Microsoft Windows, it is recommended to install our VS Code extension. This extension bundles OpenOCD, ARM toolchains, GDB, and register definitions for Pico-series microcontrollers.
 
-You need to install OpenOCD.
+See Chapter 3 of our https://datasheets.raspberrypi.com/pico/getting-started-with-pico.pdf[Getting Started with Raspberry Pi Pico] guide.
 
-To install OpenOCD, run the following command in a terminal:
+Alternatively, tools can be manually installed by following Appendix C in the guide.
 
-[source,console]
-----
-$ sudo apt install openocd
-----
+NOTE: Manual installation of the tools on Windows is not recommended.
 
-To run OpenOCD, use the `openocd` command in your terminal.
-
-==== Install OpenOCD on macOS
-
-First, install the https://brew.sh/[Homebrew] package manager:
-
-[source,console]
-----
-$ /bin/bash -c "$(curl -fsSL https://raw.githubusercontent.com/Homebrew/install/master/install.sh)"
-----
-
-To install OpenOCD on macOS, run the following commands:
-
-[source,console]
-----
-$ brew install openocd
-----
-
-To run OpenOCD, use the `openocd` command in your terminal.
-
-=== Install GDB
-
-We also need to install the GNU debugger (GDB).
-
-==== Linux
-
-Install `gdb-multiarch`:
-
-[source,console]
-----
-$ sudo apt install gdb-multiarch
-----
-
-==== macOS
-
-Run the following command to install `gdb`:
-
-[source,console]
-----
-$ brew install arm-none-eabi-gdb
-----
-
-You can safely ignore the request for "special privileges" messages on installation.
-
-==== MS Windows
-
-GDB is available as part of our https://github.com/raspberrypi/pico-setup-windows/releases/latest[Pico setup for Windows installer]. It is also included in the https://developer.arm.com/downloads/-/arm-gnu-toolchain-downloads[Arm GNU Toolchain Downloads].
-
-Alternatively information about manual installation can be found in Chapter 9 and Appendix A of our https://datasheets.raspberrypi.com/pico/getting-started-with-pico.pdf[Getting Started with Raspberry Pi Pico] book. 
-
-NOTE: Manual installation of GDB on Windows is not recommended.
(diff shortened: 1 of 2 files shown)


documentation/asciidoc/microcontrollers/debug-probe/swd-connection.adoc
```diff
@@ -1,10 +1,12 @@
-== Serial Wire Debug (SWD)
+== Starting a Debug Session
 
-Serial Wire Debug (SWD) is a two-pin interface (https://developer.arm.com/documentation/101761/1-0/Debug-and-trace-interface/Serial-Wire-Debug-signals[SWDIO and SWCLK]) alternative to the JTAG four- or five-pin debugging interface standard.
+The Debug Probe will let you load binaries via the SWD port and OpenOCD: you will not need to unplug, and then push-and-hold, the BOOTSEL button every time you push a new binary to your Pico. Using the Debug Probe to upload new binaries is an entirely hands-off affair.
```
lurch, line 3: nitpick: IMHO "then push-and-hold, the BOOTSEL button every" ought to be changed to "then push-and-hold the BOOTSEL button, every" (because you're not unplugging the BOOTSEL button :wink: ).

documentation/asciidoc/microcontrollers/debug-probe/swd-connection.adoc
```diff
@@ -35,10 +37,13 @@ $ make -j4
 ----
 
 In a debug build you will get more information when you run it under the debugger, as the compiler builds your program with the information to tell GDB what your program is doing.
-
-See Chapter 6 of https://datasheets.raspberrypi.com/pico/getting-started-with-pico.pdf[Getting started with Raspberry Pi Pico] for more information.
 ======
 
+NOTE: For devices that are _not_ Raspberry Pis, a variant of GDB that can debug ARM processors is required. Use one of the following alternatives depending on your operating system and device:
```
lurch, line 42: "For devices that are _not_ Raspberry Pis" - might be worth changing "devices" to "computers", in order to avoid possible confusion with Raspberry Pi Pico devices?

lurch:
On the whole, LGTM. I guess this addresses some of the concerns raised in #4003

documentation/asciidoc/microcontrollers/debug-probe/swd-connection.adoc
P33M: This copy-pastes the prior bad grammar.

documentation/asciidoc/microcontrollers/debug-probe/installing-tools.adoc
```diff
@@ -1,66 +1,12 @@
 == Install tools
 
-To use the Debug Probe, install the following tools.
+To use the Debug Probe, OpenOCD and the GNU Project Debugger (GDB) are required. An Integrated Development Environment (IDE) may also be useful.
 
-=== Install OpenOCD
+On Raspberry Pi OS, most Linux variants, and Microsoft Windows, it is recommended to install our VS Code extension. This extension bundles OpenOCD, ARM toolchains, GDB, and register definitions for Pico-series microcontrollers.
```
lurch, line 5: Also macOS ?

documentation/asciidoc/microcontrollers/debug-probe/swd-connection.adoc
```diff
@@ -35,10 +37,13 @@ $ make -j4
...
+* On Linux devices, use `gdb-multiarch`.
+* On macOS devices, use `arm-none-eabi-gdb`.
+* On Windows devices, use `gdb` bundled with the Raspberry Pi VSCode extension.
```
will-v-pi, line 45: Maybe say use `arm-none-eabi-gdb` for both Windows and macOS, as the extension is just downloading that from Arm's website (https://developer.arm.com/downloads/-/arm-gnu-toolchain-downloads)?

documentation/asciidoc/microcontrollers/debug-probe/swd-connection.adoc
lurch: Ahh, I hadn't noticed that. Will leave this for future copy-editors then.

documentation/asciidoc/microcontrollers/debug-probe/installing-tools.adoc
will-v-pi: Excluding x86_64 macOS, where the extension is currently missing OpenOCD

documentation/asciidoc/microcontrollers/debug-probe/swd-connection.adoc
P33M: IMO console commands all belong in the getting-started guide, but I guess the site should be less wrong.

documentation/asciidoc/microcontrollers/debug-probe/installing-tools.adoc
P33M: So what do x86_64 macOS users have to do?

will-v-pi: Probably use the homebrew OpenOCD head, which should have RP2350 support? The extension doesn't support debugging on x86_64 macOS

P33M: Seems to me like that needs to go in the getting-started guide and not here.

lurch: I agree that the distinction between x86_64 macOS and aarch64 macOS deserves to live in the PDF documentation, but IMHO if we're mentioning "most Linux variants, and Microsoft Windows" on this page, we also ought to mention macOS here.

lurch (approved):

lurch:
> It's not within the scope of this PR. I'm not doing wholesale rearchitecting.

Fair enough.

> The config files (and options) mutate between downstream openocd and upstream and are out of scope here.

Perhaps in a future PR it might be worth also removing the example `openocd` commands in the HTML documentation, and again refer the user to the getting-started PDF?

JamesH65:
Best to add an issue so the extra work is not forgotten when the new docs person arrives.

PR #4117, opened 2025-06-05
Add keyboard models to the setup guide

lurch:


documentation/asciidoc/computers/getting-started/setting-up.adoc
```diff
@@ -142,10 +139,11 @@ All variants of the Raspberry Pi 1, 2, 3, and 4 include a 3.5mm auxiliary http:/
 The following Raspberry Pi models come with Wi-Fi and Bluetooth connectivity:
 
 * Flagship models since Raspberry Pi 3 Model B
+* All Keyboard models
 * All Zero W models
```
JamesH65, line 143: Do we need Zero 2 in here?

JamesH65 (changes requested):

documentation/asciidoc/computers/getting-started/setting-up.adoc
```diff
@@ -26,19 +26,19 @@ You can use any high-quality power supply that provides the correct power mode.
 |Recommended power supply (voltage/current)
 |Raspberry Pi power supply
 
-|Raspberry Pi 5
+|Raspberry Pi 5 & Raspberry Pi 500
```
JeuneseRPi: Can we replace ampersands (&) with the word “and”?

documentation/asciidoc/computers/getting-started/setting-up.adoc
lurch: https://www.raspberrypi.com/documentation/computers/raspberry-pi.html#zero-series lumps in the Zero 2 with "Zero series" ?

documentation/asciidoc/computers/getting-started/setting-up.adoc
lurch: Done :+1:

documentation/asciidoc/computers/getting-started/setting-up.adoc
JamesH65: Perhaps  "All Zero W series models."

But adding a second line for Zero 2 is minimal effort and a lot clearer.

lurch: Looking at it again, the full list says:

The following Raspberry Pi models come with Wi-Fi and Bluetooth connectivity:

* Flagship models since Raspberry Pi 3 Model B
* All Keyboard models
* All Zero W models
* All Pico W models
* Compute Modules configured with wireless (available since CM4)


So if we were to add a separate entry for Zero 2 W (rather than assuming that "All Zero W models" covers it), that sort of implies that we'd also need to create a separate entry for Pico 2 W? And IMHO that's beyond the scope of this PR, so that should probably be discussed in a separate PR / issue.

PR #4126, opened 2025-06-26
Add new "Software sources" page

mudge:
Taken from a document provided by @ghollingworth, convert it to Asciidoc and add it to the "Computers" section of the documentation site (alongside other software-related sections such as "Raspberry Pi OS", "The Linux kernel", "Camera software", etc.).

Note the use of Asciidoc's formal passthrough macro to render `+rpt` and `#` as literals, see https://docs.asciidoctor.org/asciidoc/latest/text/literal-monospace/

Full-size screenshot from Chrome:

127 0 0 1_4000_documentation_computers_software-sources html(iPad Pro)


mudge:
And a full-size screenshot of the updated documentation landing page:

127 0 0 1_4000_documentation(iPad Pro)

JamesH65:
Couple of possible typos. Debian needs a capital D. The Applications section has some odd capitalisations which I don't think are right.

mudge:
> Couple of possible typos. Debian needs a capital D. The Applications section has some odd capitalisations which I don't think are right.

I've capitalised Debian and removed the capitals from "Camera". I'll leave other copy changes for now so we can get the initial version of this live and then edit from there (unless there is anything obviously wrong is a showstopper).

ghollingworth:
It should go through copy-edit...  I didn't go through it at all...  It's currently just "type as fast as I can" content...

Gordon

ghollingworth:
But, I agree that CE'ing shouldn't delay the output of documentation...  Even Gordon's writing is better than no writing...

:)

PR #4141, opened 2025-07-20
Changed regex for a cp command, to not copy all half built files, tha…

henwist:
…t are not needed for the kernel (`*.dtb*` to `*.dtbo`).


lurch:
I assume that the original was probably written like that for a good reason? Pinging @pelwell

pelwell:
Your suggested change will not copy overlay_map.dtb. Also, it's not a regex, it's a file glob.

henwist:
I tested to copy the files in the instructions, as per the current instructions on the kernel building page. And I did the wrong assumptions based on file globbing ( a leading . in the filename means it is not copied, as an exemption to what  a * expands to). So my apologies for bringing this unnessesary pull request up :-) .

lurch:
> ( a leading . in the filename means it is not copied, as an exemption to what a * expands to)

I think you're simply misunderstanding how file globbing works - a `.` has no special significance, and just matches to a literal `.` in the filename. So a file-glob of `*.dtb*` is roughly equivalent to the regex `^[^\.].*\.dtb.*$`

EDIT: Updated regex after Phil's feedback

pelwell:
Not quite, @lurch  - that doesn't exclude filenames starting with a `.`.

PR #4146, opened 2025-07-28
Update the audio board pinout diagram

JamesH65:


documentation/asciidoc/accessories/audio/hardware-info.adoc
```diff
@@ -39,7 +39,7 @@ If appropriate then the following are also used:
 
 === DAC PRO, DAC{plus}, DigiAMP{plus}, Codec Zero
 
-image::images/pin_table_new.jpg[width="80%"]
```
lurch, line 42: If you named the new image the same as the old image, then you wouldn't need to edit this adoc file, and GitHub would show you a nice little side-by-side view of the old and new images :slightly_smiling_face:

JamesH65: One is a JPG one is a PNG, didn't want to convert them.

lurch: LOL, I didn't notice that :facepalm:  PNG definitely makes more sense than a JPG here.
In which case, could you update this PR to also delete `pin_table_new.jpg` please, in order to avoid future confusion?

JamesH65: Can do, but I suspected there are a number of unused images in there (with suffix _new) so I was reluctant to delete as it appeared others had not been deleted either. But I will delete in this case once I have checked it not used elsewhere.

JamesH65: Done

PR #4157, opened 2025-08-17
Add instructions for building and installing kernel header .deb packages

MhmRhm:


lurch:
ping @XECDesign @popcornmix and @pelwell for feedback.

XECDesign:
The `deb-pkg` target is not something I would claim to support or recommend.

MhmRhm:
since documentation has not said anything about installing headers from source, using dep packages seems like a good option. Otherwise how a reader who is new to this topic will know where to get headers?
Installing kernel with dep packages doesn't work but headers does. Maybe we can add a disclaimer about this fact?

documentation/asciidoc/computers/linux_kernel/headers.adoc
```diff
@@ -20,4 +20,18 @@ If you use a 32-bit version of Raspberry Pi OS, run the following command to ins
...
 ----
 
+If you choose to build your own kernel, add the `deb-pkg` target to the build command:
+
+[source,console]
+----
+$ make -j6 Image.gz modules dtbs deb-pkg
+----
+
+This will generate three `.deb` files in the parent directory of the kernel source. If the kernel is cross-compiled, copy the `linux-headers-{version}.deb` file to your board, then run the following command to install:
```
katshann-raspberry:
```suggestion
This command generates three `.deb` files in the parent directory of the kernel source. If the kernel is cross-compiled, copy the `linux-headers-{version}.deb` file to your board, then run the following command to install:
```

documentation/asciidoc/computers/linux_kernel/headers.adoc
```diff
@@ -20,4 +20,18 @@ If you use a 32-bit version of Raspberry Pi OS, run the following command to ins
 $ sudo apt install linux-headers-rpi-{v6,v7,v7l}
 ----
 
+If you choose to build your own kernel, add the `deb-pkg` target to the build command:
```
katshann-raspberry:
```suggestion
If you choose to build your own kernel, add the `deb-pkg` target to the build command.

NOTE: Raspberry Pi doesn't test or support using the  `deb-pkg` target. If you choose this approach, proceed with caution.
```

katshann-raspberry:
Added a disclaimer. @MhmRhm @XECDesign is there more to do here to get this ready to publish?

katshann-raspberry:
@learmj I'm told you might have some thoughts about the workflow being added in this PR. If you have a chance to cast your eye over, that'd be grand.

learmj:
> @learmj I'm told you might have some thoughts about the workflow being added in this PR. If you have a chance to cast your eye over, that'd be grand.

Thanks. Sorry for the late reply. I'll post back once I've reviewed.

learmj:
Our Documentation encompasses what is provided by official Raspberry Pi Debian packages and how this can be customised (where applicable) for interoperability with Raspberry Pi hardware devices. The `linux-image-rpi-*` Debian packages released by Raspberry Pi provide linux kernel images that are built, tested and supported by Raspberry Pi.

The linux kernel is a piece of software that, depending on uses/needs, may necessitate customisation. One size does not  fit all. However, Raspberry Pi kernel images are designed to effectively be a drop-in, i.e. the general principle is that while kernels may need customisation, Raspberry Pi kernel images suit many use-cases.

Where the kernel becomes more significant is when the intended use of the device has specific requirements which may make it challenging to be supported by the default configuration. Examples include:

- Performance optimisation for specific workloads
- Hardware support for non-standard or custom peripherals not included by default
- Security hardening with custom configurations, reduced attack surface, or specialised security modules
- Resource constraints requiring a minimal kernel footprint
- Development and testing of kernel features, drivers, or patches before they're available in official releases
- Compliance requirements mandating specific kernel versions or configurations

For these scenarios, building a custom kernel is a requirement and doing so allows bespoke and fine-grained control over kernel features, modules, and optimisations that the `linux-image-rpi-*` packages may not be able to accommodate. However, since the Documentation is primarily concerned with officially released Raspberry Pi software it's reasonable to question whether further information regarding a custom kernel is required over and above what's in place today on the existing pages.

It's probably fair to say that if a user is comfortable with building and installing a custom kernel, they likely won't need guidance over and above what's currently available in order to do so. Furthermore, because of the different ways a custom kernel can be installed, it seems out of scope for the Documentation to advocate one method in particular over others, especially since there are at least two methods of creating Debian packages from a mainline linux source build.

katshann-raspberry:
@learmj Thank you for the review and comprehensive explanation. Based on this recommendation, I'm going to close this PR without merging.

PR #4164, opened 2025-08-23
Delete non-existent item entries

ykla:
404


lurch:
These repos all exist, but they're currently private. Should they be made public @spl237 ?

spl237:
Repos made public - no need for this change.

lurch:
> Repos made public - no need for this change.

Thanks!

lurch:
Thank you also to @ykla for actually noticing this :+1:

PR #4167, opened 2025-08-27
Add super-old (and unsupported) PXE scripts from datasheets site

lurch:


documentation/asciidoc/computers/remote-access/old/prepare_pxetools.sh
```diff
@@ -0,0 +1,102 @@
...
+
+echo "Getting latest Raspberry Pi OS lite image to use as NFS root"
+# Get latest Raspberry Pi OS lite image
+sudo mkdir -p /nfs/bases
+cd /nfs/bases
+sudo wget -O raspios_latest.zip https://downloads.raspberrypi.org/raspios_lite_armhf_latest
+sudo unzip raspios_latest.zip
+sudo rm raspios_latest.zip
+
+sudo wget  -O /usr/local/sbin/pxetools https://datasheets.raspberrypi.org/soft/pxetools.py
```
mudge, line 99: Should this now refer to wherever `pxetools.py` will end up on the documentation site?

lurch: I was assuming that I'd be able to point the redirect on datasheets to this new location? (when the time comes :wink: )

mudge:
After running the full build with `make`, it doesn't look like these files end up in `documentation/html` which is what will be deployed to the static web server. Do you also need to make sure they are copied to `build/jekyll` first?

lurch:
> After running the full build with `make`, it doesn't look like these files end up in `documentation/html` which is what will be deployed to the static web server. Do you also need to make sure they are copied to `build/jekyll` first?

Ah, yes. Well spotted. I'll have to go away and have a think... :brain:

lurch:
Change of plan... :wastebasket:

PR #4168, opened 2025-08-27
Use of DISP0 on CMIO requires correct dtoverlay override

6by9:
Add ,dsi0 override when using DISP0 on CMIO


6by9:
Noted from https://forums.raspberrypi.com/viewtopic.php?p=2334288 where someone was following the docs and using the wrong dtoverlay line.

JamesH65 (approved):

PR #4174, opened 2025-09-07
Update getting-started.adoc

copeland3300:
Added wire colors for GND, SC and SD for clarity


lurch (approved):
Approving, as this matches up with the colours mentioned at https://www.raspberrypi.com/documentation/microcontrollers/debug-probe.html#the-debug-probe

PR #4178, opened 2025-09-10
Update Connect logging section

lurch:
Tweak journalctl commands to work with the volatile logging used in recent versions of Raspberry Pi OS

ping @clowder and @roliver-rpi


roliver-rpi:
LGTM

PR #4189, opened 2025-09-22
Fix GPU clock frequencies

lurch:
See the default `core_freq` frequencies listed at https://www.raspberrypi.com/documentation/computers/config_txt.html#overclocking for reference.

ping @popcornmix


popcornmix:
so on Pi5 gpu_freq (core, hevc, isp) is 910MHz. v3d_freq is 960MHz.
This section is referring to v3d, so I guess 960 is more correct.

lurch:
Updated and force-pushed.

PR #4194, opened 2025-09-25
Additional mentions of the 500+

katshann-raspberry:


documentation/asciidoc/computers/raspberry-pi/power-supplies.adoc
```diff
@@ -90,6 +90,11 @@ All Raspberry Pi models with a PoE-capable Ethernet jack require a HAT to draw p
 | 1.6A (600mA if using a 3A power supply)
 | 800mA
 
+| Pi 500+
+| 5.0A
+| 1.6A
+| 800mA
```
lurch: @egycm3 Presumably the "Typical bare-board active current consumption" figure is higher for Pi 500+ than it is for Pi 500?

documentation/asciidoc/computers/config_txt/conditional.adoc
```diff
@@ -32,13 +32,13 @@ The conditional model filters apply according to the following table.
 | Model 4B, Pi 400, Compute Module 4, Compute Module 4S
 
 | `[pi5]`
-| Raspberry Pi 5, Compute Module 5, Pi 500
+| Raspberry Pi 5, Compute Module 5, Pi 500, Pi 500+
```
lurch: Super minor nitpick, but for consistency with the other entries in this table, maybe it's worth moving CM5 to the end of the list?

documentation/asciidoc/computers/raspberry-pi/eeprom-bootloader.adoc
```diff
@@ -190,7 +190,7 @@ Default: `0`
 [[SD_OVERCURRENT_CHECK]]
 ==== `SD_OVERCURRENT_CHECK`
 
-Raspberry Pi 4, Raspberry Pi 5 and Raspberry Pi 500 only.
+Raspberry Pi 4, Raspberry Pi 5, Raspberry Pi 500, and Raspberry Pi 500+ only.
```
lurch: @timg236 Can you confirm that this doesn't apply to Pi 400?

timg236: Correct

documentation/asciidoc/computers/config_txt/boot.adoc
```diff
@@ -5,13 +5,13 @@
...
-`fixup_file` specifies the file used to fix up memory locations used in the `start_file` to match the GPU memory split. 
+`fixup_file` specifies the file used to fix up memory locations used in the `start_file` to match the GPU memory split.
 
 The `start_file` and the `fixup_file` are a matched pair - using unmatched files will stop the board from booting. This is an advanced option, so we advise that you use `start_x` and `start_debug` rather than this option.
 
 NOTE: Cut-down firmware (`start*cd.elf` and `fixup*cd.dat`) cannot be selected this way - the system will fail to boot.  The only way to enable the cut-down firmware is to specify `gpu_mem=16`. The cut-down firmware removes support for codecs, 3D and debug logging as well as limiting the initial early-boot framebuffer to 1080p @16bpp - although KMS can replace this with up to 32bpp 4K framebuffer(s) at a later stage as with any firmware.
 
-NOTE: The Raspberry Pi 5, Compute Module 5, and Raspberry Pi 500 firmware is self-contained in the bootloader EEPROM.
+NOTE: The Raspberry Pi 5, 500, 500+, and Compute Module 5 firmware is self-contained in the bootloader EEPROM.
```
timg236, line 14: This should refer to Pi4 / Pi5 family because it applies to all devices including CM4.

documentation/asciidoc/computers/configuration/boot_folder.adoc
```diff
@@ -85,12 +85,12 @@ Various xref:linux_kernel.adoc#kernel[kernel] image files that correspond to Ras
...
 | `kernel8.img`
 | BCM2837, BCM2711, BCM2712
-| Pi Zero 2 W, Pi 2 (later revisions), Pi 3, CM3, Pi 3+, CM3+, Pi 4, CM4, CM4S, Pi 400, CM5, Pi 5, Pi 500
+| Pi Zero 2 W, Pi 2 (later revisions), Pi 3, CM3, Pi 3+, CM3+, Pi 4, CM4, CM4S, Pi 400, CM5, Pi 5, Pi 500, Pi 500+
 | xref:config_txt.adoc#boot-options[64-bit kernel]. Earlier revisions of Raspberry Pi 2 (with BCM2836) do not support 64-bit kernels.
 
 | `kernel_2712.img`
 | BCM2712
-| Pi 5, CM5, Pi 500
+| Pi 5, CM5, Pi 500, Pi 500+
```
timg236, line 93: Also - Pi4 / Pi5 family rather than listing every possible model

documentation/asciidoc/computers/linux_kernel/building.adoc
```diff
@@ -93,7 +93,7 @@ $ cd linux
 $ KERNEL=kernel_2712
 $ make bcm2712_defconfig
 ----
-| Pi 500
+| Pi 500/500+
```
timg236: This is relevant to entire product line

documentation/asciidoc/computers/raspberry-pi/power-supplies.adoc
```diff
@@ -90,6 +90,11 @@ All Raspberry Pi models with a PoE-capable Ethernet jack require a HAT to draw p
 | 1.6A (600mA if using a 3A power supply)
 | 800mA
 
+| Pi 500+
+| 5.0A
+| 1.6A
+| 800mA (with LEDs off and SSD inactive)
```
timg236, line 96: It's wrong to mention LEDs / SSD here because this is referring to the supply current and not the consumed current

documentation/asciidoc/computers/raspberry-pi/eeprom-bootloader.adoc
timg236: This should be a specific list btw

documentation/asciidoc/computers/config_txt/boot.adoc
lurch: In this sentence, I think "firmware" is referring to `start.elf` / `start4.elf`, which is still a separate file for the Pi4 family?

timg236: Yes, Pi4 family use start4.elf by default and Pi5 family is self-contained or perhaps more precise the BCM2712 bootrom family because that's the significant thing rather than the board design.

documentation/asciidoc/computers/linux_kernel/building.adoc
lurch: It looks like this change is adding Pi 500+ to [this table](https://www.raspberrypi.com/documentation/computers/linux_kernel.html#native-build-configuration).
But @katshann-raspberry it seems that the [other table](https://www.raspberrypi.com/documentation/computers/linux_kernel.html#cross-compiled-build-configuration) is missing Pi 500, Pi 500+ and CM5 :facepalm:  (And you're of course free to simplify these tables if you can think of a good way of doing so; but perhaps that's outside the scope of this PR!)

documentation/asciidoc/computers/raspberry-pi/power-supplies.adoc
lurch: This is the table at https://www.raspberrypi.com/documentation/computers/raspberry-pi.html#typical-power-requirements where the 4th column is titled "Typical bare-board active current consumption".

documentation/asciidoc/computers/config_txt/boot.adoc
lurch: Yeah, it's always hard to get the balance right in technical documentation like this, but I'm happy to bow to @katshann-raspberry and @JeuneseRPi 's best judgement.

katshann-raspberry: It seems like what this note wants to emphasise is the self-contained nature of the BCM2712 family firmware.  I'm unclear as to why, though. What does the information in this note change about what the customer would choose or do after reading it?

documentation/asciidoc/computers/configuration/boot_folder.adoc
katshann-raspberry: I don't think we're hard up for space in this table. So it doesn't cost us anything to explicitly list. 
If a reader is fast-scanning the text for 500 info, they might jump over a mention of 5 on first reading.

documentation/asciidoc/computers/linux_kernel/building.adoc
katshann-raspberry: Good spot of the second table. Will update.

documentation/asciidoc/computers/config_txt/boot.adoc
lurch: This appears at https://www.raspberrypi.com/documentation/computers/config_txt.html#start_file-fixup_file so perhaps the NOTE is trying to explain that the BCM2712 (i.e. Pi5 family) doesn't use the `start_file` and `fixup_file` options? And yes, perhaps that could/should be made clearer :slightly_smiling_face:

documentation/asciidoc/computers/raspberry-pi/eeprom-bootloader.adoc
lurch: Yes, perhaps "Raspberry Pi 4" needs to be clarified to "Raspberry Pi 4 Model B"?

documentation/asciidoc/computers/linux_kernel/building.adoc
lurch: I've also just spotted that the first table uses e.g. "Compute Module 3" whereas the second table uses "Raspberry Pi Compute Module 3". (IMHO the former is preferable, as it's shorter)

katshann-raspberry: Throughout the docs, our choice of when to say Raspberry Pi X vs Pi X vs just X is not super consistent. I think the quest to make this consistent is going to be a long and ongoing on. 
I'll fix these ones to be consistent, but for the sake of getting this PR published before the weekend, let's leave the wider quest for another PR.

lurch: Yep, sorry.

documentation/asciidoc/computers/config_txt/boot.adoc
katshann-raspberry: Thanks for the info on this one, folks. I'll pick up improving this note as a separate action for next week.

documentation/asciidoc/computers/linux_kernel/building.adoc
katshann-raspberry: Never apologise for high quality proofreading!

documentation/asciidoc/computers/configuration/boot_folder.adoc
timg236: In general, I think listing every possible product tends to result in accidental omissions and implies that there it's specific to the board rather than the chip.
As someone building the kernel you are probably more interested in the architecture that it's supporting. So kernel_2712 is any 2712 device.

lurch: I agree with what you're saying Tim (and there are many other large parts of the documentation that are technically SoC-specific rather than Pi-model-specific), but how does somebody with a newly-purchased Pi 500+ know that it uses a BCM2712 chip?

timg236: My suggestion is that where there is a broad definition for an entire class of products (Pi5 family) it's clearer, more concise and less prone to bitrot (and new product launches) to say that.
If there's a a need for an explicit list then that normal is specific to the board and it's actually a good signpost that it's an explicit list.
A short section somewhere, which explains the flagship model for the family is probably useful and probably already documented e.g. Pi5 has a BCM2712 processor, this performance, 64-bit only kernel, firmware entirely within SPI flash, PMIC

katshann-raspberry: I think Jeunese already has some thoughts in progress about taxonomy and how we consistently refer to these things in docs. So I'll wait for her before wading too deeply into this.

There's also some understanding I need to do with regards to our audience and how it might differ depending on which page of the docs the content is on. There might be some articles where it's fine to just say which chip and trust that anyone reading that page only needs that info to understand. However, there might be others where the target audience  doesn't need to or want to understand that deeply and just wants to know if a particular statement applies to the product they bought without having to delve into the versioning of its components.

timg236: Sounds good. Something copy-and-pasteable :)

documentation/asciidoc/computers/config_txt/boot.adoc
JeuneseRPi: As a nit-pick, there should be no "The" at the beginning of the sentence.

lurch: In this case I believe it's talking about "The firmware..." - so perhaps the sentence should be rearranged to:
"The firmware for Raspberry Pi 5, 500, 500+, and Compute Module 5 is self-contained in the bootloader EEPROM." ?

JeuneseRPi: You are correct. Forgive me. It's my first day back. I'll come up with a new excuse for mistakes I make tomorrow. ;-)

PR #4195, opened 2025-09-26
500+ updates and change of dependency

katshann-raspberry:


katshann-raspberry:
@mudge Adding you to review just for sanity check that it's OK to take this change that includes a dependency update through to the production site.

lurch (approved):

PR #4206, opened 2025-10-07
Accessibility tweaks

katshann-raspberry:


jekyll-assets/_layouts/docs.html
```diff
@@ -11,19 +11,19 @@
         <div id="toc-container">
           <div id="toc-inner">
             <div id="docs-header">
-              <h1 id="docs-header-title">
+              <span id="docs-header-title">
```
mudge: Is this a semantic change to [avoid using more than one `<h1>` on a page](https://developer.mozilla.org/en-US/docs/Web/HTML/Reference/Elements/Heading_Elements#avoid_using_multiple_h1_elements_on_one_page)? Given it _is_ a heading, should we use an `<h2>`, etc. instead?

jekyll-assets/_layouts/docs.html
```diff
@@ -11,19 +11,19 @@
...
               <label class="mobile-menu-toggle" for="mobile-toggle">
                 <div class="mobile-menu-toggle-inner"></div>
               </label>
             </div>
-            <div id="docsearch">
+            <div id="docsearch" role="search">
             </div>
           </div>
-          <div class="toc">
+          <div class="toc" role="navigation">
```
mudge, line 26: If this is truly a navigation element, should we switch the `<div>` for a [`<nav>`](https://developer.mozilla.org/en-US/docs/Web/HTML/Reference/Elements/nav)?

jekyll-assets/_layouts/docs.html
```diff
@@ -154,14 +154,14 @@ <h1 id="docs-header-title">
       </div>
       <div id="docs-container">
         <div id="main-window">
-          <section id="content">
+          <section id="content" role="main">
```
mudge, line 157: As above, if this is truly the dominant element, should we switch from `<section>` to [`<main>`](https://developer.mozilla.org/en-US/docs/Web/HTML/Reference/Elements/main)?

jekyll-assets/css/style.css
```diff
@@ -754,7 +756,7 @@ h6 .anchor::before {
 
 #content p {
   font-size: 0.95em;
-  font-weight: 300;
+  font-weight: 400;
```
mudge, line 763: If we're no longer using the "Light" weight of 300, can we remove it from the [Google Fonts include](https://github.com/raspberrypi/documentation/blob/4d43c3f789b556311d59fb4d87b41fc00fc7d922/jekyll-assets/_includes/head.html#L17)?

katshann-raspberry: We're still using it in some headers and copyright text. I only made the change for body text.

jekyll-assets/_layouts/docs.html
katshann-raspberry: Same reasoning as above. Element name changes have a greater need for CSS testing at light/dark and different viewport sizing to ensure nothing got messed up. With the risk that if it does change thigs there's the unknown amount of time and effort from me (a CSS amateur) to fix up what gets broken.

jekyll-assets/_layouts/docs.html
katshann-raspberry: My reasoning behind adding an aria role instead was to avoid unintended styling effects that might emerge depending on how the CSS is set up.
Get the accessibility benefit without the CSS refactor risk.

jekyll-assets/_layouts/docs.html
katshann-raspberry: Can do. I was just trying to avoid CSS faff.

jekyll-assets/_layouts/docs.html
mudge: That's fair: I was hoping the styling was based on classes rather than element type but it looks like we have selectors like `div.content`.

jekyll-assets/css/style.css
```diff
@@ -102,6 +104,10 @@ h1 {
   background: transparent;
 }
 
+h2 {
```
mudge: If we want to be a bit defensive here, is it worth scoping to `#docs-header` or even putting this on `#docs-header-title` directly?

katshann-raspberry: I've put it on docs-header. I didn't want to be too specific as there is an overriding 1.9em on the more specific item that I want to apply, but my browser was multiplying this with a 1.5em from somewhere. I think I needed the initial to clear that 1.5em.

mudge: Yeah, `em` is a relative unit so it is affected by the parent font size, see https://developer.mozilla.org/en-US/docs/Learn_web_development/Core/Styling_basics/Values_and_units#numbers_lengths_and_percentages. (This is why we prefer using `rem` for the rest of the site.)

mudge (approved):

PR #4221, opened 2025-10-17
Add a background to the transparent images for dark mode compatibility.

katshann-raspberry:


lurch:
Not _directly_ related to this PR, but:
* `all_audio_boards_gpio_pinouts.png` probably doesn't need to be "W: 3556px | H: 2528px" ! (I guess that's probably a "print resolution" rather than a "web resolution"?)
* I'm not sure if it makes sense for the CM5-related technical drawings (`cm4-cm5-antenna-physical.png`, `cm5-case-physical.png` and `cm5-cooler-physical.png`) to be in our HTML docs; it might make more sense to just link to the relevant product-briefs instead? (Links can be found on https://www.raspberrypi.com/products/compute-module-4-antenna-kit/ https://www.raspberrypi.com/products/io-case-cm5/ and https://www.raspberrypi.com/products/cooler/ ). That way if the product-briefs get updated in future, there's less risk of the information in the HTML docs slipping out of date (IMHO).

katshann-raspberry:
@lurch I've rescaled the audio image. The rest, I'll take as a separate piece of work.

PR #4265, opened 2025-12-06
Update window manager options in raspi-config documentation

Not-a-bug:
For raspi-config nonint do_wayland,
W2 option is now labwc
W3 option has been removed


lurch:
Thanks. I guess https://www.raspberrypi.com/documentation/computers/configuration.html#wayland should also be updated to remove the "and choose a window manager" text, because in Raspberry Pi OS Trixie we no longer offer a choice of switching to the wayfire window manager.

Not-a-bug:
Good point, done.

lurch:
Thank you!

Blackmvmba88 (commented):
Perfect

katshann-raspberry:
Changes now reflected on the docs website.

PR #4272, opened 2026-01-02
Update revision-codes.adoc

JamesH65:
Add the CM0 to the revision code type list.


katshann-raspberry:
Progressing through internal repo.

PR #4274, opened 2026-01-05
Update conditional.adoc to add CM0

JamesH65:


katshann-raspberry:
Progressing through the internal repo.

PR #4277, opened 2026-01-24
Clarify Wi-Fi instructions to explain how to skip.

iressler:
Currently if you don't want to setup wi-fi on your Raspberry Pi, but are using the imager from a device with wi-fi the Image will prefill the wi-fi SSID and make it appear that wi-fi is required to continue. There is an [issue on the rpi-imager](https://github.com/raspberrypi/rpi-imager/issues/1373)  to improve this, however until that is resolved I believe the documentation can be improved.

This not being clear, even as I followed the instructions, nearly made me stop using the imager/skip all customisation since I do not want my Raspberry Pi to have any network access.

This change addresses the above concerns by updating the wi-fi customisation instructions to explain how to skip setting up wifi.


katshann-raspberry:
Thanks for this PR. The page has now been updated.

PR #4278, opened 2026-01-30
fix: typo in hailo examples link

LarsFronius:


katshann-raspberry:
Update now live on public site.

PR #4138, opened 2025-07-07
Update power-supplies.adoc

JamesH65:
Add power figures for CM4/CM5

These numbers were taken from the respective datasheets.

PR #4279, opened 2026-01-31
Update power-supplies.adoc

Royalsuitking:
Add power figures for CM4/CM5


lurch:
This is a duplicate of #4138 , which was already closed.

PR #4280, opened 2026-02-09
Update conditional.adoc to support the cust_otp variable

martamomotko:
The cust_otpN boot variable, where N ranges from 0–7, correspond to the eight customer OTP rows, and then can be used as a conditional variable in config.txt to select different configurations.


katshann-raspberry:
Progressing on internal repository

PR #4294, opened 2026-04-13
Fix order of mechanical drawings links for Raspberry Pi 1

kduvekot:
The PDF and DXF links were in switched


katshann-raspberry:
Many thanks for this PR. The changes are now live.

PR #4295, opened 2026-04-15
setting-up: Use "computer" instead of "circuit board"

JohnVeness:
"Circuit board" may be a little too jargon-y, and we use "computer" elsewhere on the page.


katshann-raspberry:
Thank you for this PR. The changes are now live on the documentation website.

PR #4296, opened 2026-04-17
Add GitHub link to footer

n3orma:
@mudge  
@katshann-raspberry


katshann-raspberry:
Hi @n3orma, thanks for this PR. We're not going to accept it at this time as the documentation site footer needs to match the footer in other parts of the site. We also have multiple GitHub organisations that could equally be linked here. 
As this raises a question of wider changes across our sites, we're going to take this into internal discussions about what the right thing to do here is. 
Cheers,
Kat

lurch:
> We also have multiple GitHub organisations that could equally be linked here.

See https://raspberrypi.github.io/ and https://www.raspberrypi.com/documentation/computers/software-sources.html for more information.

n3orma:
Hi @katshann-raspberry, @lurch,

Thanks for the clarification — that makes sense regarding consistency across Raspberry Pi sites and multiple GitHub organizations.

Instead of closing the PR, would it be possible to convert it into a draft? Since this topic is being considered for broader internal discussion, keeping it as a draft could help track context and continue iteration if a unified approach is decided later.

Happy to adapt the changes based on what direction is taken.

Thanks!

katshann-raspberry:
I've created a related issue instead.

PR #4298, opened 2026-04-28
Add a link to a CM5 testpoints visualisation

KoviRobi:
I wanted to use the UART testpoints (accidentally torn off the JST header I soldered, by trying to unplug the connector), and it was hard to know which one is which.

Thought it might be useful for others, feel free to just vendor the SVG or otherwise request changes (including license changes or repo moves). It's more or less just a scatter plot example using the table from the schematic PDF plus a photo. But might be useful for others.

### DIFF
--- a/documentation/asciidoc/computers/compute-module/datasheet.adoc
+++ b/documentation/asciidoc/computers/compute-module/datasheet.adoc
@@ -6,6 +6,7 @@ To learn more about Compute Module 5 (CM5) and its corresponding IO Board, see t
 
 * https://pip.raspberrypi.com/documents/RP-008180-DS[CM5 datasheet]
 * https://rpltd.co/cm5-design-files[CM5 design files]
+* https://kovirobi.github.io/rpi-cm5-testpoints[CM5 testpoints visualisation]
 
 === Compute Module 5 IO Board datasheet
 


katshann-raspberry:
@KoviRobi Thanks for this suggested contribution. I'll check with our subject matter experts about whether and how to include this information. 
Looking at your SVG in Chrome, there's some superposition of text:

image

Fine in Firefox though.

ETA: I'm inclined to think that we might instead put a diagram of this nature in https://pip.raspberrypi.com/documents/RP-008180-DS-6 Appendix A, rather than the HTML docs.

KoviRobi:
Ah thanks for checking, turns out while the `dominant-baseline` attribute itself is widely supported, the values I used (`text-bottom` and `text-top`) aren't. Fixed using the `dy` attribute instead.

And yes I'd be very happy with a direct image in the PDF (as long as it is searchable/not just a screenshot). I've used MIT no attribution as I think that makes it easiest to just use however but happy to change to whatever, if you just want to directly use the SVG.

But also since you have the PCB layout you could just as easily use that to generate a render of the bottom of the PCB with the testpoints labelled (probably).

katshann-raspberry:
@KoviRobi Apologies for the delay in response (out on hols). Thanks for the update to the SVG. I'll look at progressing this into the datasheet, thought we might go with a diagrammatic/rendered version instead. 

I'll keep this PR open for now to track that there's outstanding work and let you know when the datasheet is updated.

Cheers!

katshann-raspberry:
@KoviRobi We ended up cooking up our own version from the design files instead. Thank you for the inspiration! The diagram is now included after the testpoints table in the CM5 datasheet PDF.

PR #4307, opened 2026-05-16
Fix link to RP2040 Datasheet in Pico documentation

ErfanTech:
Updated the RP2040 datasheet link to point to the correct document (`RP-008371-DS`) instead of the previous `RP-008373-DS`, which incorrectly referenced the RP2350 datasheet. This fixes the documentation mismatch and ensures the link now directs to the proper RP2040 reference.


katshann-raspberry:
Many thanks, @ErfanTech , for your fix. It should now be in and live on the website.

Issue #3938, opened 2024-11-25
How to change the primary UART of a Raspberry Pi


nllho:
Related to https://github.com/raspberrypi/documentation/issues/3239.

Context:
- The primary UART is where the Linux console appears.
- On most Raspberry Pis, the primary UART appears on the GPIO header.
- For Raspberry Pi 5, the primary UART appears on the Debug header.

Use case:
When switching to Pi 5 from older versions, people get confused because the primary UART has changed, and they do not see the console as expected. This confusion may persist into the future with new Pi models and as people upgrade their hardware.

Current state:
The bullet points above already exist in the documentation. However, we do not currently document how to change the primary UART of a Raspberry Pi.

Possible docs solutions:

1. Create a troubleshooting section for "the console does not appear" somewhere under initial configuration or UART configuration. This section could be expanded with multiple solutions, including the information here about using the correct primary UART for a given Raspberry Pi model.

2. Add "Changing the primary UART" to https://www.raspberrypi.com/documentation/computers/configuration.html#configure-uarts. [This comment by Phil](https://forums.raspberrypi.com/viewtopic.php?t=359132&start=25#p2154397) explains how to change the primary UART of the Pi 5 to the GPIO header. If we create this section, we would need to confirm instructions for all Pi models.

Issue #3239, opened 2023-10-26
Need to update UART documentation for Pi5


metachip:
The [existing documentation](https://www.raspberrypi.com/documentation/computers/configuration.html#configuring-uarts) makes no mention of Pi5 or how it differs in terms of UART usage and [we've reported at least one new bug](https://github.com/RPi-Distro/raspberrypi-sys-mods/issues/84#issuecomment-1780175158) (affecting older models but introduced by code changes to support Pi5). While we get the big picture apropos [the new connector](https://www.raspberrypi.com/documentation/computers/raspberry-pi-5.html#uart-connector) (all good!) it appears the logical meaning or intended usage of the alias names (serial0 and serial1) and the [confounding bluetooth complexities](https://docs.bitscope.com/pi-serial/#bluetooth) of the past may be changed in Pi5, or there are unresolved bugs, or both?

For example, on Pi4 with bullseye, specifying **console=serial0,115200** in **cmdline.txt** when **uart_enable=1** and/or **dtoverlay=disable-bt** is asserted in **config.txt** results in **serial0** being linked as primary (i.e. **ttyAMA0**) and the console being enabled on pins **14/15** on **J8** enabling firmware boot and kernel messages on boot and the system console via the J8 pins. In terms of aliases we have **/dev/serial0 -> ttyAMA0** and **/dev/serial1 -> ttyS0**. All good.

Not so on Pi5.

Instead on Pi5 with bookworm this configuration seems to enable all this on new UART connector. A different setup is needed to replicate backward compatible behaviour which setup seems to be: specify **console=ttyAMA0,115200** in **cmdline.txt** when **dtparam=uart0=on** is asserted in **config.txt**. All good, but confusing as now **/dev/serial0 -> ttyAMA10** where (I'm guessing ttyAMA10 is the new port) and **/dev/serial1** is not assigned at all? If we also assert **dtoverlay=disable-bt** in **config.txt** we get aliases **/dev/serial0 -> ttyAMA10** and **/dev/serial1 -> ttyS0**.

So there are changes and bugs here and we're having difficulty disambiguating between them.

**Bottom line**: updated documentation so we know what the intended behaviour is, even if bugs may get in the way :)

aallan:
You can expect the documentation to be update to cover Pi 5 over the next few month. If you're interested you can keep track of what's going on by looking at the project [milestone](https://github.com/raspberrypi/documentation/milestone/4).

pelwell:
[...]

PR #4310, opened 2026-06-08
fix: BOOT_UART on Raspberry Pi 5 does not use GPIO 14/15

christopher-besch:
Related to https://github.com/raspberrypi/documentation/issues/3239 and https://github.com/raspberrypi/documentation/issues/3938 .

We were pulling our hair out at getting UART to work on the Raspberry Pi 5 Compute Module. After a lot of trying around we concluded that there are mistakes in:
- the docs, which this PR fix,
- the [datasheet](https://pip-assets.raspberrypi.com/categories/944-raspberry-pi-compute-module-5/documents/RP-008180-DS-6-cm5-datasheet.pdf) and
- raspi-config.

I wrote an article explaining this in-depth: https://chris-besch.com/articles/raspberry_pi_5_uart
We're not entirely sure our understanding of all this is completely correct and we base our explanation on some tests we did only with a Raspberry Pi 5 Compute Module IO Board. We'd of course be very interested in any mistakes we might have done.

### DIFF
--- a/documentation/asciidoc/computers/configuration/interfaces.adoc
+++ b/documentation/asciidoc/computers/configuration/interfaces.adoc
@@ -225,7 +225,7 @@ The available UART interfaces vary by Raspberry Pi model. The following table su
   `UART2` through to `UART5`: PL011 (disabled by default)
 
 | Raspberry Pi 5, 500, 500+, and Compute Module 5
-| `UART0` through to `UART4`: PL011 (disabled by default) +
+| `UART0` through to `UART4` and `UART10` (also called `debug uart`): PL011 (disabled by default) +
   No mini UART
 |===
 
--- a/documentation/asciidoc/computers/raspberry-pi/eeprom-bootloader.adoc
+++ b/documentation/asciidoc/computers/raspberry-pi/eeprom-bootloader.adoc
@@ -27,7 +27,7 @@ This section describes all the configuration items available in the bootloader. 
 [[BOOT_UART]]
 ==== `BOOT_UART`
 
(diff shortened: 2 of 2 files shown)


lurch:
There's also https://www.raspberrypi.com/documentation/computers/config_txt.html#enable_rp1_uart and https://github.com/raspberrypi/rpi-eeprom/blob/master/firmware-2712/release-notes.md#2025-07-03-enable-firmware-uart-output-on-the-40-pin-header-latest (which doesn't appear to be mentioned in the docs yet).

katshann-raspberry:
@pelwell Can you have a look at this PR and the linked article and confirm if the understanding and suggested docs changes are correct? 
Cheers!

pelwell:
I think that looks OK. I'm not so keen on the use of "primary UART" to refer to the UART header on Pi 5, since this is more of a configuration-dependent software concept. It's defined as the thing that `/dev/serial0` ends up pointing at, which is often the debug UART on Pi 5, but see below.

As suggested in the PR, there are some omissions:
1. `enable_rp1_uart` will route the later firmware debug messages out through GPIOs 14 & 15 - enough to include all of the file accesses (initramfs, overlays etc.).
2. If config.txt contains the old familiar `enable_uart=1` _and_ there is no UART cable plugged into the 3-pin header, the firmware will automatically enable debug output on the 14 & 15, in time to get all of the kernel logging.

In both cases, /dev/serial0 ends up pointing at /dev/ttyAMA0 (UART0), which is the primary in this configuration.

katshann-raspberry:
@christopher-besch  Thank you for your contribution and your investigation of this matter. We'll work on publishing it soon.

katshann-raspberry:
Updates based on this PR are now published to the live site. Many thanks!

PR #4318, opened 2026-07-08
Update link to conditional filters page

iraklis10:
previous link was dead


lurch:
Nope, the original is correct. After going through the build-magic, this link actually ends up as https://www.raspberrypi.com/documentation/computers/config_txt.html#conditional-filters (which works).

And the actual "conditionals section" link you're editing here is displayed at https://www.raspberrypi.com/documentation/computers/config_txt.html#conditional-filtering and as you can see, that link also works.

PR #4323, opened 2026-08-06
Never suggest "sudo -E"

MichaIng:
Passing the full environment from the user session to the root session is seen as potential privacy and security issue.

If a particular environment variable is intended to be passed through, `sudo --preserve-env=VAR` can be used, or `sudo VAR=value` to set it explicitly. The prior is used in the two cases, where the `EDITOR` environment variable is mentioned in the context.


katshann-raspberry:
HI @MichaIng Thank you for your contribution. I'll check with our engineers which variables need to be passed in some of these cases and we'll progress these updates to publish shortly.

lurch:
See also https://github.com/raspberrypi/rpi-eeprom/blob/master/rpi-eeprom-config#L532

MichaIng:
Thanks for the hint, which confirms that it is about the `EDITOR` variable. Instead of setting `export EDITOR=vi` and then passing it via `--preserve-env=EDITOR`, maybe it is easier to instead suggest
```sh
sudo EDITOR=vi rpi-eeprom-config --edit
```
to set it directly? Wording could be aligned and changed to something like

> ...
> ```sh
> sudo rpi-eeprom-config --edit
> ```
> The default text editor is `nano` and may be overridden by setting the 'EDITOR'
> environment variable, e.g. to use Vim in next example.
> ```sh
> sudo EDITOR=vi rpi-eeprom-config --edit
> ```

katshann-raspberry:
There are some other places in the docs where we already listed this command without -E.  So I'm wondering if telling folks how to set their editor is beyond our scope here. Perhaps best to just have the simplest version of the command consistently throughout? (`sudo rpi-eeprom-config --edit`)

MichaIng:
Being a `nano` guy, I am not sure how common it is that CLI tools respect `$EDITOR`, but generally I think it is good to know that `rpi-eeprom-config` does. But should be sufficient in the `--help` output/manpage and maybe in the most canonical page of the docs about the `rpi-eeprom-config` command? I wonder which one it is, as the sections in the first 3 files edited by this PR are almost identical, even the headings 😄:
```
==== Edit the bootloader configuration
==== Editing the current bootloader configuration
=== Edit the configuration
```
Regarding how to pass the environment variable through `sudo`, I agree that it is probably out of scope: Anyone who has multiple editors installed, and prefers one over another, will have likely faced and know about this already. And `nano` is a safe default: faster to grasp than `vi` for noobs, and those who can use and prefer `vi`, can deal with `nano` as well, especially for such a tiny bootloader config. That it is about ~3-5 key=value lines is probably also the best argument, that fiddling with `$EDITOR` in the docs is kinda overkill.

katshann-raspberry:
Updates from this PR are now published on the live site. Many thanks for your contribution.

MichaIng:
For completeness linking the commit here: https://github.com/raspberrypi/documentation/commit/8136a09
Hence the simple way without `EDITOR` variable mention, which should be fine. Thanks!

PR #4326, opened 2026-09-08
AB: Slot update artefacts now have a basename

learmj:
rpi-image-gen now creates the update as <basename>.update.tar.zst so reflect that in the docs.
Ref: https://github.com/raspberrypi/rpi-image-gen/pull/323


stevedavies-rp:
Reviewed and merged internally.

